PR #1, opened 2019-03-01
Add registry manifests, Dockerfile, and playbook

djzager:
This is meant to be a functional starting point for review/feedback.


Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+# docker build -t docker.io/djzager/kubevirt-operators -f Dockerfile .
+FROM quay.io/openshift/origin-operator-registry
+
+# Point this at kubevirt/kubevirt when ready
+# This doesn't work because go isn't installed
+# Need root in order to be able to install git
+# Only need this if the manifests are stored elsewhere
+#USER root
+#RUN yum -y install git
+#RUN git clone --branch olm https://github.com/slintes/kubevirt.git && \
```
djzager: I tried automating the process of grabbing the manifests but since the `operator-registry` image has no Go environment that was a dead end. I could probably do something with multi-stage builds but that was too involved for this initial push.

playbook.yaml
```diff
@@ -0,0 +1,47 @@
+---
```
djzager: This really exists for the sole purpose of starting with a OpenShift 4.0 cluster and adding the stuff to test KubeVirt with OLM.

registry/cdi/1.5.2/cdioperator.v1.5.2.clusterserviceversion.yaml
```diff
@@ -0,0 +1,127 @@
...
+          - apps
+          resources:
+          - deployments
+          - daemonsets
+          - replicasets
+          - statefulsets
+          verbs:
+          - "*"
+      deployments:
+      - name: cdi-operator
```
djzager: This operator creates a deployment (when user creates a CR) that fails because it doesn't have enough permissions. We'll likely need to include that in either the CSV definition or as additional manifests.

registry/kubevirt-group/kubevirtgroup.package.yaml
```diff
@@ -0,0 +1,4 @@
+packageName: kubevirtgroup
```
djzager: kubevirt being the project name and the umbrella name was confusing. So I just made up this name for my sanity.

registry/virt/0.14.0/virtoperator.clusterrolebinding.yaml
```diff
@@ -0,0 +1,16 @@
+---
```
djzager: All of the virtoperator stuff should come from kubevirt/kubevirt but that automated process can be dealt with later IMO.

djzager:
Another potential option for this project is for us to create a single unified CSV with the RBAC/Deployments/CRDs for all of KubeVirts components. Then you could provide KubeVirt as a single item in OLM and operatorhub.

README.md
```diff
@@ -0,0 +1,46 @@
...
+# Ansible Requirements
+
+If you want to run the playbook you will need a recent version of Ansible (`>=
+2.6`) and the [`openshift`
+client](https://github.com/openshift/openshift-restclient-python/).
+
+# Example Run
+
+1. Get a running OpenShift 4.0 Cluster.
+1. Build the registry image: `docker build -t docker.io/djzager/kubevirt-operators:$(git rev-parse --short HEAD) -f Dockerfile .`
```
alexxa: before this step, there should be `git clone this repo`. Then, "From the root of this directory"

README.md
```diff
@@ -0,0 +1,46 @@
...
+1. Push the image: `docker push docker.io/djzager/kubevirt-operators:$(git rev-parse --short HEAD)`
```
alexxa: ?! `denied: requested access to the resource is denied`

README.md
```diff
@@ -0,0 +1,46 @@
...
+1. Run the Playbook `ansible-playbook playbook.yaml -e registry_image=$(git rev-parse --short HEAD)`
```
alexxa: @djzager , Even after I have installed OpenShift Python client, I see

`TASK [Kubevirt namespace state=present]`
`fatal: [localhost]: FAILED! => {"changed": false, "msg": "This module requires the OpenShift Python client. Try `pip install openshift`"}`

advise please.

README.md
```diff
@@ -0,0 +1,66 @@
...
+# Cluster Requirements
+
+The [`registry`](registry/) directory makes use of OLM's bundle support that is
+provided only via the `grpc` type `CatalogSource`. You can see us create this
+`CatalogSource` in the provided [`playbook.yaml`](playbook.yaml). This means
+that you **must** have a relatively new **OpenShift 4.0** cluster.
+
+# Ansible Requirements
+
+If you want to run the playbook you will need a recent version of Ansible (`>=
```
davidvossel: whoa... playbooks? what's this about. Can't all of this be managed by posting manifests using oc/kubectl tooling?

djzager: You absolutely could use kubectl|oc. I only wrote the playbook because I was tired of running `sed "s|image: .*|image: docker.io/djzager/kubevirt-operators:sometag|" | kubectl (create|apply|delete) -f -` and a simple playbook was the easiest method available.

It's totally optional and can be removed, but I do believe that it makes iterating on this stuff easier.

davidvossel: This sounds like you're looking for code to generate manifests using a template. We have to handle the same thing in kubevirt.

Take a look at the https://github.com/kubevirt/kubevirt/tree/master/tools/manifest-templator file. We use that to inject dynamic data into templated manifests from variables to do exactly what you're talking about. 

It should be a goal to remove dependency of ansible from the dev workflow

djzager: Are you using manifests as a general reference to YAML files with one or more Kubernetes object specs inside or the components to an OLM CSV? If you are talking about the latter, I think updating this project to generate a KubeVirt CSV from all of the component projects is absolutely something we should do. In that respect I would need a little more guidance on how we proceed down that route (ie. do I include all the other projects as submodules in this one, grab them by github URL, or grab them out of a released image of KubeVirt|CDI|etc).

Writing a manifest-templator so that I can template out a Namespace, OperatorGroup, and CatalogSource for the sole purpose of deploying it into my development cluster strikes me as an odd use of time. If removing the dependency on Ansible is important, I would just make a single yaml file and go back to my `sed + kubectl` method.

> It should be a goal to remove dependency of ansible from the dev workflow

I guess that all of my working with Ansible has biased me somehow. Your saying that it should be a goal to remove this Ansible dependency makes me think there is more to this than just a preference. For my edification, would you be able to elaborate?

registry/kubevirt/0.0.1/kubevirt.v0.0.1.clusterserviceversion.yaml
```diff
@@ -0,0 +1,252 @@
+apiVersion: operators.coreos.com/v1alpha1
```
davidvossel: where did this file come from?  I'd expect this to be vendored from the KubeVirt packages once the kubevirt olm manifest generator is merged.

registry/kubevirt/0.0.1/virtoperator.clusterrole.yaml
```diff
@@ -0,0 +1,23 @@
+---
```
davidvossel: this needs to be vendored from kubevirt as well.

Basically any manifest that comes from kubevirt, cdi, etc... needs to be vendored from those projects. That's going to be the best way to keep all of this in sync properly.

README.md
davidvossel: > I guess that all of my working with Ansible has biased me somehow. Your saying that it should be a goal to remove this Ansible dependency makes me think there is more to this than just a preference. For my edification, would you be able to elaborate?

we've been moving toward a pattern where manifests are generated by code. Take https://github.com/kubevirt/kubevirt/pull/2062 for example. The kubevirt olm manifest originates from golang code.

Ideally this project venders the work kubevirt has done to generate manifests from code and uses that to generate the hyperconverged olm manifest. 

For example, look at the OLM manifest in KubeVirt in that pr I referenced. https://github.com/kubevirt/kubevirt/blob/66abe1ad92614c066412d056b508478c15de03a3/manifests/release/kubevirt.VERSION.csv.yaml.in

You'll see things like ```{{.OperatorDeploymentSpec}}``` sprinkled into that file. The manifest templator is injecting the kubevirt operator deployment spec into that template file, and the deployment spec is actually generated by golang code.

In order to leverage all this work in KubeVirt, this package needs to be following a similar pattern.

djzager:
Now I understand. That is exactly the guidance I was looking for. I will
execute to this on Monday and likely create a separate Pull Request.

Thank you for the explanation.

PR #7, opened 2019-03-20
Add Makefile

arachmani:


Makefile
```diff
@@ -0,0 +1,49 @@
...
+	# Create projects
+	oc new-project kubevirt;
+	oc new-project cdi;
+	# Switch project to kubevirt
+	oc project kubevirt;
+endif
+	# Deploy HCO manifests
+	${CMD} create -f deploy/;
+	${CMD} create -f deploy/crds/hco_v1alpha1_hyperconverged_crd.yaml;
+	${CMD} create -f deploy/crds/hco_v1alpha1_hyperconverged_cr.yaml;
```
rthallisey: Have to be careful here because the API server may not create the CRD right away and this could fail.  It would be easier to put this into a bash script under `hack/` so you can add some logic.

Makefile
```diff
@@ -0,0 +1,49 @@
...
+CDI_OPERATOR_URL=$(shell curl --silent "https://api.github.com/repos/kubevirt/containerized-data-importer/releases/latest" | grep browser_download_url | grep "cdi-operator.yaml\"" | cut -d'"' -f4)
+KUBEVIRT_OPERATOR_URL=$(shell curl --silent "https://api.github.com/repos/kubevirt/kubevirt/releases/latest" | grep browser_download_url | grep "kubevirt-operator.yaml\"" | cut -d'"' -f4)
+CURRENT_CONTEXT=$(shell kubectl config current-context)
+
+ifneq (${KUBECTL},)
+	CMD=oc
+endif
+
+deploy:
+ifeq (${CMD},kubectl)
```
rthallisey: I would do this in a bash script.  It's hard to do logic correctly in a Makefile.  Bash is a little better at this.  Let's try to cut these recipes down to 1 line.

Makefile
```diff
@@ -0,0 +1,49 @@
...
+remove:
+ifeq (${CMD},kubectl)
+	# Switch namespace to kubevirt
+	kubectl config set-context ${CURRENT_CONTEXT} --namespace=kubevirt;
+else
+	# Switch project to kubevirt
+	oc project kubevirt;
+endif
+	# Delete kubevirt-operator
+	${CMD} delete -f ${KUBEVIRT_OPERATOR_URL};
```
rthallisey: Double check that this is going to work.  I think you're going to end up with the kubevirt namespace stuck in terminating.   I think this will help fix it: ```oc delete -n kubevirt apiservice v1alpha3.subresources.kubevirt.io```.

hack/common.sh
```diff
@@ -0,0 +1,30 @@
...
+# limitations under the License.
+#
+# Copyright 2017 Red Hat, Inc.
+#
+
+CDI_OPERATOR_URL=$(curl --silent "https://api.github.com/repos/kubevirt/containerized-data-importer/releases/latest" \
+ | grep browser_download_url | grep "cdi-operator.yaml\"" | cut -d'"' -f4)
+KUBEVIRT_OPERATOR_URL=$(curl --silent "https://api.github.com/repos/kubevirt/kubevirt/releases/latest" \
+ | grep browser_download_url | grep "kubevirt-operator.yaml\"" | cut -d'"' -f4)
+KUBECTL=$(which kubectl 2> /dev/null)
```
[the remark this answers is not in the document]
irosenzw: IMO, we should deal with other cases later on, this is enough for now.

PR #9, opened 2019-03-24
Added option to set imagePullPolicies for KubeVirt and CDI

lveyde:
This patch allows to set imagePullPolicy for KubeVirt and/or CDI by
setting the appropriate variable(s) in the HCO CR under the Spec.
section: KubeVurtImagePullPolicy and/or CDIImagePullPolicy

Please note that currently the CR(s) will NOT be re-created on
the policy change, thus one needs to edit the HCO CR manifest
file.


pkg/apis/hco/v1alpha1/hyperconverged_types.go
```diff
@@ -13,6 +13,8 @@ type HyperConvergedSpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
 	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
+    KubeVirtImagePullPolicy	string	`json:"KubeVirtImagePullPolicy"`
```
rthallisey: I think you'll need to add `omitempty` to each of these vars to prevent `KubeVirtImagePullPolicy: ""`, which is currently a valid input.

lveyde: Fixed, even though I'm checking in the code for empty string, and thus empty string has basically the same effect as not including the key.

PR #10, opened 2019-03-25
add network addons operator

phoracek:


phoracek:
@rthallisey could you please take a look?

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -136,6 +138,24 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
 		return result, err
 	}
 
+	// Define a new CDI object
```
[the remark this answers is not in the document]
phoracek: Done

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -192,3 +212,21 @@ func newCDIForCR(cr *hcov1alpha1.HyperConverged) *cdi.CDI {
...
+
+// newNetworkAddonsForCR returns a NetworkAddonsConfig CR
+func newNetworkAddonsForCR(cr *hcov1alpha1.HyperConverged) *networkaddons.NetworkAddonsConfig {
+	labels := map[string]string{
+		"app": cr.Name,
+	}
+	return &networkaddons.NetworkAddonsConfig{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      networkaddonsnames.OPERATOR_CONFIG,
+			Namespace: networkaddonsnames.APPLIED_NAMESPACE,
```
djzager: The hyperconverged-cluster-operator is not (or at least should not) have permissions to be creating components across namespaces. Maybe I don't understand the direction that we are going but I would expect _User subscribes to the HCO and gets the HCO operator and the component operators (KubeVirt, CDI, ClusterNetworkAddons, etc.). Then a HCO CR would result in all of the components installed in the namespace where the CR is created._

djzager: @rthallisey you may be most qualified to direct here.

phoracek: Sorry this is a mistake, NetworkAddonsConfig is cluster-wide, non-namespaced resource. Will fix it.

There are two issues connected to your request:

1) Since network addons alter hosts, they can be deployed only once, therefore we have the resource cluster-wide and with a single acknowledged name "cluster". IIUIC, this is the same for CDI, therefore we could copy CDI leader election.
2) Network addons operator deploys third-party components (multus and linux bridge CNI) into their own namespaces since they are self-container and separated projects.

README.md
```diff
@@ -41,6 +41,14 @@ Launch the CDI operator.
 oc create -f https://github.com/kubevirt/containerized-data-importer/releases/download/v1.6.0/cdi-operator.yaml
 ```
 
+Launch the Cluster Network Addons operator.
+```bash
+oc create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/v0.1.0/cluster-network-addons-operator_00_namespace.yaml
+oc create -f https://github.com/kubevirt/cluster-network-addons-operator/releases/download/v0.1.0/cluster-network-addons-operator_01_crd.yaml
```
lveyde: Shouldn't the HCO Operator itself install these CRDs?

phoracek: I think that it should eventually. However, now it has to be done manually. Correct me if I'm wrong please, I base this only on the readme.

rthallisey: That's out of scope for the HCO.  OLM creates the CRDs.

phoracek:
I was not able to verify this PT as it is, since there seems to be a bug in CDI. However, after commenting out CDI parts from the operator, it deployed KubeVirt and network components OK.

phoracek:
/hold

Hold this PR until we make it possible to run it namespaced.

phoracek:
/hold cancel

We cannot run in the same namespace as HCO. Network addons operator (and IIUIC CDI as well) needs to create a ClusterRoleBinding. The ClusterRoleBinding specifies ServiceAccount and its namespace. The namespace has to be hardcoded in operator's RBAC manifest. Therefore, until we are integrated with OLM which can adjust those manifests, we have no other choice but to deploy Network Addons operator in its own namespace.

@lveyde @djzager Please take a look and tell me whether there is something I should change in this PR.

phoracek:
@lveyde @rthallisey could you please take another look? I added network addons operator deployment to the hack/ directory and added support for imagePullPolicy

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -210,3 +245,21 @@ func newCDIForCR(cr *hcov1alpha1.HyperConverged) *cdi.CDI {
...
+		"app": cr.Name,
+	}
+	return &networkaddons.NetworkAddonsConfig{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:   networkaddonsnames.OPERATOR_CONFIG,
+			Labels: labels,
+		},
+		Spec: networkaddons.NetworkAddonsConfigSpec{
+			Multus:          &networkaddons.Multus{},
+			LinuxBridge:     &networkaddons.LinuxBridge{},
```
rthallisey: Why not expose Multus and LinuxBridge as configurable options too?

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -154,6 +171,24 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
...
+	networkAddonsCR := newNetworkAddonsForCR(instance)
+
+	// Set HyperConverged instance as the owner and controller
+	if err := controllerutil.SetControllerReference(instance, networkAddonsCR, r.scheme); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// Check if this NetworkAddonsConfig CR already exists
+	foundNetworkAddons := &networkaddons.NetworkAddonsConfig{}
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: networkaddonsnames.OPERATOR_CONFIG, Namespace: ""}, foundNetworkAddons)
```
rthallisey: They are the same, but I'd prefer to use the `networkAddonsCR` to get the CR Name instead of grabbing it from the networkaddon code.  That way if we ever move away from the `OPERATOR_CONFIG` name, we'll have an abstraction that will pickup the change.

hack/operator-test.sh
```diff
@@ -27,23 +27,26 @@ done
...
 kubevirt_sed
 cdi_sed
+network_addons_sed
 echo "Replaced image strings"
 
 set +e
 set -x
 
-oc create -f ${TEMP_DIR}/
+./oc create -f ${TEMP_DIR}/
```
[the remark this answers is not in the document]
phoracek: Oh right, forgot this from testing

pkg/controller/hyperconverged/hyperconverged_controller.go
phoracek: HCO should be opinionated. In my opinion Linux Bridge and Multus should be always deployed with this hyperconverged infrastructure.

hack/operator-test.sh
```diff
@@ -27,23 +27,26 @@ done
...
+./oc wait pod $CDI_POD --for condition=Ready -n cdi
+./oc wait pod $NETWORK_ADDONS_POD --for condition=Ready -n cluster-network-addons-operator
 
-oc create -f ${TEMP_DIR}/crs
+./oc create -f ${TEMP_DIR}/crs
 
 echo "Let the API server process the CRs"
 sleep 10
-oc wait kubevirt kubevirt --for condition=Ready -n kubevirt
+./oc wait kubevirt kubevirt --for condition=Ready -n kubevirt
```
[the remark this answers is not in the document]
phoracek: I am working on status reporting for network addons. It's not ready yet though.

pkg/controller/hyperconverged/hyperconverged_controller.go
phoracek: Done

hack/operator-test.sh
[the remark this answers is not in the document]
phoracek: Done

pkg/controller/hyperconverged/hyperconverged_controller.go
rthallisey: That's fine, but we can default these values behind the scenes.  I'm wondering if we should give the user the opportunity to configure them to be something other than the defaults.

phoracek: Could we start with this and introduce configurable values in a different PR? I think it should be discussed with wider audience from the whole project (and it may also affect kubevirt and CDI). Where is the line between flexibility and "the right way to use hyperconverged cluster".

rthallisey: @phoracek sounds good to me

phoracek:
All comments should be addressed now.

@djzager can we merge this PR before #8 or do we need to wait for it?

phoracek:
Thank you! :)

PR #11, opened 2019-03-25
Added KubeVirt/CDI imagePullPolicy verification code

lveyde:
If invalid imagePullPolicy is supplied then the system is never reconciled.
Thus it's important to verify that the values for both KubeVirt and CDI are correct.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -85,6 +85,14 @@
   version = "v2.9.3"
 
 [[projects]]
+  digest = "1:448bf60e9dce6d5fd9dfa8fd9eb9a23163424ff4e4f142456405781e942e3998"
+  name = "github.com/evanphx/json-patch"
+  packages = ["."]
+  pruneopts = "NT"
+  revision = "026c730a0dcc5d11f93f1cf1cc65b01247ea7b6f"
+  version = "v4.5.0"
+
+[[projects]]
   digest = "1:81466b4218bf6adddac2572a30ac733a9255919bc2f470b4827a317bd4ee1756"
   name = "github.com/ghodss/yaml"
   packages = ["."]
@@ -1026,6 +1034,7 @@
     "rest",
     "rest/watch",
     "restmapper",
+    "testing",
     "third_party/forked/golang/template",
     "tools/auth",
     "tools/cache",
@@ -1196,6 +1205,7 @@
     "pkg/client",
     "pkg/client/apiutil",
     "pkg/client/config",
+    "pkg/client/fake",
     "pkg/controller",
     "pkg/controller/controllerutil",
     "pkg/event",
@@ -1255,6 +1265,7 @@
     "github.com/MarSik/kubevirt-ssp-operator/pkg/apis",
     "github.com/MarSik/kubevirt-ssp-operator/pkg/apis/kubevirt/v1",
     "github.com/ghodss/yaml",
+    "github.com/go-logr/logr",
     "github.com/go-openapi/spec",
     "github.com/kubevirt/cluster-network-addons-operator/pkg/apis",
     "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/v1alpha1",
@@ -1265,6 +1276,7 @@
     "github.com/kubevirt/web-ui-operator/pkg/components",
     "github.com/onsi/ginkgo",
     "github.com/onsi/gomega",
+    "github.com/openshift/api/security/v1",
     "github.com/operator-framework/operator-sdk/pkg/k8sutil",
     "github.com/operator-framework/operator-sdk/pkg/leader",
     "github.com/operator-framework/operator-sdk/pkg/log/zap",
@@ -1281,6 +1293,7 @@
     "k8s.io/apimachinery/pkg/runtime",
     "k8s.io/apimachinery/pkg/runtime/schema",
     "k8s.io/apimachinery/pkg/types",
+    "k8s.io/client-go/kubernetes/scheme",
     "k8s.io/client-go/plugin/pkg/client/auth/gcp",
     "k8s.io/code-generator/cmd/client-gen",
     "k8s.io/code-generator/cmd/conversion-gen",
@@ -1300,6 +1313,7 @@
     "kubevirt.io/kubevirt/pkg/virt-operator/creation/rbac",
     "sigs.k8s.io/controller-runtime/pkg/client",
     "sigs.k8s.io/controller-runtime/pkg/client/config",
+    "sigs.k8s.io/controller-runtime/pkg/client/fake",
     "sigs.k8s.io/controller-runtime/pkg/controller",
     "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil",
     "sigs.k8s.io/controller-runtime/pkg/handler",
--- a/deploy/converged/olm-catalog/kubevirt-hyperconverged/0.0.1/kubevirt-hyperconverged-operator.v0.0.1.clusterserviceversion.yaml
+++ b/deploy/converged/olm-catalog/kubevirt-hyperconverged/0.0.1/kubevirt-hyperconverged-operator.v0.0.1.clusterserviceversion.yaml
@@ -1368,7 +1368,7 @@ spec:
                       - name: WATCH_NAMESPACE
                         value: ""
                       - name: VIRT_LAUNCHER_TAG
-                        value: v0.17.0
+                        value: v0.18.1
                       - name: OPERATOR_NAME
                         value: "kubevirt-ssp-operator"
 
--- a/deploy/converged/operator.yaml
+++ b/deploy/converged/operator.yaml
@@ -239,7 +239,7 @@ spec:
             - name: WATCH_NAMESPACE
               value: ""
             - name: VIRT_LAUNCHER_TAG
-              value: v0.17.0
+              value: v0.18.1
             - name: OPERATOR_NAME
               value: "kubevirt-ssp-operator"
 
--- a/deploy/converged/ssp-op
+++ b/deploy/converged/ssp-op
@@ -153,7 +153,7 @@ spec:
             - name: WATCH_NAMESPACE
               value: ""
             - name: VIRT_LAUNCHER_TAG
-              value: v0.17.0
+              value: v0.18.1
             - name: OPERATOR_NAME
               value: "kubevirt-ssp-operator"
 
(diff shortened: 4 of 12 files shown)


pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -2,13 +2,14 @@ package hyperconverged
 
 import (
 	"context"
+        "errors"
```
phoracek: this looks malformed, spaces instead of \t?

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -109,13 +110,25 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
         if instance.Spec.KubeVirtImagePullPolicy != "" {
            reqLogger := log.WithValues("imagePullPolicy", instance.Spec.KubeVirtImagePullPolicy)
            reqLogger.Info("HCO CR contains KubeVirt Image Pull Policy")
-           KubeVirtImagePullPolicy = instance.Spec.KubeVirtImagePullPolicy
+           _, err = verifyPullPolicyType(instance.Spec.KubeVirtImagePullPolicy)
+           if err != nil {
+               reqLogger := log.WithValues("HCO CR imagePullPolicy", instance.Spec.KubeVirtImagePullPolicy, "default imagePullPolicy", KubeVirtImagePullPolicy)
+               reqLogger.Info("HCO CR KubeVirt Pull Policy is invalid. Reverting to default policy")
```
phoracek: i don't see the code "reverting" the policy. also, shouldn't we fail the operation rather then try to "fix" it? what if i pass "Alwayss" and this defaults me to "IfNotPresent".

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -210,3 +223,18 @@ func newCDIForCR(cr *hcov1alpha1.HyperConverged) *cdi.CDI {
 		},
 	}
 }
+
+// Verify if the value is a valid PullPolicy
+func verifyPullPolicyType(policy string) (bool, error) {
```
phoracek: why do you return the bool if you never use it?

pkg/apis/hco/v1alpha1/hyperconverged_types.go
```diff
@@ -13,7 +13,9 @@ type HyperConvergedSpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
 	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
+// +kubebuilder:validation:Enum=Always,Never,IfNotPresent
     KubeVirtImagePullPolicy	string	`json:"KubeVirtImagePullPolicy,omitempty"`
+// +kubebuilder:validation:Enum=Always,Never,IfNotPresent
```
djzager: shouldn't these comments be lined up?

djzager: Maybe you should run `go fmt` on this stuff?

pkg/controller/hyperconverged/hyperconverged_controller.go
djzager: I agree. I think `go fmt` should be run on this file.

pkg/controller/hyperconverged/hyperconverged_controller.go
lveyde: Simply because I prefer to use a regular convention, returning error as well.

pkg/controller/hyperconverged/hyperconverged_controller.go
lveyde: How exactly you think we can propagate this failure back to the user?

djzager: Status on the CR

lveyde: We basically have 2 options here: we either stop processing - which is actually the same as if no verification code would have been added, still leaving the user clueless as to what have happened, and why the Operators haven't been started - even though the HCO CR manifest was imported.

OR, we can verify the values and disregard invalid ones, backing off to the default values.
While still properly mentioning this in the log.

IMHO the second option is much better - the system is still functioning, and if the use of default policy still creates some issue then the user will look up the logs and find the message there.

pkg/controller/hyperconverged/hyperconverged_controller.go
lveyde: +1 go fmt it is then :-)

pkg/apis/hco/v1alpha1/hyperconverged_types.go
lveyde: the same - will run the go fmt and let it format the code.

generally I did this intentionally to differentiate between the placeholder comments, which provide instructions to the user and bring no functional value and the comments that have the functional value.

pkg/controller/hyperconverged/hyperconverged_controller.go
djzager: I don't understand. Putting failure status messages on the CR is a complete and totally valid response to an invalid spec.

lveyde: > 
> 
> I don't understand. Putting failure status messages on the CR is a complete and totally valid response to an invalid spec.

So you think that we'll actually be better off if we halt/stop the processing at this point?

From my point of view it's better to try and continue running if we can fix the situation, especially as we don't have an option to note the user directly during or just after the CR manifest was created - and our only options are posting something to log and/or CR status.

Also right now the CR is not re-processed on the change (edit), thus in case of i.e. typo/invalid policy one will need to actually delete the CR, fix the issue in the file and re-import again.

phoracek: Trying to be smart, in case of a failure, can lead to unexpected behaviour. If the configuration is wrong. Also, I can imagine it could be problematic once we need to upgrade to newer version of the spec. Something invalid used to work, but it does not anymore. I think we should be strict with this, but I won't block this PR.

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -106,17 +107,29 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
 		return reconcile.Result{}, err
 	}
 
-        if instance.Spec.KubeVirtImagePullPolicy != "" {
-           reqLogger := log.WithValues("imagePullPolicy", instance.Spec.KubeVirtImagePullPolicy)
-           reqLogger.Info("HCO CR contains KubeVirt Image Pull Policy")
-           KubeVirtImagePullPolicy = instance.Spec.KubeVirtImagePullPolicy
-        }
+	if instance.Spec.KubeVirtImagePullPolicy != "" {
```
rthallisey: I'm wondering if we can improve the abstraction here for validating input.   We are going to need to add code like this for every configuration we support, so this func will fill up fast.

pkg/controller/hyperconverged/hyperconverged_controller.go
rthallisey: I think there are some additional risks we take by reverting to defaults.  For example, if one parameter fails, do we revert all parameters back to their defaults? Or only the one failed param?  If we only do the failed param, we run the risk if using a default that's not compatible with the user's imputed config.  We also aren't sure if our defaults will work in all environments.

IMO, we should report status on the CR like `invalid input for parameter ImagePullPolicy` and let the user take action.

pkg/apis/hco/v1alpha1/hyperconverged_types.go
```diff
@@ -23,6 +25,8 @@ type HyperConvergedStatus struct {
 	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
 	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
 	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
+	KubeVirt string `json:"KubeVirt"`
```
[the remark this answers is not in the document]
lveyde: > 
> 
> Shouldn't status contain well structured information about the state? As it is in KubeVirt with state, reason, conditions etc?

I prefer to follow the KISS principle, not sure we need a very complicated status structure.
Also, at least currently there is not much info to show.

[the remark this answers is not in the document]
phoracek: fair enough :) my understanding was that we need that complicated structure to integrate with whatever will be the higher layer, but there is no need to have it in this very PR

pkg/controller/hyperconverged/hyperconverged_controller.go
lveyde: > 
> 
> I'm wondering if we can improve the abstraction here for validating input. We are going to need to add code like this for every configuration we support, so this func will fill up fast.

It's surely something to think about for the future, i.e. creating some sort of parameter validation framework, but this out of the scope of the current task IMHO.

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -106,23 +109,54 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
...
+
+	}
 
 	// Define a new KubeVirt object
+	reqLogger.Info("Creating a new KubeVirt object")
 	virtCR := newKubeVirtForCR(instance)
 
 	// Set HyperConverged instance as the owner and controller
 	if err := controllerutil.SetControllerReference(instance, virtCR, r.scheme); err != nil {
+		reqLogger.Info(err.Error())
```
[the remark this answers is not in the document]
lveyde: > 
> 
> why don't you reqLogger.Error?

Thought about it and decided that there is no much difference in this case.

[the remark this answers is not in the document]
phoracek: I think that if we are logging something that is an error ("failed to set controller reference"), it should not go to info ("XY was created"). Maybe it does not change anything with the current logging library, but if we switch to something that actually writes ERROR or INFO in the beginning of the log, it should go to ERROR. Also, it will make sure that it will be always logged even with higher level of logging. It will be obvious which event caused the reconcile to repeat etc.

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -182,6 +249,7 @@ func newKubeVirtForCR(cr *hcov1alpha1.HyperConverged) *kubevirt.KubeVirt {
 	labels := map[string]string{
 		"app": cr.Name,
 	}
+	log.Info("Creating a new KubeVirt object")
```
phoracek: this log should be on a different place. it implies (at least to me) that the object is being created on kube api, while you only initialize the structure here.

PR #15, opened 2019-04-01
Add Makefile

arachmani:


hack/deploy.sh
```diff
@@ -0,0 +1,52 @@
...
+    oc new-project kubevirt
+    oc new-project cdi
+    
+    # Switch project to kubevirt
+    oc project kubevirt;
+fi
+# Deploy HCO manifests
+"${CMD}" create -f deploy/
+"${CMD}" create -f deploy/crds/hco_v1alpha1_hyperconverged_crd.yaml
+"${CMD}" create -f deploy/crds/hco_v1alpha1_hyperconverged_cr.yaml
```
rthallisey: I would move the HCO CR creation to last in this file.  It will give the cluster some time to process the CRD and it will have the HCO create the components after everything is healthy and running.

hack/clean.sh
```diff
@@ -0,0 +1,33 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Copyright 2017 Red Hat, Inc.
+#
+
+source hack/common.sh
+
```
rthallisey: Need to let the operators cleanup their components before we delete the operators.  I don't think the HCO is capable of cleaning up component operator CRs yet, so we'll have to remove all the CRs individually.
```bash
"${CMD}" delete -f <cdi-cr.yaml>
"${CMD}" delete -f <kubevirt-cr.yaml>
"${CMD}" delete -f <hco-cr.yaml>

# Wait for all crs to be gone
```

hack/clean.sh
```diff
@@ -0,0 +1,33 @@
...
+# Delete kubevirt-operator
+"${CMD}" delete crd kubevirts.kubevirt.io
```
rthallisey: I think the kubevirt manifest should cover this?

hack/clean.sh
rthallisey: I tested this yesterday.  The hco will remove the component CRs.  I think you only need to remove the hco-cr and hco-crd

arachmani: Once the hco-crd is removed the hco-cr is removed as well.

hack/clean.sh
```diff
@@ -0,0 +1,33 @@
...
+"${CMD}" delete -n kubevirt apiservice v1alpha3.subresources.kubevirt.io
+"${CMD}" delete -f "${KUBEVIRT_OPERATOR_URL}"
+
+# Remove HCO manifests
+"${CMD}" delete -f deploy/
+"${CMD}" delete -f deploy/crds/hco_v1alpha1_hyperconverged_crd.yaml
```
rthallisey: The HCO needs to be removed first in order for the HCO to cleanup the kubevirt & cdi CRs.

irosenzw:
@rthallisey please review

hack/clean.sh
rthallisey: My comment here isn't very clear.  We need to remove the HCO's CR first and give the operator time to clean up kubevirt and cdi CRs.

Also if you haven't yet, rebase this patch.  The content in `deploy` has changed.

```bash
"${CMD}" delete -f deploy/standard/crds/hco.cr.yaml

# Some command to wait until object is removed
# "${CMD}" wait hco/hyperconverged-cluster

"${CMD}" delete -f deploy/standard/crds/hco.crd.yaml
"${CMD}" delete -f deploy/
```

rthallisey:
This is very close.  I think it will need one more update to merge.

rthallisey:
Using the makefile should cover #30 

fixes https://github.com/kubevirt/hyperconverged-cluster-operator/issues/30

PR #19, opened 2019-04-02
WIP use dynamic rest mapper

phoracek:
Default operator mapper is not able to use resources that were
created after its start. With dynamic REST mapper it will be.

Without this PR, operator restarts and fails with:

```
{"level":"error","ts":1554243680.2445314,"logger":"kubebuilder.source","msg":"if kind is a CRD, it should be installed before calling Start","kind":"KubeVirt.kubevirt.io","error":"no matches for kind \"KubeVirt\" in version \"kubevirt.io/v1alpha3\"","stacktrace":"github.com/kubevirt/hyperconverged-cluster-operator
...
```

**Not yet sure if the solution works**


phoracek:
Looks like this doesn't help.

PR #23, opened 2019-04-09
Hook up the HCO's configuration to all integrated CRs

rthallisey:
related https://github.com/kubevirt/hyperconverged-cluster-operator/pull/22

### DIFF
--- a/deploy/standard/olm-catalog/kubevirt-hyperconverged/0.0.1/kubevirt-hyperconverged-operator.v0.0.1.clusterserviceversion.yaml
+++ b/deploy/standard/olm-catalog/kubevirt-hyperconverged/0.0.1/kubevirt-hyperconverged-operator.v0.0.1.clusterserviceversion.yaml
@@ -16,8 +16,9 @@ metadata:
             "namespace":"kubevirt-hyperconverged"
           },
           "spec": {
-            "CDIImagePullPolicy":"IfNotPresent",
-            "KubeVirtImagePullPolicy":"IfNotPresent"
+            "Version":"2.0",
+            "ContainerRegistry":"quay.io/kubevirt",
+            "ImagePullPolicy":"IfNotPresent"
           }
         }
       ]
--- a/pkg/apis/hco/v1alpha1/hyperconverged_types.go
+++ b/pkg/apis/hco/v1alpha1/hyperconverged_types.go
@@ -1,30 +1,26 @@
 package v1alpha1
 
 import (
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
-
-// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
-// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
 // HyperConvergedSpec defines the desired state of HyperConverged
 // +k8s:openapi-gen=true
 type HyperConvergedSpec struct {
-	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
-	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
-	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
-	KubeVirtImagePullPolicy      string `json:"KubeVirtImagePullPolicy,omitempty"`
-	CDIImagePullPolicy           string `json:"CDIImagePullPolicy,omitempty"`
-	NetworkAddonsImagePullPolicy string `json:"NetworkAddonsImagePullPolicy,omitempty"`
+	// Version of all the components under the HCO
+	Version string `json:"Version"`
+
+	// Always, IfNotPresent, Never
+	ImagePullPolicy v1.PullPolicy `json:"ImagePullPolicy,omitempty"`
+
+	// Registry container are pulled from: quay.io/kubevirt
+	ContainerRegistry string `json:"ContainerRegistry"`
 }
 
 // HyperConvergedStatus defines the observed state of HyperConverged
 // +k8s:openapi-gen=true
-type HyperConvergedStatus struct {
-	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
-	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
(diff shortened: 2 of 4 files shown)


pkg/config/config.go
```diff
@@ -0,0 +1,60 @@
...
+		ImagePullPolicy: hcoSpec.ImagePullPolicy,
+	}
+}
+
+// NewNetworkAddonsSpec - returns a NetworkAddonsSpec object populated by HCO variables
+func NewNetworkAddonsSpec(hcoSpec *hcov1alpha1.HyperConvergedSpec) networkaddons.NetworkAddonsConfigSpec {
+	return networkaddons.NetworkAddonsConfigSpec{
+		Multus:          &networkaddons.Multus{},
+		LinuxBridge:     &networkaddons.LinuxBridge{},
+		ImagePullPolicy: string(hcoSpec.ImagePullPolicy),
```
rthallisey: @phoracek networkaddons should use `k8s.io/api/core/v1.PullPolicy` as a type instead of a string.

phoracek: thanks! I will tackle it with our upcoming release. Then I open a PR on HCO bumping the version and fixing this

pkg/apis/hco/v1alpha1/hyperconverged_types.go
```diff
@@ -1,30 +1,26 @@
...
 // +k8s:openapi-gen=true
 type HyperConvergedSpec struct {
-	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
-	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
-	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
-	KubeVirtImagePullPolicy      string `json:"KubeVirtImagePullPolicy,omitempty"`
-	CDIImagePullPolicy           string `json:"CDIImagePullPolicy,omitempty"`
-	NetworkAddonsImagePullPolicy string `json:"NetworkAddonsImagePullPolicy,omitempty"`
+	// Version of all the components under the HCO
+	Version string `json:"Version"`
```
djzager: Is the idea here that each of the components is going to have a version, imagePullPolicy, and containerRegistry (maybe under a different name) configurable in their CR Spec?

deploy/standard/olm-catalog/kubevirt-hyperconverged/0.0.1/kubevirt-hyperconverged-operator.v0.0.1.clusterserviceversion.yaml
```diff
@@ -16,8 +16,9 @@ metadata:
             "namespace":"kubevirt-hyperconverged"
           },
           "spec": {
-            "CDIImagePullPolicy":"IfNotPresent",
-            "KubeVirtImagePullPolicy":"IfNotPresent"
+            "Version":"2.0",
```
djzager: We may need to remove the deploy directory entirely to prevent people from modifying generated files...or remove the generator. You must modify the template `templates/olm-catalog/kubevirt-hyperconverged/VERSION/kubevirt-hyperconverged-operator.VERSION.clusterserviceversion.yaml.in` for this to not get blown away by the generator.

pkg/config/config.go
```diff
@@ -0,0 +1,60 @@
...
+		Version:           HcoVersion,
+		ContainerRegistry: HcoContainerRegistry,
+		ImagePullPolicy:   HcoImagePullPolicy,
+	}
+}
+
+// NewKubeVirtSpec - returns a KubeVirtSpec object populated by HCO variables
+func NewKubeVirtSpec(hcoSpec *hcov1alpha1.HyperConvergedSpec) kubevirt.KubeVirtSpec {
+	return kubevirt.KubeVirtSpec{
+		ImageTag:        hcoSpec.Version,
```
djzager: So KubeVirt has to be able to respect the HCO versions? Out of the box, if the HCO version were `0.1.0` wouldn't you end up with some old  version of KubeVirt...or worse if the version is `1.0.0` you would be trying to install a KubeVirt version that doesn't exist.

djzager:
I personally would prefer for the HCO to have no spec at all and each version of the HCO would have the versions of components encoded inside.

pkg/apis/hco/v1alpha1/hyperconverged_types.go
rthallisey: Yes, each component operator should express these variables in some way on their CR.  https://github.com/kubevirt/hyperconverged-cluster-operator/pull/22#discussion_r273484482

deploy/standard/olm-catalog/kubevirt-hyperconverged/0.0.1/kubevirt-hyperconverged-operator.v0.0.1.clusterserviceversion.yaml
rthallisey: ah thanks, I went to the wrong dir.

pkg/config/config.go
rthallisey: Yes.  See https://github.com/kubevirt/hyperconverged-cluster-operator/pull/22#discussion_r273214077.  Downstream, this will work.  Upstream, this is a problem which we are going to have a follow up discussion on.

rthallisey:
> I personally would prefer for the HCO to have no spec at all and each version of the HCO would have the versions of components encoded inside.
>
The trouble with that exposing nothing is `containerRegistry` needs to be configurable so users can pull from a private registry.  `version` has the potential to he hidden on the hco.

djzager:
> The trouble with that exposing nothing is `containerRegistry` needs to be configurable so users can pull from a private registry. `version` has the potential to he hidden on the hco.

Why does this need to be configured by the end user? Are you referring to the downstream versus upstream case? If you mean something other than a Red Hat owned private registry, how would you know that all of the necessary images are in that containerRegistry?

rthallisey:
>Why does this need to be configured by the end user? Are you referring to the downstream versus upstream case? If you mean something other than a Red Hat owned private registry, how would you know that all of the necessary images are in that containerRegistry?
>
The issue is downstream.  Someone may have an on prem registry they use to hold downstream images pulled from the rhcc.  I'm curious how openshift is dealing with this because they should have the same problem.

PR #33, opened 2019-04-16
add SSP operator

fromanirh:
The SSP operator is unique because
- it is a pure ansible-operator, so we added (mostly)autogenerated
  go APIs to facilitate the integration with HCO
- it manages and wants to manage more than one CR


fromanirh:
some TODOs still pending

rthallisey:
also add ssp to `deploy.sh` and `clean.sh` so it gets picked up by the Makefile

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -30,6 +32,7 @@ var (
 	KubeVirtImagePullPolicy      = "IfNotPresent"
 	CDIImagePullPolicy           = "IfNotPresent"
 	NetworkAddonsImagePullPolicy = "IfNotPresent"
```
rthallisey: The vars are going to be removed, so need to add and `ImagePullPolicy`

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -135,6 +161,12 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
 		NetworkAddonsImagePullPolicy = instance.Spec.NetworkAddonsImagePullPolicy
 	}
 
+	if instance.Spec.SSPImagePullPolicy != "" {
```
rthallisey: Also no longer need this if.

pkg/controller/hyperconverged/hyperconverged_controller.go
fromanirh: Not sure I understood, could you please rephrase? keeping the hunk temporarily.

pkg/controller/hyperconverged/hyperconverged_controller.go
fromanirh: Removed.

pkg/controller/hyperconverged/hyperconverged_controller.go
rthallisey: The HCO is not going to expose any component configuration right now.  We're removing it in #40.  So remove `SSPImagePullPolicy` and any code that uses it.

fromanirh: ack.

fabiand:
@tiraboschi can you pick up this PR as @fromanirh will not be able to work on this during the next week,

PR #38, opened 2019-04-18
allow customization of network addons sriov type

phoracek:
On D/S, we want to keep sriov plugin with `cnv-` prefix, to make it
clear it is a part of CNV distribution.


phoracek:
@rthallisey @lukas-bednar @myakove

phoracek:
@lukas-bednar correct

lukas-bednar:
LGTM

hack/defaults
```diff
@@ -59,6 +60,7 @@ function network_addons_sed {
     sed -i "s|value: .*sriov-device-plugin.*|value: ${NETWORK_ADDONS_SRIOV_DP_IMAGE}|g" ${TEMP_DIR}/operator.yaml
     sed -i "s|value: .*sriov-cni.*|value: ${NETWORK_ADDONS_SRIOV_CNI_IMAGE}|g" ${TEMP_DIR}/operator.yaml
     sed -i "s|value: .*mac-controller.*|value: ${NETWORK_ADDONS_KUBEMACPOOL_IMAGE}|g" ${TEMP_DIR}/operator.yaml
+    sed -i "s|value: sriov$|value: ${NETWORK_ADDONS_SRIOV_NETWORK_TYPE}|g" ${TEMP_DIR}/operator.yaml
```
rthallisey: @phoracek I'm assuming you'll use the downstream value when you ship the network-operator?

phoracek: Yes, as Lukáš commented https://github.com/kubevirt/hyperconverged-cluster-operator/pull/38#pullrequestreview-228211842

rthallisey: ok then `deploy/converged/operator.yaml` should change since it's downstream only
```
        - name: SRIOV_NETWORK_TYPE
          value: sriov
```

phoracek: Yes, we will need to add flags to manifests generator to cover all network addons images and this new variable.

rthallisey: created an issue: #39

PR #40, opened 2019-04-18
Remove HCO configuration

lveyde:


deploy/converged/crds/hco.cr.yaml
```diff
@@ -4,5 +4,4 @@ kind: HyperConverged
 metadata:
   name: hyperconverged-cluster
 spec:
-  CDIImagePullPolicy: IfNotPresent
-  KubeVirtImagePullPolicy: IfNotPresent
+  Version: 2.0
```
rthallisey: No version.  Version will be linked to the image version.

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -257,10 +227,9 @@ func newNetworkAddonsForCR(cr *hcov1alpha1.HyperConverged) *networkaddons.Networ
 			Labels: labels,
 		},
 		Spec: networkaddons.NetworkAddonsConfigSpec{
-			Multus:          &networkaddons.Multus{},
-			LinuxBridge:     &networkaddons.LinuxBridge{},
-			KubeMacPool:     &networkaddons.KubeMacPool{},
-			ImagePullPolicy: v1.PullPolicy(NetworkAddonsImagePullPolicy),
+			Multus:      &networkaddons.Multus{},
```
rthallisey: Are these required?  Preferably, we should have nothing.  If they are required, @phoracek can you add `omitempty` and have the network-operator default?

phoracek: They are required. Our operator gives users a way to deploy additional components based on their needs. Addons operator won't deploy anything if the CR is empty. We cannot set them enabled by default because of otherwise there will be no way to disable them. Why is it a problem?

rthallisey: Makes sense.  We'll leave it then

deploy/standard/olm-catalog/kubevirt-hyperconverged/0.0.1/kubevirt-hyperconverged-operator.v0.0.1.clusterserviceversion.yaml
```diff
@@ -18,8 +18,8 @@ metadata:
           "kind": "HyperConverged",
           "metadata": {
             "name": "kubevirt-hyperconverged",
-            "namespace": "kubevirt-hyperconverged"
```
rthallisey: This change should be in `templates/olm-catalog/kubevirt-hyperconverged/VERSION/kubevirt-hyperconverged-operator.VERSION.clusterserviceversion.yaml.in` otherwise it will get overwritten when templates get regenerated.

PR #41, opened 2019-04-18
Use cnv-sriov for converged operator manifest

### LINKED ISSUE #39: change sriov-network-type to d/s value in converged
`deploy/converged/operator.yaml` should change to it's downstream value singe `converged` is downstream only.
```
        - name: SRIOV_NETWORK_TYPE
          value: sriov
```

booxter:
The manifest is for CNV, and we should always use cnv- prefix for sriov
plugin there. Since there is no need to override the type for regular
deployments, general ability to override it is removed from
hack/defaults.

Fixes #39

### DIFF
--- a/hack/defaults
+++ b/hack/defaults
@@ -28,7 +28,6 @@ NETWORK_ADDONS_LINUX_BRIDGE_CNI_IMAGE="${NETWORK_ADDONS_LINUX_BRIDGE_CNI_IMAGE:-
 NETWORK_ADDONS_SRIOV_DP_IMAGE="${NETWORK_ADDONS_SRIOV_DP_IMAGE:-quay.io/booxter/sriov-device-plugin:latest}"
 NETWORK_ADDONS_SRIOV_CNI_IMAGE="${NETWORK_ADDONS_SRIOV_CNI_IMAGE:-docker.io/nfvpe/sriov-cni:latest}"
 NETWORK_ADDONS_KUBEMACPOOL_IMAGE="${NETWORK_ADDONS_KUBEMACPOOL_IMAGE:-quay.io/schseba/mac-controller:latest}"
-NETWORK_ADDONS_SRIOV_NETWORK_TYPE="${NETWORK_ADDONS_SRIOV_NETWORK_TYPE:-sriov}"
 
 if echo "${CDI_CONTAINER_REGISTRY}" | grep 'brew'; then
     CONTROLLER_IMAGE="virt-${CONTROLLER_IMAGE}"
@@ -60,7 +59,6 @@ function network_addons_sed {
     sed -i "s|value: .*sriov-device-plugin.*|value: ${NETWORK_ADDONS_SRIOV_DP_IMAGE}|g" ${TEMP_DIR}/operator.yaml
     sed -i "s|value: .*sriov-cni.*|value: ${NETWORK_ADDONS_SRIOV_CNI_IMAGE}|g" ${TEMP_DIR}/operator.yaml
     sed -i "s|value: .*mac-controller.*|value: ${NETWORK_ADDONS_KUBEMACPOOL_IMAGE}|g" ${TEMP_DIR}/operator.yaml
-    sed -i "s|value: sriov$|value: ${NETWORK_ADDONS_SRIOV_NETWORK_TYPE}|g" ${TEMP_DIR}/operator.yaml
 }
 
 # Deploy upstream operators
--- a/deploy/converged/operator.yaml
+++ b/deploy/converged/operator.yaml
@@ -190,7 +190,7 @@ spec:
         - name: SRIOV_NETWORK_NAME
           value: sriov-network
         - name: SRIOV_NETWORK_TYPE
-          value: sriov
+          value: cnv-sriov
         - name: KUBEMACPOOL_IMAGE
           value: quay.io/schseba/mac-controller:latest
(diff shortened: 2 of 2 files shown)


booxter:
Please keep me honest here. I *think* I am right to clean up hack/defaults but maybe I miss the goal of having it driven by a shell script.

booxter:
@phoracek

hack/defaults
```diff
@@ -28,7 +28,6 @@ NETWORK_ADDONS_LINUX_BRIDGE_CNI_IMAGE="${NETWORK_ADDONS_LINUX_BRIDGE_CNI_IMAGE:-
 NETWORK_ADDONS_SRIOV_DP_IMAGE="${NETWORK_ADDONS_SRIOV_DP_IMAGE:-quay.io/booxter/sriov-device-plugin:latest}"
 NETWORK_ADDONS_SRIOV_CNI_IMAGE="${NETWORK_ADDONS_SRIOV_CNI_IMAGE:-docker.io/nfvpe/sriov-cni:latest}"
 NETWORK_ADDONS_KUBEMACPOOL_IMAGE="${NETWORK_ADDONS_KUBEMACPOOL_IMAGE:-quay.io/schseba/mac-controller:latest}"
-NETWORK_ADDONS_SRIOV_NETWORK_TYPE="${NETWORK_ADDONS_SRIOV_NETWORK_TYPE:-sriov}"
```
phoracek: Place for discussion. My original idea was to keep this "sriov" by default, but make it possible to modify it (to "cnv-sriov") for D/S testing.

Do we want to make it "sriov" on U/S or "cnv-sriov"? I would vote for the former, since that would keep it consistent with network addons operator. @booxter

phoracek:
@booxter commented in the code. AFAIK the shell script is a temporary simple way to deploy all components together (we should converge to converged HCO, containing all sub-operators). With the shell script you can test operators on u/s with their default images and on d/s, overriding image and version variables.

hack/defaults
booxter: @phoracek if we keep the custom type set in manifest for d/s and no override for u/s then what's the reason to have a way to override it using the `defaults` script?

phoracek: @booxter d/s does not contain any manifests anymore. The sriov plugin name is kept only in manifests. Currently, this HCO code (and code generating HCO converged manifests) is responsible for setting the name.

booxter: Perhaps I miss something.

My understanding is that `deploy/converged` is for d/s manifests and `deploy/standard` is for upstream. The `defaults` script consumes one of those and spills out the result with some substitutions applied (via `sed`). There is no need to dynamically change network type if we know which manifests we want to target. Hence there is no need to have it in `defaults`; instead we can keep corresponding values hardcoded in respective manifests. (In this case, we can leave override out for `standard` because default value works, and override with `cnv-sriov` in `converged`.)

Now, I feel like this understanding is wrong. Could you please tell me where?

booxter: And to clarify, the reason why I touched the script is because once I change the `converged` manifest to use `cnv-sriov`, the current `sed` rule is no longer working because it expects `sriov` not `cnv-sriov` in the input template. I felt that instead of making the rule to operate with the new type name I can drop it since there is no clear reason for this attribute to be dynamically set independently from respective product target (meaning, one would not need to use a type name different from `cnv-sriov` in CNV setups; as for `standard` manifest, it doesn't have the type listed right now so the `sed` rule is ineffective there anyway).

phoracek: @booxter 

> My understanding is that `deploy/converged` is for d/s manifests and `deploy/standard` is for upstream.

Not sure that is true. Converged is simply one big CSV for all components, should work on both U/S and D/S. Standard it the other "mode", where you deploy all components using their own CSVs. Not sure if it is used anywhere.

> The `defaults` script consumes one of those and spills out the result with some substitutions applied (via `sed`).

The hack folder contains set of scripts using by QE to deploy all operators at once, IIUIC it should be replaced by converged CSV. This script should work on both U/S and D/S, although it is not an official way to deploy operators on U/D, hence, it doesn't really matter if we keep cnv- prefix there ot not.

> There is no need to dynamically change network type if we know which manifests we want to target. Hence there is no need to have it in `defaults`; instead we can keep corresponding values hardcoded in respective manifests.

You might be right here. I was trying to keep it so we have strictly split U/S and D/S. But since these scripts are in practice more or less for D/S only, it is not that big of a deal.

> And to clarify, the reason why I touched the script is because once I change the converged manifest to use cnv-sriov, the current sed rule is no longer working because it expects sriov not cnv-sriov in the input template.

Maybe I'm getting lost too. Converged manifests and these scripts (with `sed`) are not related. `sed` is applied on released scripts from respective operator's repo. Converged manifests are generated using vendored operators' components.

> I felt that instead of making the rule to operate with the new type name I can drop it since there is no clear reason for this attribute to be dynamically set independently from respective product target (meaning, one would not need to use a type name different from cnv-sriov in CNV setups; as for standard manifest, it doesn't have the type listed right now so the sed rule is ineffective there anyway).

We could keep cnv-sriov only in these temporary scripts. However, one day, this code will converge into something consumable by users on both U/S and D/S, maybe it would be worth it distinguishing between the two from the very beginning.

Your call, we can keep cnv-sriov only and if needed, change it later.

rthallisey:
This is going to get picked up by the upstream/downstream automation work.  Closing.

PR #43, opened 2019-04-22
Remove output bin

lveyde:


deploy/standard/olm-catalog/kubevirt-hyperconverged/0.0.1/kubevirt-hyperconverged-operator.v0.0.1.clusterserviceversion.yaml
```diff
@@ -18,8 +18,8 @@ metadata:
           "kind": "HyperConverged",
           "metadata": {
             "name": "kubevirt-hyperconverged",
-            "namespace": "kubevirt-hyperconverged"
-          }
+            "namespace": "kubevirt-hyperconverged",
+          },
```
tiraboschi: AFAIK any comma on the last element will cause a validation error

djzager:
We need to do this but it has to wait for now.

PR #44, opened 2019-04-22
bump network addons operator version to 0.5.0

phoracek:
Should fix kubemacpool on u/s.


phoracek:
@rthallisey @djzager @SchSeba could you please take a look?

PR #37, opened 2019-04-17
Make operator-courier happy

tiraboschi:
Various changes to prevent
'operator-courier verify --ui_validate_io'
from complaining.

Shared as a draft for discussion.

PR #45, opened 2019-04-22
Add missing courier code to templates

rthallisey:
The code to make operator-courier verify our CSVs needs to get added
to the manifest generation code.

Follow up to https://github.com/kubevirt/hyperconverged-cluster-operator/pull/37


rthallisey:
@tiraboschi I couldn't get this to work with `operator-courier verify --ui_validate_io deploy/converged/`.  Let me know if you see anything wrong.

tiraboschi:
The first error is "Courier requires valid CSV, CRD, and Package files" and indeed deploy/converged/cluster_role_binding.yaml has "kind: ClusterRoleBinding" and so it will fail the validation.

Then deploy/converged/operator.yaml is a multi document yaml file and so it will also cause the operator courier to fail as for: https://github.com/operator-framework/operator-courier/issues/107

PR #47, opened 2019-04-24
Create KubeVirt's configuration ConfigMap

slintes:
Just took the same hardcoded values as we have in kubevirt-ansible for now, see https://github.com/kubevirt/kubevirt-ansible/blob/master/roles/kubevirt/templates/kubevirt-config.yaml

Renamed `manageComponentCR` to `manageComponentResource`, so that it fits other resources than CRs too

### DIFF
--- a/pkg/controller/hyperconverged/hyperconverged_controller.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller.go
@@ -405,13 +405,6 @@ func newKubevirtTemplateValidatorForCR(cr *hcov1alpha1.HyperConverged) *sspv1.Ku
 	}
 }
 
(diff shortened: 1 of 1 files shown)


pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -250,11 +270,26 @@ func manageComponentCR(err error, o runtime.Object, kind string, c client.Client
 	}
 
 	// Object CR already exists - don't requeue
```
djzager: nit. Update comment.

slintes: fixed , thx

PR #49, opened 2019-04-25
web-ui: Hardocde branding to okdvirt

mareklibra:


mareklibra:
Nice to see #24 merged. This is a follow-up to address related topics being discussed there or elsewhere.

Let me rise following questions:
- What is the last decision, how the operator's CR will get it's config options?
  For web-ui:
    - `KWebUIMasterDefaultSubdomain`, `KWebUIPublicMasterHostname` are taken from HCO CR - who/what will set them so the code within #24 can forwards these options to the kwebui's CR?
    - How can be branding set if needed? Per request, it's hardocded to the upstream value `okdvirt` within golang code.
    - The kwebui's operator *requires* `Version` (docker tag) and `RegistryUrl` (URL including context path) to be set *in its CR*. Is the #22 approach/implementation still the way to get these values?

Cc @djzager , @rthallisey

mareklibra:
Merged without responding to the open questions above.
Without addressing these issues, the operator/**HCO will not work**.

@rthallisey , @djzager , can you please comment?

rthallisey:
> The kwebui's operator requires Version (docker tag) and RegistryUrl (URL including context path) to be set in its CR. Is the #22 approach/implementation still the way to get these values?
>
The CR variables are tied to the operator that launched them.  Meaning, the `Version` and `RegistryURL` on the CR should be provided by your component operator as a default.

>How can be branding set if needed? Per request, it's hardocded to the upstream value okdvirt within golang code.
>
Does branding need to be exposed as a param?  Or is it that it's different from upstream to downstream?

If you need it to be configurable, it would be the same as:  https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/pkg/controller/hyperconverged/hyperconverged_controller.go#L422-L423

> What is the last decision, how the operator's CR will get it's config options? For web-ui:
- KWebUIMasterDefaultSubdomain, KWebUIPublicMasterHostname are taken from HCO CR - who/what will set them so the code within #24 can forwards these options to the kwebui's CR?
>
Admin will set them when they create the HCO CR.  We will need to document the presence of these vars and when they should be used.

cc @jelkosz

PR #53, opened 2019-04-25
Install common templates to the openshift namespace

MarSik:
The common templates need to be in openshift namespace so all users can read them.


MarSik:
@rthallisey Can you please check if this is all that is needed to change the namespace?

PR #56, opened 2019-04-29
Align NMO to version v0.2.0

yanirq:
Align node maintenance operator image link to the latest official version v0.2.0


yanirq:
@rthallisey please note that also templates were aligned with the version. 
in the longer term we will need device a way to generate the converged csv also from nmo and ssp without these hard coded values (unless the method will change in the near future)

rthallisey:
@yanirq thanks.

I agree, we'll need template generate code in each SSP & NMO to do that.

PR #57, opened 2019-04-29
Fix nmo format

rthallisey:
Follow up to https://github.com/kubevirt/hyperconverged-cluster-operator/pull/52


yanirq:
@rthallisey what is the change here actually ?  only indentation ?

rthallisey:
The indentation of nmo was wrong in the converged CSV creating an invalid yaml.

PR #72, opened 2019-05-03
Create crs in the correct namespace based on scope

rthallisey:
Cluster scoped CRDs should be using `Namesapce: ""` and namespace scoped CRDs should be given a ns.  

Also update vendor dir.


rthallisey:
Closing in favor of https://github.com/kubevirt/hyperconverged-cluster-operator/pull/73

PR #73, opened 2019-05-03
Instead of GETing resources, always attempt to CREATE them

rthallisey:
During our reconcile loop, we Get a resource we're watching, check to see if it already exists, then create it if it doesn't exists.  Instead, let's always create the resource and catch IfAlreadyExists errors.


pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -209,10 +202,8 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
 		return reconcile.Result{}, err
```
djzager: You should probably only requeue when the error is something other than it already exists.

djzager: Now I see that `manageComponentResource` already does that. :+1:

PR #80, opened 2019-05-08
Update Kubevirt to v0.17.0

slintes:


deploy/converged/olm-catalog/kubevirt-hyperconverged/0.0.1/kubevirt-hyperconverged-operator.v0.0.1.clusterserviceversion.yaml
```diff
@@ -613,6 +611,12 @@ spec:
               - get
               - list
               - watch
+            - apiGroups:
```
tiraboschi: Wasn't the one one line 485 enough?

slintes: the RBAC rules of the KubeVirt operator are a concatenation of what's needed by every single virt component (apiserver, controller, handler, and the operator itself). It's not worth the effort to merge this together, because k8s is fine with it. So there might be some duplicate rules indeed.

PR #87, opened 2019-05-12
Add the option to build and test locally

irosenzw:
Add recipies to the makefile in order to build HCO
and test it using a local cluster.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,9 @@ QUAY_USERNAME ?=
 QUAY_PASSWORD ?=
 SOURCE_DIRS   = cmd pkg
 SOURCES       := $(shell find . -name '*.go' -not -path "*/vendor/*")
+IMAGE_REGISTRY ?= docker.io
+IMAGE_TAG ?= latest
+OPERATOR_IMAGE ?= rthallisey/hyperconverged-cluster-operator
 
 build: $(SOURCES) ## Build binary from source
 	go build -i -ldflags="-s -w" -o _out/hyperconverged-cluster-operator ./cmd/manager > /dev/null
@@ -9,12 +12,33 @@ build: $(SOURCES) ## Build binary from source
 clean: ## Clean up the working environment
 	@rm -rf _out/
 
-# TODO: maybe we don't need make targets for stuff already in shell scripts?
-hack-start: ## Run .hack/deploy.sh
+start:
 	./hack/deploy.sh
 
 hack-clean: ## Run ./hack/clean.sh
 	./hack/clean.sh
+
+docker-build: docker-build-operator
+
+docker-build-operator:
+	docker build -f build/Dockerfile -t $(IMAGE_REGISTRY)/$(OPERATOR_IMAGE):$(IMAGE_TAG) .
+
+docker-push: docker-push-operator
+
+docker-push-operator:
+	docker push $(IMAGE_REGISTRY)/$(OPERATOR_IMAGE):$(IMAGE_TAG)
+
+cluster-up:
+	./cluster/up.sh
+
+cluster-down:
+	./cluster/down.sh
+
+cluster-sync:
+	./cluster/sync.sh
+
+cluster-clean:
+	CMD="./cluster/kubectl.sh" ./hack/clean.sh
 
 stageRegistry:
 	@REGISTRY_NAMESPACE=redhat-operators-stage ./hack/quay-registry.sh $(QUAY_USERNAME) $(QUAY_PASSWORD)
@@ -28,4 +52,19 @@ help: ## Show this help screen
 		awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-20s\033[0m %s\n", $$1, $$2}'
 	@echo ''
 
-.PHONY: build clean hack-start hack-clean
+
+.PHONY: start \
+		clean \
+		build \
+		help \
+		hack-clean \
+		docker-build \
+		docker-build-operator \
+		docker-push \
+		docker-push-operator \
+		cluster-up \
+		cluster-down \
+		cluster-sync \
+		cluster-clean \
+		stageRegistry
+
--- a/README.md
+++ b/README.md
@@ -121,3 +121,32 @@ Create an HCO CustomResource, which creates the KubeVirt CR, launching KubeVirt.
 ```bash
 kubectl create -f deploy/converged/crds/hco.cr.yaml
 ```
+
+## Launching the HCO on a local cluster
+
+Lunch the HCO locally for testing, experimenting and developing.
+
+**NOTE:** no need to install any type of kubenetes cluster as a prerequisite.
+
+1. Navigate to the project's directory
+```bash
+$ cd <path>/hyperconverged-cluster-opertor
+```
+2. Remove an old cluster
+```bash
+$ make cluster-down
+```
+3. Create a new cluster
+```bash
+$ make cluster-up
+```
+4. Clean previous HCO deployment and re-deploy HCO \
+   (When making a change, execute only this command - no need to repeat steps 1-3)
+```bash
+$ make cluster-sync
+```
+### Command-Line Tool
+Use `./cluster/kubectl.sh` as the command-line tool. for example:
+```bash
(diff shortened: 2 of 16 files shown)


irosenzw:
@rthallisey @tiraboschi @djzager @arachmani please review

cluster/sync.sh
```diff
@@ -0,0 +1,26 @@
...
+IMAGE_REGISTRY=$registry make docker-build-operator docker-push-operator
+
+for i in $(seq 1 ${CLUSTER_NUM_NODES}); do
+    ./cluster/cli.sh ssh "node$(printf "%02d" ${i})" 'sudo docker pull rthallisey/hyperconverged-cluster-operator'
+    # Temporary until image is updated with provisioner that sets this field
+    # This field is required by buildah tool
+    ./cluster/cli.sh ssh "node$(printf "%02d" ${i})" 'sudo sysctl -w user.max_user_namespaces=1024'
+done
+
+CMD="./cluster/kubectl.sh" ./hack/deploy.sh
```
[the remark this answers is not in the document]
irosenzw: done

README.md
```diff
@@ -120,3 +120,32 @@ Create an HCO CustomResource, which creates the KubeVirt CR, launching KubeVirt.
...
+```
+$ make cluster-up
+```
+4. Clean previous HCO deployment and re-deploy HCO \
+   (When making a change, execute only this command - no need to repeat steps 1-3)
+```
+$ make cluster-sync
+```
+### Command-Line Tool
+Use `./cluster/kubectl.sh` as the command-line tool. fo example:
```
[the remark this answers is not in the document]
irosenzw: done

Makefile
```diff
@@ -1,12 +1,38 @@
...
+IMAGE_TAG ?= latest
+OPERATOR_IMAGE ?= rthallisey/hyperconverged-cluster-operator
+
 start:
 	./hack/deploy.sh
 
 clean:
 	./hack/clean.sh
 
+docker-build: docker-build-operator
```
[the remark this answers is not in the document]
irosenzw: done

Makefile
```diff
@@ -1,13 +1,49 @@
...
+	./cluster/sync.sh
+
+cluster-clean:
+	CMD="./cluster/kubectl.sh" ./hack/clean.sh
+
 stageRegistry:
 	@REGISTRY_NAMESPACE=redhat-operators-stage ./hack/quay-registry.sh $(QUAY_USERNAME) $(QUAY_PASSWORD)
 
-.PHONY: start clean
+.PHONY: start \
```
djzager: This is a nit, I don't care if you do anything about it, but the readability of the `.PHONY` isn't really important...maybe just put it all on one line?

build/Dockerfile
```diff
@@ -1,15 +1,18 @@
-FROM registry.access.redhat.com/ubi7-dev-preview/ubi-minimal:7.6
+FROM golang:1.10 AS builder
```
djzager: :tada:

cluster/cli.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash -e
```
djzager: I really don't want for us to expand into another top-level directory. Can we put this in `hack/`?

djzager: Same question for everything else in `cluster/`.

cluster/sync.sh
```diff
@@ -0,0 +1,26 @@
...
+# Cleanup previously generated manifests
+#rm -rf _out/
+
+# Copy release manifests as a base for generated ones, this should make it possible to upgrade
+#cp -r deploy/converged _out/
+
+# Sed from docker.io to registry:5000
+#sed -i 's/image: docker\.io\/rthallisey\/hyperconverged-cluster-operator:latest/image: registry:5000\/rthallisey\/hyperconverged-cluster-operator:latest/g' _out/operator.yaml
+
+make cluster-clean
```
djzager: I think if we have make targets running shell scripts it should be a one-way interaction. Either:

- `cluster-clean` should be a dependency on the `cluster-sync` target
- `cluster/sync.sh` calls the appropriate function provided by `cluster/clean.sh` (keeping in mind that I think these should move into the `hack/` directory.

build/Dockerfile
```diff
@@ -1,15 +1,18 @@
...
+ENV GOPATH=/go
+WORKDIR /go/src/github.com/kubevirt/hyperconverged-cluster-operator/
+COPY . .
+RUN CGO_ENABLED=0 GOOS=linux go build -o /hyperconverged-cluster-operator github.com/kubevirt/hyperconverged-cluster-operator/cmd/manager
 
+FROM centos:centos7
```
djzager: Maybe we should be using the UBI here though now that it's released?

tiraboschi: I think that all the other kubevirt subprojects are going to move to ubi8

Makefile
irosenzw: The convention is to separate each target in a different line.
Other projects in CNV does that too.

cluster/cli.sh
irosenzw: This is because those scripts are placed also in the `cluster/` directory on the other CNV projects (kubevirt, CNA, CDI ....) and I think it's important the directory tree will be the same (more or less) between all of these projects, so it will be easier for others to contribute.

build/Dockerfile
irosenzw: Good thought... I will look into it

irosenzw:
@djzager @tiraboschi @rthallisey please review

deploy/standard/operator.yaml
```diff
@@ -31,7 +31,7 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
-        image: rthallisey/hyperconverged-cluster-operator:latest
+        image: docker.io/rthallisey/hyperconverged-cluster-operator:latest
```
djzager: You can't just change this here. You'll either need to make changes [here](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/pkg/components/components.go#L16) **or, probably easier,** update the [manifest-templator](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/tools/manifest-templator/manifest-templator.go#L149). After you update the templator (or components.go) you'll need to re-run `./hack/build-manifests.sh`

build/Dockerfile
```diff
@@ -1,15 +1,18 @@
-FROM registry.access.redhat.com/ubi7-dev-preview/ubi-minimal:7.6
+FROM golang:1.11 AS builder
+ENV GOPATH=/go
+WORKDIR /go/src/github.com/kubevirt/hyperconverged-cluster-operator/
+COPY . .
+RUN CGO_ENABLED=0 GOOS=linux go build -o /hyperconverged-cluster-operator github.com/kubevirt/hyperconverged-cluster-operator/cmd/manager
 
+FROM registry.access.redhat.com/ubi8/ubi
```
djzager: Is there a reason you didn't use the `ubi-minimal`?

build/Dockerfile
```diff
@@ -1,15 +1,18 @@
-FROM registry.access.redhat.com/ubi7-dev-preview/ubi-minimal:7.6
+FROM golang:1.11 AS builder
+ENV GOPATH=/go
+WORKDIR /go/src/github.com/kubevirt/hyperconverged-cluster-operator/
+COPY . .
+RUN CGO_ENABLED=0 GOOS=linux go build -o /hyperconverged-cluster-operator github.com/kubevirt/hyperconverged-cluster-operator/cmd/manager
```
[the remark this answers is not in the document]
tiraboschi: On my opinion a multistage build inside the container is a good idea; we can evaluate the idea of removing the capability to build the binary outside the container.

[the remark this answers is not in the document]
djzager: What I mean is that he can `RUN make build` instead of `RUN go build ...`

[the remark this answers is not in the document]
tiraboschi: Yes, right, sorry!

irosenzw:
@djzager @tiraboschi @rthallisey please review

PR #89, opened 2019-05-13
Enable CPU feature discovery for kubevirt

MarSik:
This patch enables the CPUNodeDiscovery feature gate that enables cpu model and feature based scheduling in kubevirt.


MarSik:
@fromanirh @rthallisey @ksimon1 Can you please review this?

fabiand:
@tiraboschi @rthallisey we need this, fyi

MarSik:
The feature gate name is correct according to https://github.com/kubevirt/kubevirt/blob/63c038a4490a44562703d915388c55038ea284ef/pkg/virt-config/feature-gates.go#L38

PR #94, opened 2019-05-15
SSP: don't specify component versions

fromanirh:
Just request the sub-components presence and let the operator
decide which version should be pulled.

### DIFF
--- a/pkg/controller/hyperconverged/hyperconverged_controller.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller.go
@@ -357,9 +357,6 @@ func newKubevirtCommonTemplateBundleForCR(cr *hcov1alpha1.HyperConverged) *sspv1
 			Labels:    labels,
 			Namespace: "openshift",
 		},
-		Spec: sspv1.VersionSpec{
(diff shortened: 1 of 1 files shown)


fromanirh:
SSP-operator (v1.0.1) can handle this, testing

djzager:
cc @tiraboschi

fromanirh:
please not we will need to refresh the kubevirt-ssp-operator too

djzager:
@fromanirh we are waiting on https://github.com/MarSik/kubevirt-ssp-operator/pull/39 ?

fromanirh:
> @fromanirh we are waiting on [MarSik/kubevirt-ssp-operator#39](https://github.com/MarSik/kubevirt-ssp-operator/pull/39) ?

No, not needed for the moment. We verified that the operator does the right thing if the SSP CRs do _not_ have the version specified, like this PR does. So this PR should be everything we need.

djzager:
> No, not needed for the moment. We verified that the operator does the right thing if the SSP CRs do _not_ have the version specified, like this PR does. So this PR should be everything we need.

Awesome. Thank you.

PR #95, opened 2019-05-15
Add basic ci

djzager:
It'd be nice if we, at a minimum, didn't accept PRs that wouldn't build.


djzager:
This PR is a follow-up to #94 . Subsequent PRs will remove the binary from the tree and use multi-stage builds.

PR #99, opened 2019-05-21
Add unit tests for creating component operators' CRs

rwsu:
Added test target to Makefile.
Also modified travis config to invoke test suite.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ services:
 
 script:
 - make build
+- make test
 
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -4,11 +4,18 @@ go:
 sudo: required
 dist: trusty
 
+install:
+    - go get -v github.com/onsi/ginkgo/ginkgo
```
djzager: til about BDD testing in go. Nice.

PR #100, opened 2019-05-22
Add a basic functional test

arachmani:
- Deploy k8s
- Build deploy and test the HCO deployment

### DIFF
--- a/automation/test.sh
+++ b/automation/test.sh
@@ -1,0 +1,7 @@
+#!/bin/bash -xe
+
+make cluster-down
+make cluster-up
+make cluster-sync
+make functest
+make docker-push
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,9 @@ cluster-sync:
 cluster-clean:
 	CMD="./cluster/kubectl.sh" ./hack/clean.sh
 
+functest:
+	./hack/functest.sh
+
 stageRegistry:
 	@REGISTRY_NAMESPACE=redhat-operators-stage ./hack/quay-registry.sh $(QUAY_USERNAME) $(QUAY_PASSWORD)
 
@@ -70,4 +73,5 @@ test: test-unit
 		cluster-down \
 		cluster-sync \
 		cluster-clean \
-		stageRegistry
+		stageRegistry \
+		functest
--- a/hack/functest.sh
+++ b/hack/functest.sh
@@ -1,0 +1,41 @@
+#!/bin/bash
+
+source hack/common.sh
+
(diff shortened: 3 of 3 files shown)


cluster/test.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash -xe
+
+make cluster-down
```
djzager: Most of these targets don't currently exist. Also, I may be able to be convinced otherwise but I think humans and travis ci are the only things off-hand that should be calling `make`, scripts should include/execute other scripts instead.

cluster/functest.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
```
djzager: In kubevirt/kubevirt the functests go in the hack dir. Should this go there?

djzager: I'm also curious if bash, in the end, is going to be robust enough to do this well. I know that the operator-sdk (which this project is built using) has some ideas on how to do functional testing, maybe we should be looking into that?

cluster/test.sh
arachmani: This PR depends on PR#87, the targets exist there.

cluster/functest.sh
arachmani: I will move it to hack dir, and take a look at the functional test in the Operator-sdk, though I believe there shouldn't be any problem using bash.

arachmani:
@rthallisey @djzager @tiraboschi @irosenzw can you please review?

automation/test.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash -xe
+
+make cluster-down
```
djzager: I don't see where this script is used but I think it should be calling other shell scripts instead of `make cluster-down ...`

rthallisey: this could be plugged into the Makefile.  Maybe a new recipe that ties them all together? `make test` or something?

gbenhaim:
ci test please

automation/test.sh
arachmani: The files that will be placed under automation directory, will be executed by STDCI.
I'm not sure we need a recipe for it.

arachmani:
ci test please

ovirt-infra:

ovirt-infra:
All tests passed

rthallisey:
ci add to whitelist

ovirt-infra:
All tests passed

rthallisey:
ci is hooked up in https://github.com/kubevirt/hyperconverged-cluster-operator/pull/131

PR #104, opened 2019-05-31
Refactor HCO controller reconcile function

rwsu:
Broke up the CR creation code in reconcile into functions.
The functions are now called in parallel.

### DIFF
--- a/pkg/controller/hyperconverged/factory.go
+++ b/pkg/controller/hyperconverged/factory.go
@@ -1,0 +1,149 @@
+package hyperconverged
+
+import (
+	sspv1 "github.com/MarSik/kubevirt-ssp-operator/pkg/apis/kubevirt/v1"
+	networkaddons "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/v1alpha1"
+	networkaddonsnames "github.com/kubevirt/cluster-network-addons-operator/pkg/names"
+	hcov1alpha1 "github.com/kubevirt/hyperconverged-cluster-operator/pkg/apis/hco/v1alpha1"
+	kwebuis "github.com/kubevirt/web-ui-operator/pkg/apis/kubevirt/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	cdi "kubevirt.io/containerized-data-importer/pkg/apis/core/v1alpha1"
+	kubevirt "kubevirt.io/kubevirt/pkg/api/v1"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// The set of resources managed by the HCO
+func (r *ReconcileHyperConverged) getAllResources(cr *hcov1alpha1.HyperConverged, request reconcile.Request) []runtime.Object {
+	return []runtime.Object{
+		newKubeVirtConfigForCR(cr, request.Namespace),
+		newKubeVirtForCR(cr, request.Namespace),
+		newCDIForCR(cr, request.Namespace),
+		newNetworkAddonsForCR(cr, ""),
+		newKubevirtCommonTemplateBundleForCR(cr, ""),
+		newKubevirtNodeLabellerBundleForCR(cr, request.Namespace),
+		newKubevirtTemplateValidatorForCR(cr, request.Namespace),
+		newKWebUIForCR(cr, ""),
+	}
+}
+
+func newKubeVirtConfigForCR(cr *hcov1alpha1.HyperConverged, namespace string) *corev1.ConfigMap {
+	labels := map[string]string{
+		"app": cr.Name,
+	}
+	return &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "kubevirt-config",
+			Labels:    labels,
+			Namespace: namespace,
+		},
+		Data: map[string]string{
+			"feature-gates": "DataVolumes,SRIOV,LiveMigration,CPUManager,CPUNodeDiscovery",
+		},
+	}
+}
+
+// newKubeVirtForCR returns a KubeVirt CR
+func newKubeVirtForCR(cr *hcov1alpha1.HyperConverged, namespace string) *kubevirt.KubeVirt {
+	labels := map[string]string{
+		"app": cr.Name,
+	}
+	return &kubevirt.KubeVirt{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "kubevirt-" + cr.Name,
+			Labels:    labels,
+			Namespace: namespace,
+		},
+	}
+}
+
+// newCDIForCr returns a CDI CR
+func newCDIForCR(cr *hcov1alpha1.HyperConverged, namespace string) *cdi.CDI {
+	labels := map[string]string{
+		"app": cr.Name,
+	}
+	return &cdi.CDI{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "cdi-" + cr.Name,
+			Labels:    labels,
+			Namespace: namespace,
+		},
+	}
+}
+
+// newNetworkAddonsForCR returns a NetworkAddonsConfig CR
+func newNetworkAddonsForCR(cr *hcov1alpha1.HyperConverged, namespace string) *networkaddons.NetworkAddonsConfig {
+	labels := map[string]string{
+		"app": cr.Name,
+	}
+	return &networkaddons.NetworkAddonsConfig{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      networkaddonsnames.OPERATOR_CONFIG,
+			Labels:    labels,
+			Namespace: namespace,
+		},
+		Spec: networkaddons.NetworkAddonsConfigSpec{
+			Multus:      &networkaddons.Multus{},
+			LinuxBridge: &networkaddons.LinuxBridge{},
+			KubeMacPool: &networkaddons.KubeMacPool{},
+		},
+	}
+}
+
+func newKubevirtCommonTemplateBundleForCR(cr *hcov1alpha1.HyperConverged, namespace string) *sspv1.KubevirtCommonTemplatesBundle {
+	labels := map[string]string{
+		"app": cr.Name,
+	}
+	return &sspv1.KubevirtCommonTemplatesBundle{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "common-templates-" + cr.Name,
+			Labels:    labels,
+			Namespace: "openshift",
+		},
+	}
+}
+
+func newKubevirtNodeLabellerBundleForCR(cr *hcov1alpha1.HyperConverged, namespace string) *sspv1.KubevirtNodeLabellerBundle {
+	labels := map[string]string{
+		"app": cr.Name,
+	}
+	return &sspv1.KubevirtNodeLabellerBundle{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "node-labeller-" + cr.Name,
+			Labels:    labels,
+			Namespace: namespace,
+		},
+	}
+}
+
+func newKubevirtTemplateValidatorForCR(cr *hcov1alpha1.HyperConverged, namespace string) *sspv1.KubevirtTemplateValidator {
+	labels := map[string]string{
+		"app": cr.Name,
+	}
+	return &sspv1.KubevirtTemplateValidator{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "template-validator-" + cr.Name,
+			Labels:    labels,
+			Namespace: namespace,
+		},
+	}
+}
+
+func newKWebUIForCR(cr *hcov1alpha1.HyperConverged, namespace string) *kwebuis.KWebUI {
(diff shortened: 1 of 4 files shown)


rwsu:
I would like some feedback on this refactoring. Whether parallelism is desired. There is still a lot of duplicate code. I have not found a common interface/type between the objects that are returned by the newX functions. Having one would simplify the code quite a bit.

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -302,6 +225,25 @@ func newKubeVirtConfigForCR(cr *hcov1alpha1.HyperConverged) *corev1.ConfigMap {
 	}
 }
 
+func createKubeVirtConfigCRAndSetOwnership(instance *hcov1alpha1.HyperConverged, request reconcile.Request, r *ReconcileHyperConverged, c chan CreateCRResult) {
+	kvConfig := newKubeVirtConfigForCR(instance)
```
djzager: is it possible to simply have one function, like `createCrAndSetOwnership` that (in addition to what I see here) takes the name (ie. `KubeVirtConfig`) and the CR generator function (ie. `newKubeVirtConfigForCR`)?

rwsu: Let me look into this again. I had tried before, but didn't find a compatible type. Looking at the add function simplification you submitted, I wonder if we can also use runtime.Object here.

rwsu:
Reworked the code to follow some of the patterns that I saw being used in the CDI operator. 
Added tests to verify that the resources being created by HCO are actually present after a reconcile.

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -107,261 +109,91 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
...
+		wg.Wait()
+		close(c)
+	}()
+
+	var result reconcile.Result
+	var crErr error
+	var hasError bool
+	// Loop through the results, until an error is found. Once an error is found, don't change result or crErr, but
+	// drain the results channel. If no errors are found, return the last result.
+	for r := range c {
```
rthallisey: Will `c` always have all the results from the channel when this is executed?  I'm not sure it will because `wg.Wait()` is run in a thread.

rwsu: Yes, c will always have all the results. wg.Wait() doesn't run until the WaitGroup is drained. That occurs when the last deferred wg.Done() is called when the last goroutine in the loop above completes.

To be honest, I think making this parallel may be overkill at this point. I don't know if we gain that much or at all by making it parallel. It was difficult to assess the difference in speed when I observed the CRs being created before and after this refactor. This may be premature optimization. 

Would you be ok if I rip the parallelism out and but leave the rest in place?

rthallisey: I'm not as concerned with speed as I am with resiliency.  As long as we don't continuously fail on the first error we see, it's a big upgrade.  Also, 👍  to more tests.

rthallisey: Makes sense to drop the parallelism.  We can come back to it later.

rwsu: Ok removed.

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -119,79 +118,58 @@ type ResultFromCRCreate struct {
...
+		if err != nil {
+			if !errors.IsNotFound(err) {
+				results = append(results, ResultFromCRCreate{result: reconcile.Result{}, err: err})
+			}
+
+			if err = controllerutil.SetControllerReference(cr, desiredMetaObj, r.scheme); err != nil {
+				results = append(results, ResultFromCRCreate{result: reconcile.Result{}, err: err})
+			}
+
+			// TODO: common-templates and cdi fails the Get check above and appears to still be missing.
```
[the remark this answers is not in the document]
lveyde: I noticed in the past that the code that queries for CDI fails, unless an empty string ("") is used for the namespace. It may be related.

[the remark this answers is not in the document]
rwsu: I like the flow of check and if not exist, create. It would be good to figure out why the GET requests aren't working correctly. @rthallisey, do you know where they are globally scoped? 

thanks @lveyde, I checked and setting namespace to "" for CDI fixes it. Is empty string ok as the namespace for CDI?

Need to figure out what is going on with common-templates.

ovirt-infra:

pkg/controller/hyperconverged/hyperconverged_controller.go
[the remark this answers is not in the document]
rthallisey: Ya an empty string is how you're suppose to reference a globally scoped CRD.  We'll need keep a list of those and have different check for them. 

>@rthallisey, do you know where they are globally scoped?
>
Not sure I understand the question.

The empty string didn't also GET the common-templates?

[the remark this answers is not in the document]
rwsu: @rthallisey let me rephrase, how do you know if something is globally scoped?

The empty string doesn't work for common-templates.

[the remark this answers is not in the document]
djzager: If the something you are referring to is a CRD. Then you can look at the `spec.scope` of the CRD.

For example, [cluster network addons](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/deploy/converged/crds/cna.crd.yaml#L13) or [kubevirt](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/deploy/converged/crds/kubevirt.crd.yaml#L24)

pkg/controller/hyperconverged/factory.go
```diff
@@ -0,0 +1,149 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	cdi "kubevirt.io/containerized-data-importer/pkg/apis/core/v1alpha1"
+	kubevirt "kubevirt.io/kubevirt/pkg/api/v1"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// The set of resources managed by the HCO
+func (r *ReconcileHyperConverged) getAllResources(cr *hcov1alpha1.HyperConverged, request reconcile.Request) []runtime.Object {
+	return []runtime.Object{
+		newKubeVirtConfigForCR(cr, request.Namespace),
```
djzager: do we really need for these to be function calls when we can just list them here?

```
func () getAllResources(...) []runtime.Object {
	labels := map[string]string{
		"app": cr.Name,
	}
	return []runtime.Object{
		&corev1.ConfigMap{
				Name:      "kubevirt-config",
				Labels:    labels,
				Namespace: namespace,
			},
			Data: map[string]string{
				"feature-gates": "DataVolumes,SRIOV,LiveMigration,CPUManager,CPUNodeDiscovery",
			},
		},
	       ...
	}
}
```

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -121,325 +121,72 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
...
-	}
+func (r *ReconcileHyperConverged) reconcileUpdate(logger logr.Logger, cr *hcov1alpha1.HyperConverged, request reconcile.Request) (reconcile.Result, error) {
+	results := []ResultFromCRCreate{}
 
-	labels := resource.GetLabels()
-	if app, labelExists := labels["app"]; !labelExists || app != cr.Name {
-		log.Info("Existing resource wasn't deployed by HCO, ignoring", "Kind", resource.GetObjectKind())
-		return reconcile.Result{}, nil
-	}
+	resources := r.getAllResources(cr, request)
```
djzager: To be honest, I find [this portion of the knicluster_controller reconcile](https://github.com/mhrivnak/kni-operator/blob/master/pkg/controller/knicluster/knicluster_controller.go#L272-L281) to be similar to what you are trying to accomplish here while being somewhat easier to follow and reason about.

Maybe we can update the `newKubevirtTemplateValidatorForCR` type functions to instead ensure that a given resource is as it should be. Then in the reconcile we just loop through those functions.

pkg/controller/hyperconverged/factory.go
djzager: This contradicts my recommendation below :smile: 

I think having a collection of functions that individually reconcile their respective resources to the state they should be in is the correct way to go.

pkg/controller/hyperconverged/hyperconverged_controller.go
djzager: Another advantage of this approach is that if it gets more complex than "create if it doesn't exist" then that complexity is in the function where that logic belongs.

rwsu: @djzager I submitted an update. It should look more like the knicluster_controller now.

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -121,325 +121,115 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
...
+func (r *ReconcileHyperConverged) ensureKWebUI(instance *hcov1alpha1.HyperConverged, logger logr.Logger, request reconcile.Request) error {
+	kWebUI := newKWebUIForCR(instance, request.Namespace)
+	return r.ensureResourceExists(instance, logger, request, kWebUI)
 }
 
-func newKWebUIForCR(cr *hcov1alpha1.HyperConverged) *kwebuis.KWebUI {
-	labels := map[string]string{
-		"app": cr.Name,
+func (r *ReconcileHyperConverged) ensureResourceExists(instance *hcov1alpha1.HyperConverged, logger logr.Logger, request reconcile.Request, desiredRuntimeObj runtime.Object) error {
+	desiredMetaObj := desiredRuntimeObj.(metav1.Object)
```
rthallisey: I'm not familiar with what you're doing in `desiredRuntimeObj.(metav1.Object)`.   Specially, how to the parenthesis around `metav1.Object` effect this call?

pkg/controller/hyperconverged/factory.go
```diff
@@ -0,0 +1,149 @@
...
+// newCDIForCr returns a CDI CR
+func newCDIForCR(cr *hcov1alpha1.HyperConverged, namespace string) *cdi.CDI {
+	labels := map[string]string{
+		"app": cr.Name,
+	}
+	return &cdi.CDI{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "cdi-" + cr.Name,
+			Labels:    labels,
+			Namespace: namespace,
```
rthallisey: Shouldn't this always be `""`?  Line 154 calls `cdi := newCDIForCR(instance, request.Namespace)`, which may not always be correct.  Maybe we should create some `const`s with NS values and spread them around?

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -121,325 +121,115 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
...
+		if !errors.IsNotFound(err) {
+			return err
+		}
+
+		// set the reference for the object that is about to be created
+		if err = controllerutil.SetControllerReference(instance, desiredMetaObj, r.scheme); err != nil {
+			return err
+		}
+
+		// TODO: common-templates fails the Get check above and appears to still be missing.
```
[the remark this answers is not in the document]
rwsu: Appears so.

The CR is in the openshift ns:

````
[rwsu@localhost converged]$ kubectl get kubevirtcommontemplatesbundle -n openshift
NAME                                      AGE
common-templates-hyperconverged-cluster   4m51s
````

request.Namespace is kubevirt-hyperconverged, but key.namespace is openshift. This may be why the Get is not able to find it.

What is interesting is KWebUI which doesn't specify a namespace in the CR can be queried/Get using any namespace.

File: pkg/controller/hyperconverged/hyperconverged_controller.go (hunk 6 above)
Comment (rwsu) [reply pkg/controller/hyperconverged/hyperconverged_controller.go]: The underlying object implements multiple interfaces. It comes in as runtime.Object. It is "cast" as metav1.Object here so that the Name and Namespace functions are accessible.

File: pkg/controller/hyperconverged/factory.go (hunk 7 above)
Comment (rwsu) [reply pkg/controller/hyperconverged/factory.go]: Good catch. Yes, constants are a good idea. Added.

**Comment by rthallisey** [comment]
ci test please

**Comment by rthallisey** [comment]
ci add to whitelist

File: pkg/controller/hyperconverged/hyperconverged_controller.go (hunk 8 above)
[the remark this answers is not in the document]
Comment (rthallisey) [reply pkg/controller/hyperconverged/hyperconverged_controller.go]: `Kwebui`, `CDI`, and `NetworkAddons` are all cluster scoped, so that makes sense.
```
...
  names:
    kind: KWebUI
    listKind: KWebUIList
    plural: kwebuis
    singular: kwebui
  scope: Cluster
  version: v1alpha1
```

Do you think this `TODO` is no longer needed?

rthallisey:
@rwsu I think this is ready to go.  Just one last nit

rthallisey:
nice job @rwsu

PR #105, opened 2019-05-31
Simplify add function in hco controller

djzager:
Iterate over the slice of runtime.Objects we want to watch and simply
create the watch.

### DIFF
--- a/pkg/controller/hyperconverged/hyperconverged_controller.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller.go
@@ -52,58 +52,23 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


djzager:
cc @rwsu

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -52,61 +52,26 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+			IsController: true,
+			OwnerType:    &hcov1alpha1.HyperConverged{},
+		})
+		if err != nil {
+			return err
+		}
 	}
 
-	return nil
+	return err
```
rthallisey: I think your logic is right in that `err` will always be `nil` here, but I think it reads better as `nil`.  WDYT?

rwsu:
Nice code clean up. lgtm.

PR #106, opened 2019-06-02
bump CDI version to 1.9.2

annastopel:
Bumps CDI vendored code to version 1.9.2

### DIFF
--- a/deploy/converged/olm-catalog/kubevirt-hyperconverged/0.0.1/kubevirt-hyperconverged-operator.v0.0.1.clusterserviceversion.yaml
+++ b/deploy/converged/olm-catalog/kubevirt-hyperconverged/0.0.1/kubevirt-hyperconverged-operator.v0.0.1.clusterserviceversion.yaml
@@ -78,7 +78,7 @@ metadata:
           "spec": {
             "nodeName": "nodename",
             "reason": "node maintenance reason"
-          }          
+          }
         }
       ]
     description: Creates and maintains a HyperConverged KubeVirt Deployment
@@ -1184,11 +1184,13 @@ spec:
             selector:
               matchLabels:
                 name: cdi-operator
+                operator.cdi.kubevirt.io: ""
             strategy: {}
             template:
               metadata:
                 labels:
                   name: cdi-operator
+                  operator.cdi.kubevirt.io: ""
               spec:
                 containers:
                 - env:
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,421 +2,342 @@
 
 
 [[projects]]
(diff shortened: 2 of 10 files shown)


deploy/converged/olm-catalog/kubevirt-hyperconverged/0.0.1/kubevirt-hyperconverged-operator.v0.0.1.clusterserviceversion.yaml
```diff
@@ -1160,7 +1160,7 @@ spec:
                   - hyperconverged-cluster-operator
                   env:
                   - name: OPERATOR_IMAGE
-                    value: rthallisey/hyperconverged-cluster-operator:latest
+                    value: docker.io/hyperconverged-cluster-operator:latest
```
djzager: Something isn't right here with this image value.

ohadlevy: its right with not using @rthallisey private one ;-)

djzager:
[This is where](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/43ccbf50d4ba261b85a285ac99ed985fcbc7b9c1/pkg/components/components.go#L16-L18) the strange behavior comes from. 

I don't see anything in your PR that would cause this but the manifests need to be correct before merged. Maybe, you could:

1.  [modify this code](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/43ccbf50d4ba261b85a285ac99ed985fcbc7b9c1/pkg/components/components.go#L16-L18). This would make it so, from the manifest-templator you would have to be explicit where the image is coming from (which I think is probably better in the long run).
1. Update [the manifest-templator](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/43ccbf50d4ba261b85a285ac99ed985fcbc7b9c1/tools/manifest-templator/manifest-templator.go#L148-L152) to explicitly specify `docker.io/rthallisey` for the repository string. Once we start pushing these images to the kubevirt org...then we can update the manifests (and manifest-templator).

djzager:
#111 fixes the issue. If you rebase and regenerate the manifests I believe the issue will have gone away.

rthallisey:
@annastopel Does this need a backport to release-2.0?

annastopel:
Yes, it fixes labels on cdi pods to cdi.kubevirt.io when deployed via olm

rthallisey:
@annastopel can you file the backport?  Will review when the PR is up.

PR #103, opened 2019-05-30
bump cna to 0.9.0

phoracek:

PR #108, opened 2019-06-03
Align with master

tiraboschi:
Align with master.

https://github.com/kubevirt/hyperconverged-cluster-operator/pull/103 introduced NMSTATE_STATE_HANDLER_IMAGE env variable for the cluster-network-addons-operator.
nmstate state handler is a 2.1 only feature so we have definitively to start working with two branches.


rthallisey:
align release-2.0 with HEAD-1

PR #112, opened 2019-06-05
Fix cluster-sync script after moving to quay.io

irosenzw:


irosenzw:
Changing the image repo to quay.io broke the testing env.

PR #114, opened 2019-06-05
Create CSV 0.0.2 to start upgrade testing 

rthallisey:
Create a new CSV that we can use to start testing upgrades.  Master changes should go into the 0.0.2 CSV file, which will now be used by default for all HCO changes.


deploy/converged/olm-catalog/kubevirt-hyperconverged/0.0.2/kubevirt-hyperconverged-operator.v0.0.2.clusterserviceversion.yaml
```diff
@@ -0,0 +1,1451 @@
...
+                containers:
+                - command:
+                  - virt-operator
+                  - --port
+                  - "8443"
+                  - -v
+                  - "2"
+                  env:
+                  - name: OPERATOR_IMAGE
+                    value: kubevirt/virt-operator:latest
```
rwsu: Should the kubevirt image be on latest? I wonder if we should be using actual version numbers. CSV v0.0.1 also has OPERATOR_IMAGE set to kubevirt/virt-operator:latest. So an upgrade is attempted from v0.0.1 to v0.0.2, we may not see an upgrade occur for the virt-operator.

I have the same concern for other images listed using latest below.

rthallisey: it shouldn't be set to latest, but I'm not sure what it should be set to. We aren't locked with kubevirt's release cycle so it's really up to us what we want to set it to for testing.  Regardless of versions, we should still be able to test the upgrade procedure through OLM.

This exposes a larger problem though that we'll need to sort out for CI...

rthallisey: After testing a few days ago, I saw OLM upgrade the component operators.  What's odd is that with latest, all the pods restarted when I don't think any new containers were pulled.  It could be that OLM assumes that a new CSV means every container must be updated regardless?

rthallisey:
This was added in a different PR

PR #118, opened 2019-06-10
[WIP] Add phases 

irosenzw:
[WIP] Add phases that represent HCO status


pkg/apis/hco/v1alpha1/hyperconverged_types.go
```diff
@@ -14,17 +14,27 @@ type HyperConvergedSpec struct {
...
 	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
 	KWebUIMasterDefaultSubdomain string `json:"KWebUIMasterDefaultSubdomain,omitempty"` // optional, required if openshift-console project is missing
-	KWebUIPublicMasterHostname   string `json:"KWebUIPublicMasterHostname,omitempty"` // optional, required if openshift-console project is missing
+	KWebUIPublicMasterHostname   string `json:"KWebUIPublicMasterHostname,omitempty"`   // optional, required if openshift-console project is missing
 }
 
+// HyperConvergedPhase is a label for the phase of a HCO deployment at the current time.
+// ---
+// +k8s:openapi-gen=true
+type HyperConvergedPhase string
```
rthallisey: I'm not sure we need `+k8s:openapi-gen=true` because openapi-gen will catch this string in `type HyperConvergedStatus struct`.

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -113,6 +116,7 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
 
 	// Set HyperConverged instance as the owner and controller
 	if err := controllerutil.SetControllerReference(instance, kvConfig, r.scheme); err != nil {
+		instance.Status.Phase = hcov1alpha1.FailedPhase
```
djzager: You'll have to write the status back with a `r.client.Status().Update(context.TODO(), instance)` if you want more than the in-memory version of the instance to have the Phase.

rthallisey:
Didn't realize phase was a deprecated pattern.  We should still be able to report in a similar way except using `Reason` instead of `Phase`.

irosenzw:
@rthallisey @djzager @tiraboschi please review.
This patch is _almost_ ready. it has few things that I need to fix, but in the overall it is working.

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -106,13 +125,19 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
...
+		r.updatePhase(instance, hcov1alpha1.ReconcilingPhase, request, "")
+	}
 
 	// Define KubeVirt's configuration ConfigMap first
 	kvConfig := newKubeVirtConfigForCR(instance)
 	kvConfig.ObjectMeta.Namespace = request.Namespace
 
 	// Set HyperConverged instance as the owner and controller
 	if err := controllerutil.SetControllerReference(instance, kvConfig, r.scheme); err != nil {
+		r.updatePhase(instance, hcov1alpha1.FailedPhase, request, "Set HyperConverged instance as the owner and controller")
```
rthallisey: Instead of having to add phase on each err, we could create a shim in front of the `Reconcile` func which captures the return from `Reconcile` and processes it.

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -106,13 +125,19 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
 		// Error reading the object - requeue the request.
 		return reconcile.Result{}, err
 	}
+	if string(instance.Status.Phase) == "" {
+		r.updatePhase(instance, hcov1alpha1.InstallingPhase, request, "")
+	} else {
+		r.updatePhase(instance, hcov1alpha1.ReconcilingPhase, request, "")
```
rthallisey: Doesn't look like you're processing `err` from the `updatePhase` call.

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -76,6 +77,24 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 
 var _ reconcile.Reconciler = &ReconcileHyperConverged{}
 
+func (r *ReconcileHyperConverged) updatePhase(instance *hcov1alpha1.HyperConverged, phase hcov1alpha1.HyperConvergedPhase, request reconcile.Request, failedOperation string) error {
+	err := r.client.Get(context.TODO(), request.NamespacedName, instance)
+	if err != nil {
+		return err
+	}
+	instance.Status.Phase = phase
+	err = r.client.Status().Update(context.TODO(), instance)
```
phoracek: Does this retrigger reconciliation? Maybe you need to add a predicate to reconciliation, to ignore status updates https://github.com/kubevirt/cluster-network-addons-operator/blob/master/pkg/controller/networkaddonsconfig/networkaddonsconfig_controller.go#L107

djzager: My understanding is that using the [status subresource on the CRD](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/deploy/converged/crds/hco.crd.yaml#L23-L24) makes it so we can independently update the status w/o triggering a reconcile.

phoracek: I see, right!

phoracek:
We can expose the condition when listing HyperConverged object. Add `additionalPrinterColumns` to HyperConverged CRD.

```yaml
  additionalPrinterColumns:
  - JSONPath: .status.conditions[?(@.type=="Ready")].status
    name: Ready
    type: string
  - JSONPath: .status.conditions[?(@.type=="Ready")].message
    name: Status
    type: string
    priority: 1
```

Should work.

(Or if you stick to deprecated Phase):

```yaml
    additionalPrinterColumns:
    - JSONPath: .status.phase
        name: Phase
        type: string
```

ovirt-infra:
All tests passed

PR #122, opened 2019-06-12
use finalizers for NetworkAddonsConfig cleanup

phoracek:
Kubernetes garbage collector doesn't support setting namespaced
resource as an owner reference of a cluster-wide resource. Due to
that, we have to perform cleanup of NetworkAddonsConfig via
a finalizer.

### DIFF
--- a/pkg/controller/hyperconverged/hyperconverged_controller.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller.go
@@ -2,6 +2,7 @@ package hyperconverged
 
 import (
 	"context"
+	"time"
 
 	sspv1 "github.com/MarSik/kubevirt-ssp-operator/pkg/apis/kubevirt/v1"
 	networkaddons "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/v1alpha1"
@@ -11,10 +12,13 @@ import (
 	cdi "kubevirt.io/containerized-data-importer/pkg/apis/core/v1alpha1"
 	kubevirt "kubevirt.io/kubevirt/pkg/api/v1"
 
+	"encoding/json"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
@@ -26,6 +30,16 @@ import (
 )
 
 var log = logf.Log.WithName("controller_hyperconverged")
+
+const (
+	// We cannot set owner reference of cluster-wide resources to namespaced HyperConverged object. Therefore,
+	// use finalizers to manage the cleanup.
+	FinalizerName = "hyperconvergeds.hco.kubevirt.io"
+
+	// Foreground deletion finalizer is blocking removal of HyperConverged until explicitly dropped.
+	// TODO: Research whether there is a better way.
+	foregroundDeletionFinalizer = "foregroundDeletion"
+)
 
 // Add creates a new HyperConverged Controller and adds it to the Manager. The Manager will set fields on the Controller
 // and Start it when the Manager is Started.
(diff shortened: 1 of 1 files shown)


phoracek:
/cc @djzager @rthallisey 

Please review, it is urgent as a fix for https://bugzilla.redhat.com/show_bug.cgi?id=1712851. Thanks

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -107,6 +118,27 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
 		return reconcile.Result{}, err
 	}
 
+	// Check if the instance was marked to be deleted, and if it is, perform cleanup
+	deleted := instance.GetDeletionTimestamp() != nil
```
djzager: This block looks similar to what I find in the [user-guide](https://github.com/operator-framework/operator-sdk/blob/master/doc/user-guide.md#handle-cleanup-on-deletion) but for some reason this is hard for me to read. Can we simplify?

```
func Reconcile(...) (...) {
        ...
        if instance.ObjectMeta.DeletionTimestamp.IsZero() {
               // add the finalizer if it's not there                                                          
                if !contains(instance.ObjectMeta.Finalizers, FinalizerName) {                                  
                        instance.ObjectMeta.Finalizers = append(instance.ObjectMeta.Finalizers, FinalizerName)       
                        return reconcile.Result{}, r.client.Update(context.TODO(), instance)                         
                } 
        } else {
                if contains(instance.ObjectMeta.Finalizers, FinalizerName) {                                   
                        // do the cleanup
                        // which can be it's own function if you wish, but for now only really
                        // removes the network-addons-config if certain conditions are met
                        ...
         
                        // remove the finalizer                            
                        instance.ObjectMeta.Finalizers = drop(instance.ObjectMeta.Finalizers, FinalizerName) 
                        return reconcile.Result{}, r.client.Update(context.TODO(), instance) 
                }                                         
        }
        ....
}
```

djzager: I still think that we should only be worrying about an `hyperconvergeds.hco.kubevirt.io` finalizer...not sure what requires the `foregroundDeletionFinalizer`.

djzager: I believe this would be easier to follow/maintain.

phoracek: If we don't remove foregroundDeletion, nobody will, it stucks. I can't say I understand where it comes from though.

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -117,11 +149,11 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
 	}
 
 	// Create the KubeVirt ConfigMap if it doesn't already exist
-	result, err := manageComponentResource(kvConfig, "KubeVirtConfig", r.client)
+	err = manageComponentResource(kvConfig, "KubeVirtConfig", r.client)
```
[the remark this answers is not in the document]
phoracek: I added a code to the end of Reconcile function. Then this result was not really needed anymore. Also, I prefer not to expose edge API (Reconciliation result here, HTTP return codes in other projects) to anyone but the top level function. The function either fails or passes, but it should not decide whether we requeue, or not, or we do, but only after some time.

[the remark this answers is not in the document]
phoracek: (Will add it in a separate PR, was lazy)

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -230,26 +257,82 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
...
 	// KWebUI failed to create, requeue
 	if err != nil {
-		return result, err
+		return reconcile.Result{}, err
 	}
-	return result, nil
+
+	// Everything went fine, automatically reconcile after after a minute without observed activity to
+	// make sure that even deployed objects without owner reference will be re-created if removed.
+	return reconcile.Result{RequeueAfter: time.Minute}, nil
```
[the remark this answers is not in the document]
phoracek: Understood and agreed.

pkg/controller/hyperconverged/hyperconverged_controller.go
phoracek: Was trying to be granular with those, for case where we successfully clean-up only half of our CRs. However, that is duplicate of does-CR-still-exist check. I will change it to use a single finalizer.

phoracek:
Hello @igoihman, your operator's resource is cluster-wide too, isn't it? Not sure how comes you don't hit the same issue as we do (maybe you do, but you don't know yet). Please check this PR and https://bugzilla.redhat.com/show_bug.cgi?id=1712851.

In short, if you set owner reference of cluster-wide (e.g. CDI) resource to namespaced (e.g. HyperConverged), bad stuff happens.

ovirt-infra:
All tests passed

phoracek:
@djzager your comments should be addressed. Could you please take another look?

ovirt-infra:
All tests passed

igoihman:
@phoracek we don't experience this issue, you can check cdi operator log.

igoihman:
@phoracek I think it's a matter of missing operator RBAC permissions

phoracek:
@igoihman yes, for us only network addons operator was failing on OKD with HCO. CDI was fine. Just wanted you to be aware of it if it ever happens.

Interesting, not sure how RBAC would help, but I can't say I understand Kubernetes garbage collector. Lemme check, thanks.

igoihman:
Thanks @phoracek . keep me in the loop if you find something.

phoracek:
@djzager I tested patched HCO, if recreated NetworkAddonsConfig when needed and it removed it during the cleanup.

PR #124, opened 2019-06-13
fix: no need for nodemaintenance namespace

### LINKED ISSUE #123: deployment script creates the node-maintenance-operator
Since 666b7fe3a9c65687156784d1ba0a5dde4bbb3db0, the node maintenance operator uses the HCO's namespace, but the script still creates the *node-maintenance-operator* one, which AFAIK, it's not required.

tripledes:
fixes #123


ovirt-infra:
All tests passed

PR #127, opened 2019-06-13
use finalizers for NetworkAddonsConfig cleanup

phoracek:
Kubernetes garbage collector doesn't support setting namespaced
resource as an owner reference of a cluster-wide resource. Due to
that, we have to perform cleanup of NetworkAddonsConfig via
a finalizer.


ovirt-infra:
All tests passed

PR #147, opened 2019-06-28
NMO was using wrong api group

rthallisey:
kubevirt-hyperconverged also shouldn't have a comma.


rthallisey:
ci test please

yanirq:
lgtm 
@rthallisey  will this handle the alm-example to be shown properly as well ?

rthallisey:
@yanirq yes.  Looks like OLM noticed NMO was using the wrong API group so it ignored the spec changes.

PR #154, opened 2019-07-03
Fix template files.

lveyde:
Due to the macro processing issue, an empty line was inserted as the very first line of CRB files.
This was causing issues with tools that use strict checking of the YAML file syntax.
Specifically it was causing the build using the latest origin-operator-registry image to fail.
This patch fixes the templates.


lveyde:
OK, replaced by another set of patches.

Please see PR#155.

lveyde:
Per request added re-based set of patches to this PR as well.

rthallisey:
ci test please

djzager:
Which PR is winning here, this or #155 ?

rthallisey:
I think #155 can be closed.

PR #157, opened 2019-07-03
Align the upstream CSV with d/s and expand description

rthallisey:
Let move this CSV closer to downstream format.  Also, expand the description so we can eventually merge the converged CSV to the _community_ operatorhub directory.  Kubevirt is in _upstream-community_ so we won't have a conflict of API ownership.

### DIFF
--- a/tools/manifest-templator/manifest-templator.go
+++ b/tools/manifest-templator/manifest-templator.go
@@ -27,6 +27,7 @@ import (
 	"os"
 	"strings"
 	"text/template"
+	"time"
 
 	"github.com/ghodss/yaml"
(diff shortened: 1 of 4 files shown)


tools/manifest-templator/manifest-templator.go
```diff
@@ -475,6 +477,7 @@ func main() {
 		SSP:      &operatorData{Tag: *sspTag},
 		NMO:      &operatorData{Tag: *nmoTag},
 	}
+	data.CreatedAt = time.Now().String()
```
djzager: :+1:

PR #163, opened 2019-07-05
Standarize conditions across component operators

rthallisey:
Conditions are how component operators will report status to the HCO.  We want to pass that status on to the user and use a readiness probe to inform OLM of sensitive lifecycle operations.

Conditions will be reported on the component operator CRs.

### DIFF
--- a/docs/conditions.md
+++ b/docs/conditions.md
@@ -8,15 +8,18 @@ Conditions are..
 Kubernetes conditions [documentation](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status).
 
 The HCO’s CR is a representation of the all the underlying component operators'
-state.  If the the HCO's CR exists, then all component CRs exist, and all
-applications exist.  If the object doesn’t exist, all component CRs don’t exist,
-and all applications don't exist.  However, the CR existence doesn’t help us with the
-state where the operators exist, but are they healthy?  This is where conditions
-on the HCO’s CR and component operator CRs will answer this question by
-reflecting the last known condition of the underlying applications.
+state.  In theory, if the HCO's CR exists, then all component CRs _should_
+exist, and all applications _should_ exist.  If the object doesn’t exist, then
+all component CRs _should not_ exist, and all applications _should not_ exist.
+However, the CR existence can only can tell us if the application should exist and
+doesn't help us observe the application's health. This is where the HCO and
+component operators will use conditions on their CRs to reflect the health of
+the underlying application.  Component operators store conditions that are
+watched by the HCO and the HCO will store conditions that reflect the
+[worst state](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/docs/conditions.md#hco-conditions) of all component operator conditions.
 
 ## Condition Struct
-We can use some of the CVO's [conditions](https://github.com/openshift/api/blob/master/config/v1/types_cluster_operator.go#L121-L133) to standardize across components.
+We can use some of the CVO's [conditions](https://github.com/openshift/api/blob/b1bcdbc/config/v1/types_cluster_operator.go#L123-L134) to standardize across components.
 
 Here's how the Condition struct will look...
 
--- a/docs/conditions.md
+++ b/docs/conditions.md
@@ -7,12 +7,13 @@ Conditions are..
 
 Kubernetes conditions [documentation](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status).
 
-The HCO’s CR is a representation of the underlying component operators.  If the
-object exists, then all components exist.  If the object doesn’t exist, all
-components don’t exist.  However, the CR existence doesn’t help us with the
+The HCO’s CR is a representation of the all the underlying component operators'
+state.  If the the HCO's CR exists, then all component CRs exist, and all
+applications exist.  If the object doesn’t exist, all component CRs don’t exist,
+and all applications don't exist.  However, the CR existence doesn’t help us with the
 state where the operators exist, but are they healthy?  This is where conditions
-on the HCO’s CR will answer this question by providing the observed health of
-the underlying components.
+on the HCO’s CR and component operator CRs will answer this question by
+reflecting the last known condition of the underlying applications.
 
 ## Condition Struct
 We can use some of the CVO's [conditions](https://github.com/openshift/api/blob/master/config/v1/types_cluster_operator.go#L121-L133) to standardize across components.
@@ -20,8 +21,9 @@ We can use some of the CVO's [conditions](https://github.com/openshift/api/blob/
 Here's how the Condition struct will look...
 
 ```go
-type OperatorStatusCondition struct {
-   // type specifies the state of the operator's reconciliation functionality.
+type ApplicationStatusCondition struct {
+   // type specifies the state of the operator's reconciliation functionality,
+   // which reflects the state of the application
    Type ConditionType `json:"type"`
 
    // status of the condition, one either True or False.
@@ -40,18 +42,18 @@ type OperatorStatusCondition struct {
 ```
 
 ## ConditionType
-`ConditionType` _specifies the state of the operator's reconciliation functionality_.
-`ConditionType`s use `ConditionStatus` to report state.  The `ConditionStatus`es
-we will use are either `True` or `False`.  The `ConditionStatus` object can also
-be `Unknown`, but only the HCO will use `Unknown` because it's not clear what
-`Unknown` means in terms of an application's lifecycle.  The HCO can assume
-`Unknown` for conditions while operators are starting up.
+`ConditionType` _specifies the state of the operator's reconciliation functionality,
+which reflects the state of the application_. `ConditionType`s use `ConditionStatus`
+to report state.  The `ConditionStatus`es we will use are either `True` or `False`.
+The `ConditionStatus` object can also be `Unknown`, but only the HCO will use
+`Unknown` because it's not clear what `Unknown` means in terms of an application's
+lifecycle.  The HCO can assume `Unknown` for conditions, while operators are starting up.
 
-#### OperatorAvailable
+#### ApplicationAvailable
 ```
-	OperatorAvailable ClusterStatusConditionType = "Available"
+	ApplicationAvailable ClusterStatusConditionType = "Available"
 ```
-OperatorAvailable indicates that the binary maintained by the operator
+ApplicationAvailable indicates that the binary maintained by the operator
 (eg: openshift-apiserver for the openshift-apiserver-operator), is functional
 and available in the cluster.
 
@@ -63,11 +65,11 @@ Progressing indicates that the operator is actively making changes to the binary
 maintained by the operator (eg: openshift-apiserver for the
 openshift-apiserver-operator).
 
-#### OperatorDegraded
+#### ApplicationDegraded
 ```
-	OperatorDegraded ClusterStatusConditionType = "Degraded"
+	ApplicationDegraded ClusterStatusConditionType = "Degraded"
 ```
-Degraded indicates that the component operator is not functioning completely.
+Degraded indicates that the application is not functioning completely.
 An example of a degraded state would be if there should be 5 copies of the
 component running but only 4 are running. It may still be available, but it is
 degraded.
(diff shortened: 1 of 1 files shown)


rthallisey:
cc @mhrivnak @jarrpa

docs/conditions.md
```diff
@@ -0,0 +1,73 @@
...
+Conditions are..
+	   _the latest available observations of an object's state. They are
+	   an extension mechanism intended to be used when the details of an
+	   observation are not a priori known or would not apply to all
+	   instances of a given Kind._
+
+The HCO’s CR is a representation of the underlying component operators.  If the
+object exists, then all components exist.  If the object doesn’t exist, all
+components don’t exist.  However, the CR existence doesn’t help us with the
+state where the operators exist, but are they healthy?  This is where conditions
```
djzager: I believe it's worth pointing out the `Reason` and `Message` (and maybe the whole Condition struct) to make it clear that we should be giving a reason and message for transitions into (and out of) the different ConditionTypes.

Copying the [CVO types](https://github.com/openshift/api/blob/master/config/v1/types_cluster_operator.go):

```
type ClusterOperatorStatusCondition struct {
	// type specifies the state of the operator's reconciliation functionality.
	Type ClusterStatusConditionType `json:"type"`

	// status of the condition, one of True, False, Unknown.
	Status ConditionStatus `json:"status"`

	// lastTransitionTime is the time of the last update to the current status object.
	LastTransitionTime metav1.Time `json:"lastTransitionTime"`

	// reason is the reason for the condition's last transition.  Reasons are CamelCase
	Reason string `json:"reason,omitempty"`

	// message provides additional information about the current condition.
	// This is only to be consumed by humans.
	Message string `json:"message,omitempty"`
}
```

docs/conditions.md
```diff
@@ -0,0 +1,73 @@
+# CR Conditions and Readiness Probe
+Conditions are..
```
rmohr: They are also backward compatible. State machine changes (e.g. via status) are not.

docs/conditions.md
rmohr: :+1: Let's explicitly point out that the conditions are the same condition type like in k8s (and CVO).

rmohr: Oh, and let's maybe also standardize some of the `Reson` strings: e.g. `Healing` or `Updating`. I guess we can find some in the okd repo.

djzager: Just keep in mind that, assuming we follow the pattern set by CVO, there would be at least 3 ConditionTypes `Progressing|Available|Degraded` that can be `True|False|Unkown`. It may be difficult to standardize on the reasons for all of those but I would be interested in the proposal.

rthallisey: I only talked about below `True and False` and left out `Unknown` because I wasn't sure how to write the `Meaning` section.

For instance, if you're `OperatorAvailable = true` `OperatorProgressing = unknown` and `OperatorDegraded = false`, what definitive meaning can we gather?  Does `Unknown` imply the operator doesn't know what's going on?  If that's the case then `Unknown` could be the "yellow warning light" to call for help?   Maybe that can be useful during failed upgrades...

djzager: My comment about "Just keep in mind" was more directed at the standardizing on `Reason` string.

I don't have an immediate understanding of a scenario where the `Progressing` Condition would be `Unknown`...maybe when we first create the conditions slice? But I agree this proposal should focus on what `True|False` mean.

docs/conditions.md
```diff
@@ -0,0 +1,73 @@
...
+#### OperatorDegraded
+```
+	OperatorDegraded ClusterStatusConditionType = "Degraded"
+```
+Degraded indicates that the component operator is not functioning completely.
+An example of a degraded state would be if there should be 5 copies of the
+component running but only 4 are running. It may still be available, but it is
+degraded.
+
+#### Condition Matrix
```
phoracek: Maybe mention what is the default value of (missing) condition.

docs/conditions.md
rmohr: > Just keep in mind that, assuming we follow the pattern set by CVO, there would be at least 3 ConditionTypes `Progressing|Available|Degraded` that can be `True|False|Unkown`. It may be difficult to standardize on the reasons for all of those but I would be interested in the proposal.

Yes you are right. Let me rephrase, maybe some suggestions would be good. For instance k8s uses `FailedCreate`, `FailedUpdate`, to indicate REST issues. But I guess it is more important to have the conditions standardized.

docs/conditions.md
rthallisey: What do you mean by 
> default value of (missing) condition
>
Are you asking what will happen if the HCO notices a CR hasn't implemented conditions or a particular condition?

phoracek: Every CR is initially without any state - conditions. If HCO sees this initial state, how does it evaluates condition value? True, False or Unknown.

rthallisey: If the HCO sees the CR with no value for the conditions, then the HCO will assume the Operator's condition is `OperatorAvailable = false`, `OperatorProgressing = unknown` and `OperatorDegraded = unknown`.  This is a good place for `unknown`

docs/conditions.md
```diff
@@ -0,0 +1,126 @@
...
+See this [issue](https://github.com/operator-framework/operator-lifecycle-manager/issues/922) for why we only want to report a readiness probe on the HCO
+instead of on all component operators.
+
+## Reason
+`Reason` is _a one-word CamelCase reason for the condition's last transition_.
+
+We'll use a series of lifecycle inspired prefixes paired with postfixes to
+standardize values for `Reason`.
+
+|         | -Failed  | -Succeeded | -Invalid | -InProgress |
```
djzager: Table isn't rendering, making this hard to review.

slintes:
Just crossed my mind: how do we deal with operators running in HA mode, so multiple instances? How should HCO find the leader? Maybe with another condition?

docs/conditions.md
```diff
@@ -0,0 +1,150 @@
...
+| :------------- |:-------------:|:-----:|:-----:|:-----:|
+| Install- | InstallFailed | InstallSucceeded | InstallInvalid | InstallInProgress |
```
djzager: It is not clear to me which Reason you would use when transitioning to `Available: true`. I'm not really sure what standardizing on the `Reason` field really buys us at this point. IMO prescribing that operators populate the Reason is sufficient.

docs/conditions.md
```diff
@@ -0,0 +1,150 @@
...
+`Message` is a _human-readable message indicating details about last transition_.
+
+Explain why your CR has `Reason`.
+
+
+## HCO conditions
+It's important to point out that the `ConditionTypes` on the HCO don't represent
+the condition the HCO is in, rather the condition of the component CRs.
+
+The HCO will use the same `ConditionType`s and `Reason`s, but it will be the
```
djzager: Again, the scenarios in which an operator could be `Degraded` and `Unavailable` can't be pinned down to one Reason. I think we should skip Reason for now.

docs/conditions.md
```diff
@@ -0,0 +1,150 @@
...
+
+The HCO's `ConditionType`s will always represent the _worst_ `Status` and the
+_most recently viewed_ `Reason`.
+
+The _worst_ `Status` for each `ConditionType`:
+
+| Condition   | Status  |
+| :------------- |:-------------:|
+| OperatorAvailable | False |
+| OperatorProgressing | False |
```
djzager: I think that if any operator is Progressing...then the HCO is progressing.

docs/conditions.md
rmohr: Works for me. Let's drop reasons for now.

rthallisey:
Going to change the names of the conditions to replace `Operator` with `Application` to make it more clear what the condition refers to.   Does anyone object?

docs/conditions.md
```diff
@@ -0,0 +1,152 @@
...
+Conditions are..
+	   _the latest available observations of an object's state. They are
+	   an extension mechanism intended to be used when the details of an
+	   observation are not a priori known or would not apply to all
+	   instances of a given Kind._
+
+Kubernetes conditions [documentation](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status).
+
+The HCO’s CR is a representation of the all the underlying component operators'
+state.  If the the HCO's CR exists, then all component CRs exist, and all
```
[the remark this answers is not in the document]
rthallisey: That's true, but I think this is a good example.  I'll change the language around to be more accurate.

docs/conditions.md
rthallisey: Do you think this should be renamed on the HCO?

rthallisey:
@all I want to wrap up the design by tomorrow, Wednesday.  If you have any additional comments or if anything is unclear, please speak up and we can work to resolve it.

docs/conditions.md
djzager: This was more to point out that when you say "worst" for a given condition type...that doens't really work for the OperatorProgressing (or ApplicationProgressing). Since, what we want to communicate, is that if any of the components are progressing, then the HCO is also progressing.

This wasn't really about renaming anything. Just that the "worst" for progressing should be "True"

rthallisey: I agree.  If Progressing is shown on the HCO as false and one operator has "True", it should be shown as "True" or that will be confusing.

"Worst" doesn't apply uniformly for all the conditions because the worst case is that we're failed and we're not doing anything about it.  OperatorProgressing is really an outlier here.

docs/conditions.md
```diff
@@ -44,6 +44,13 @@ type ApplicationStatusCondition struct {
 }
 ```
 
+## Library
+We're going to use a shared library to provide the condition types to operators.
+This will ensure the code is not specific to any operator and it will allow
+products outside of CNV to also consume it.
+
+https://github.com/openshift-kni/operator-status
```
mhrivnak: @markmc WDYT? We talked to operator-sdk about owning this, but they don't want to commit to it until they get past some other priorities. All operators in the CNV and OCS ecosystems will need to share these conditions, so openshift-kni seems like a reasonable org to be their home for now.

markmc: I see openshift-kni as downstream of openshift, kubevirt, etc. - suggest doing this under github.com/openshift

docs/conditions.md
```diff
@@ -0,0 +1,150 @@
...
+state.  In theory, if the HCO's CR exists, then all component CRs _should_
+exist, and all applications _should_ exist.  If the object doesn’t exist, then
+all component CRs _should not_ exist, and all applications _should not_ exist.
+However, the CR existence can only can tell us if the application should exist and
+doesn't help us observe the application's health. This is where the HCO and
+component operators will use conditions on their CRs to reflect the health of
+the underlying application.  Component operators store conditions that are
+watched by the HCO and the HCO will store conditions that reflect the
+[worst state](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/docs/conditions.md#hco-conditions) of all component operator conditions.
+
```
markmc: FWIW, in a drain controller design discussion this week, I saw Doug say:

> All the advice I've seen recently says that "conditions" are no longer the preferred way to express overall status, and that explicit fields should be used instead. The reason I've seen is that having a list of conditions makes it harder to make (or maybe keep) API promises than having regular fields on the CRD.

(I don't know have any further background on this line of thinking, and Doug is away this week. So again, just FWIW)

djzager: Interesting that you mention that. So the [api-conventions doc](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties) is slightly older than these issues talking about Status and Conditions:

- https://github.com/kubernetes/kubernetes/issues/50798
- https://github.com/kubernetes/kubernetes/issues/7856

There do seem to be some reasonable arguments on both sides of this discussion. At the same time, this is exactly what the Cluster Version Operator is doing. Without having a convincing argument to not use Conditions or clearly established alternative, I still think Conditions are the way to move forward.

rthallisey: This might be the issue where it was discussed https://github.com/kubernetes-sigs/cluster-api/pull/1081 ?

rthallisey: I think what's outlined in this doc follows the "outlook" approach written [here](https://github.com/kubernetes/kubernetes/issues/7856#issuecomment-99667941).  Since this topic is long running discussion, I'm in favor of implementing this approach and following whatever the resolution of https://github.com/kubernetes/kubernetes/issues/7856 is at a later date.

PR #179, opened 2019-07-19
drop crs from deploy.sh

phoracek:
CRs should be created by HCO, not via the deploy script.

### DIFF
--- a/hack/deploy.sh
+++ b/hack/deploy.sh
@@ -46,12 +46,10 @@ fi
 # Create cluster-network-addons-operator
 "${CMD}" create -f "${CNA_URL_PREFIX}"/network-addons-config.crd.yaml
 "${CMD}" create -f "${CNA_URL_PREFIX}"/operator.yaml
-"${CMD}" create -f "${CNA_URL_PREFIX}"/network-addons-config-example.cr.yaml
 
 # Create ssp-operator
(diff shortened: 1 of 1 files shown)


phoracek:
@rthallisey @djzager please take a look

phoracek:
@rthallisey maybe there is a misunderstanding on my side. How I understand it is, that we install operators (KubeVirt, cdi, cna, ...) by creating their respective deployment, CRD, RBAC etc. Them HCO is installed, HyperConverged object created and HCO controller creates KubeVirt, NetworkAddonsConfiguration and all the other CRs. in reality however, we create some Cars as a part of this script, that deploys network components before HCO is even started.

This is not a matter of OLM, the script is doing something (creating Cars) that should be done by running HCO.

rthallisey:
@phoracek I read the title a little to quickly, so I thought it said 'crds'.  Can you rebase?

phoracek:
@rthallisey @djzager what is the reason to have download and create CRs in `hack/operator-test.sh`? CRs should be created by HCO, so it tests everything but HCO's functionality. Can we drop it (now when we test using combination of `hack/deploy.sh` and `tests/test.sh`)?

rthallisey:
@phoracek I dropped it in another PR https://github.com/kubevirt/hyperconverged-cluster-operator/pull/180.  I forgot to circle back and close this.

phoracek:
Cool, thanks :)

PR #182, opened 2019-07-24
Add missing double-quotes

irosenzw:


hack/defaults
```diff
@@ -96,9 +96,9 @@ https://github.com/kubevirt/cluster-network-addons-operator/releases/download/${
 https://github.com/kubevirt/cluster-network-addons-operator/releases/download/${NETWORK_ADDONS_MANIFEST_VERSION}/namespace.yaml
 https://github.com/kubevirt/cluster-network-addons-operator/releases/download/${NETWORK_ADDONS_MANIFEST_VERSION}/operator.yaml
 https://raw.githubusercontent.com/MarSik/kubevirt-ssp-operator/${SSP_MANIFEST_VERSION}/cluster/${SSP_MANIFEST_VERSION:1}/kubevirt-ssp-operator-crd.yaml
-https://raw.githubusercontent.com/MarSik/kubevirt-ssp-operator/${SSP_MANIFEST_VERSION}/cluster/${SSP_MANIFEST_VERSION:1}/kubevirt-ssp-operator.yaml
+https://raw.githubusercontent.com/MarSik/kubevirt-ssp-operator/${SSP_MANIFEST_VERSION}/cluster/${SSP_MANIFEST_VERSION:1}/kubevirt-ssp-operator.yaml"
```
djzager: I'm not sure I understand why we still need to be pulling these things from their releases page. We vendor and build their manifests. We have the deploy directory with all of the rbac, crds, and deployment info we need. These are all in addition to installing it via OLM.

Maybe this isn't the PR to address this...but worth pointing out.

PR #194, opened 2019-08-01
Overwrite the CatalogSourceConfig created by the OperatorSource

rthallisey:
When you create an OperatorSource, it creates a CatalogSourceConfig listing all the operators from that source.  Let's overwrite it to only use the HCO's content to avoid any packaging errors caused by other operators.


rthallisey:
this is proving to be a challenge d/s because an operator source grabs all the content from a registry.  This can cause other csv to prevent our csv from working, so we'll keep this as is for now...

PR #197, opened 2019-08-01
WIP: HCO conditions

djzager:
The purpose of this pull request is to add Conditions to the HCO's Custom Resource. In this initial pass, the path is laid to evaluate each component CR's Status.Conditions in order to consolidate the HyperConverged's status.

Out of the box, this PR simply handles the initial status, when reconcile succeeds, and when reconcile fails.


djzager:
@mhrivnak @sacharya

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -119,107 +117,139 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
...
+	})
+	conditionsv1.SetStatusCondition(conditions, conditionsv1.Condition{
+		Type:    conditionsv1.ConditionUpgradeable,
+		Status:  corev1.ConditionUnknown,
+		Reason:  reason,
+		Message: message,
+	})
 }
 
-func (r *ReconcileHyperConverged) ensureResourceExists(instance *hcov1alpha1.HyperConverged, logger logr.Logger, request reconcile.Request, desiredRuntimeObj runtime.Object) error {
```
djzager: I feel bad for having to move away from this. It was nice to not repeat over and over for all that we are managing...but I needed more granular control and access to the Status through the runtime.Object wasn't possible.

pkg/apis/hco/v1alpha1/hyperconverged_types.go
```diff
@@ -21,6 +24,11 @@ type HyperConvergedStatus struct {
 	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
 	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
 	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
+
+	// Conditions is a list of conditions related to operator reconciliation
+	Conditions []conditionsv1.Condition `json:"conditions,omitempty"  patchStrategy:"merge" patchMergeKey:"type"`
+	// RelatedObjects is a list of objects that are "interesting" or related to this operator.
```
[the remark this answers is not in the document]
djzager: When you say user-facing docs. My guess is that you mean a new file in the docs directory describing all that we are doing with respect to status updates?

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -119,107 +117,139 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
...
-		r.ensureKubeVirtCommonTemplatebundle,
+		r.ensureKubeVirtCommonTemplateBundle,
 		r.ensureKubeVirtNodeLabellerBundle,
 		r.ensureKubeVirtTemplateValidator,
 	} {
 		err = f(instance, reqLogger, request)
 		if err != nil {
+			initReconcileFailureConditions(&instance.Status.Conditions)
+			// don't want to overwrite the actual reconcile failure
+			uErr := r.client.Status().Update(context.TODO(), instance)
```
rthallisey: For a first pass context.TODO will be ok, but we should quickly turn around to fix.  Reconcile can be executed much faster than the API server can process the requests and we don't want to have any races with our state.

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -119,107 +117,139 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
...
+
+		err = r.client.Status().Update(context.TODO(), instance)
+		if err != nil {
+			reqLogger.Error(err, "Failed to add conditions to status")
+			return reconcile.Result{}, err
+		}
+	}
+
+	// Create happy conditions
+	initHappyConditions(&r.conditions)
```
rthallisey: Our state isn't always happy here.  If the previous Reconcile was a Failure, why would we set it to Happy before we have fixed the issue?

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -119,107 +117,139 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
...
 		r.ensureKubeVirtTemplateValidator,
 	} {
 		err = f(instance, reqLogger, request)
 		if err != nil {
+			initReconcileFailureConditions(&instance.Status.Conditions)
+			// don't want to overwrite the actual reconcile failure
+			uErr := r.client.Status().Update(context.TODO(), instance)
+			if uErr != nil {
+				reqLogger.Error(uErr, "Failed to update conditions")
+			}
```
rthallisey: The HCO will be stuck reconciling the first failure it sees.   Can we avoid that?

deploy/olm-catalog/kubevirt-hyperconverged/0.0.2/kubevirt-hyperconverged-operator.v0.0.2.clusterserviceversion.yaml
```diff
@@ -1161,7 +1161,7 @@ spec:
                   - name: WATCH_NAMESPACE
                     valueFrom:
                       fieldRef:
-                        fieldPath: metadata.namespace
+                        fieldPath: ""
```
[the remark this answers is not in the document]
djzager: Yeah. Unfortunately, we need to watch all namespaces in order to be able to "GET" the things we create in the openshift namespace for example.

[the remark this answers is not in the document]
djzager: Outside the scope of this PR but we should probably make sure that namespace exists before we create something in it :sunglasses:

pkg/controller/hyperconverged/hyperconverged_controller.go
djzager: Admittedly, if we are failing to make it through the `Reconcile()` function something is seriously wrong. That was why I added a new condition type for us `ReconcileComplete` to make it possible to see cases like this where we aren't even making it through the whole Reconcile loop.

rthallisey: I don't disagree with you on that point, but do we have to block?  If we do, it can have adverse affects, like not recreating a deleted component cr.  Maybe instead of `ReconcileComplete` there can be `ComponentErrored`, `ComponentMissing`, or `ComponentUknown`?

pkg/controller/hyperconverged/hyperconverged_controller.go
djzager: I know that I'm not using the `initHappyConditions()` anymore but the concept is still the same. We have an in-memory representation `r.conditions` and the persistent representation `instance.Status.Conditions`. What you'll see in my latest commit is that I always set `r.conditions = nil` and as we go through the different `ensureXYZ` they will set conditions on `r.conditions` (notably not `instance.Status.Conditions`) and the last step of `Reconcile()` is to  handle the case where `r.conditions != nil` (means something like KubeVirt is not available) or `r.conditions == nil` (means that everyone is in a happy state).

pkg/controller/hyperconverged/hyperconverged_controller.go
djzager: I don't think it's obvious but this isn't really a change in behavior. The only addition is to set the `ReconcileCondition` type to false if we encounter an error.

rthallisey: Let's revisit in another PR

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -2,30 +2,34 @@ package hyperconverged
...
-	sspv1 "github.com/MarSik/kubevirt-ssp-operator/pkg/apis/kubevirt/v1"
 	"github.com/go-logr/logr"
-	networkaddons "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/v1alpha1"
-	hcov1alpha1 "github.com/kubevirt/hyperconverged-cluster-operator/pkg/apis/hco/v1alpha1"
-	kubevirt "kubevirt.io/client-go/api/v1"
-	cdi "kubevirt.io/containerized-data-importer/pkg/apis/core/v1alpha1"
-
-	"encoding/json"
-
+	// "github.com/operator-framework/operator-sdk/pkg/ready"
```
rthallisey: you can uncomment and add the readiness probe now

deploy/crds/hco.cr.yaml
```diff
@@ -0,0 +1,6 @@
+---
+apiVersion: hco.kubevirt.io/v1alpha1
```
rthallisey: This should be removed.  I think this got picked up in a rebase.

docs/status.md
```diff
@@ -0,0 +1,99 @@
...
+1. Keep it simple. The logic for how the conditions on the `HyperConverged`'s
+   resource are being set should be easy to follow.
+1. Avoid timeouts. It's tempting to say if `KubeVirt` has been progressing for
+   an hour that we should allow OLM to upgrade us (by setting the readiness
+   probe to succeed). However, OLM should provide a mechanism for overriding
+   operators and we want this to be simple.
+
+**NOTE**
+
+Conditions consist of number of fields including Type, Status, etc. For the
```
rthallisey: `consist of a number...`

pkg/components/components.go
```diff
@@ -76,12 +76,13 @@ func GetDeployment(repository string, tag string, imagePullPolicy string) *appsv
...
 								{
-									Name: "WATCH_NAMESPACE",
-									ValueFrom: &corev1.EnvVarSource{
-										FieldRef: &corev1.ObjectFieldSelector{
-											FieldPath: "metadata.namespace",
-										},
-									},
+									Name:  "WATCH_NAMESPACE",
+									Value: "",
+									// ValueFrom: &corev1.EnvVarSource{
```
rthallisey: Comment still needed?

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -120,113 +137,744 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
 		return reconcile.Result{}, err
 	}
 
+	// Add conditions if there are none
```
rthallisey: I think `Readiness=false` is accurate always during initialization of the HCO, so I think we should it before the first reconcile occurs.   Maybe somewhere in `cmd/`

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -120,113 +137,744 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
...
+		// and here we are simply writing it back to the server.
+		// One shortcoming is that only one failure of a particular condition can be
+		// captured at one time (ie. if KubeVirt and CDI are both reporting !Available,
+		// you will only see CDI as it updates last).
+		for _, condition := range r.conditions {
+			conditionsv1.SetStatusCondition(&instance.Status.Conditions, condition)
+		}
+
+		// If for any reason we marked ourselves !upgradeable...then unset readiness
+		if conditionsv1.IsStatusConditionFalse(instance.Status.Conditions, conditionsv1.ConditionUpgradeable) {
```
rthallisey: Could using going out to the API server be racy?  Can we check the in memory condition?

pkg/controller/hyperconverged/hyperconverged_controller.go
djzager: Since the readiness probe is checking for that file the `ready` package creates. We could simply remove [these lines](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/cmd/hyperconverged-cluster-operator/main.go#L78-L85) to have the behavior you specify.

pkg/components/components.go
djzager: Nope.

pkg/controller/hyperconverged/hyperconverged_controller.go
djzager: We only get it once [here](https://github.com/kubevirt/hyperconverged-cluster-operator/pull/197/files#diff-64e1e2b9251fa5137c91d5a7c8427b88R127) for the rest of the Reconcile() we just make updates to it and occasionally write things like Status back (ie. `r.client.Status().Update(context.TODO(), instance)`). When we access `instance.Status.Conditions` we aren't having to reach out to the API server.

pkg/controller/hyperconverged/hyperconverged_controller.go
rthallisey: 👍 Let's remove those lines.

rthallisey: Do you agree/disagree that initialization should always be `Readiness=false`?

rthallisey:
any unit tests we can add to /hyperconverged_controller_test.go ?

pkg/controller/hyperconverged/hyperconverged_controller.go
djzager: I think it makes the most sense to start "unready" and not change that until you are "ready".

rthallisey:
ci test please

PR #199, opened 2019-08-05
Metrics aggregation

fromanirh:
This PR enables the metrics aggregation deployment rules.
To do that, we need to jump from ssp-operator client from v1.0.4 to v1.0.7


Gopkg.toml
```diff
@@ -96,9 +96,7 @@ required = [
 
 [[constraint]]
   name = "github.com/operator-framework/operator-sdk"
-  # The version rule is used for a specific release and the master branch for in between releases.
-  branch = "master" #osdk_branch_annotation
-  # version = "=v0.5.0" #osdk_version_annotation
+  version = "=v0.8.2"
```
djzager: Based on the comment I wonder where we are with respect to branching? @rthallisey

fromanirh: Sure thing. From HCO perspective, this PR just wants to enable the SSP operator to install our metric aggregations rule. In turn, metrics aggregation rules are just `PrometheusRule`s (prometheus operator CRDs). See: https://github.com/MarSik/kubevirt-ssp-operator/blob/master/roles/KubevirtMetricsAggregation/templates/aggregation-rule-vmi-count.yaml.j2

djzager:
I wonder @fromanirh what your thoughts are about adding the creation of:

```
apiVersion: monitoring.coreos.com/v1
kind: PrometheusRule
metadata:
  labels:
    prometheus: k8s
    role: cnv-rules
  name: prometheus-k8s-rules-cnv
  namespace: "{{ meta.namespace }}"
spec:
  groups:
  - name: cnv.rules
    rules:
    - expr: |
        sum(kubevirt_vmi_phase_count{phase="running"}) by (node)
record: cnv:vmi_status_running:count
```

directly in the HCO instead of creating another CRD. We haven't really yet established the boundaries between operators|controllers|reconcile steps (ie. when you should just create it vs. having a dedicated controller for it vs. creating a whole new operator for it) but this looks like a good candidate for just an additional reconcile step in the HCO.

fromanirh:
> I wonder @fromanirh what your thoughts are about adding the creation of:
> 
> ```
> apiVersion: monitoring.coreos.com/v1
> kind: PrometheusRule
> metadata:
>   labels:
>     prometheus: k8s
>     role: cnv-rules
>   name: prometheus-k8s-rules-cnv
>   namespace: "{{ meta.namespace }}"
> spec:
>   groups:
>   - name: cnv.rules
>     rules:
>     - expr: |
>         sum(kubevirt_vmi_phase_count{phase="running"}) by (node)
> record: cnv:vmi_status_running:count
> ```
> 
> directly in the HCO instead of creating another CRD. We haven't really yet established the boundaries between operators|controllers|reconcile steps (ie. when you should just create it vs. having a dedicated controller for it vs. creating a whole new operator for it) but this looks like a good candidate for just an additional reconcile step in the HCO.

The change per se looks easy. We need to undo some work on ssp-operator and do a release, but no big deal.

The problem is the ownership of the `PrometheusRule`. IMO it belongs to either virt-operator or ssp-operator, more the former than the latter. Moving it "one layer up" in the ssp-operator is still OK, but it has no place on HCO. Too up to the chain.

Last, please note that yes, at *this* moment we are shipping just one trivial PrometheusRule, but we are already discussing the future expansion with more rules/aggregation incoming.

djzager:
> but it has no place on HCO

maybe, but does this really deserve it's own CRD + controller? Even if there are eventually 1000 Prometheus Rules I'm not sure that we need to expose an API for interacting with these metrics rules. Based on what you are saying it sounds like this is something that should be included as part of the virt-operator, when I create a `KubeVirt` resource I would also want all of these `PrometheusRules`.

fromanirh:
> > but it has no place on HCO
> 
> maybe, but does this really deserve it's own CRD + controller? Even if there are eventually 1000 Prometheus Rules I'm not sure that we need to expose an API for interacting with these metrics rules. Based on what you are saying it sounds like this is something that should be included as part of the virt-operator, when I create a `KubeVirt` resource I would also want all of these `PrometheusRules`.

I tend to agree it doesn't deserve a CRD. It's just easy enough (actually, any other option is much harder) in the ansible-based SSP operator.

Adding in kubevirt resource is tempting, but there is the issue that the `PrometheusRule` can be deployed only if `prometheus-operator` is already deployed. That's something that we can take for granted in the SSP operator (we require to run on openshift anyway because of templates), but that kubevirt can't, because it wants to run on vanilla k8s.

mhrivnak:
> Adding in kubevirt resource is tempting, but there is the issue that the `PrometheusRule` can be deployed only if `prometheus-operator` is already deployed. That's something that we can take for granted in the SSP operator (we require to run on openshift anyway because of templates), but that kubevirt can't, because it wants to run on vanilla k8s.

It's easy enough to detect what APIs are available at runtime, so you could optionally create these rules when the required API is available.

fromanirh:
> > Adding in kubevirt resource is tempting, but there is the issue that the `PrometheusRule` can be deployed only if `prometheus-operator` is already deployed. That's something that we can take for granted in the SSP operator (we require to run on openshift anyway because of templates), but that kubevirt can't, because it wants to run on vanilla k8s.
> 
> It's easy enough to detect what APIs are available at runtime, so you could optionally create these rules when the required API is available.

Good point. The only remaining issue is we don't have the capacity to do that in the current sprint :)
@djzager would it work for you if we re-evaluate the deployment options for `PrometheusRules` in the next cycle? From HCO perspective you should see little changes.

Gopkg.toml
rthallisey: We'll cut the 2.1 branch after we have some more features merged.

deploy/operator.yaml
```diff
@@ -238,7 +238,7 @@ spec:
       serviceAccountName: kubevirt-ssp-operator
       containers:
         - name: kubevirt-ssp-operator
-          image: quay.io/fromani/kubevirt-ssp-operator-container:v1.0.4
+          image: quay.io/fromani/kubevirt-ssp-operator-container:v1.0.7
```
rthallisey: would like to move away from storing images in personal repos.

deploy/crds/metrics-aggregation.crd.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
rthallisey: Your CRD should be added to the `templates/crds/` dir.

deploy/operator.yaml
fromanirh: Agreed, Unfortunately we didn't make it in time, the ssp-operator is not yet under the kubevirt umbrella.

deploy/crds/metrics-aggregation.crd.yaml
fromanirh: Ack, fixing.

fromanirh:
@rthallisey sorry, I don't get what you want me to do wrt the Gopkg.toml dep update. Should I keep the `operator-sdk` dep to the master branch and update our code to follow their changes?

rthallisey:
@fromanirh I'd prefer we keep it tied to master until a 2.1 branch is cut then lock it down.

fromanirh:
> @fromanirh I'd prefer we keep it tied to master until a 2.1 branch is cut then lock it down.

OK, so unless I'm missing something, we'll need some update to HCO code - the operator-sdk API changed after v0.8.2, we need some fixes to our side unrelated to this patch.
I'll need either another PR and make this on depending on the new, or update this PR - I'd go for the latter approach.
Would that be OK for you?

rthallisey:
@fromanirh do you know which version of the sdk is targeted for 4.2? My understanding was they didn't cut a 4.2 targeting branch, but if operator-sdk has, then we should lock on to it.

fromanirh:
> @fromanirh do you know which version of the sdk is targeted for 4.2? My understanding was they didn't cut a 4.2 targeting branch, but if operator-sdk has, then we should lock on to it.

nope, I have no idea. I'll try to investigate ASAP.

fromanirh:
I'm delaying a bit the update to pick up the last SSP operator versions with the latest and greatest feature we need to deliver.

deploy/operator.yaml
fromanirh: IOW, that's the only u/s operator image available :\

fromanirh:
Looks like we need to rebase on https://github.com/kubevirt/hyperconverged-cluster-operator/pull/203 to address some of the comments from @rthallisey

fromanirh:
rebased on top of https://github.com/kubevirt/hyperconverged-cluster-operator/pull/203 , please check again

PR #203, opened 2019-08-07
Update Operator SDK dependency

djzager:


cmd/hyperconverged-cluster-operator/main.go
```diff
@@ -108,39 +113,50 @@ func main() {
 	log.Info("Registering Components.")
 
 	// Setup Scheme for all resources
-	if err := apis.AddToScheme(mgr.GetScheme()); err != nil {
-		log.Error(err, "")
-		os.Exit(1)
+	for _, f := range []func(*apiruntime.Scheme) error{
+		apis.AddToScheme,
+		cdiv1alpha1.AddToScheme,
```
djzager: It's weird that we aren't adding kubevirt to the Scheme.

rthallisey: I think it's because kubevirt adds itself when you import the pkg.   In other words, there's an `init` somewhere in kubevirt that does it.

PR #217, opened 2019-08-13
Fix CRDs, with versions sub-section

lveyde:
Some CRDs miss this sub-section, which causes issues with the OLM.

This patch fixes the issue for the generated files.

### DIFF
--- a/deploy/olm-catalog/kubevirt-hyperconverged/0.0.2/kubevirt.crd.yaml
+++ b/deploy/olm-catalog/kubevirt-hyperconverged/0.0.2/kubevirt.crd.yaml
@@ -25,3 +25,7 @@ spec:
     singular: kubevirt
   scope: Namespaced
   version: v1alpha3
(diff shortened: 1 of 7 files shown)


deploy/olm-catalog/kubevirt-hyperconverged/0.0.2/kubevirt.crd.yaml
```diff
@@ -25,3 +25,7 @@ spec:
     singular: kubevirt
   scope: Namespaced
   version: v1alpha3
+  versions:
```
rthallisey: These changes are going to get wiped out when we regenerate.   Components will have to add these

rthallisey:
I think we should wait for https://github.com/operator-framework/operator-lifecycle-manager/pull/973 to be fixed instead.  Closing this

rthallisey:
Reopen Lev if you disagree

Issue #206, opened 2019-08-09
HyperConverged Controller needs more test coverage


djzager:
A majority of the HCO's code lives in the controller and at the time of writing we are at 42% code coverage. This may easily be improved by following the [sample memcached operator](https://github.com/operator-framework/operator-sdk-samples/blob/master/memcached-operator/pkg/controller/memcached/memcached_controller_test.go) from operator-sdk-samples.

PR #223, opened 2019-08-15
wip: test(reconcile): improve testing of Reconcile func

djzager:
The purpose of this PR is to bring the code coverage back to more
acceptable figures after the HCO Conditions PR was merged.


rthallisey:
fixes: https://github.com/kubevirt/hyperconverged-cluster-operator/issues/206

pkg/controller/hyperconverged/hyperconverged_controller_test.go
```diff
@@ -2,230 +2,993 @@ package hyperconverged
...
+					Status:  corev1.ConditionTrue,
+					Reason:  "KubeVirtConditions",
+					Message: "KubeVirt resource has no conditions",
+				})))
+				Expect(r.conditions).To(ContainElement(RepresentCondition(conditionsv1.Condition{
+					Type:    conditionsv1.ConditionUpgradeable,
+					Status:  corev1.ConditionFalse,
+					Reason:  "KubeVirtConditions",
+					Message: "KubeVirt resource has no conditions",
+				})))
```
rwsu: Why isn't conditionsv1.ConditionDegraded checked here? I see it used in line 227.

pkg/controller/hyperconverged/hyperconverged_controller_test.go
```diff
@@ -2,230 +2,993 @@ package hyperconverged
...
+				// Check conditions
+				Expect(r.conditions).To(ContainElement(RepresentCondition(conditionsv1.Condition{
+					Type:    conditionsv1.ConditionAvailable,
+					Status:  corev1.ConditionFalse,
+					Reason:  "KubeVirtConditions",
+					Message: "KubeVirt resource has no conditions",
+				})))
+				Expect(r.conditions).To(ContainElement(RepresentCondition(conditionsv1.Condition{
+					Type:    conditionsv1.ConditionProgressing,
+					Status:  corev1.ConditionTrue,
```
rwsu: If no conditions were found for KubeVirt, how do we know if it is actually progressing? Should this be ConditionUnknown?

pkg/controller/hyperconverged/hyperconverged_controller_test.go
```diff
@@ -2,230 +2,993 @@ package hyperconverged
...
+						Reason:  "Foo",
+						Message: "Bar",
+					},
+					kubevirtv1.KubeVirtCondition{
+						Type:    kubevirtv1.KubeVirtConditionDegraded,
+						Status:  corev1.ConditionTrue,
+						Reason:  "Foo",
+						Message: "Bar",
+					},
+				}
```
rwsu: Here it is testing one combination of conditions. Any thoughts about testing the other possible combinations in https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/docs/conditions.md?

pkg/controller/hyperconverged/hyperconverged_controller_test.go
djzager: We don't set the degraded condition when the kubevirt resource has no conditions. On 227 I had specifically crafted a KubeVirt resource with the degraded condition set to true.

pkg/controller/hyperconverged/hyperconverged_controller_test.go
djzager: They start adding conditions really on in their reconcile. That's why I think it's fair to assume that the operator is doing work if there are no conditions.

pkg/controller/hyperconverged/hyperconverged_controller_test.go
djzager: My focus in this pass was to cover all of the different paths through the `Reconcile()` it may be worth going through all of the possible combinations.

djzager:
> Question related to upgrades. When a new installPlan is created for an upgrade, shouldn't there be a status somewhere that tells us that an upgrade was kicked off irregardless of the sub component condition statuses? When a OCP cluster is upgraded, you can see from the UI that it is in-progress. Do we have a status like that at the HCO level?

In the ClusterOperator case (the resource being watched by CVO) they have this idea of [versions](https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md#version) and this is how they are able to deduce what's going on [during an upgrade](https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md#version-reporting-during-an-upgrade).

Having the concept of version and/or reconciled version and all that stuff seems useful but I think we should keep it simple for now.

djzager:
> And is there a need to differentiate between a heal and an upgrade when a the application is in degraded state?

I don't know if there is a need, but I think the primary benefit to having the conditions is that between it and log messages you can more easily drill down into what's going on. We provide the reference to the KubeVirt resource we create in `HCO.Status.RelatedObjects` so when we mark ourselves as degraded because of `KubeVirtDegraded` then it's easy to know where to look.

PR #225, opened 2019-08-20
Remove blank line from mro.crd.yaml

rwsu:
Causes "make bundleRegistry" to fail.


deploy/olm-catalog/kubevirt-hyperconverged/0.0.2/mro.crd.yaml
```diff
@@ -1,4 +1,3 @@
-
```
rthallisey: This will get wiped out when the csv gets regenerated.  Edit ` templates/crds/mro.crd.yaml.in` and put `{{if .Converged}}` on the same line as `---`

rwsu: Ok! Done.

rthallisey:
@rwsu also run `hack/build-manifests.sh` so the CR gets the change.

PR #227, opened 2019-08-20
controller: create storage configmap on reconcile

danielerez:
Added a new ConfigMap for defining default values for
storage according to the following logic:

- For a BareMetal infrastructure:
  - volumeMode: Block
  - accessMode: ReadWriteMany
- For all other deployments:
  - volumeMode: File
  - accessMode: ReadWriteMany

### DIFF
--- a/pkg/apis/hco/v1alpha1/hyperconverged_types.go
+++ b/pkg/apis/hco/v1alpha1/hyperconverged_types.go
@@ -17,6 +17,12 @@ type HyperConvergedSpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
 	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
+
+	// BareMetalPlatform indicates whether the infrastructure is baremetal.
+	BareMetalPlatform bool `json:"BareMetalPlatform,omitempty"`
+
+	// LocalStorageClassName the name of the local storage class.
+	LocalStorageClassName string `json:"LocalStorageClassName,omitempty"`
 }
 
 // HyperConvergedStatus defines the observed state of HyperConverged
--- a/pkg/controller/hyperconverged/hyperconverged_controller.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller.go
@@ -190,6 +190,7 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
 
 	for _, f := range []func(*hcov1alpha1.HyperConverged, logr.Logger, reconcile.Request) error{
 		r.ensureKubeVirtConfig,
+		r.ensureKubeVirtStorageConfig,
 		r.ensureKubeVirt,
 		r.ensureCDI,
 		r.ensureNetworkAddons,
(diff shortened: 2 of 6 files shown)


danielerez:
@rthallisey @mareklibra @aglitke @awels @jelkosz @rollandf - please review

docs/configmaps.md
```diff
@@ -0,0 +1,15 @@
...
+
+### Example
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: kubevirt-storage-class-defaults
+  namespace: openshift
+data:
+  accessMode: ReadWriteMany
+  volumeMode: Filesystem (Block for BareMetal infrastructure)
```
mareklibra: nit: the example should be copy&paste, so I suggest to add the hint into a valid comment

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -821,6 +826,77 @@ func (r *ReconcileHyperConverged) ensureKubeVirtTemplateValidator(instance *hcov
...
+		"app": cr.Name,
+	}
+	return &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "kubevirt-config-storage-class-defaults",
+			Labels:    labels,
+			Namespace: namespace,
+		},
+		Data: map[string]string{
+			"accessMode": "ReadWriteMany",
```
mareklibra: With this set, no VM created by UI (Wizard) can be started when local SC is used.

docs/configmaps.md
```diff
@@ -0,0 +1,15 @@
+# ConfigMaps
```
rthallisey: This is specifically for RHHI right?  I would call this doc `docs/kni.md`

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -821,6 +826,77 @@ func (r *ReconcileHyperConverged) ensureKubeVirtTemplateValidator(instance *hcov
...
+			"volumeMode": volumeMode,
+		},
+	}
+}
+
+func (r *ReconcileHyperConverged) isBareMetalPlatform(request reconcile.Request) bool {
+	infra := &apiconfigv1.Infrastructure{}
```
rthallisey: I would rather the user set a RHHI or Baremetal flag instead of us auto detecting.   You can add the flag [here](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/pkg/apis/hco/v1alpha1/hyperconverged_types.go#L19).

danielerez:
> This patch is sort of hardcoding the defaults, no matter a `ConfigMap` is used.
> 
> Is automatic detection of StorageClass capabilities still in plan?
> The ConfigMap should contain sc-specific defaults if these generic (Filesystem/RWM) can not be used (like for the local storage).
> 
> There is no way to hardcode the defaults right for a generic cluster, as recently implemented here.
> If detection of SC capabilities will not land shortly, there should be a way how the administrator can change values in this ConfigMap. Via HCO configuration or manually while not-breaking the reconciliation logic.

In 'ensureKubeVirtStorageConfig' func we create the ConfigMap only if doesn't exist, so admin should be able to modify it.

docs/configmaps.md
danielerez: sure, done.

pkg/controller/hyperconverged/hyperconverged_controller.go
danielerez: done

djzager:
> I would rather the user set a RHHI or Baremetal flag instead of us auto detecting. You can add the flag here.

I'm curious @rthallisey why we are against auto detecting. Originally I would have thought we would want to auto detect where possible.

pkg/apis/hco/v1alpha1/hyperconverged_types.go
```diff
@@ -17,6 +17,9 @@ type HyperConvergedSpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
 	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
+
+	// IsBareMetalPlatform indicates whether the infrastructure is baremetal.
+	IsBareMetalPlatform bool `json:"IsBareMetalPlatform,omitempty"`
```
rthallisey: How about `BareMetalPlatform`?

danielerez: sounds good, changed.

rthallisey:
> I'm curious @rthallisey why we are against auto detecting. Originally I would have thought we would want to auto detect where possible.
>
My thinking is that I want to avoid having code that checks for OpenShift specific things in tree wherever possible.  Having a flag for the user to set isn't too much of a trade off IMO.

djzager:
> My thinking is that I want to avoid having code that checks for OpenShift specific things in tree wherever possible. Having a flag for the user to set isn't too bad of a trade-off IMO.

Interesting, if I may offer a different perspective. I would suggest that "we", as the maintainers of the HCO, know best how to install all of the different KubeVirt components and the HyperConverged resource is our API. Every field in our HyperConverged's spec, we accept all of the supported values and support them -- consider a memcached-operator that accepts a version (but only 1, 2, 3, 4) then a user would expect to be able to put any of 1-4 in the spec.version (there could be issues where downgrades are not allowed but you get the idea). In our current scenario, we have no way of supporting `BareMetalPlatform` on anything but a bare metal platform...so we would need to have some validation on that field before doing work anyway. We could simply detect the appropriate APIs before doing work without asking the user for anything...that would prevent the user from mucking around with fields and values that we never intend to support (like `BareMetalPlatform: true` on a non-baremetal cluster).

danielerez:
> > My thinking is that I want to avoid having code that checks for OpenShift specific things in tree wherever possible. Having a flag for the user to set isn't too bad of a trade-off IMO.
> 
> Interesting, if I may offer a different perspective. I would suggest that "we", as the maintainers of the HCO, know best how to install all of the different KubeVirt components and the HyperConverged resource is our API. Every field in our HyperConverged's spec, we accept all of the supported values and support them -- consider a memcached-operator that accepts a version (but only 1, 2, 3, 4) then a user would expect to be able to put any of 1-4 in the spec.version (there could be issues where downgrades are not allowed but you get the idea).
> 
> In our current scenario, we have no way of supporting `BareMetalPlatform` on anything but a bare metal platform...so we would need to have some validation on that field before doing work anyway. We could simply detect the appropriate APIs before doing work without asking the user for anything...that would prevent the user from mucking around with fields and values that we never intend to support (like `BareMetalPlatform: true` on a non-baremetal cluster).

@rthallisey So what's your take on that then? Should we keep the flag or revert to the previous solution (detect automatically by fetching the infrastructure's platform).

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -821,6 +822,66 @@ func (r *ReconcileHyperConverged) ensureKubeVirtTemplateValidator(instance *hcov
...
+	return &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "kubevirt-config-storage-class-defaults",
+			Labels:    labels,
+			Namespace: namespace,
+		},
+		Data: map[string]string{
+			"accessMode":          "ReadWriteMany",
+			"volumeMode":          volumeMode,
+			"local-sc.accessMode": "ReadWriteOnce",
```
mareklibra: @danielerez , this assumes that the target cluster has a storage class called `local-sc`. It's up to cluster's admin how (s)he calls it and what particular SC is behind that name.

We can either
- autodetect
- or provide a way for manual intervention

danielerez: The admin can change it by modifying the ConfigMap (reconcile won't change it once already created). @rthallisey do we want to add the local class name to HyperConvergedSpec as well or is it too specific?

danielerez: @mareklibra @rthallisey Added 'LocalStorageClassName' key to HyperConverged spec, so it could be configured on deployment.

pkg/apis/hco/v1alpha1/hyperconverged_types.go
```diff
@@ -17,6 +17,12 @@ type HyperConvergedSpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
 	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
+
+	// BareMetalPlatform indicates whether the infrastructure is baremetal.
+	BareMetalPlatform bool `json:"BareMetalPlatform,omitempty"`
```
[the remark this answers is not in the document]
danielerez: Added.

PR #232, opened 2019-08-20
Update CNA deployment name as a temp. fix

lveyde:


lveyde:
> 
> 
> Should we change the names for every deployment?

Generally it will only be required in case we change some immutable fields, e.g. labels etc.

We should either generate random names for each release for all containers, or only do some local fixes, and as currently we await to see the progress on the OLM side, I think that local fixes should do.

What do you think?

rthallisey:
Let's randomly generate the names of all the containers.  How about using git hash?

rthallisey:
Replacing with https://github.com/kubevirt/hyperconverged-cluster-operator/pull/238

PR #242, opened 2019-08-23
ssp: bump dependency to 1.0.9

fromanirh:
We want to ship a new image with improved condition reporting.


fromanirh:
This change seems a bit too big @rthallisey @tiraboschi @djzager . What I did:
1. bump dep in `Gopkg.toml`
2. run `dep update -ensure`
3. run `hack/build-manifests.sh`
4. added to the patch only the hunks related to the SSP

MarSik:
The version bump is fine, but 380 files.. ugh...

djzager:
> The version bump is fine, but 380 files.. ugh...

Yeah. It does get kind of insane. For PRs, sometimes it helps to put all of the vendor stuff in a specific commit so that a reviewer can look at the actual changes separately. Not sure if that's needed here though.

djzager:
The other option is to only update one package with the `dep ensure -u github.com/MarSik/kubevirt-ssp-operator`

Issue #246, opened 2019-08-24
The CSV 0.0.1 uses "latest" floating tag


fromanirh:
Looking at the 0.0.1 CSV, we see that the operator image tag is "latest":
https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/deploy/olm-catalog/kubevirt-hyperconverged/0.0.1/kubevirt-hyperconverged-operator.v0.0.1.clusterserviceversion.yaml#L73

This means that if we try to deploy v0.0.1 (e.g. to test the upgrade flow) we will get actually a 0.0.2 image. To run, this image will need additional CRDs, like `machine remediation operator` and `metrics aggregation`, which are 0.0.2 features.

The suggested fix is push a 0.0.1 image (stream) and fix the CSV accordingly.
I see this image published on quay.io: https://quay.io/repository/kubevirt/hyperconverged-cluster-operator?tab=tags `v0.0.1`  but I'm not sure it is good.

fromanirh:
Looks like the aforementioned image has some issues:
```bash
[root@shinji2 hcotest01]# oc get pods
NAME                                               READY   STATUS             RESTARTS   AGE
cdi-operator-57768c48dc-bghgx                      1/1     Running            0          8m8s
cluster-network-addons-operator-7bdd7579bb-mbxqt   1/1     Running            0          8m8s
hco-operator-858766d9-vgfkb                        0/1     CrashLoopBackOff   6          8m8s
kubevirt-ssp-operator-d787b7dd8-xdp4b              1/1     Running            0          8m8s
node-maintenance-operator-85bbf7f9b8-8fzt9         1/1     Running            0          8m8s
virt-operator-c4866bdcb-22fzv                      1/1     Running            0          8m8s
virt-operator-c4866bdcb-lqv5q                      1/1     Running            0          8m8s
[root@shinji2 hcotest01]# oc describe pod hco-operator-858766d9-vgfkb | grep image
  Normal   Pulled     7m8s (x4 over 8m1s)    kubelet, test1-8xtjc-worker-0-ddsc9  Successfully pulled image "quay.io/kubevirt/hyperconverged-cluster-operator:v0.0.1"
  Normal   Pulling    6m15s (x5 over 8m8s)   kubelet, test1-8xtjc-worker-0-ddsc9  Pulling image "quay.io/kubevirt/hyperconverged-cluster-operator:v0.0.1"
[...]

PR #250, opened 2019-08-26
Use HCO v0.0.1 in CSV v0.0.1

lveyde:
Make sure we use HCO v0.0.1 in CSV v0.0.1, instead of the latest version,
to prevent issues that arise from that.


fromanirh:
/lgtm
fixes: https://github.com/kubevirt/hyperconverged-cluster-operator/issues/246

PR #257, opened 2019-08-27
Fix CRD templates

lveyde:
This PR includes patches that fix the CRD templates.


lveyde:
> 
> 
> @lveyde did you run `hack/build-manifests.sh`?

Yes, in order verify the patches. Will send updated manifests in a separate PR.

lveyde:
> 
> 
> > Yes, in order verify the patches. Will send updated manifests in a separate PR.
> 
> No need. Might as well include them here so it's easier to track how template changes affect the files.

Adding patch w/ fixed CRD manifests.

PR #261, opened 2019-08-28
Expand the jq command to print out all versions

rthallisey:
The last item in the list isn't always the newest bundle.  We might as well print out all the bundles that exist.


rthallisey:
The while was making sure the user doesn't input the same value as the latest bundle.  It's hard in bash to figure out the latest `x.y.z` value.  However, the user isn't completely out in the cold, the push step will fail if a duplicate bundle already exists.

PR #262, opened 2019-08-29
Add missing nmstate and ovs to networkaddonsconfig

oshoval:
was removed as part of 

https://github.com/kubevirt/hyperconverged-cluster-operator/commit/77be1be9ebd4e18c1850cd42a2b494ae7b351ccd#diff-64e1e2b9251fa5137c91d5a7c8427b88L12


oshoval:
adding
/cc @djzager

is it safe the return just those 2 lines out of the commit or more is needed please?

PR #290, opened 2019-09-16
bump CSV for 0.0.3

phoracek:


phoracek:
OKD check failed with:

```
./cluster-up/kubectl.sh get pods -n openshift-operator-lifecycle-manager | grep hco-catalogsource
hco-catalogsource-example-t57mj     1/1     Terminating   0          3m25s
error: resource(s) were provided, but no name, label selector, or --all flag specified
```

Any idea what could be wrong there? Is it possible that we have outdated `oc`? I saw similar issue when I used `oc` 3.11 instead of 4.

phoracek:
ci test please

rwsu:
@phoracek I think it is picking up the old hco-catalogsource pod after the patch in https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/hack/upgrade-test.sh#L188. I'll post a patch to exclude "Terminating" pods. Thanks for pointing this out.

PR #310, opened 2019-09-26
[release-2.1] Drop v2v support on release 2.1

tiraboschi:


djzager:
This must come after #306 since it actually stops the reconcile of the configmap

deploy/operator.yaml
```diff
@@ -29,9 +29,7 @@ spec:
               fieldPath: metadata.name
         - name: WATCH_NAMESPACE
         - name: CONVERSION_CONTAINER
```
djzager: Can probably remove the environment variable entirely for now. We should probably also drop the `ensure_ims_config` from the hyperconverged controller.

mareklibra:
lgtm in general - to disable the v2v in the UI, it's sufficient to remove the `V2VVmware` CRD which I can see to happen in this PR.

I would need to see this PR executed (which I can't) to be sure that other changes do not break anything else.

PR #110, opened 2019-06-04
Adding script to deploy HCO in one line

mmirecki:

PR #314, opened 2019-09-29
HCO Upgrade test validates HCO component operators

MoserMichael:
upon completion of upgrade testing with make upgrade-test : check if cluster is up and running

Signed-off-by: MichaelMoser <mmoser@localhost.localdomain>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,7 @@ endif
 build: $(SOURCES) ## Build binary from source
 	go build -i -ldflags="-s -w" -o _out/hyperconverged-cluster-operator ./cmd/hyperconverged-cluster-operator
 	go build -i -ldflags="-s -w" -o _out/csv-merger tools/csv-merger/csv-merger.go
+	go build -i -ldflags="-s -w" -o _out/test-hco-utils tools/test-hco-utils/test-hco-utils.go
 
 install:
 	go install ./cmd/...
@@ -41,6 +42,12 @@ container-build: container-build-operator container-build-operator-courier
 container-build-operator:
 	docker build -f build/Dockerfile -t $(IMAGE_REGISTRY)/$(OPERATOR_IMAGE):$(IMAGE_TAG) .
 
+container-copy-test-hco:
+	bash -cx "mkdir -p test-out || ls -al test-out || true"
+	docker run --entrypoint cat $(IMAGE_REGISTRY)/$(OPERATOR_IMAGE):$(IMAGE_TAG) /usr/local/bin/test-hco-utils > ./test-out/test-hco-utils
+	stat ./test-out/test-hco-utils
+	chmod +x ./test-out/test-hco-utils
+ 
 container-build-operator-courier:
 	docker build -f tools/operator-courier/Dockerfile -t hco-courier .
 
@@ -79,6 +86,9 @@ stageRegistry:
 
 bundleRegistry:
 	REGISTRY_NAMESPACE=$(REGISTRY_NAMESPACE) IMAGE_REGISTRY=$(IMAGE_REGISTRY) ./hack/build-registry-bundle.sh
+
+container-clusterserviceversion:
+	REGISTRY_NAMESPACE=$(REGISTRY_NAMESPACE) IMAGE_REGISTRY=$(IMAGE_REGISTRY) ./hack/upgrade-test-clusterserviceversion.sh
 
 build-push-all: container-build-operator container-push-operator container-build-operator-courier bundle-push
 
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,6 @@ endif
 build: $(SOURCES) ## Build binary from source
 	go build -i -ldflags="-s -w" -o _out/hyperconverged-cluster-operator ./cmd/hyperconverged-cluster-operator
 	go build -i -ldflags="-s -w" -o _out/csv-merger tools/csv-merger/csv-merger.go
-	go build -i -ldflags="-s -w" -o _out/test-hco-utils tools/test-hco-utils/test-hco-utils.go
 
 install:
 	go install ./cmd/...
@@ -42,12 +41,6 @@ container-build: container-build-operator container-build-operator-courier
 container-build-operator:
 	docker build -f build/Dockerfile -t $(IMAGE_REGISTRY)/$(OPERATOR_IMAGE):$(IMAGE_TAG) .
 
-container-copy-test-hco:
-	bash -cx "mkdir -p test-out || ls -al test-out || true"
-	docker run --entrypoint cat $(IMAGE_REGISTRY)/$(OPERATOR_IMAGE):$(IMAGE_TAG) /usr/local/bin/test-hco-utils > ./test-out/test-hco-utils
-	stat ./test-out/test-hco-utils
-	chmod +x ./test-out/test-hco-utils
- 
 container-build-operator-courier:
 	docker build -f tools/operator-courier/Dockerfile -t hco-courier .
 
--- a/hack/upgrade-test-build-images.sh
+++ b/hack/upgrade-test-build-images.sh
@@ -1,4 +1,4 @@
-#!/bin/bash -e
+#!/bin/bash -ex
 #
 # This file is part of the KubeVirt project
 #
@@ -47,8 +47,14 @@ export REGISTRY_IMAGE_NAME="hco-registry-upgrade"
 export REGISTRY_EXTRA_BUILD_ARGS="--build-arg KUBEVIRT_PROVIDER=$KUBEVIRT_PROVIDER"
 make bundleRegistry
 
+pwd
+make container-clusterserviceversion
+ls -al ./test-out
+
+
 # check images are accessible
 CLUSTER_NODES=$(./cluster-up/kubectl.sh get nodes | grep Ready | cut -d ' ' -f 1)
 for NODE in $CLUSTER_NODES; do
     ./cluster-up/ssh.sh $NODE 'sudo podman pull registry:5000/kubevirt/hco-registry-upgrade:latest'
 done
+
--- a/tools/test-hco-utils/test-hco-utils.go
+++ b/tools/test-hco-utils/test-hco-utils.go
@@ -1,0 +1,243 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/ghodss/yaml"
+	//"gopkg.in/yaml.v2"
+	"io/ioutil"
+	"os"
+	"os/exec"
+)
+
+var (
+	kubectlCmd                    string
+	clusterServiceVersionFileName string
+)
+
+// yaml parsing structs (for parsing of cluster service version yaml file)
+type hcoType struct {
+	Spec hcoSpec `yaml:"spec"`
+}
+
+type hcoSpec struct {
+	Install hcoInstall `yaml:"install"`
+}
+
+type hcoInstall struct {
+	Spec hcoInstallSpec `yaml:"spec"`
+}
+
+type hcoInstallSpec struct {
+	Deployments []hcoDeployments `yaml:"deployments"`
+}
+
+type hcoDeployments struct {
+	Name string            `yaml:"name"`
+	Spec hcoDeploymentSpec `yaml:"spec"`
+}
+
+type hcoDeploymentSpec struct {
+	Template hcoDeploymentSpecTemplate `yaml:"template"`
+}
+
+type hcoDeploymentSpecTemplate struct {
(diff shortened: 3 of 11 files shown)


hack/check-state.sh
```diff
@@ -0,0 +1,132 @@
...
+   while [ "$retry_count" -lt "$MAX_APP_AVAILABLE_CHECKS" ]; do
+     get_hco_json
+    
+     get_reconcile_json
+     if [ "x$RECONCILE_COMPLETED" == 'xTrue' ]; then
+
+        get_application_available_json
+        get_application_degraded_json
+        
+        if [ "x$APPLICATION_AVAILABLE" == 'xTrue' ] && [ "x$APPLICATION_DEGRADED" == 'xFalse' ]; then
```
rwsu: I'm curious why "OPERATION_PROGRESSING" isn't checked here for false? Isn't there a case where the application is functioning but is in the middle of upgrading or healing?

MoserMichael: OPEATION_PROGRESSING is checked in line #87, this way I can put out a detailed print out in line #110 that we are still progressing.

MoserMichael: Just a question: what should happen if after the global timeout we have APPLICATION_AVAILABLE=True && APPLICATION_DEGRADED=False && OPERATION_PROGRESSING=True ? Is that considered a failure or success of the upgrade?

MoserMichael: I misunderstood that OPERATION_PROGRESSING==True while APPLICATION_AVAILABLE==True && APPLICATION_DEGRADED==False would mean that the upgrade finished successfully. Ok, will fix that.

hack/check-state.sh
```diff
@@ -0,0 +1,132 @@
...
+MAX_APP_AVAILABLE_CHECKS=500
+MAX_GET_HCO_RETRY=500
+
+function get_hco_json() {
+    retry_count_hco=1
+    
+    while [ true ]; do
+
+
+        HCO_JSON=`./cluster-up/kubectl.sh get hyperconvergeds.hco.kubevirt.io hyperconverged-cluster -n kubevirt-hyperconverged -o go-template='{{range .status.conditions }}{{ if eq .type "ReconcileComplete" }}{{ printf " ReconcileComplete: \"%s\" \"%s\" \"%s\" "  .status .reason .message  }}{{else if eq .type "Progressing" }}{{ printf " ApplicationProgressing: \"%s\" \"%s\" \"%s\" "  .status .reason .message  }}{{ else if eq .type "Available" }}{{ printf "ApplicationAvailable: \"%s\" \"%s\" \"%s\""  .status .reason .message  }}{{ else if eq .type "Degraded" }}{{ printf "ApplicationDegraded: \"%s\" \"%s\" \"%s\""  .status .reason .message  }}{{ end }}{{ end }}'`
```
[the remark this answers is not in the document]
MoserMichael: the deploy.sh script sources hack/common.sh where CMD Is set to the wrong value for STDCI.
However in the Makefile you are setting the CMD variable for the cluster-clean target, that's what I did for the new check-state target.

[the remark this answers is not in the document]
MoserMichael: fixed in commit 9719329237c6e0adc65fc2f1bc34dc152926dcbb

hack/check-state.sh
MoserMichael: fixed in commit 9719329237c6e0adc65fc2f1bc34dc152926dcbb

ifireball:
ci test please

hack/check-state.sh
```diff
@@ -0,0 +1,210 @@
...
+
+cat <<EOF
+
+Checking operator kind: $kind name: $name 
+
+EOF
+
+   retry_count=0
+   while [ "$retry_count" -lt "$MAX_APP_AVAILABLE_CHECKS" ]; do
+      get_operator_json $kind $name $namespace
```
djzager: I think this script would benefit from eliminating some of the functions (I personally think all of the `get_*_json` functions can go away), and explicitly getting the condition you want with something like:

```
${CMD} get $kind $name $namespace -o go-template='{{range .status.conditions }}{{ if eq .type "Available" }}{{ .status }}{{ end }}'
```

You only really need the `reason` and `message fields to tell us humans whats broken and why...in which case you would probably want to just do a `kubectl describe` on the object so we can see the whole thing in a glance.

hack/check-state.sh
```diff
@@ -0,0 +1,210 @@
...
+    
+      if [ "x$APPLICATION_AVAILABLE" == 'xTrue' ] && [ "x$APPLICATION_DEGRADED" == 'xFalse' ]; then
+        if [ "x$OPERATION_PROGRESSING" == 'xFalse' ]; then
+          echo "Success: Operator kind: $kind name: $name ${HAS_RECONCILE}is fully available"
+          return
+        fi
+        REASON="Operator $kind available & not degraded, however still progressing"    
+      else
+        if [ "x$OPERATION_PROGRESSING" == 'xFalse' ]; then
+            set +x
```
djzager: not sure you need this here.

hack/check-state.sh
MoserMichael: I will remove the get_*_json functions; However I don't think we want to have call to kubectl/oc get for every condition check, that's why I have one kubectl get to retrieve all the data and then parse it for each condition (previously I got the json for the whole condition and extracted each field by means of jq - however jq is not installed on all platforms so I switched it to go-template)

djzager: > However I don't think we want to have call to kubectl/oc get for every condition check (these take a long time and things may change between runs)

My guess is that this is because of the performance hit of using `kubectl` to talk to an api server to grab the conditions that we already have? If that is the case, my argument would be that since this is a test and not an application itself + the bottleneck is in how long the upgrade takes not these commands keeping up that the improved readability is worth the performance impact given the mitigating circumstances.

hack/check-state.sh
```diff
@@ -0,0 +1,210 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
MoserMichael: ok

hack/check-state.sh
```diff
@@ -0,0 +1,210 @@
...
+    local namespace=$3
+    local retry_count_hco=1
+ 
+    if [ "$namespace" != "." ]; then
+        namespace="-n $namespace"
+    else
+        namespace=""
+    fi
+    
+    while [ true ]; do
```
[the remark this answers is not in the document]
MoserMichael: ok

hack/check-state.sh
MoserMichael: I think we need that.
If OperationAvailable==False || OperationDegraded==True then 
     if OperationProgressing==False then 
        error - no more hope for change
     fi

hack/check-state.sh
```diff
@@ -0,0 +1,210 @@
...
+        HCO_JSON=`${CMD} get $kind $name $namespace -o go-template='{{range .status.conditions }}{{ if eq .type "ReconcileComplete" }}{{ printf " ReconcileComplete: \"%s\" \"%s\" \"%s\" "  .status .reason .message  }}{{else if eq .type "Progressing" }}{{ printf " ApplicationProgressing: \"%s\" \"%s\" \"%s\" "  .status .reason .message  }}{{ else if eq .type "Available" }}{{ printf "ApplicationAvailable: \"%s\" \"%s\" \"%s\""  .status .reason .message  }}{{ else if eq .type "Degraded" }}{{ printf "ApplicationDegraded: \"%s\" \"%s\" \"%s\""  .status .reason .message  }}{{ end }}{{ end }}'`
+
+        if [ "x$?" == "x0" ]; then
```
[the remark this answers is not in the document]
MoserMichael: done

hack/check-state.sh
MoserMichael: I am more concerned that the state of the operator may change in between runs of kubectl, so that the script would make an incorrect decision based on old data obtained from a previous run; I think it is therefore reasonable to get the state of all conditions in one run of kubectl.

hack/check-state.sh
djzager: I was referring to the `set +x`

hack/check-state.sh
djzager: While I'm not totally on board with the concern, I think what you have is fairly easy to reason about so I'm good with it.

Makefile
```diff
@@ -68,6 +68,9 @@ build-push-all: container-build-operator container-push-operator container-build
 upgrade-test:
 	./hack/upgrade-test.sh
  
+check-state:
+	timeout 10m bash -c 'export CMD="./cluster-up/kubectl.sh";exec ./hack/check-state.sh'
```
rwsu: ./cluster-up/kubectl.sh will fail on Prow since it only works with STDCI.

rthallisey:
@MoserMichael what else is needed to finish this PR up?

rthallisey:
ci test please

MoserMichael:
Please review, the feature is now working as expected.

rwsu:
Hi @MoserMichael. This PR is getting more confusing. I see you created a new PR for checking the condition statuses, https://github.com/kubevirt/hyperconverged-cluster-operator/pull/336, which was the original intent of this PR. 

And in this PR, you've added changes to validate the image versions to check that it matches the CSV. The subject of this PR no longer matches the content.

There are also merge issues in upgrade-test.sh. I see old lines that have now been eliminated.

This PR is not reviewable now.

I would suggest closing this PR and starting a new one for validating the image versions.

hack/check-state.sh
```diff
@@ -0,0 +1,229 @@
+#!/usr/bin/env bash
```
rwsu: This file appears to duplicate the same file in this PR https://github.com/kubevirt/hyperconverged-cluster-operator/pull/336. I think 336 should be reviewed and merged first and then rebase this PR on top of it. Or did you have a different merging process in mind?

tools/test-hco-utils/test-hco-utils.go
```diff
@@ -0,0 +1,243 @@
+package main
```
rwsu: This should be written as a BDD-style tests using ginkgo to match the other existing functional style tests. See  https://github.com/onsi/ginkgo
https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/pkg/controller/hyperconverged/hyperconverged_controller_test.go
https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/tests/func-tests/virtual_machines_test.go

MoserMichael: I need to pass parameters to the test; currently the test gets two command line parameters: 1) the path of the clusterservice yaml file 2) the command name of the kubectl command (so as not to hardcode it). I am still learning ginko, so my questions is: how should I pass these values to the ginko test?

hack/check-state.sh
MoserMichael: done.

tools/test-hco-utils/test-hco-utils.go
MoserMichael: I have a problem with ginkgo: the test executable is compiled in the build docker, but copied from the docker to the shell where upgrade-test.sh is running, that is outside of the docker. Here we don't have go and don't have ginkgo so I can't invoke the test other than running the test executable (can't run go test or ginkgo as mentioned in https://onsi.github.io/ginkgo/ )

Makefile
MoserMichael: removed these lines. now it is no longer hard coded.

MoserMichael:
> This PR still has the same subject name as #336. I find this confusing.
PR's have been renamed - same title as corresponding Jira issues.

tools/test-hco-utils/test-hco-utils.go
MoserMichael: done, now using ginkgo

hack/upgrade-test-build-images.sh
```diff
@@ -47,8 +47,14 @@ export REGISTRY_IMAGE_NAME="hco-registry-upgrade"
 export REGISTRY_EXTRA_BUILD_ARGS="--build-arg KUBEVIRT_PROVIDER=$KUBEVIRT_PROVIDER"
 make bundleRegistry
 
+pwd
+make container-clusterserviceversion
```
rwsu: This script only runs on the STDCI. We'll need to have the component operator validation also occur on the OpenShift CI.

MoserMichael:
Do you want to merge it, given that all requests have been addressed as of now?

rwsu:
@rthallisey and @tiraboschi, is this ok to merge? With this PR the component validation only works on STDCI. There will need to be another PR to add support for OpenShift CI.

PR #315, opened 2019-10-02
[CNV-3175] HCO add debugging information when an upgrade fails

MoserMichael:
Also add make dump-state target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,6 +67,11 @@ build-push-all: container-build-operator container-push-operator container-build
 
 upgrade-test:
 	./hack/upgrade-test.sh
+
+dump-state:
+	CMD="./cluster-up/kubectl.sh" ./hack/dump-state.sh error
+
+
  
 help: ## Show this help screen
 	@echo 'Usage: make <OPTIONS> ... <TARGETS>'
--- a/Makefile
+++ b/Makefile
@@ -69,7 +69,7 @@ upgrade-test:
 	./hack/upgrade-test.sh
 
 dump-state:
-	CMD="./cluster-up/kubectl.sh" ./hack/dump-state.sh error
+	CMD="./cluster-up/kubectl.sh" ./hack/dump-state.sh 
 
 
  
--- a/hack/dump-state.sh
+++ b/hack/dump-state.sh
@@ -2,11 +2,31 @@
 
 
 function RunCmd {
-    cmd="$@"
+    cmd=$@
     echo "Command: $cmd"
-    $cmd
-    if [ "x$?" != "x0" ]; then
-        echo "Command failed: $cmd Status: $?"
+    echo ""
+    bash -c "$cmd"
+    stat=$?
+    if [ "$stat" != "0" ]; then
+        echo "Command failed: $cmd Status: $stat"
+    fi
+}
+
+function ShowOperatorSummary {
+
+    local kind=$1
+    local name=$2
+    local namespace=$3
+
+    echo ""
+    echo "Status of Operator object: kind=$kind name=$name"
+    echo ""
+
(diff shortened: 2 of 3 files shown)


hack/dump-state.sh
```diff
@@ -0,0 +1,64 @@
...
+
+========
+HCO Pods
+========
+
+EOF
+
+RunCmd "$CMD get pods -n kubevirt-hyperconverged -o json"
+
+cat <<EOF
```
rthallisey: I would also add a section for that HCO's status. You can use `oc get hco -n kubevirt-hyperconveged hyperconverged-cluster -o=jsonpath='{range .status.conditions[*]}{.type}{"\t"}{.status}{"\t"}{.message}{"\n"}{end}'` to get the conditions.

hack/dump-state.sh
```diff
@@ -0,0 +1,64 @@
...
+
+RunCmd "$CMD get deployments -n kubevirt-hyperconverged -o json"
+
+cat <<EOF
+========================
+HCO operator related CRD
+========================
+EOF
+
+RELATED_OBJECTS=`${CMD} get hyperconvergeds.hco.kubevirt.io hyperconverged-cluster -n kubevirt-hyperconverged -o go-template='{{range .status.relatedObjects }}{{if .namespace }}{{ printf "%s %s %s\n" .kind .name .namespace }}{{ else }}{{ printf "%s %s .\n" .kind .name }}{{ end }}{{ end }}'`
```
rthallisey: You may be able to use the command I mentioned above to get the component operators statuses too (haven't tried yet).

Makefile
```diff
@@ -67,6 +67,11 @@ build-push-all: container-build-operator container-push-operator container-build
 
 upgrade-test:
 	./hack/upgrade-test.sh
+
+dump-state:
+	CMD="./cluster-up/kubectl.sh" ./hack/dump-state.sh error
```
rthallisey: 👍 very useful script.  I think this script could be helpful outside of cluster-up, so we may want to avoid setting `CMD` here and have `upgrade-test.sh` set it.

Here's a place where it could also be used: https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/hack/deploy.sh#L118-L125

hack/upgrade-test.sh
```diff
@@ -43,6 +43,18 @@
 # to verify that it is updated to the new operator image from 
 # the local registry.
 
+function cleanup() {
+    rv=$?
+    if [ "x$rv" != "x0" ]; then
```
rthallisey: Why `"x$rv" != "x0"` and not `"$rv" != "0"`?  What's the purpose of the x character?

rthallisey:
Thanks for the PR @MoserMichael!  This will be a nice addition.

hack/upgrade-test.sh
MoserMichael: That used to be a trick to guard against leading spaces in string comparison. Probably not required when comparing an exit code though.

Makefile
MoserMichael: I tried this before in https://github.com/kubevirt/hyperconverged-cluster-operator/pull/314/commits but  hack/common.sh is setting up the wrong value for CMD (at least for the environment that does the CI testing in this environment)

hack/dump-state.sh
MoserMichael: ok

hack/dump-state.sh
MoserMichael: ok will add that.

MoserMichael:
Thank you! glad to help.

Makefile
rthallisey: ok we can table that addition for now.  Would like to come back to it later though

PR #316, opened 2019-10-03
Update to KubeVirt 0.20.8 with new image-prefix flag

slintes:
heads up, if you call virt-operator's csv-generator directly in its image, the flag name is `-imagePrefix`

/cc @tiraboschi


rthallisey:
ci test please

PR #318, opened 2019-10-07
Bug 1757784: Use predicates to prevent noisy reconcile

djzager:
We want to ignore updates to the metadata/status fields of the
HyperConverged resource as we will be the one making that change. This
is especially important as we populate the LastHeartbeatTime on the
resource's conditions...so without this change the operator will
constantly reconcile.


djzager:
ci test please

rthallisey:
the cherrypick will have to wait until 2.1 goes out.

PR #328, opened 2019-10-29
Fix vendor issue

arachmani:


rthallisey:
Should fix vendor issue in #313

PR #332, opened 2019-10-30
JSON bool can't be represented as a string

rthallisey:
JSON bools should not be in quotes, otherwise it will get
translated to 'true' or 'false', which will fail.


rthallisey:
ci test please

PR #333, opened 2019-10-31
build: Use component's csv-generator to build CSV

djzager:
Update the build-manifests, components, and csv-merger to support
grabbing component manifests from their built container images.


djzager:
Checklist:

- [ ] Clean up build-manifests so it doesn't just plop a bunch of intermediate files into the project's root directory
- [ ] Need to handle CRDs, putting them where they belong.
- [ ] Need to be able to populate the entire deploy directory not just deploy/olm-catalog

hack/build-manifests.sh
```diff
@@ -18,96 +18,190 @@ CONTAINER_PREFIX="${CONTAINER_PREFIX:-kubevirt}"
 CNA_CONTAINER_PREFIX="${CNA_CONTAINER_PREFIX:-quay.io/kubevirt}"
 IMAGE_PULL_POLICY="${IMAGE_PULL_POLICY:-IfNotPresent}"
 
-#CNV IMS Images
+# Component Images
+KUBEVIRT_IMAGE="${KUBEVIRT_IMAGE:-docker.io/kubevirt/virt-operator:v0.20.8}"
```
rthallisey: Can we query the respective repos in git for this value?

djzager: I think that's something we could pursue but I think we should create an issue/card for that as an improvement to make.

djzager: I don't know if the docker/quay apis have a good way of getting the latest released tag...plus I think we would want to have reproducible builds (not ones that change based on when you run it like in the case KubeVirt does a new release).

Basically, I think it requires enough discussion and thought that I believe it's worth tackling separately from this PR.

rwsu:
Having not worked on the manifest-templator before, it is hard to follow what this proposed change is doing. Can you add some documentation to describe how build_manifest.sh works in conjunction with manifest-templator?

djzager:
@rwsu comments added to both `build-manifests` and `manifest-templator`

djzager:
ci test please

hack/build-manifests.sh
```diff
@@ -1,113 +1,214 @@
...
+# a combined CSV and 2) other development related manifests (like the
+# operator deployment + rbac).
+#
+# Second, we pass all of the component CSVs off to the manifest-templator
+# that handles the deployment specs, service account names, permissions, and
+# clusterPermissions by converting them into their corresponding Kubernetes
+# manifests (ie. permissions + serviceAccountName = role + service account
+# + role binding) before writing them to disk.
+#
+# Lastly, we take give the component CSVs to the csv-merger that combines all
```
rwsu: Thanks for adding this David, very helpful. One question, why does build-manifest.sh create a _merger file now? But in 0.0.1/0.0.2/0.0.3 directories under deploy/olm-catalog/kubevirt-hyperverged there isn't one? Is there an extra step one needs to do to create a new version?

djzager: I left the `templates/` directory alone for this Pull Request to keep the size down but that should be removed in a follow-up.

Previously, the `templates/olm-catalog/kubevirt-hyperconverged/VERSION/kubevirt-hyperconverged-operator.VERSION.clusterserviceversion_merger.yaml.in` was taken and put into the container image (downstream) for use with merging the CSVs together.

Now, we are just putting all of the HCO related objects in source and building the CSV from the `components` package.

To create a new version, you would just change the `CSV_VERSION` (and probably also the `REPLACES_VERSION`) and run `hack/build-manifests.sh`. The directory should be created for you.

tiraboschi:
ci please test

djzager:
ci test please

tiraboschi:
ci test please

tiraboschi:
OVS issues are going to be solved on kubevirt/ovs-cni#94

djzager:
ci test please

PR #335, opened 2019-11-04
Add upgrade process documentation.

lveyde:

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,0 +1,71 @@
+### Kubevirt HCO upgrade procedure
+
+##### Summary
+
+In order to tell the OLM to start the upgrade process, one needs to point the relevant operator subscription to the updated CSV (ClusterServiceVersion).
+
+It may be as simple as updating the subscription's **channel**, inside the currently used **catalog**. Or, in other cases, it may require updating the **catalogsource** first, to point to the updated image, one that will contain a new catalog, with a CSV for this new channel.
+
+Once OLM will recognize that the subscription points to a new CSV, it will attempt to peform an update. By default this happens automatically, but this behaviour can be modified for each operator, by modifying the relevant **subscription**, in case one wants to manually approve each update.
+
+Steps 1 and 4 describe how to modify the subscription update policy to manual, and how to approve the update, once one becomes available. These steps are not required for the default, automatic update flow.
+
+##### 1. Modify the subscription update policy to manual (ONLY for manual approval flow)
+
+Using UI:
+Go to Operators -> Installed Operators
+Select kubevirt-hyperconverged project
+Click on the KubeVirt HyperConverged Cluster Operator
+Click on the Subscription
+Click on the Automatic under Approval
+Select Manual and click on the Save button
+
+Using CLI:
+	
+	# kubectl edit subs hco-subscription
+
+Add the following line under the spec: section:
+
+    installPlanApproval: Manual
+
+##### 2. Update the CatalogSource to point to the registry-bundle that contains a more recent version of KubeVirt HCO CSV (ONLY if 2 registry-bundle images are used)
+
+NOTE: This step of course is only required if 2 different registry-bundle images are used, one that contains the initial version and a second image that includes an updated version.
+
+	# kubectl edit catalogsources -n openshift-marketplace hco-catalogsource
+
+	Edit the image to point to the updated one (image: under spec: section).
+	I.e. to the docker.io/lveyde/hco-registry:CRMyRLkM
+
+OR
+
+    # kubectl patch catalogsource hco-catalogsource -n openshift-marketplace -p '{"spec":{"image": "docker.io/lveyde/hco-registry:CRMyRLkM"}}' --type merge
+
+##### 3. Update the Subscription’s channel
+	# kubectl edit subs hco-subscription
+
+	Edit the subscription to point to the new channel (channel: under spec: section).
+I.e. to the channel 0.0.2
+
+OR
+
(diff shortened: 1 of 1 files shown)


UPGRADE.md
```diff
@@ -0,0 +1,139 @@
+### Kubevirt HCO upgrade procedure
+
+Please note that the instructions were tested w/ OpenShift 4.2.
```
rthallisey: The instructions should work regardless of OCP release. I don't think you need to include this note.

UPGRADE.md
```diff
@@ -0,0 +1,139 @@
...
+Known issues related to the upgrade:
+
+[https://docs.google.com/document/d/1BM8NudsavT4Yh0Mvqv0D-vpqufYhWKAPradphbBB5KQ/edit?usp=sharing](https://docs.google.com/document/d/1BM8NudsavT4Yh0Mvqv0D-vpqufYhWKAPradphbBB5KQ/edit?usp=sharing "https://docs.google.com/document/d/1BM8NudsavT4Yh0Mvqv0D-vpqufYhWKAPradphbBB5KQ/edit?usp=sharing")
```
rthallisey: Can you list them out?  This is an internal doc.

UPGRADE.md
```diff
@@ -0,0 +1,139 @@
...
+
+
+###### 1. Start OpenShift 4.2 environment and make sure the KUBECONFIG is properly exported to point to the correct configuration files, so that kubectl command works properly.
```
rthallisey: The assumption is that the user has already installed a cluster and CNV.  Let's only mention the steps it takes to go from an deployed release to a new one.

UPGRADE.md
lveyde: The note was there to make sure nobody tries it w/ OpenShift version <4.2, i.e. 4.1.
But I agree it's mostly irrelevant now.
Removing.

UPGRADE.md
lveyde: I wonder if we may still want to share these known issues (current and future ones).
Removing the reference for now.

UPGRADE.md
lveyde: OK, will update the document.

lveyde: Updated. Should we remove the manual flow as well (leaving only automatic one)?

UPGRADE.md
```diff
@@ -0,0 +1,61 @@
+### Kubevirt HCO upgrade procedure
+
```
rthallisey: Let's create a section to describe what happens when you set upgrade to automatic.  Then, the next few sections will be for manual upgrade.

lveyde: Pushed an updated version.

UPGRADE.md
```diff
@@ -0,0 +1,71 @@
...
+Using the UI:
+Go to Operators -> Installed Operators
+Select kubevirt-hyperconverged project
+Click on the 3 dots and then click on the Edit Subscription from the popup menu
+Click on Subscriptions
+Click on the hco-subscription
+Once the system gets aware of the update it will show a link which says: 1 requires approval
+Click on it, then click on the Preview Install Plan button
+Click on the Approve button
```
rwsu: Is there a way to approve an install plan via CLI?

djzager: All you need to do is patch the install plan and set `.spec.approved = true`.

UPGRADE.md
```diff
@@ -0,0 +1,71 @@
+### Kubevirt HCO upgrade procedure
```
djzager: Go ahead and make it a proper heading.

```suggestion
# Kubevirt HCO Upgrade Procedure
```

UPGRADE.md
```diff
@@ -0,0 +1,71 @@
...
+
+##### Summary
```
djzager: nit, make it one level below 

```suggestion
## Summary
```

UPGRADE.md
djzager: I think we should reference [the OLM docs](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/how-to-update-operators.md) when talking about upgrade and avoid duplication where possible.

UPGRADE.md
```diff
@@ -0,0 +1,71 @@
...
+
+In order to tell the OLM to start the upgrade process, one needs to point the relevant operator subscription to the updated CSV (ClusterServiceVersion).
```
djzager:
```suggestion
In order to tell OLM to start the upgrade process, one needs to point the relevant operator **subscription** to an updated ClusterServiceVersion (CSV).
```

UPGRADE.md
```diff
@@ -0,0 +1,71 @@
...
+
+It may be as simple as updating the subscription's **channel**, inside the currently used **catalog**. Or, in other cases, it may require updating the **catalogsource** first, to point to the updated image, one that will contain a new catalog, with a CSV for this new channel.
```
djzager:
```suggestion
It may be as simple as updating the subscription's **channel**, inside the currently used **catalog**. In other cases, it may require updating the **catalogsource** first, to point to an updated image, one that will contain a new catalog, with a CSV for this new channel.
```

UPGRADE.md
```diff
@@ -0,0 +1,71 @@
...
+
+Once OLM will recognize that the subscription points to a new CSV, it will attempt to peform an update. By default this happens automatically, but this behaviour can be modified for each operator, by modifying the relevant **subscription**, in case one wants to manually approve each update.
```
djzager: This seems a bit off to me. My understanding was that OLM is constantly searching through the different operator catalogs looking for updates to operators. Once it finds an update, it starts going through the process of updating that operator.

PR #338, opened 2019-11-04
Update to KubeVirt 0.23.0

slintes:
Heads up, running `./hack/build-manifests.sh` removed the "KVM_EMULATION" env var from the hco operator in deploy/operator.yaml. Not sure if that is intentional?


deploy/operator.yaml
```diff
@@ -19,8 +19,6 @@ spec:
       - command:
         - hyperconverged-cluster-operator
         env:
-        - name: KVM_EMULATION
```
rthallisey: @arachmani it looks like this wasn't added to the CSV template.  Can you open a PR to add it?

rthallisey:
@slintes without it, the prow job will fail.  Thanks for pointing it out.

deploy/operator.yaml
rthallisey: https://github.com/kubevirt/hyperconverged-cluster-operator/pull/343

rthallisey:
#343 has been merged; can you rebase @slintes ?

slintes:
> #343 has been merged; can you rebase @slintes ?

done, thx for the heads up

PR #341, opened 2019-11-05
add missing permission for ssp operator

ksimon1:
Add missing permission for ssp operator
This permission is needed due to garbage collection (https://github.com/operator-framework/operator-sdk/blob/ee54f0cf84a845bca7a92ac86e6e8f062fe268ed/doc/user/metrics/README.md#garbage-collection)

Signed-off-by: Karel Šimon <ksimon@redhat.com>
/cc @fromanirh, @rthallisey

### DIFF
--- a/deploy/cluster_role.yaml
+++ b/deploy/cluster_role.yaml
@@ -908,6 +908,7 @@ rules:
   - apps
   resources:
   - deployments
+  - deployments/finalizers
   - replicasets
   - daemonsets
   - statefulsets
--- a/templates/cluster_role.yaml.in
+++ b/templates/cluster_role.yaml.in
@@ -54,6 +54,7 @@ rules:
   - apps
   resources:
   - deployments
+  - deployments/finalizers
   - replicasets
   - daemonsets
   - statefulsets
--- a/templates/olm-catalog/kubevirt-hyperconverged/VERSION/kubevirt-hyperconverged-operator.VERSION.clusterserviceversion.yaml.in
+++ b/templates/olm-catalog/kubevirt-hyperconverged/VERSION/kubevirt-hyperconverged-operator.VERSION.clusterserviceversion.yaml.in
@@ -267,6 +267,7 @@ spec:
             - apps
             resources:
             - deployments
+            - deployment/finalizers
             - replicasets
             - daemonsets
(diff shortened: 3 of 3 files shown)


rthallisey:
ci test please

rthallisey:
ci test please

fabiand:
@ksimon1 any comment on Ryan's change request?

ksimon1:
@rthallisey both files are updated, tests passed without error. Can it be merged?

deploy/cluster_role.yaml
```diff
@@ -908,6 +908,7 @@ rules:
   - apps
   resources:
   - deployments
+  - deployments/finalizers
```
rthallisey: This also needs to be changed in templates/olm-catalog/kubevirt-hyperconverged/VERSION/kubevirt-hyperconverged-operator.VERSION.clusterserviceversion.yaml.in.

rthallisey:
we'll get these ssp changes merged today.

deploy/cluster_role.yaml
MarSik: Ryan, that looks like a third place already (not counting the original project repository).. have you considered automating this? Even something simple like git precommit-hook or a manually executed script to refresh all the copies...

rthallisey: Back when we were building the CSV, we asked components to write the templates as code.  This feature didn't make it into SSP for 2.0, so we carry a copy of the CSV in tree.

The downstream automation is making it's way upstream in https://github.com/kubevirt/hyperconverged-cluster-operator/pull/333 and should replace the current csv generation.

fromanirh: It is possible that our generator code got out of date somehow. This is surprising, since it is supposed to use the same YAMLs we used as source. I think we need to check the flow on our side.

ksimon1:
@rthallisey i updated the last file you mentioned

deploy/cluster_role.yaml
MarSik: @rthallisey Yeah, I have no issue in having the copy in tree, but I was wondering why you have three of them..

rthallisey:
ci test please

PR #344, opened 2019-11-05
Fix OpenShift CI Upgrade Test

rwsu:
Should not source ./hack/common.sh as it defaults a value
for KUBEVIRT_PROVIDER, which breaks the upgrade-test in
OpenShift CI.

Export CMD so that "make dump-state" works correctly
in OpenShift CI.

"make cluster-clean" should only be called in STDCI.

Hard-coding of ./cluster-up/kubectl.sh in "make dump-state"
makes it incompatible with clusters that are not
started by cluster-up.

### DIFF
--- a/hack/upgrade-test.sh
+++ b/hack/upgrade-test.sh
@@ -46,25 +46,11 @@
 echo "KUBEVIRT_PROVIDER: $KUBEVIRT_PROVIDER"
 
 if [ -n "$KUBEVIRT_PROVIDER" ]; then
-  # okd-* provider, STDCI
-  REGISTRY_IMAGE="registry:5000/kubevirt/hco-registry"
-  REGISTRY_IMAGE_UPGRADE="registry:5000/kubevirt/hco-registry-upgrade"
-  REGISTRY_IMAGE_URL_PREFIX="registry:5000/kubevirt"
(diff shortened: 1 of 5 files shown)


hack/upgrade-test.sh
```diff
@@ -42,14 +42,15 @@
...
-if [ -n "${KUBEVIRT_PROVIDER}" ]; then
+echo "KUBEVIRT_PROVIDER: $KUBEVIRT_PROVIDER"
+
+if [ -n "$KUBEVIRT_PROVIDER" ]; then
   # okd-* provider, STDCI
   REGISTRY_IMAGE="registry:5000/kubevirt/hco-registry"
   REGISTRY_IMAGE_UPGRADE="registry:5000/kubevirt/hco-registry-upgrade"
   REGISTRY_IMAGE_URL_PREFIX="registry:5000/kubevirt"
-  CMD="./cluster-up/kubectl.sh"
+  export CMD="./cluster-up/kubectl.sh"
```
rthallisey: What do you think about putting the if else in a config file somewhere that is `source`'d by files that need it?

rwsu: Sure, let me revise this PR.

PR #352, opened 2019-11-11
Add minimal script to rotate certs for CNV

rmohr:
Add a secript which rotates certificates for:

 * kubemacpool
 * CDI
 * KubeVirt
 * SSP

The following test-scenarios are added:

* Check that kubevirt components serve new certs after the rotation
* Check that CDI components serve new certs after the rotation
* Check that kubemacpool components serve new certs after the rotation
* Check that SSP components serve new certs after the rotation
* Check that a VM can be created and started successfully after a rotation. This implicitly checks that
  - CDI validationwebhookcs and mutatingwebhooks are still working with the new certificates
  - kubevirt validationwebhooks and mutatingwebhooks are still working with the new certificate
  - kubemacpool mutatingwebhooks are still working
  - SSP validationwebhooks are still working

The following is untested:

* CDI upload route uses re-encrypt. The rotation script deletes it, but no test scenario checks if it got updated.

As documentation on how to use the rotation script, an example `Dockerfile` and an example `CronJob` are included in the PR.


rmohr:
@MarSik I think we also need to rotate secrets for SSP, right?

tools/rotate-certs.sh
```diff
@@ -0,0 +1,39 @@
...
+    esac
+    shift
+done
+shift $((OPTIND-1))
+
+echo "# Rotating kubemacpool certificates ..."
+${_kubectl} --namespace "${namespace}" delete pods -l app==kubemacpool
+
+echo "# Rotating cdi certificates ..."
+${_kubectl} delete secrets --namespace "${namespace}" -l cdi.kubevirt.io
```
rthallisey: @djzager cc.  PR https://github.com/kubevirt/hyperconverged-cluster-operator/pull/333 makes changes to the labels.  Should they change?

rmohr:
> Is this the kind of thing that needs to be packaged into the container image?
> Should I be able to run this script against an existing deployment? How would I verify it didn't blow up?

I am in the process of adding a test. It is supposed to be run as a CronJob (yaml will follow too).

tools/rotate-certs.sh
djzager: We don't create any secrets so they shouldn't be impacted but I'm looking into the impact of the label changes now.

rmohr: Could you clarify what you mean by "changes to the labels"? I don't think that you can change these labels, since that would mean that you change labels on objects which are created by the operators managed by hco.

rthallisey: @djzager my point in mentioning it is you're changing the labels, but there's an expectation to still use the "old" labels e.g. `kubevirt.io = ""`.  So while it may not cause a problem in this script, it could cause a problem for us somewhere else.   I want to make sure we cover our bases.

@rmohr here's an example: https://github.com/kubevirt/hyperconverged-cluster-operator/pull/333/files#diff-412c4fc972d3983c4e02e20a248c6421L15 .

rmohr: Looks like this changes the labels of operator-level objects. Let me check if that may be an issue for SSP.

djzager: When these operators are deployed using OLM. You don't have the luxury of setting those labels.

All of the labels I see that are impacted in #333 are ones that you can't set when using OLM anyway. I'm looking for one where this is not the case.

djzager: Everything that we can get from a component operator's CSV should be safe (ie. https://github.com/kubevirt/hyperconverged-cluster-operator/blob/fcb0bbc05213c6967fc30228828e06c41721a19e/deploy/olm-catalog/kubevirt-hyperconverged/0.0.3/kubevirt-hyperconverged-operator.v0.0.3.clusterserviceversion.yaml#L1444-L1460 )

The things that we can get from each component's CSV include the Deployment Spec, (Cluster)Rules, and Service Account Names so the fact that we don't have a `"kubevirt.io: ""` label on the `cdi-operator` service account should not be a breaking change.

Hope that helps.

tools/rotate-certs.sh
```diff
@@ -37,3 +37,7 @@ ${_kubectl} delete pods --namespace "${namespace}" -l cdi.kubevirt.io
 echo "# Rotating kubevirt certificates ..."
 ${_kubectl} delete secrets --namespace "${namespace}" -l kubevirt.io
 ${_kubectl} delete pods --namespace "${namespace}" -l kubevirt.io
+
+echo "# Rotating SSP certificates ..."
+${_kubectl} delete secrets --ignore-not-found=true --namespace "${namespace}" virt-template-validator-certs
+${_kubectl} delete pods --namespace "${namespace}" -l kubevirt.io=virt-template-validator
```
MarSik: I am not sure the validator regenerates them on first start, better check that with @fromanirh

rmohr:
@tiraboschi could you have a look?

rmohr:
Hm, the openshift CI jobs are failing with:
```
could not wait for build: the build hco-registry-upgrade failed after 3m33s with reason DockerBuildFailed: Docker build strategy has failed.

++ CSV_VERSION=100.0.0
++ ./hack/build-manifests.sh
Gopkg.lock is out of sync with imports and/or Gopkg.toml. Run `dep check` for details.
input-digest mismatch
error: build error: running 'export UPGRADE_VERSION=${UPGR...hack/upgrade-create-new-version.sh' failed with exit code 1
```

but `dep status` shows that everyting is fine. Probably need a `dep` version which has the `check` command.

tests/func-tests/certificates_test.go
```diff
@@ -0,0 +1,376 @@
+package tests_test
```
djzager: :+1:

tools/README.md
```diff
@@ -0,0 +1,9 @@
+# Tools
+
+## Rotating Certificates
+
+Run
+
+```
+
```
djzager: should their be code here?

rmohr: Ha.

rmohr:
> Question, outside of the scope of this PR, but I wonder if the certificate rotation is something we could handle inside the reconcile loop of the HCO. Probably not managing a CronJob, but maybe we could on some interval rotate the certificates by just launching a Pod and verifying that everything went well (and including it in our status). Thoughts?

Hm, the main purpose here is to offer people the script. I think that the CronJob is one example on how it can be automated. I think that  we should asap change our operators (kubevirt, cdi, ...) to natively handle cert rotation without downtime. What do you think?

djzager:
> Hm, the main purpose here is to offer people the script. I think that the CronJob is one example on how it can be automated. I think that we should asap change our operators (kubevirt, cdi, ...) to natively handle cert rotation without downtime. What do you think?

Seems reasonable to me. As I understand it then, this PR, is to provide a means for consumers to rotate certificates if they need to with the long-term solution being each of the operators (kubevirt, cdi, ...) rotating certificates periodically on their own. If that is correct, :+1:

tools/README.md
rmohr: Fixed.

rmohr:
@tiraboschi @djzager I had to refactor the tests, to make them suite my needs. Would be great if you can do another review.

rmohr:
ci test please

rmohr:
Ci test please

rmohr:
Ci test please

rmohr:
Ci test please

rmohr:
@djzager, is the 4.1 lane normally reliable? I only even managed once so far to see HCO reaching the deploy phase.

rmohr:
hm, tests succeeded on 4.1 but we ran into a timeou: https://kubevirt-master-stdci-production.apps.ovirt.org/job/kubevirt_hyperconverged-cluster-operator_standard-check-pr/922/artifact/ci_build_summary.html

tiraboschi:
ci test please

rmohr:
ci test please

djzager:
> is the 4.1 lane normally reliable? I only even managed once so far to see HCO reaching the deploy phase.

I'm not certain. I know that a lot of tests have been added to ci over the past few weeks but I think @rwsu and @MoserMichael may know more about the timeout issue you saw.

In general it's looking more and more difficult to get a clean run through CI.

rthallisey:
ci test please

rthallisey:
Looks like the cert rotation test is failing

```
Certificates
[90m/go/src/github.com/kubevirt/hyperconverged-cluster-operator/tests/func-tests/certificates_test.go:28[0m
  [91m[1mshould rotate kubemacpool certificates [It][0m
  [90m/go/src/github.com/kubevirt/hyperconverged-cluster-operator/tests/func-tests/certificates_test.go:42[0m

  [91mExpected success, but got an error:
      <*exec.ExitError | 0xc00000c5c0>: {
          ProcessState: {
              pid: 105159,
              status: 256,
              rusage: {
                  Utime: {Sec: 0, Usec: 492002},
                  Stime: {Sec: 0, Usec: 98214},
                  Maxrss: 63664,
                  Ixrss: 0,
                  Idrss: 0,
                  Isrss: 0,
                  Minflt: 19192,
                  Majflt: 1,
                  Nswap: 0,
                  Inblock: 8,
                  Oublock: 0,
                  Msgsnd: 0,
                  Msgrcv: 0,
                  Nsignals: 0,
                  Nvcsw: 2584,
                  Nivcsw: 82,
              },
          },
          Stderr: nil,
      }
      exit status 1[0m

  /go/src/github.com/kubevirt/hyperconverged-cluster-operator/tests/func-tests/certificates_test.go:49
```

rmohr:
> Looks like the cert rotation test is failing

There fail always different ones. Most of the time after the first or second test, pods don't become ready anymore ...

rmohr:
also note  that openshift ci did not fail a single time so far.

rmohr:
ci test please

tools/rotate-certs.sh
```diff
@@ -0,0 +1,55 @@
...
+${_kubectl} delete pods --namespace "${namespace}" -l cdi.kubevirt.io
+# then delete registrations
+${_kubectl} delete validatingwebhookconfigurations --ignore-not-found=true  --namespace "${namespace}" cdi-api-datavolume-validate
+${_kubectl} delete mutatingwebhookconfigurations --ignore-not-found=true  --namespace "${namespace}" cdi-api-datavolume-mutate
+if [[ "$(${_kubectl} api-resources | grep routes)" =~ "routes" ]]
+then 
+    ${_kubectl} delete routes --ignore-not-found=true --namespace "${namespace}" cdi-uploadproxy
+fi
+# finally restart again, so that all registrations get recreated
+${_kubectl} delete pods --namespace "${namespace}" -l cdi.kubevirt.io
```
mhenriks: do this for all namespaces not just the namespace CDI is installed in.  this will get all import/upload/clone pods in progress

rmohr: Done. Also made the route-detection simpler and more robust. If any add-on is not discoverable, `kubectl get api-resources` just returns an error instead of the discovered list.

tests/func-tests/certificates_test.go
```diff
@@ -0,0 +1,349 @@
...
+		Expect(RotateCeritifcates(testscore.KubeVirtInstallNamespace)).To(Succeed())
+		By("waiting for all pods to become ready again")
+		WaitForPodsToBecomeReady(testscore.KubeVirtInstallNamespace)
+
+		By("getting the ceritifcate again after doing the rotation")
+		newCert, err := GetCertForService("kubemacpool-service", testscore.KubeVirtInstallNamespace, "443")
+		Expect(newCert).ToNot(BeEmpty())
+
+		By("verifying that the ceritificate indeed changed")
+		Expect(newCert).ToNot(Equal(oldCert))
```
aglitke: Should we check the expiration date?

tools/README.md
```diff
@@ -0,0 +1,20 @@
...
+To rotate certificates for all CNV components, type:
+
+```
+export KUBECONFIG=<wherever>
+export KUBECTL_BINARY=kubectl
+./rotate-certs.sh -n kubevirt-hyperconverged
+```
+
+During the rotation, the following things may occure:
```
aglitke: occur

tools/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+ * Migrations will be cancelled
+ * Image uploads will be cancelled
+ * VNC and Console connections will be closed
+
+After the rotation is done, all opperations will continue as usual.
+VirtualMachine and VirtualMachineInstance workloads will not be affected.
```
aglitke: CDI import and clone operations will also be unaffected.

tiraboschi:
ci test please

rmohr:
@tiraboschi should we just disable the 4.1 lane? We have openshift CI covering openshift anyway ...

Alternatively I can skip my tests for this lane.

tiraboschi:
> @tiraboschi should we just disable the 4.1 lane? We have openshift CI covering openshift anyway 

Done, please rebase.

tools/README.md
rmohr: Done.

tools/README.md
rmohr: Are you sure? @mhenriks asked me to delete all cdi pods in all namespaces.

rmohr:
@tiraboschi does it look good now?

tiraboschi:
Yes, thanks.
Waiting only for @phoracek approval

PR #353, opened 2019-11-12
Enable kvm emulation in ci

rwsu:


deploy/Dockerfile.registry.upgrade
```diff
@@ -13,6 +13,9 @@ ARG UPGRADE_VERSION=100.0.0
 COPY --from=builder /go/src/github.com/kubevirt/hyperconverged-cluster-operator/deploy/olm-catalog /registry
 USER root
 
+# enable KVM_EMULATION for CI, needed by kubevirt-node-labeller on AWS
+RUN find /registry/kubevirt-hyperconverged/ -type f -exec sed -i 's|##-enable-only-in-CI-##|"true"|' {} \; || :
```
rthallisey: Can you use the sed from `hack/deploy.sh` instead? (https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/hack/deploy.sh#L106)  That way, you won't need to change the CSV.

rwsu: Yes, done.

rthallisey:
ci test please

PR #364, opened 2019-12-04
add test for subscription channel

MoserMichael:
Signed-off-by: MichaelMoser <mmoser@localhost.localdomain>

### DIFF
--- a/hack/upgrade-test.sh
+++ b/hack/upgrade-test.sh
@@ -57,7 +57,13 @@ function Msg {
     set -x
 }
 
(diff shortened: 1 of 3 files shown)


hack/upgrade-test.sh
```diff
@@ -57,7 +57,13 @@ function Msg {
     set -x
 }
 
-echo "KUBEVIRT_PROVIDER: $KUBEVIRT_PROVIDER"
+#UPGRADE_METHOD=${UPGRADE_METHOD:-catalog_source}
```
rthallisey: is this comment needed?

PR #384, opened 2019-12-18
Update okd-4.1 image hash

tiraboschi:


tiraboschi:
okd-4.1 is disabled on master

PR #389, opened 2019-12-23
Add cluster-monitoring label to the deployment namespace

danielBelenky:
The openshift monitoring-operator is looking for rules
under namespaces that are labeled with the cluster-monitoring
label. Since the components that we deploy require this label,
we should set this label on our namespace.

* When deployed on Kubernetes, this label will have no effect.

Signed-off-by: Daniel Belenky <dbelenky@redhat.com>

### DIFF
--- a/deploy/hco.yaml
+++ b/deploy/hco.yaml
@@ -2,6 +2,8 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: kubevirt-hyperconverged
+  labels:
+    openshift.io/cluster-monitoring: "true"
 
 ---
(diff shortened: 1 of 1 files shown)


danielBelenky:
/hold

Probably going to solve it in `virt-operator`
https://github.com/kubevirt/kubevirt/pull/2952

rthallisey:
Is there a way to create this from the UI?  This yaml file is not used in production deployments.  Maybe this also needs a doc change?

tiraboschi:
Solved on virt-operator side with: https://github.com/kubevirt/kubevirt/pull/2952
We can close this one.

PR #380, opened 2019-12-17
WIP: Create HyperConverged resource on startup

djzager:
- Lock down the name and namespace of the HyperConverged resource
- Create the HyperConverged resource on startup
- Ignore requests of a different name and namespace than configured on Deployment
- Map all events on our secondary watches to requests on our HyperConverged resource

PR #393, opened 2019-12-31
Don't create problematic ownerReferences for CDI and NetworkAddonsConfig

maya-r:
if HCO operator (scope namespace) owns CDI (scope cluster) and
the kubernetes garbage collector finds out, it will delete CDI.

In an update, if an ownerReference exists, remove it.

### DIFF
--- a/pkg/controller/hyperconverged/hyperconverged_controller.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller.go
@@ -211,6 +211,7 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
 			for _, f := range []func(c client.Client, cr *hcov1alpha1.HyperConverged) error{
 				ensureCDIDeleted,
 				ensureNetworkAddonsDeleted,
+				ensureKubeVirtCommonTemplateBundleDeleted,
 			} {
 				err = f(r.client, instance)
 				if err != nil {
@@ -837,9 +838,6 @@ func newKubeVirtCommonTemplateBundleForCR(cr *hcov1alpha1.HyperConverged, namesp
 
 func (r *ReconcileHyperConverged) ensureKubeVirtCommonTemplateBundle(instance *hcov1alpha1.HyperConverged, logger logr.Logger, request reconcile.Request) error {
 	kvCTB := newKubeVirtCommonTemplateBundleForCR(instance, OpenshiftNamespace)
-	if err := controllerutil.SetControllerReference(instance, kvCTB, r.scheme); err != nil {
-		return err
-	}
 
 	key, err := client.ObjectKeyFromObject(kvCTB)
 	if err != nil {
@@ -855,6 +853,20 @@ func (r *ReconcileHyperConverged) ensureKubeVirtCommonTemplateBundle(instance *h
 
 	if err != nil {
 		return err
+	}
+
+	existingOwners := found.GetOwnerReferences()
+
+	// Previous versions used to have HCO-operator (namespace: kubevirt-hyperconverged)
+	// as the owner of kvCTB (namespace: OpenshiftNamespace).
+	// It's not legal, so remove that.
+	if (len(existingOwners) > 0) {
+		logger.Info("kvCTB has owners, removing...")
+		found.SetOwnerReferences([]metav1.OwnerReference{})
+		err = r.client.Update(context.TODO(), found)
+		if err != nil {
+			logger.Error(err, "Failed to remove kvCTB's previous owners")
+		}
 	}
 
 	logger.Info("KubeVirt Common Templates Bundle already exists", "bundle.Namespace", found.Namespace, "bundle.Name", found.Name)
@@ -1254,6 +1266,25 @@ func ensureNetworkAddonsDeleted(c client.Client, instance *hcov1alpha1.HyperConv
 
 	if err != nil {
 		log.Error(err, "Failed to get NetworkAddonsConfig from kubernetes")
+		return err
+	}
+
+	return componentResourceRemoval(found, c, instance)
+}
+
+func ensureKubeVirtCommonTemplateBundleDeleted(c client.Client, instance *hcov1alpha1.HyperConverged) (error) {
+	kvCTB := newKubeVirtCommonTemplateBundleForCR(instance, OpenshiftNamespace)
+
+	key, err := client.ObjectKeyFromObject(kvCTB)
+	if err != nil {
+		log.Error(err, "Failed to get object key for KubeVirt Common Templates Bundle")
+	}
+
+	found := &sspv1.KubevirtCommonTemplatesBundle{}
+	err = c.Get(context.TODO(), key, found)
+
+	if err != nil {
+		log.Error(err, "Failed to get KubeVirt Common Templates Bundle from kubernetes")
 		return err
 	}
 
--- a/pkg/controller/hyperconverged/hyperconverged_controller.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller.go
@@ -8,9 +8,13 @@ import (
 	"reflect"
 	"strings"
 
+	"encoding/json"
 	"github.com/go-logr/logr"
 	"github.com/operator-framework/operator-sdk/pkg/ready"
+	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/reference"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
@@ -41,9 +45,7 @@ const (
 	// use finalizers to manage the cleanup.
 	FinalizerName = "hyperconvergeds.hco.kubevirt.io"
 
-	// Foreground deletion finalizer is blocking removal of HyperConverged until explicitly dropped.
-	// TODO: Research whether there is a better way.
-	foregroundDeletionFinalizer = "foregroundDeletion"
+	HyperConvergedName = "hyperconverged-cluster"
 
 	// UndefinedNamespace is for cluster scoped resources
 	UndefinedNamespace string = ""
@@ -83,6 +85,11 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
+	hco, err := getHyperconverged()
+	if err != nil {
+		return err
+	}
+
 	// Watch secondary resources
 	for _, resource := range []runtime.Object{
 		&kubevirtv1.KubeVirt{},
@@ -93,9 +100,14 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		&sspv1.KubevirtTemplateValidator{},
 		&sspv1.KubevirtMetricsAggregation{},
 	} {
-		err = c.Watch(&source.Kind{Type: resource}, &handler.EnqueueRequestForOwner{
-			IsController: true,
-			OwnerType:    &hcov1alpha1.HyperConverged{},
+		err = c.Watch(&source.Kind{Type: resource}, &handler.EnqueueRequestsFromMapFunc{
+			ToRequests: handler.ToRequestsFunc(
+				// always enqueue the same HyperConverged object, since there should be only one
+				func(a handler.MapObject) []reconcile.Request{
+					return []reconcile.Request{
+						{NamespacedName: hco},
+					}
+				}),
 		})
 		if err != nil {
 			return err
@@ -186,6 +198,35 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
 	// in-memory conditions should start off empty. It will only ever hold
 	// negative conditions (!Available, Degraded, Progressing)
 	r.conditions = nil
+
+	// Handle finalizers
+	if instance.ObjectMeta.DeletionTimestamp.IsZero() {
+		// Add the finalizer if it's not there
+		if !contains(instance.ObjectMeta.Finalizers, FinalizerName) {
+			instance.ObjectMeta.Finalizers = append(instance.ObjectMeta.Finalizers, FinalizerName)
+			err = r.client.Update(context.TODO(), instance)
+		}
+	} else {
+		if contains(instance.ObjectMeta.Finalizers, FinalizerName) {
+			for _, f := range []func(c client.Client, cr *hcov1alpha1.HyperConverged) error{
+				ensureCDIDeleted,
+				ensureNetworkAddonsDeleted,
+			} {
+				err = f(r.client, instance)
+				if err != nil {
+					reqLogger.Error(err, "Failed to manually delete objects")
+					return reconcile.Result{}, err
+				}
+			}
+
+			// Remove the finalizer
+			instance.ObjectMeta.Finalizers = drop(instance.ObjectMeta.Finalizers, FinalizerName)
+			err = r.client.Update(context.TODO(), instance)
+
+			// Need to requeue because finalizer update does not change metadata.generation
+			return reconcile.Result{Requeue: true}, err
+		}
+	}
 
 	for _, f := range []func(*hcov1alpha1.HyperConverged, logr.Logger, reconcile.Request) error{
 		r.ensureKubeVirtConfig,
@@ -470,9 +511,6 @@ func newCDIForCR(cr *hcov1alpha1.HyperConverged, namespace string) *cdiv1alpha1.
 
 func (r *ReconcileHyperConverged) ensureCDI(instance *hcov1alpha1.HyperConverged, logger logr.Logger, request reconcile.Request) error {
 	cdi := newCDIForCR(instance, UndefinedNamespace)
-	if err := controllerutil.SetControllerReference(instance, cdi, r.scheme); err != nil {
-		return err
-	}
 
 	key, err := client.ObjectKeyFromObject(cdi)
 	if err != nil {
@@ -481,6 +519,7 @@ func (r *ReconcileHyperConverged) ensureCDI(instance *hcov1alpha1.HyperConverged
 
 	found := &cdiv1alpha1.CDI{}
 	err = r.client.Get(context.TODO(), key, found)
+
 	if err != nil && apierrors.IsNotFound(err) {
 		logger.Info("Creating CDI")
 		return r.client.Create(context.TODO(), cdi)
@@ -491,6 +530,20 @@ func (r *ReconcileHyperConverged) ensureCDI(instance *hcov1alpha1.HyperConverged
 	}
 
 	logger.Info("CDI already exists", "CDI.Namespace", found.Namespace, "CDI.Name", found.Name)
+
+	existingOwners := found.GetOwnerReferences()
+
+	// Previous versions used to have HCO-operator (scope namespace)
+	// as the owner of CDI (scope cluster).
+	// It's not legal, so remove that.
+	if (len(existingOwners) > 0) {
+		logger.Info("CDI has owners, removing...")
+		found.SetOwnerReferences([]metav1.OwnerReference{})
+		err = r.client.Update(context.TODO(), found)
+		if err != nil {
+			logger.Error(err, "Failed to remove CDI's previous owners")
+		}
+	}
 
 	// Add it to the list of RelatedObjects if found
 	objectRef, err := reference.GetReference(r.scheme, found)
@@ -590,9 +643,6 @@ func newNetworkAddonsForCR(cr *hcov1alpha1.HyperConverged, namespace string) *ne
 
 func (r *ReconcileHyperConverged) ensureNetworkAddons(instance *hcov1alpha1.HyperConverged, logger logr.Logger, request reconcile.Request) error {
 	networkAddons := newNetworkAddonsForCR(instance, UndefinedNamespace)
-	if err := controllerutil.SetControllerReference(instance, networkAddons, r.scheme); err != nil {
-		return err
-	}
 
 	key, err := client.ObjectKeyFromObject(networkAddons)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


AsherShoshan:
HCO Operator creates the CDI cr, and puts HCO cr as the owner. so the new lines of code is redundant; only the removal of the setownerreference is required.  AFAIK, no one else puts additional owner reference in the CDI kind cr.
But then , when you delete the HCO cr (undeploying CNV), CDI will not be undeployed..

dankenigsberg:
@AsherShoshan the added code is not redundant - it is needed when upgrading from an older cluster.

You are right about this patch not handing the deletion use case - it should.

AlonaKaplan:
> @AsherShoshan the added code is not redundant - it is needed when upgrading from an older cluster.
> 
> You are right about this patch not handing the deletion use case - it should.

Why do we need to add a special caring for the deletion use case?
Is there a use case for removing the hco cr directly and not due to unregistration from the olm?
If not, I think we don't need the finalizer. When unregistering from the olm, the cdi cr will be removed anyway when the olm will remove the cdi crd.

dankenigsberg:
> > You are right about this patch not handing the deletion use case - it should.
> 
> Why do we need to add a special caring for the deletion use case?
> Is there a use case for removing the hco cr directly and not due to unregistration from the olm?
> If not, I think we don't need the finalizer. When unregistering from the olm, the cdi cr will be removed anyway when the olm will remove the cdi crd.

@AlonaKaplan good point. When the cdis.cdi.kubevirt.io is removed, the CDI object would be removed with it even if the latter has no owner.

AsherShoshan:
Alona,
There is a UI use case of deleting+redeploying the HCO cr, without removing
all OLM operators
Asher

AlonaKaplan:
> Alona, There is a UI use case of deleting+redeploying the HCO cr, without removing all OLM operators Asher

@AsherShoshan can you please elaborate? Do you mean there is an option from the ui to remove the hco cr? Why would the user want to do it? Maybe we should block it.

AsherShoshan:
Yes.
After operator is installed (vis operator hub page).
You can remove/create hco.
More than that - in UI you can explicitly create each cr (kubevirt, cdi,
etc).. without being owned by HCO cr.

dankenigsberg:
Sorry @AsherShoshan, I do not get your point. How is editing various CRs related to our discussion?

AsherShoshan:
Dan,
I didn't write editing...
In UI, you can delete the kind-HCO cr (without removing OLM operators), and
recreate it.
Also you can, for example, create kind-CDI cr - without going via kind-HCO
cr.

dankenigsberg:
@AsherShoshan what you are saying is not clear to me. What do you expect to happen when the kind-hco rc is deleted? Does it happen now? Would it happen with this PR?

AsherShoshan:
On Wed, Jan 1, 2020 at 10:18 PM Dan Kenigsberg <notifications@github.com>
wrote:

> @AsherShoshan <https://github.com/AsherShoshan> what you are saying is
> not clear to me. What do you expect to happen when the kind-hco rc is
> deleted? Does it happen now? Would it happen with this PR?
>
I replied Alona regarding a use case of HCO cr deletion without HCO CRD
deletion.
In UI, you can delete only the HCO-kind cr, then all dependent cr are
deleted, but OLM operators are not deleted.
Also, in UI, when only OLM operators installed (i.e kubevirt-hyperconverged
operator installed via OperatorHub page), you can create each cr (kubevirt,
cdi, etc), without creating the HCO cr.  It means kubevirt-kind has no
owner.

IIUC, after merging the PR, if you delete the HCO cr via UI (without OLM
removal), dependant cr's will not be deleted, and VM, VMI will not be
deleted.
with the second use case, with the fix of Omer, (to put HCO CRD as owner of
VM,VMI CRD), deleting kubevirt-kind cr will not delete VM, VMI either.

dankenigsberg:
Yes, this pr changes the behavior if kind-hco is deleted (either by ui or
command line, it does not matter).

However, I don't think that deleting kind-hco cr is a valid use case. It
was not an acceptable method to cleanly remove hco before this pr, and
neither after.

rthallisey:
I don't think it's a good idea to change the documentation behavior of cleanup without much notice.  There's been some discussion of changing the behavior in the past, but there should be a formal agreement from all (PM & ENG) and a controlled rollout, with plenty of time for testing, of the change.  In the near term, the underlying problem of garbage collection should be the priority and can be fixed without changing the cleanup behavior.   See https://github.com/kubevirt/hyperconverged-cluster-operator/pull/127/files#diff-64e1e2b9251fa5137c91d5a7c8427b88R35.  Long term, it would be worth discussing cleanup.

dankenigsberg:
@rthallisey I am not aware of any change to any documented behavior incurred by this PR. Could you point me to any?

maya-r:
Asking for clarification, there are two kinds of cleanup:
- Cleanup on deletion, discussed in the replies, not done in the PR
- Cleanup on update, which I've attempted in the PR

Would you prefer not to cleanup on update yet, too?

rthallisey:
>Asking for clarification, there are two kinds of cleanup:
>
>    Cleanup on deletion, discussed in the replies, not done in the PR
>
Cleanup of CDI's CR when HCO's CR is deleted will no longer happen after this PR is merged.  I'm saying that this changes the expectation for cleanup and that I don't think that's the right approach.  IMO there should be a discussion before changing how things are cleaned up, so we don't end up bundling a feature together with a bug fix.

I think we should look if it's still possible to solve with finalizes as @phoracek did here ( https://github.com/kubevirt/hyperconverged-cluster-operator/pull/127/files#diff-64e1e2b9251fa5137c91d5a7c8427b88R35)

Docs: https://docs.openshift.com/container-platform/4.2/cnv/cnv_install/uninstalling-container-native-virtualization.html 

>    Cleanup on update, which I've attempted in the PR
>
>Would you prefer not to cleanup on update yet, too?
>

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -492,6 +489,18 @@ func (r *ReconcileHyperConverged) ensureCDI(instance *hcov1alpha1.HyperConverged
...
+	existingOwners := found.GetOwnerReferences()
+
+	// Previous versions used to have HCO-operator (scope namespace)
+	// as the owner of CDI (scope cluster).
+	// It's not legal, so remove that.
+	if (len(existingOwners) > 0) {
+		logger.Info("CDI has owners, removing...")
+		noOwnerRefs := []metav1.OwnerReference{}
+		found.SetOwnerReferences(noOwnerRefs)
+		r.client.Update(context.TODO(), found)
```
djzager: Please handle the error.

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -492,6 +489,18 @@ func (r *ReconcileHyperConverged) ensureCDI(instance *hcov1alpha1.HyperConverged
...
 	logger.Info("CDI already exists", "CDI.Namespace", found.Namespace, "CDI.Name", found.Name)
 
+	existingOwners := found.GetOwnerReferences()
+
+	// Previous versions used to have HCO-operator (scope namespace)
+	// as the owner of CDI (scope cluster).
+	// It's not legal, so remove that.
+	if (len(existingOwners) > 0) {
+		logger.Info("CDI has owners, removing...")
+		noOwnerRefs := []metav1.OwnerReference{}
```
djzager:
```suggestion
```

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -492,6 +489,18 @@ func (r *ReconcileHyperConverged) ensureCDI(instance *hcov1alpha1.HyperConverged
...
+		found.SetOwnerReferences(noOwnerRefs)
```
djzager:
```suggestion
		found.SetOwnerReferences([]metav1.OwnerReference{})
```

pkg/controller/hyperconverged/hyperconverged_controller.go
djzager:
```suggestion
		return r.client.Update(context.TODO(), found)
```

Seems sufficient to me.

dankenigsberg:
@rthallisey https://docs.openshift.com/container-platform/4.2/cnv/cnv_install/uninstalling-container-native-virtualization.html#cnv-deleting-kubevirt-hyperconverged-custom-resource_uninstalling-container-native-virtualization says that one one remove the hco.cr, but I am not completely sure why. Wouldn't the next step https://docs.openshift.com/container-platform/4.2/cnv/cnv_install/uninstalling-container-native-virtualization.html#cnv-deleting-hco-subscription_uninstalling-container-native-virtualization trigger the removal of cnv's CRDs, and by that the removal of cdi.cr ?

However, since HCO is the one creating cdi.cr, it makes sense that HCO would also be the one to delete it. @maya-r would you add a finalizer to remove cdi.cr on hco.cr deletion?

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -470,9 +470,6 @@ func newCDIForCR(cr *hcov1alpha1.HyperConverged, namespace string) *cdiv1alpha1.
 
 func (r *ReconcileHyperConverged) ensureCDI(instance *hcov1alpha1.HyperConverged, logger logr.Logger, request reconcile.Request) error {
 	cdi := newCDIForCR(instance, UndefinedNamespace)
-	if err := controllerutil.SetControllerReference(instance, cdi, r.scheme); err != nil {
```
dankenigsberg: line 89 in this file depends on CDI being owned by the hco.cr. With the suggested code, HCO would not be notified if someone edits CDI.cr. For completeness, please make this controller watch the cdi.cr regardless of its owner.

maya-r:
Merged chunks of #380 and #122

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -601,6 +675,18 @@ func (r *ReconcileHyperConverged) ensureNetworkAddons(instance *hcov1alpha1.Hype
 
 	found := &networkaddonsv1alpha1.NetworkAddonsConfig{}
 	err = r.client.Get(context.TODO(), key, found)
+
+	// Handle HCO finalizer
+	if !instance.ObjectMeta.DeletionTimestamp.IsZero() {
```
djzager: Should go away and be centrally located in the main `Reconcile()`

coypoop:
Is this roughly what you had in mind? https://github.com/kubevirt/hyperconverged-cluster-operator/compare/master...maya-r:cdi-owner6 (having some issues with setting it up while at home, so not tested).

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -601,6 +675,18 @@ func (r *ReconcileHyperConverged) ensureNetworkAddons(instance *hcov1alpha1.Hype
...
+		errRemoval := manageComponentResourceRemoval(found, r.client, instance)
+		if errRemoval != nil {
+			logger.Error(err, "Failed during CDI removal")
```
dankenigsberg: wrong error string?

maya-r:
This resembles the existing reconcile logic, hopefully this is your intention.
In the long run I feel compelled to have a "getCDI" that gets the actual CDI instance.

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -601,13 +649,28 @@ func (r *ReconcileHyperConverged) ensureNetworkAddons(instance *hcov1alpha1.Hype
...
 		return err
 	}
 
+	existingOwners := found.GetOwnerReferences()
+
+	// Previous versions used to have HCO-operator (scope namespace)
+	// as the owner of NetworkAddons (scope cluster).
+	// It's not legal, so remove that.
+	if (len(existingOwners) > 0) {
+		logger.Info("NetworkAddons has owners, removing...")
```
ormergi: consider clearer log message for example: Removing NetworkAddonsConfigs owner references

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -187,6 +199,33 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
...
+	// Handle finalizers
+	if instance.ObjectMeta.DeletionTimestamp.IsZero() {
+		// Add the finalizer if it's not there
+		if !contains(instance.ObjectMeta.Finalizers, FinalizerName) {
+			instance.ObjectMeta.Finalizers = append(instance.ObjectMeta.Finalizers, FinalizerName)
+			// Need to requeue because finalizer update does not change metadata.generation
+			return reconcile.Result{}, r.client.Update(context.TODO(), instance)
+		}
+	} else {
+		for _, f := range []func(c client.Client, cr *hcov1alpha1.HyperConverged) error{
```
ormergi: I think there should be another check for finalizer existence in case something went wrong and the finalizer is no set.

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -187,6 +199,33 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
...
+			err = f(r.client, instance)
+			if err != nil {
+				reqLogger.Error(err, "Failed to manually delete objects")
+				return reconcile.Result{}, err
+			}
+		}
+
+		// Remove the finalizer
+                instance.ObjectMeta.Finalizers = drop(instance.ObjectMeta.Finalizers, FinalizerName)
+
```
ormergi: Please add foreground finalizer removal in case there is one, so it wont block the HCO deletion process:
```
if contains(instance.ObjectMeta.Finalizers, foregroundDeletionFinalizer) &&
    len(instance.ObjectMeta.Finalizers) == 1 {
    instance.ObjectMeta.Finalizers = drop(instance.ObjectMeta.Finalizers, foregroundDeletionFinalizer)
}
```

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -1085,3 +1148,112 @@ func isKVMAvailable() bool {
 	log.Info("Running with KVM available")
 	return true
 }
+
+// GetNamespacedName returns the name/namespace of the HyperConverged resource
+func GetNamespacedName() (types.NamespacedName, error) {
```
ormergi: I think this function should be private

ormergi:
Please consider those changes:
1. Refactoring the functions GetNamespacedName, contains, drop, toUnstructured, 
componentResourceRemoval. to be more readable (i.e: meaningful variables names) and more testable (i.e: in GetNamespacedName the function using external variable instead of pass it as argument).

2. I think GetNamespacedName should be private 

3. At reconcile function when handling finalizers in case there is a deletionTimeStamp add another check for finalizer existence in case something went wrong and the finalizer is no set.

4. At reconcile function when handling finalizers, when removing the finalizer please add handling for removing foregroundDeletionFinalizer.

maya-r:
>     2. I think GetNamespacedName should be private

It isn't private because it's taken from #380 where it's used as a public function. I picked a minimal part of #380 to make this pull request work, trying not to conflict with further development there and make sure this pull request can be merged whether #380 is merged before it or not.

A name like "GetHyperconverged" would be good, but I think it makes more sense to make this change in #380

>     3. At reconcile function when handling finalizers in case there is a deletionTimeStamp add another check for finalizer existence in case something went wrong and the finalizer is no set.
> 
>     4. At reconcile function when handling finalizers, when removing the finalizer please add handling for removing foregroundDeletionFinalizer.

I don't think we should remove any finalizers we are not currently adding, someone may have their own use case for adding finalizers on top of of ours, and we'll be fighting them.

dankenigsberg:
> > ```
> > 2. I think GetNamespacedName should be private
> > ```
> 
> It isn't private because it's taken from #380 where it's used as a public function. I picked a minimal part of #380 to make this pull request work, trying not to conflict with further development there and make sure this pull request can be merged whether #380 is merged before it or not.

Each PR should be self-contained. Right now we do not know if #380 is to be merged nor when. Therefore I would vote to keep the function private. When it has to be exposed, this can easily be done.

maya-r:
I couldn't think of better names for the variables in 'toUnstructured', suggestions welcome.
I think it's very important to have a working fix in-tree so we can test it and become happy by the expected release time. we can create additional pull requests for additional cleanup later.

pkg/controller/hyperconverged/hyperconverged_controller.go
djzager: seems reasonable.

djzager:
> Each PR should be self-contained. Right now we do not know if #380 is to be merged nor when. Therefore I would vote to keep the function private. When it has to be exposed, this can easily be done.

I agree. Go ahead and make it private.

pkg/controller/hyperconverged/hyperconverged_controller.go
djzager: There is some good information in the operator-sdk documentation regarding [finalizers](https://github.com/operator-framework/operator-sdk/blob/master/doc/user-guide.md#handle-cleanup-on-deletion).

Checking for the finalizer and not removing the finalizer unless the cleanup proceeds successfully sounds like a good idea.

djzager:
You've changed the way the Reconcile function works. You will need to update the tests to reflect these changes.

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -187,6 +199,33 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
...
+				reqLogger.Error(err, "Failed to manually delete objects")
+				return reconcile.Result{}, err
+			}
+		}
+
+		// Remove the finalizer
+                instance.ObjectMeta.Finalizers = drop(instance.ObjectMeta.Finalizers, FinalizerName)
+
+                // Need to requeue because finalizer update does not change metadata.generation
+                return reconcile.Result{}, r.client.Update(context.TODO(), instance)
```
djzager: need to set `Requeue: true` if you actually want to force requeu.

pkg/controller/hyperconverged/hyperconverged_controller.go
djzager: are we even using the foregroundDeletionFinalizer? I don't see anywhere where we would be adding that finalizer to a HyperConverged resource.

If not, I would prefer us not adding extra logic to cleanup a finalizer we know we aren't adding.

maya-r:
What's a way of running the tests locally? `make test-unit` doesn't have the intended effect.

djzager:
> What's a way of running the tests locally? `make test-unit` doesn't have the intended effect.

`make test`

djzager:
I see there is some wonkiness between `make test` and `make test-unit` ... odd ... but I don't think now is the time to play around with the Makefile.

djzager:
I think code coverage (coveralls) should be taken up in a follow-up PR. Focus on getting the existing tests to pass.

pkg/controller/hyperconverged/hyperconverged_controller.go
ormergi: I searched in hco, kubevit, cnao, cdi, ssp operators repository's and there is  no usage of foreground deletion finalizer for HCO CR, so its looks like its safe not to add it.

maya-r:
Changes since previous version:
- Fix whitespace
- Combine commits into one (rather than create a function and immediately modify it)
- Only attempt to remove the finalizer if it exists, suggested by ormergi
- Don't requeue another reconcile after installing the finalizer, there's no problem
  with just continuing our current run, and it upsets the tests that expect a certain
  state after the first reconcile.

ormergi:
lgtm

rthallisey:
ci test please

orenc1:
@maya-r please also handle ownerReference removal (& cleanup at finalizer) of KubevirtCommonTemplatesBundle, whose owner/creator is also the HCO.

maya-r:
> @maya-r please also handle ownerReference removal (& cleanup at finalizer) of KubevirtCommonTemplatesBundle, whose owner/creator is also the HCO.

done

djzager:
@maya-r Wow. That stinks. All green except for code coverage and I'm asking for a change.

I still think you need to handle the err for getting the kubevirt common templates bundle before handing the owner reference stuff. Everything else looks great.

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -41,9 +45,7 @@ const (
 	// use finalizers to manage the cleanup.
 	FinalizerName = "hyperconvergeds.hco.kubevirt.io"
 
-	// Foreground deletion finalizer is blocking removal of HyperConverged until explicitly dropped.
-	// TODO: Research whether there is a better way.
-	foregroundDeletionFinalizer = "foregroundDeletion"
+	HyperConvergedName = "hyperconverged-cluster"
```
tiraboschi: I think it will be better to let the user express a custom value with an environmental variable keeping this as a sane default.

djzager: my reason for not pursuing this route is that it's going to be backed into the CSV already so the user will not be able to configure it when installing via OLM. Why not remove the variation and simply encode the value in the source?

maya-r:
> @maya-r Wow. That stinks. All green except for code coverage and I'm asking for a change.
> 
> I still think you need to handle the err for getting the kubevirt common templates bundle before handing the owner reference stuff. Everything else looks great.

thanks, fixed!

pkg/controller/hyperconverged/hyperconverged_controller.go
tiraboschi: On the other side we are going to fail if, by mistake, the user creates the CR with a different name.
We can set consume this a default, set an override via env variable in the CSV and compose the alm-example in the CSV according to what we have in the environment variable.
See: https://github.com/kubevirt/hyperconverged-cluster-operator/pull/399

djzager: Why not just put the "hyperconverged-cluster" in the alm-example, done. How does using an environment variable solve the problem of

> On the other side we are going to fail if, by mistake, the user creates the CR with a different name.

This will still be a problem if we use an environment variable. The facts are, regardless of hardcoded in source or using an environment variable, that:

1. The user must create the resource in the same namespace as the operator
1. The name of that resource is determined long before the user "Subscribes" to the operator

This was the reason @dankenigsberg that I wanted to specifically handle this scenario where a user creates a HyperConverged resource that we weren't expecting, either:

1. They mistakenly created the resource with the wrong name or namespace. 
1. They mistakenly created the resource when one we are reconciling already exists.

In either case, the polite thing to do is notify the user (via status), "Hey, this CR doesn't look right."

dankenigsberg: Found it awkward to touch wrongly created CR. But when thinking about it again I suspect it would be ok for the hyperconverged controller to track all hyperconverged CRs.

Anyway, this should be discussed on another PR.

djzager:
with only code coverage remaining. I'm accepting this PR.

PR #394, opened 2020-01-02
storage class as parameter

ILpinto:
For certificate test, set storage class as parameter

### DIFF
--- a/tests/func-tests/certificates_test.go
+++ b/tests/func-tests/certificates_test.go
@@ -25,6 +25,10 @@ import (
 	testscore "kubevirt.io/kubevirt/tests"
 )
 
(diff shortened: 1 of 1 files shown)


rmohr:
/hold

I don't think that this is needed. Check the `storage-class-local` flag which I added in https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/tests/func-tests/utils.go#L27

rmohr:
you should also see it when you run `--help` on the binary.

rmohr:
can this be closed?

ILpinto:
closing parameter exists

ILpinto:
closing parameter exists

PR #398, opened 2020-01-08
KubevirtCommonTemplatesBundle CR Namespace Rectification

orenc1:
This PR addresses a single case of an issue of objects unintended deletion by GC because of cross-namepsaces ownerships.
KubevirtCommonTemplatesBundle CR's owner is HyperConverged.
Currently these are in different namespaces.
KubevirtCommonTemplatesBundle is hard-codes to be in "openshift" namespace, while other CRs, including the HyperConverged are in "request.Namespace", which is kubevirt-hyperconverged in the upstream.
Issue depicted also in BZ https://bugzilla.redhat.com/show_bug.cgi?id=1788640

### DIFF
--- a/pkg/controller/hyperconverged/hyperconverged_controller.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller.go
@@ -47,10 +47,7 @@ const (
 
 	// UndefinedNamespace is for cluster scoped resources
 	UndefinedNamespace string = ""
-
-	// OpenshiftNamespace is for resources that belong in the openshift namespace
-	OpenshiftNamespace string = "openshift"
-
+	
 	reconcileInit             = "Init"
 	reconcileInitMessage      = "Initializing HyperConverged cluster"
 	reconcileFailed           = "ReconcileFailed"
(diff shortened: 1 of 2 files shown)


ovirt-infra:

orenc1:
@djzager 
@dankenigsberg 

Please review.
This PR intends to deal with issue reported in BZ https://bugzilla.redhat.com/show_bug.cgi?id=1788640.

Thanks

dankenigsberg:
> Signed-off-by: orenc1 [oren5611@gmail.com](mailto:oren5611@gmail.com)

Please add a meaningful description. It should state why this PR is needed (and point to the BZ)

rthallisey:
ci test please

rthallisey:
What's the reason for changing the template namespace?  Can you double check that templates will be seen in an openshift-* prefixed namespace?  My understanding is that by default, OpenShift only sees templates in the "openshift" namespace.

orenc1:
@rthallisey yes, by default the command "oc new-app" searches a template located in the "openshift" namespace, but you can pass a '-n' argument to specify the namespace if the template is not there.

Anyway, the issue is that the CR of KubevirtCommonTemplatesBundle would be deleted because of cross-namespace ownership.
Can the KCTB create templates on a different namespace than itself?

djzager:
I think the correct solution for this issue is to do as we are doing for CDI and ClusterNetworkAddons. That is to simply stop adding the owner references and specify a cleanup for the KubevirtCommonTemplatesBundle when handling the finalizer. You will need to rely on the changes in #393.

rthallisey:
> @rthallisey yes, by default the command "oc new-app" searches a template located in the "openshift" namespace, but you can pass a '-n' argument to specify the namespace if the template is not there.
>
That sounds like it changes the user experience and potentially a doc change.

> Anyway, the issue is that the CR of KubevirtCommonTemplatesBundle would be deleted because of cross-namespace ownership.
>
I'm assuming ansible operators always add owner references?  We should open a bug with the ansible operator sdk if that's the case.

tiraboschi:
I tend to agree that the correct solution is to remove the owner reference and cleanup the KubevirtCommonTemplatesBundle with a finalizer.
Can we handle it in #393 instead than a specific PR because at the end the fix will be exactly the same?

orenc1:
Aborting this PR as you suggested.
The KCTB will be handled by @maya-r in #393 along with the CDI and CNAO.
Thanks.

PR #399, opened 2020-01-08
Let the user set a custom namespace in alm-example at CSV compose time

tiraboschi:
Let the user set a custom name and namespace for Hyperconverged CR at CSV compose time.
For now this is used only in the alm-example section.


dankenigsberg:
@tiraboschi can you explain why it is important to let the user specify the namespace? I thought that we are allowing installation to a single namespace. Can you add a documentation for how should the user do that?

tiraboschi:
This is just for the namespace written in alm-example section that's just a string.
Currently we always hardcode the upstream value but it would be nice to make it customizable at CSV composition time although AFAIK OLM already overwrites it with the name of the currently open project when the user tries to create his CR.

dankenigsberg:
@tiraboschi I'm asking to put this kind of information per each pr. On the first commit and in git commit message(s).

djzager:
Still LGTM

PR #400, opened 2020-01-08
Ignore invalid requests

djzager:
The chief reason for this change is that we allow only a single deployment of HCO, controlled by a single CR with a well-known name living in a well-known namespace. Any other CR is invalid, and we would like to report this fact to the user who created it.

### DIFF
--- a/pkg/controller/hyperconverged/hyperconverged_controller.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller.go
@@ -10,7 +10,6 @@ import (
 
 	"encoding/json"
 	"github.com/go-logr/logr"
-	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
 	"github.com/operator-framework/operator-sdk/pkg/ready"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -45,7 +44,8 @@ const (
 	// use finalizers to manage the cleanup.
 	FinalizerName = "hyperconvergeds.hco.kubevirt.io"
 
-	HyperConvergedName = "hyperconverged-cluster"
+	HyperConvergedName   = "hyperconverged-cluster"
+	OperatorNamespaceEnv = "OPERATOR_NAMESPACE"
 
 	// UndefinedNamespace is for cluster scoped resources
 	UndefinedNamespace string = ""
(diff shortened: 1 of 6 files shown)


djzager:
@dankenigsberg this makes it so the HyperConverged controller ignores requests we can't handle and communicates via status why it is being ignored.

dankenigsberg:
> Since all of our secondary watches are being enqueued as requests for
> name: hyperconverged-cluster namespace: ${operator_ns}. We need to let a
> user know why we are ignoring invalid requests.

The watch is a minor implementation issue. Typically, you don't want to change the API of a resource due to something like that.

Please consider using something along the following lines instead:
"""
The chief reason for this change is that we allow only a single deployment of HCO, controlled by a single CR with a well-known name living in a well-known namespace. Any other CR is invalid, and we would like to report this fact to the user who created it.
"""

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -537,7 +557,7 @@ func (r *ReconcileHyperConverged) ensureCDI(instance *hcov1alpha1.HyperConverged
 	// Previous versions used to have HCO-operator (scope namespace)
 	// as the owner of CDI (scope cluster).
 	// It's not legal, so remove that.
-	if (len(existingOwners) > 0) {
+	if len(existingOwners) > 0 {
```
dankenigsberg: this change seems unrelated to the purpose of the PR. please use another PR for code cleanup.

djzager: Done

djzager:
ci test please

djzager:
ci test please

dankenigsberg:
Thanks David for updating the description text. I still miss a description on how could a packager control the unique namespace.

djzager:
> Thanks David for updating the description text. I still miss a description on how could a packager control the unique namespace.

Not sure I follow. The namespace is controlled by where the operator is deployed. There is, intentionally, no other means for controlling this value.

dankenigsberg:
But something is reading a namespace name from and env var. I'm not qualified to code review here, but it seemed that you are adding a configurable and do not document it. (I may be very wrong)

djzager:
> But something is reading a namespace name from and env var. I'm not qualified to code review here, but it seemed that you are adding a configurable and do not document it. (I may be very wrong)

As I understand it, the only supported method for installing the HCO is via OLM and the ClusterServiceVersion we provide. This pull request uses the downward API to populate the environment variable `OPERATOR_NAMESPACE` with the namespace where the operator pod is deployed. Since there is no supported mechanism for modifying the ClusterServiceVersion, I feel comfortable with my claim that the `OPERATOR_NAMESPACE` is not a configurable.

PR #403, opened 2020-01-09
Unit tests for utility functions

maya-r:


maya-r:
"toUnstructured" doesn't like objects without Kind in them, so I had to forcibly add it.
It works in the regular code because the server fills it out.

Can I add a kind to all the getHyperconverged / newCDIForCR / ... functions in the controller?

djzager:
I think what you are saying is the very reason that [controller-runtime suggests using envtest instead of the fake client](https://github.com/kubernetes-sigs/controller-runtime/blob/master/FAQ.md#q-wheres-the-fake-client--how-do-i-use-it). If this is too problematic, we should make the move to envtest and move forward that way.

Info here https://book.kubebuilder.io/reference/testing/envtest.html?highlight=test#testing-considerations

PR #404, opened 2020-01-14
Avoid creating the OperatorSource if already there

tiraboschi:
Avoid trying to create the OperatorSource if already there:
certified-operators, community-operators, redhat-operators
are there by default on OCP.

### DIFF
--- a/deploy/deploy_marketplace.sh
+++ b/deploy/deploy_marketplace.sh
@@ -103,8 +103,9 @@ spec: {}
 EOF
 fi
 
-echo "Creating OperatorSource"
-cat <<EOF | oc create -f -
+if [ `oc get OperatorSource "${APP_REGISTRY}" -n "${MARKETPLACE_NAMESPACE}" --no-headers 2> /dev/null | wc -l` -eq 0 ]; then
+    echo "Creating OperatorSource"
+    cat <<EOF | oc create -f -
 apiVersion: operators.coreos.com/v1
 kind: OperatorSource
 metadata:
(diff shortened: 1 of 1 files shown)


deploy/deploy_marketplace.sh
```diff
@@ -91,7 +91,7 @@ EOF
 
 fi
 
-if [ `oc get operatorgroup -n "${TARGET_NAMESPACE}" 2> /dev/null | wc -l` -eq 0 ]; then
+if [ `oc get operatorgroup -n "${TARGET_NAMESPACE}" --no-headers 2> /dev/null | wc -l` -eq 0 ]; then
```
dankenigsberg: I might be wrong, but this does not seem related to this PR. Let's keep our changes tight and to the point.

rthallisey: With or without the `--no-headers`, the block is correct because the headers aren't shown if the number_of_entries == 0.    This change is to keep these checks uniform because it's a better to use `--no-headers` everywhere in case we need to look for something other than 0 later on.

dankenigsberg: I understand that `--no-headers` is advisable for scripts. I request to have this change in a PR of its own, with a proper explanation (e.g Headers are for humans, they confuse scripts. Let us drop them all over the place).

tiraboschi: Correct, this one is addressing a different issue;
moved to https://github.com/kubevirt/hyperconverged-cluster-operator/pull/405

PR #405, opened 2020-01-14
Improve OperatorGroup count check in deploy_marketplace.sh

tiraboschi:
Improve OperatorGroup count check

deploy_marketplace.sh is trying to create an OperatorGroup
only if missing according to the output of oc.
oc already skips the headers when we don't
have entries in the output.
But in general headers are for humans and they
can confuse scripts. Let's drop it in advance
to make it safer.


tiraboschi:
ci test please

PR #407, opened 2020-01-15
Additional custom values for CSV generator

tiraboschi:
Modify the CSV generator to handle additional
custom values at CSV compose time.
More specificaly:
- Label shown in OLM UI about the primary CRD
- A yaml file with puntual overrides to be applied with imdario/mergo


tiraboschi:
ci test please

PR #408, opened 2020-01-20
change HCO OperatorGroup membership to select only one namespace

orenc1:
This change is aimed to limit the visibility of HCO operator only to the target namespace, which is "kubevirt-hyperconverged" in upstream and "openshift-cnv" in downstream.
[This is according to official documentation](https://docs.openshift.com/container-platform/4.2/cnv/cnv_install/installing-container-native-virtualization.html#cnv-subscribing-to-hco-catalog_installing-container-native-virtualization).

### DIFF
--- a/deploy/olm-catalog/kubevirt-hyperconverged/1.0.0/kubevirt-hyperconverged-operator.v1.0.0.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/kubevirt-hyperconverged/1.0.0/kubevirt-hyperconverged-operator.v1.0.0.clusterserviceversion.yaml
@@ -8,6 +8,7 @@ metadata:
     capabilities: Full Lifecycle
     categories: OpenShift Optional
     certified: "false"
+    olm.targetNamespaces: kubevirt-hyperconverged
     containerImage: quay.io/kubevirt/hyperconverged-cluster-operator:latest
     createdAt: "2020-02-11 19:49:42"
     description: |-
@@ -1277,6 +1278,9 @@ spec:
                     fieldRef:
                       fieldPath: metadata.name
                 - name: WATCH_NAMESPACE
+                  valueFrom:
+                    fieldRef:
+                      fieldPath: metadata.namespace
                 - name: CONVERSION_CONTAINER
                   value: quay.io/kubevirt/kubevirt-v2v-conversion:v2.0.0
                 - name: VMWARE_CONTAINER
@@ -1554,9 +1558,9 @@ spec:
     type: OwnNamespace
   - supported: true
     type: SingleNamespace
-  - supported: true
+  - supported: false
     type: MultiNamespace
-  - supported: true
+  - supported: false
     type: AllNamespaces
   keywords:
   - KubeVirt
--- a/pkg/components/components.go
+++ b/pkg/components/components.go
@@ -123,8 +123,12 @@ func GetDeploymentSpec(image, imagePullPolicy, conversionContainer, vmwareContai
 								},
 							},
 							{
-								Name:  "WATCH_NAMESPACE",
-								Value: "",
+								Name: "WATCH_NAMESPACE",
+								ValueFrom: &corev1.EnvVarSource{
+									FieldRef: &corev1.ObjectFieldSelector{
+										FieldPath: "metadata.namespace",
+									},
+								},
 							},
 							{
 								Name:  "CONVERSION_CONTAINER",
@@ -520,15 +524,16 @@ func GetCSVBase(name, hcCRNamespace, displayName, description, image, replaces s
 			Name:      fmt.Sprintf("%v.v%v", name, version.String()),
 			Namespace: "placeholder",
 			Annotations: map[string]string{
-				"alm-examples":   string(almExamples),
-				"capabilities":   "Full Lifecycle",
-				"certified":      "false",
-				"categories":     "OpenShift Optional",
-				"containerImage": image,
-				"createdAt":      time.Now().Format("2006-01-02 15:04:05"),
-				"description":    description,
-				"repository":     "https://github.com/kubevirt/hyperconverged-cluster-operator",
-				"support":        "false",
+				"alm-examples":         string(almExamples),
+				"capabilities":         "Full Lifecycle",
+				"certified":            "false",
+				"olm.targetNamespaces": "kubevirt-hyperconverged",
+				"categories":           "OpenShift Optional",
+				"containerImage":       image,
+				"createdAt":            time.Now().Format("2006-01-02 15:04:05"),
+				"description":          description,
+				"repository":           "https://github.com/kubevirt/hyperconverged-cluster-operator",
+				"support":              "false",
 			},
 		},
 		Spec: csvv1alpha1.ClusterServiceVersionSpec{
@@ -586,11 +591,11 @@ func GetCSVBase(name, hcCRNamespace, displayName, description, image, replaces s
 				},
 				csvv1alpha1.InstallMode{
 					Type:      csvv1alpha1.InstallModeTypeMultiNamespace,
-					Supported: true,
+					Supported: false,
 				},
 				csvv1alpha1.InstallMode{
 					Type:      csvv1alpha1.InstallModeTypeAllNamespaces,
-					Supported: true,
+					Supported: false,
 				},
 			},
 			// Skip this in favor of having a separate function to get
--- a/hack/upgrade-test.sh
+++ b/hack/upgrade-test.sh
@@ -114,6 +114,9 @@ kind: OperatorGroup
 metadata:
   name: hco-operatorgroup
   namespace: kubevirt-hyperconverged
+  spec:
+    targetNamespaces:
+    - kubevirt-hyperconverged
 EOF
 
 # TODO: The catalog source image here should point to the latest version in quay.io
@@ -162,6 +165,13 @@ spec:
   source: hco-catalogsource-example
   sourceNamespace: ${HCO_CATALOG_NAMESPACE}
 EOF
+
+# temporary debug show commands
+sleep 15
+${CMD} get csv -A
+${CMD} describe csv -A
+${CMD} get subscription -A -o yaml
+${CMD} get pods -n kubevirt-hyperconverged
 
 # Allow time for the install plan to be created a for the
 # hco-operator to be created. Otherwise kubectl wait will report EOF.
--- a/templates/olm-catalog/kubevirt-hyperconverged/VERSION/kubevirt-hyperconverged-operator.VERSION.clusterserviceversion_merger.yaml.in
+++ b/templates/olm-catalog/kubevirt-hyperconverged/VERSION/kubevirt-hyperconverged-operator.VERSION.clusterserviceversion_merger.yaml.in
@@ -343,6 +343,9 @@ spec:
                       fieldRef:
                         fieldPath: metadata.name
                   - name: WATCH_NAMESPACE
+                    valueFrom:
+                      fieldRef:
+                        fieldPath: metadata.namespace
                   - name: CONVERSION_CONTAINER
                     value: quay.io/kubevirt/kubevirt-v2v-conversion:v2.0.0
                   - name: VMWARE_CONTAINER
--- a/templates/operator.yaml.in
+++ b/templates/operator.yaml.in
@@ -32,7 +32,9 @@ spec:
                 fieldRef:
                   fieldPath: metadata.name
             - name: WATCH_NAMESPACE
-              value: ""
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
             - name: VIRT_LAUNCHER_TAG
               value: {{.KubeVirt.ComponentTag}}
             - name: OPERATOR_NAME


dankenigsberg:
> This change is aimed to limit the visibility of HCO operator only to the target namespace, which is "kubevirt-hyperconverged" in upstream and "openshift-cnv" in downstream.
> [This is according to official documentation](https://docs.openshift.com/container-platform/4.2/cnv/cnv_install/installing-container-native-virtualization.html#cnv-subscribing-to-hco-catalog_installing-container-native-virtualization).

I'll add a word explaining why it is a good thing. Currently, our install doc ask the user to change the default. With your change, we can simplify the docs.

P.s. you can also suggest a PR to https://docs.openshift.com/container-platform/4.2/cnv/cnv_install/installing-container-native-virtualization.html#cnv-subscribing-to-hco-catalog_installing-container-native-virtualization

djzager:
> I remember that we already discussed this and @djzager pointed out that we have to keep it like it is because HCO operator should watch all the namespaces. David, thoughts?

Yeah and I think I was wrong. Looking through the OLM docs I could not find anything to suggest this would impact our ability to watch all namespaces.

djzager:
> I'll add a word explaining why it is a good thing. Currently, our install doc ask the user to change the default. With your change, we can simplify the docs.

Are you wanting the HCO to be installed into the `openshift-operators` namespace? This change would allow the user to install the HCO in the `openshift-operators` namespace.

djzager:
It makes more sense for this change to be allowing the user to select the namespace as @orenc1 had suggested with this screenshot.

image

The primary benefit of this change being that you don't have the ability to select "All namespaces"...I just don't see what changes in the actual documentation since the user will still have to create the namespace ahead of time and target that namespace when installing the operator through operator hub.

dankenigsberg:
> > I'll add a word explaining why it is a good thing. Currently, our install doc ask the user to change the default. With your change, we can simplify the docs.
> 
> Are you wanting the HCO to be installed into the `openshift-operators` namespace? This change would allow the user to install the HCO in the `openshift-operators` namespace.

No, I would not like to make any functional change, only to simplify installation process and doc. This PR should state if this is an outcome of it.

orenc1:
> P.s. you can also suggest a PR to https://docs.openshift.com/container-platform/4.2/cnv/cnv_install/installing-container-native-virtualization.html#cnv-subscribing-to-hco-catalog_installing-container-native-virtualization

Yes, once this PR will be merged I'll open a PR with the documentation to remove step 5a and its warning note.

orenc1:
I've got an error in the Prow job test:
```
+ oc get hco hyperconverged-cluster -n kubevirt-hyperconverged '-o=jsonpath={range .status.conditions[*]}{.type}{"\t"}{.status}{"\t"}{.message}{"\n"}{end}'
ReconcileComplete	True	Reconcile completed successfully
Available	False	KubevirtNodeLabellerBundle is not available: Node-labeller is available.
Progressing	True	KubevirtCommonTemplatesBundle is progressing: Templates progressing.
Degraded	True	KubevirtNodeLabellerBundle is degraded: Node-labeller is degraded.
Upgradeable	False	KubevirtCommonTemplatesBundle is progressing: Templates progressing.
++ oc get pods -n kubevirt-hyperconverged '--field-selector=status.phase!=Running' -o custom-columns=:metadata.name
make: *** [start] Error 1
2020/01/20 16:46:54 Container test in pod hco-e2e-aws failed, exit code 2, reason Error
```
I don't see how is this PR related to KvNLB and KvCTB being not available after deployment.

dankenigsberg:
> I don't see how is this PR related to KvNLB and KvCTB being not available after deployment.

Neither do I. But fixing flaky tests is one of our thorniest tasks. In another PR you can add code to the test to dig a bit deeper in case of an error, and supply us the data we need to fix the flakiness (e.g the current status of the unavailable components).

djzager:
> > This change would allow the user to install the HCO in the `openshift-operators` namespace.
> 
> No, I would not like to make any functional change, only to simplify installation process and doc. This PR should state if this is an outcome of it.

My original comment was wrong. This change would prevent a user from installing the HCO in the `openshift-operators` namespace. The OperatorGroup that exists in the `openshift-operators` namespace will only allow CSVs that support the `AllNamespaces` install mode. Sorry about that.

djzager:
> > P.s. you can also suggest a PR to https://docs.openshift.com/container-platform/4.2/cnv/cnv_install/installing-container-native-virtualization.html#cnv-subscribing-to-hco-catalog_installing-container-native-virtualization
> 
> Yes, once this PR will be merged I'll open a PR with the documentation to remove step 5a and its warning note.

The only thing I can understand you removing is the warning note. A user will still need to "Select A specific namespace on the cluster from the Installation Mode list and choose the `openshift-cnv` namespace."

djzager:
I do believe we need to reach a consensus on what the documentation (https://docs.openshift.com/container-platform/4.2/cnv/cnv_install/installing-container-native-virtualization.html#cnv-subscribing-to-hco-catalog_installing-container-native-virtualization) should look like after this change. Other than that, this PR looks ready to merge (once CI is green).

djzager:
> Not exactly. As I understood, once the CSV's InstallMode is configured that way, the first (currently default) option will be greyed-out, and the radio button is already set to "A specific namespace on the cluster".
> As it is in the 3scale operator I posted and you shared in your comment here.

Sure. We are in agreement here. However, the user will still need to select which namespace...and we want to be very clear that they are to select `openshift-cnv`.

The fact that the radio button will already be selected is irrelevant in my opinion. At a minimum step 5a will need to state the Operator should be installed in the `openshift-cnv` namespace (ie. not removed).

dankenigsberg:
> > Not exactly. As I understood, once the CSV's InstallMode is configured that way, the first (currently default) option will be greyed-out, and the radio button is already set to "A specific namespace on the cluster".
> > As it is in the 3scale operator I posted and you shared in your comment here.
> 
> Sure. We are in agreement here. However, the user will still need to select which namespace...and we want to be very clear that they are to select `openshift-cnv`.
> 
> The fact that the radio button will already be selected is irrelevant in my opinion. At a minimum step 5a will need to state the Operator should be installed in the `openshift-cnv` namespace (ie. not removed).

Dropping even a single warning is a worthy effort. It is best discussed on a follow-up documentation PR.

fabiand:
We can possibly limit the namespaces by using https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/operatorgroups.md#target-namespace-selection

In addition I wonder if the tests are failing because a random namespace might be sleected (that's atleast the case for some kubevirt tests)

tiraboschi:
When the user deploys from OLM UI, OLM is implicitly creating the OperatorGroup for us so we cannot really control its definition in that case.

dankenigsberg:
> When the user deploys from OLM UI, OLM is implicitly creating the OperatorGroup for us so we cannot really control its definition in that case.

But we can write our request in alm-example, and ask OLM to enforce it (i.e file an OLM RFE or send them a PR).

tiraboschi:
> But we can write our request in alm-example

Done: https://github.com/kubevirt/hyperconverged-cluster-operator/pull/399
but it's not enough by itself.

> and ask OLM to enforce it (i.e file an OLM RFE or send them a PR).

it will be handled on console side:
https://github.com/openshift/console/pull/3862

The operator will be able to recommend a namespace and eventually configure the OperatorGroup according to that.
The console will create the namespace and the operatorGroup if missing.

fabiand:
> Done: #399

@tiraboschi but this does not help us with downstream, does it?

tiraboschi:
> > Done: #399
> 
> @tiraboschi but this does not help us with downstream, does it?

The alm-example is rendered at CSV generation time and downstream (only on 2.3) we are already correctly setting the right value that gets properly wrote in the alm-example but the result is basically just cosmetic (the user still has to manually create the namespace in advance...).

The real change will come after https://github.com/openshift/console/pull/3862 and a subsequent change here to properly consume it.

pkg/components/components.go
```diff
@@ -516,6 +516,7 @@ func GetCSVBase(name, displayName, description, image, replaces string, version
 				"alm-examples":   string(almExamples),
 				"capabilities":   "Full Lifecycle",
 				"certified":      "false",
+				"olm.targetNamespaces": "kubevirt-hyperconverged",
```
tiraboschi: Please rebase this and use hcCRNamespace variable as on line 507 so that we can render it differently from upstream to downstream

djzager:
It appears you have merged master into your branch. The preferred method is to rebase your changes on top of the master branch:

```
git fetch ${upstream_remote} master
git rebase -i ${upstream_remote}/master
...fix any conflicts
git push -f ${your_remote:-origin} ${branch:-master}
```

orenc1:
the latest failure does not seems to be related to kubevirt or even openshift. It failed on AWS infrastructure setup.
Let's try again:
/retest

orenc1:
There's an error message:
`AllNamespaces InstallModeType not supported, cannot configure to watch all namespaces`
However, I see that the olm.targetNamespaces annotation did not get any value.
I'm hard-coding it for now just to see if that the issue.
/hold

orenc1:
Last test failed on AWS provisioning:
`error: could not run steps: step hco-e2e-upgrade-aws failed: cannot resolve parameter IMAGE_FORMAT into template hco-e2e-upgrade-aws: could not lazily evaluate deferred parameter: imagestreams.image.openshift.io "pipeline" is forbidden: User "system:serviceaccount:ci:ci-operator" cannot get imagestreams.image.openshift.io in the namespace "ci-op-ql6v24i5": no RBAC policy matched`
Thus retesting.
/retest

deploy/operator.yaml
```diff
@@ -34,6 +34,9 @@ spec:
             fieldRef:
               fieldPath: metadata.name
         - name: WATCH_NAMESPACE
+          valueFrom:
```
djzager: You can not do this. Our ability to get conditions from objects that are not in our namespace (ie. CDI, ClusterNetworkAddons, etc) relies on watching all namespaces.

orenc1: > You can not do this. Our ability to get conditions from objects that are not in our namespace (ie. CDI, ClusterNetworkAddons, etc) relies on watching all namespaces.

The rationale of this pull request was to grey-out the default installation mode (Operator will be available in all namespaces), so that it will be possible to install the operator only at the designated namespace, as required.
This method is implemented by many other operators (eg. 3scale, Prometheus Operator, Fuse Online Operator, etc.)
By checking the CSV of these operators, this is implemented by setting the installModes.type: AllNamespaces to supported: false.

However, that way I'm getting `AllNamespaces InstallModeType not supported, cannot configure to watch all namespaces` from OLM so I tried to limit it to one namespace.

CDI and CNAO are cluster-scoped objects, thus are not bound to any specific namespace. Do we need "AllNamespaces" watch permissions to get conditions from cluster-scoped resources?

djzager: I would be very surprised if OLM were actually looking at what is put into the `WATCH_NAMESPACE` environment varaible for any/all containers in a CSV.

Looking at the source code for OLM, [this](https://github.com/operator-framework/operator-lifecycle-manager/blob/372a68604a028eb9c0d61ce7beda0bf948ba11a6/pkg/api/apis/operators/v1alpha1/clusterserviceversion.go#L178) is where the error you are seeing comes from and [here](https://github.com/operator-framework/operator-lifecycle-manager/blob/5fd545c78a957c45d74651dab56192f00ae42b13/pkg/controller/operators/olm/operator.go#L1213-L1217) is where this function is called.

Let us look at this a little more closely. First how the function is called:

```
modeSet.Supports(out.GetNamespace(), namespaces)
```

It looks like `out.GetNamespace()` will return with the namespace where the CSV is installed (ie. where the Subscription is created). The value of `namespaces` is a slice of strings that is the result of splitting the result of this line `targets, ok := out.GetAnnotations()[v1.OperatorGroupTargetsAnnotationKey]`. `out` here is a CSV so we are looking for `olm.targetNamespaces` annotation on the CSV. Reading through [OLMs documentation on Operator Groups](https://github.com/operator-framework/operator-lifecycle-manager/blob/85375dd71d61b2db1a0a7173723dd97132b8c339/doc/design/operatorgroups.md#operatorgroup-csv-annotations) you will see that the value of the `olm.targetNamespaces` annotation comes from the spec of the OperatorGroup. 

**tl;dr**

The arguments given to the call to the `Supports(` function are:

1. The namespace where the CSV was installed (or, from a user perspective, where the Subscription was created)
1. The `spec.targetNamespaces` of the OperatorGroup

Now let us look at the `Supports(` function, trying to limit to the necessary bits:

```
func (set InstallModeSet) Supports(operatorNamespace string, namespaces []string) error {
	numNamespaces := len(namespaces)
	switch {
	case numNamespaces == 1:
		switch namespaces[0] {
		case v1.NamespaceAll:
			if !set[InstallModeTypeAllNamespaces] {
				return fmt.Errorf("%s InstallModeType not supported, cannot configure to watch all namespaces", InstallModeTypeAllNamespaces)
			}
		}
	}
```

Based on the code path we can determine that the `spec.TargetNamespaces` for the OperatorGroup was `*`.

**Solution:** You will need to be certain that the OperatorGroup, that you create or is created for you, has `spec.targetNamespaces` equivalent to the namespace where the Subscription is made.

orenc1: Hi David,
I already configured the OperatorGroup `spec.targetNamespaces` field with the proper namespace where the subscription is made ("kubevirt-hyperconverged" in our case):
`upgrade-test.sh`:
```
cat <<EOF | ${CMD} create -f -
apiVersion: operators.coreos.com/v1
kind: OperatorGroup
metadata:
  name: hco-operatorgroup
  spec:
    targetNamespaces:
    - kubevirt-hyperconverged
EOF
```
https://github.com/orenc1/hyperconverged-cluster-operator/blob/master/hack/upgrade-test.sh#L118
But still, after subscription deployment I'm getting the following fail message:
`AllNamespaces InstallModeType not supported, cannot configure to watch all namespaces`

Is there something else I'm missing?

djzager: > Is there something else I'm missing?

Nothing stands out as wrong about what you are doing. It may be appropriate to reach out to the OLM folks and see what they say. I don't see anything outside of the OperatorGroup that could cause this error message.

deploy/olm-catalog/kubevirt-hyperconverged/1.0.0/kubevirt-hyperconverged-operator.v1.0.0.clusterserviceversion.yaml
```diff
@@ -8,6 +8,7 @@ metadata:
     capabilities: Full Lifecycle
     categories: OpenShift Optional
     certified: "false"
+    olm.targetNamespaces: kubevirt-hyperconverged
```
awgreene: This is supposed to be set by olm and should not be defined in a csv. OLM will overwrite this annotation.

orenc1: > This is supposed to be set by olm and should not be defined in a csv. OLM will overwrite this annotation.

Acknowledged.
I'll remove that annotation then

tiraboschi:
#449 works as expected and it's green on all the CI lanes

orenc1:
Closing this PR as similar PR #461has been merged.

PR #409, opened 2020-01-21
simple deploy_marketplace.sh

AsherShoshan:
A simple deploy_Marketplace.sh script, which does not include the creation of operatorSource nor catalogSourceConfig.
In upstream/downstream openshift cluster creation of operatorSorce/catalogSource is not required (redhat-operators is used)
For QE/Dev environment, custom opsrc/catsrc are provided in an earlier step of cluster creation.

### DIFF
--- a/deploy/deploy_marketplace.sh
+++ b/deploy/deploy_marketplace.sh
@@ -2,205 +2,65 @@
 
 set -ex
 
-RED='\033[0;31m'
-NO_COLOR='\033[0m'
-
-globalNamespace=`oc -n openshift-operator-lifecycle-manager get deployments catalog-operator -o jsonpath='{.spec.template.spec.containers[].args[1]}'`
-echo "Global Namespace: ${globalNamespace}"
-
-APP_REGISTRY="${APP_REGISTRY:-kubevirt-hyperconverged}"
+APP_REGISTRY="${APP_REGISTRY:-redhat-operators}"
 PACKAGE="${PACKAGE:-kubevirt-hyperconverged}"
-CSC_SOURCE="${CSC_SOURCE:-hco-catalogsource-config}"
-TARGET_NAMESPACE="${TARGET_NAMESPACE:-kubevirt-hyperconverged}"
+OPERATOR="hyperconverged-cluster-operator"
+TARGET_NAMESPACE="${TARGET_NAMESPACE:-openshift-cnv}"
 CLUSTER="${CLUSTER:-OPENSHIFT}"
 MARKETPLACE_NAMESPACE="${MARKETPLACE_NAMESPACE:-openshift-marketplace}"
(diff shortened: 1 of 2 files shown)


deploy/deploy_marketplace.sh
```diff
@@ -2,195 +2,72 @@
 
 set -ex
 
-RED='\033[0;31m'
-NO_COLOR='\033[0m'
```
dankenigsberg: this change is irrelevant. If you think that this should be dropped, please do it in a separate PR.

deploy/deploy_marketplace.sh
```diff
@@ -2,195 +2,72 @@
...
-
 globalNamespace=`oc -n openshift-operator-lifecycle-manager get deployments catalog-operator -o jsonpath='{.spec.template.spec.containers[].args[1]}'`
 echo "Global Namespace: ${globalNamespace}"
+GLOBAL_NAMESPACE="${GLOBAL_NAMESPACE:-$globalNamespace}"
```
dankenigsberg: I see you do not introduce this in this PR, but the previous line printed one value, and now you are going to use another. This would confuse the reader.

deploy/deploy_marketplace.sh
AsherShoshan: Because when you have a cluster (clean one), you already have 3 catalogsources: redhat-operators, community, and certified-operators. So no need for preparation.
In case of QE/Dev, custom catalogsources are prepared earlier, and already used for other operators (not just CNV)

Should someone, needs a custom catsrc, then he can't use the "full" script as it's not supporting the creation of a plain catalogSource (ONLY by operatorSource, from QUAY).

IMO, an OLM catalogSource creation script, should be added to repo as an extra utility, and to support plain/opsrc catsrc creation.

dankenigsberg: I don't like long scripts myself. If you want to split out some logic to another script, please do it. But please take care of existing use cases

AsherShoshan: As mentioned before, for 99.99% of deploy_marketplace.sh, creation of catalogSOurce is not required.  For other cases, the split is in a new script (with old content) deploy_operatoSource.sh.

tiraboschi: Currently we already have deploy_imageregistry.sh (to consume the bundle image as a local registry) and deploy_marketplace.sh (to deploy pointing to a remote app-registry) and we want to unify them (to avoid having to duplicate each fix to both of them).
Introducing a third script is definitively not an option.

AsherShoshan: deploy_imageregistry is creating catalogSource not an operatorSource from QUAY.
Also this one should be splitted..
Sure, it's a good idea to unify them (just for catsrc/opsrc creation)

tiraboschi:
We are going to use kustomize for this, see https://github.com/kubevirt/hyperconverged-cluster-operator/pull/463

PR #411, opened 2020-01-21
increase timeout to 15 mins

guy9050:
10 minutes is not enough time for HCO deployment.
Many of our deployments are on a failed state on Jenkins because of that even though HCO is successfully deployed - it just takes a few more minutes to finish.

### DIFF
--- a/deploy/deploy_imageregistry.sh
+++ b/deploy/deploy_imageregistry.sh
@@ -127,5 +127,5 @@ spec:
 EOF
 
     echo "Waiting for HCO to get fully deployed"
-    oc wait -n ${TARGET_NAMESPACE} hyperconverged hyperconverged-cluster --for condition=Available --timeout=10m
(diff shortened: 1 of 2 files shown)


tiraboschi:
Can you please patch also deploy_imageregistry.sh ? (we should definitely unify them!)

guy9050:
> Can you please patch also deploy_imageregistry.sh ? (we should definitely unify them!)

Done @tiraboschi

tiraboschi:
@guy9050 can you please sign the two commits?

tiraboschi:
ci test please

fabiand:
The issues look like pure AWS issues ...

@tiraboschi do you think we can merge this patch anyhow?

tiraboschi:
> The issues look like pure AWS issues ...
> 
> @tiraboschi do you think we can merge this patch anyhow?

Yes, I think we tried enough.

PR #412, opened 2020-01-24
Migrate from dep to go modules

tiraboschi:
the operator-sdk doesn't support anymore dep
since v0.12.0, see:
https://github.com/operator-framework/operator-sdk/blob/master/doc/migration/version-upgrade-guide.md#v012x

Move from dep to go modules to be aligned and
ready for future changes (like upgrading the
controller-runtime in a near future).

This change attempts to be less intrusive as possible
trying to use the same versions required with dep.
Upgrades will be handled in other PRs.


go.mod
```diff
@@ -0,0 +1,80 @@
...
+
+	github.com/coreos/prometheus-operator => github.com/coreos/prometheus-operator v0.29.0
+	github.com/go-kit/kit => github.com/go-kit/kit v0.3.0
+	github.com/openshift/client-go => github.com/openshift/client-go v0.0.0-20190401163519-84c2b942258a
+	github.com/openshift/machine-api-operator => github.com/openshift/machine-api-operator v0.2.1-0.20191025120018-fb3724fc7bdf
+	github.com/operator-framework/operator-sdk => github.com/operator-framework/operator-sdk v0.10.0
+	github.com/prometheus/prometheus => github.com/prometheus/prometheus v0.0.0-20190424153033-d3245f150225
+	k8s.io/api => k8s.io/api v0.0.0-20190222213804-5cb15d344471
+	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.0.0-20190228180357-d002e88f6236
+	k8s.io/apimachinery => k8s.io/apimachinery v0.0.0-20190221213512-86fb29eff628
```
djzager: The only thing that would be nice to see is a comment detailing which version of Kubernetes we are pinning to.

tiraboschi: Done

PR #414, opened 2020-01-24
Add 'go mod vendor' to sanity tests

tiraboschi:
Add 'go mod vendor' to sanity tests
to ensure that we can only merge
PRs with up-to-date and correctly
vendored dependencies.

Configure travis to use go "1.13.x"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: go
 go:
-- "1.10.x"
+- "1.13.x"
 sudo: required
 dist: trusty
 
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -9,12 +9,12 @@ services:
 
 script:
 # Do sanity checks
-- make sanity
+- env GO111MODULE=on make sanity
```
djzager: If you bump to go `1.13.x` I think go modules are the default there (don't need the environment variable). If you don't want to move to go `1.13.x` you can either:

just add an environment section to the travis.yml

```
env:
    - GO111MODULE="on"
```

or make the first line of the script `export GO111MODULE="on"`.

tiraboschi:
ci test please

PR #421, opened 2020-02-04
Use "kubevirt-hyperconverged" as the hco resource name

tiraboschi:
Use "kubevirt-hyperconverged" as the hco resource name to promote a smoother upgrade experience

Use "kubevirt-hyperconverged" as the hco resource name
everywhere: in alm-example section (see:
https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/pkg/components/components.go#L505
in the CSV we were already suggesting "kubevirt-hyperconverged"
as the name of the hco resource so who deployed from
OLM console, without manually editing the CR, is already
using "kubevirt-hyperconverged".
Before https://github.com/kubevirt/hyperconverged-cluster-operator/pull/400
the name of the HCO resource was not that relevant but after
it the operator will ignore all the reconciliation requests
on resource with a name different from the expected.
So align it to "kubevirt-hyperconverged" everywhere to provide
a smoother upgrade experience.


tiraboschi:
ci test please

PR #424, opened 2020-02-05
Adjust rotate script for older HCO installations

rmohr:
In old HCO versions CDI and KubeVirt were not installed in the same
namespace. Adjusting the cert rotate script accordingly.

If the CDI and KubeVirt namespace differs, then one has to call the rotate script like this:

```bash
./rotate-certs.sh --namespace kubevirt --cdi-namespace cdi
```


rmohr:
/cc @tiraboschi 
/cc @ILpinto 

would you mind reviewing this?

PR #427, opened 2020-02-06
Change the configmap default to Filesystem/ReadWriteOnce

maya-r:
Adapt tests to match.

This is the lowest common denominator for many storage classes.


maya-r:
We used to be very focused on rook-ceph, which supports both, and tried to pick better defaults for that case. But now it's a problem because we're using several storage classes that don't support those two options.

jelkosz:
@mareklibra please review

mareklibra:
Vivat NFS! Curious to see it in action.
/lgtm

tiraboschi:
Ignoring slightly decreased coverage

PR #429, opened 2020-02-07
bug: do not reconcile status updates

djzager:
Using the generationChangedPredicated makes it possible for us to avoid
reconciling status updates to the HyperConverged resource.


djzager:
The advantage now is that, although we aren't receiving events for the HyperConverged resource, we are properly reconciling the events for the different components as they make their progression to deployed.
\

PR #433, opened 2020-02-11
Allow custom names for relatedImages

tiraboschi:
Enhance the csv-merger tool to let the user
explicitly pass a custom name for each
of the images in relatedImages.

--related-images-list allow the user to pass a
comma separated list of all the images referred
in the CSV for relatedImages field.
relatedImages field contains a list of images where
for each image we can specify image (the pull url)
an name (a label); currently the name is derived from image
while, for different reasons, it could be convenient
let the user pass a custom name.
With this PR --related-images-list is still a comma separated list
but each element could be eventually (if not the code will keep the
old behaviour for backward compatibility) splitted with image|name
syntax.

### DIFF
--- a/tools/csv-merger/csv-merger.go
+++ b/tools/csv-merger/csv-merger.go
@@ -74,10 +74,11 @@ var (
 	metadataDescription = flag.String("metadata-description", "", "Metadata")
 	specDescription     = flag.String("spec-description", "", "Description")
 	specDisplayName     = flag.String("spec-displayname", "", "Display Name")
-	relatedImagesList   = flag.String("related-images-list", "", "Comma separated list of all the images referred in the CSV")
 	namespace           = flag.String("namespace", "kubevirt-hyperconverged", "Namespace")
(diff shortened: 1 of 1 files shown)


tools/csv-merger/csv-merger.go
```diff
@@ -129,8 +129,15 @@ func main() {
 
 	for _, image := range strings.Split(*relatedImagesList, ",") {
 		if image != "" {
-			names := strings.Split(strings.Split(image, "@")[0], "/")
-			name := names[len(names)-1]
+			name := ""
+			if strings.Contains(image, "|") {
+				image_s := strings.Split(image, "|")
```
orenc1: If a user would like to pass `--related-images-list` parameter to `csv-merger` tool (via `build-manifest.sh`?) with both image URL **and** a label, he/she should use '|' as the seperator between url and label.
I think it should be pointed out that the seperator is this character, for example in line 77(?)

tiraboschi: Done

tiraboschi:
ci test please

orenc1:
ci test please

PR #436, opened 2020-02-12
Revert "bug: do not reconcile status updates"

tiraboschi:
Reverts kubevirt/hyperconverged-cluster-operator#429


tiraboschi:
The issue was just about consuming al older image, ignoring it

PR #437, opened 2020-02-12
Remove STDCI

gbenhaim:
Since we add Prow jobs, we don't need STDCI anymore.

Signed-off-by: gbenhaim <galbh2@gmail.com>


tiraboschi:
/release
ci infrastructure only, no need to track it in the release note

PR #438, opened 2020-02-13
Enable component validation tests for OpenShift CI

rwsu:
Moves existing csv-tests.go from test/func-tests to
test/component-validation-tests. This decouples the component
validation tests from the other functional tests so that it
can be run standalone.

The CSV file used to make comparison against running deployments
is now extracted from the HCO catalog source pod.

The component validation tests are now executed after
the upgrade-test.

Signed-off-by: Richard Su <rwsu@redhat.com>


rwsu:
Hi @tiraboschi, is this still needed?

PR #442, opened 2020-02-13
Avoid failing the whole upgrade test on "$RECONCILE_COMPLETED" = 'Unknown'

tiraboschi:
Avoid failing the whole upgrade test on "$RECONCILE_COMPLETED" = 'Unknown'

In the bootstrap of the env we have "$RECONCILE_COMPLETED" = 'Unknown'
(and not 'Unkown') which is a legitimate state.
In that case we should wait and check again after a certain
amount of time while now we are failing the whole test suite
with something like:

Checking operator kind: hyperconvergeds.hco.kubevirt.io name: kubevirt-hyperconverged
Error: ReconcileComplete not valid: Status: Unknown Reason: Init Message: Initializing HyperConverged cluster
Error: ReconcileComplete not valid: 'Unknown'

This should make the upgrade test more robust.


tiraboschi:
CI only
/release-note-none

PR #449, opened 2020-02-14
Avoid InstallModeTypeMultiNamespace and InstallModeTypeAllNamespaces

tiraboschi:
```release-note
Avoid InstallModeTypeMultiNamespace and InstallModeTypeAllNamespaces to gray out all namespaces option in OLM subscription dialog
```

See:
https://github.com/operator-framework/olm-book/blob/master/docs/operator-scoping.md


nunnatsa:
lgtm

tiraboschi:
Works as expected, I can only select a single namespace, see: https://pasteboard.co/IWQfcER.png
CI is green on all the lanes

PR #450, opened 2020-02-15
Bump CDI to v1.13.0

tiraboschi:


tiraboschi:
The human made changes are only in hack/default, hack/config, hack/build-manifests.sh; everything else is automatically generated as the result of that changes.

PR #451, opened 2020-02-17
Set default value (BlockUninstallIfWorkloadsExist) for UninstallStrategy on Kubevirt

tiraboschi:
Set default value (BlockUninstallIfWorkloadsExist) for UninstallStrategy on Kubevirt

github.com/kubevirt/kubevirt/pull/2976 introduced uninstall strategies for the virt-operator.
This PR will:
- set BlockUninstallIfWorkloadsExist as the UninstallStrategy creating Kubevirt CR
- set UninstallStrategy=BlockUninstallIfWorkloadsExist if unset on an existing Kubevirt CR
- test it

```release-note
set default value (BlockUninstallIfWorkloadsExist) for UninstallStrategy on Kubevirt on new deployments and on upgrades when missing
```

### DIFF
--- a/pkg/controller/hyperconverged/hyperconverged_controller.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller.go
@@ -417,6 +417,9 @@ func newKubeVirtForCR(cr *hcov1alpha1.HyperConverged, namespace string) *kubevir
 			Labels:    labels,
 			Namespace: namespace,
 		},
+		Spec: kubevirtv1.KubeVirtSpec{
+			UninstallStrategy: kubevirtv1.KubeVirtUninstallStrategyBlockUninstallIfWorkloadsExist,
+		},
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -444,6 +447,14 @@ func (r *ReconcileHyperConverged) ensureKubeVirt(instance *hcov1alpha1.HyperConv
 
 	logger.Info("KubeVirt already exists", "KubeVirt.Namespace", found.Namespace, "KubeVirt.Name", found.Name)
 
+	if !reflect.DeepEqual(found.Spec, virt.Spec) {
+		if found.Spec.UninstallStrategy == "" {
+			logger.Info("Updating UninstallStrategy on existing KubeVirt to its default value")
+			found.Spec.UninstallStrategy = virt.Spec.UninstallStrategy
```
orenc1: Layman's question: What if the user changed the `UninstallStrategy` on purpose to the non-default value?
Will the reconcile-loop put the default value back again?

tiraboschi: No, because in that case found.Spec.UninstallStrategy will be != "".
We are explicitly setting it only if empty.

PR #463, opened 2020-02-19
Introducing kustomize for HCO deployment

orenc1:
```release-note
Implementing Kustomization deployment for HCO, aimed to replace the "deploy_marketplace.sh" and "deploy_imageregistry.sh" scripts.
```


deploy/kustomize/overlays/downstream/hco_cr.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: hco.kubevirt.io/v1alpha1
+kind: HyperConverged
+metadata:
+  name: kubevirt-hyperconverged
+  namespace: openshift-cnv
```
djzager: IIRC I don't think things like this (ie. downstream) are supposed to be included in the upstream source.

orenc1: Agreed.
So in GitHub we'll keep only upstream-related resources, and downstream repo will contain the downstream-related resources?

deploy/kustomize/base/01_namespace.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: base-namespace
+  labels:
+    openshift.io/run-level: "1"
+    openshift.io/cluster-monitoring: "true"
```
tiraboschi: Please add all the missing newline chars at the end of files.

pkg/components/components.go
```diff
@@ -529,6 +529,7 @@ func GetCSVBase(name, hcCRNamespace, displayName, description, image, replaces s
 				"description":    description,
 				"repository":     "https://github.com/kubevirt/hyperconverged-cluster-operator",
 				"support":        "false",
+				"operatorframework.io/suggested-namespace": hcCRNamespace,
```
tiraboschi: This is already addressed in #461 : can you please rebase this on master and squash in a single commit ?

dankenigsberg:
> Can be used as a secondary method for deployment, instead of deploy/deploy_marketplace.sh

There should be only one entry point to deploy the project. You should explain the benefit of kustomize. I don't think there is any point of keeping the existing code - let me know if I am wrong.

deploy/kustomize/deploy_kustomize.sh
```diff
@@ -0,0 +1,107 @@
...
+  get_quay_token QUAY_TOKEN
+  AUTH_TOKEN=$(cat <<-END
+  authorizationToken:
+    secretName: quay-registry-${APP_REGISTRY}
+END
+)
+fi
+
+# Perform variables substitution
+MANIFESTS_DIR=output
```
tiraboschi: `MANIFESTS_DIR=$(mktemp -d) || (echo "Failed to create temp directory" && exit 1)`
and use trap to clean it

deploy/kustomize/base/06_subscription.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: hco-operatorhub
+  namespace: ${TARGET_NAMESPACE}
+spec:
+  source: ${CSC_SOURCE}
+  sourceNamespace: ${MARKETPLACE_NAMESPACE}
+  name: kubevirt-hyperconverged
+  startingCSV: kubevirt-hyperconverged-operator.v${HCO_VERSION}
+  channel: "${HCO_CHANNEL}"
+  installPlanApproval: Automatic
```
tiraboschi: Can we optionally handle KVM_EMULATION variable as we are doing here: https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/deploy/deploy_imageregistry.sh#L80 ?

deploy/kustomize/base/04_catalog_source_config.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: operators.coreos.com/v1
+kind: CatalogSourceConfig
+metadata:
+  name: ${CSC_SOURCE}
+  namespace: ${MARKETPLACE_NAMESPACE}
+spec:
+  source: ${APP_REGISTRY}
```
tiraboschi: in https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/deploy/deploy_imageregistry.sh#L57
we have also a mechanism to deploy from a local grgp registry without the need to relay on an external appregistry.
Currently we are using it for the disconnected install workaround.
Can we implement that as well?

deploy/kustomize/base/kustomization.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
+  - 01_namespace.yaml
+  - 02_secret.yaml
+  - 03_operator_source.yaml
+  - 04_catalog_source_config.yaml
```
tiraboschi: with CONTENT_ONLY mode ( https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/deploy/deploy_marketplace.sh#L21 ) we have a mechanism where we prepare everything for QE, but we let them manually trigger the deployment from UI to check what is going to happen there.
In that case the namespace and the operator group are supposed to be created by the console (see also https://github.com/kubevirt/hyperconverged-cluster-operator/issues/476 ).
Can we mimic something like that?

deploy/kustomize/base/06_subscription.yaml
orenc1: I can add the `spec.config` block on `subsciption` manifest the same way using multiline-variable substitution, or by using the `patch` feature of kustomize that will be in another yaml. What to you think is the preferred way?

deploy/kustomize/base/kustomization.yaml
orenc1: New behavior:
If we set CREATE_NAMESPACE to false prior running `deploy_kustomize.sh`, the namespace is not being created.
This flag is taken into account only if CONTENT_ONLY is true.

deploy/kustomize/base/04_catalog_source_config.yaml
orenc1: Implemented using two overlays.
- overlay **marketplace** - deploys resources related to deployment via olm marketplace (e.g. catalogsourceconfig, operatorsource and secret [if needed]).
- overlay **image_registry** - deploys resources for direct image registry deployment (e.g. catalogsource). 
overlay selection is passed as an argument to `deploy_kustomize.sh`, I'll document it.

deploy/kustomize/base/06_subscription.yaml
orenc1: Since the existence of `subscription.spec.config` section is determined by the env var KVM_EMULATION, I decided to add it using a variable substitution, to avoid creating a new overlay branch for this.

deploy/kustomize/deploy_kustomize.sh
```diff
@@ -0,0 +1,166 @@
...
+export QUAY_USERNAME="${QUAY_USERNAME:-}"
+export QUAY_PASSWORD="${QUAY_PASSWORD:-}"
+export QUAY_TOKEN=""
+export AUTH_TOKEN=""
+export CREATE_NAMESPACE="${CREATE_NAMESPACE:-}"
+
+BASE_DIR=base
+if [ "$1" = 'marketplace' ]; then
+  OVERLAY_DIR=overlays/marketplace
+elif [ "$1" = 'image_registry' ]; then
```
yuvalk: What's image_registry in that sense?

orenc1: - `marketplace` option is deployment using CatalogSourceConfig, which creates the CatalogSource from the OperatorSource. This is equivalent to `deploy_marketplace.sh`.
- `image_registry` option is deployment using CatalogSource, where you specify the bundle image of the operator directly. This is equivalent to `deploy_imageregistry.sh`.

deploy/kustomize/base/subscription.yaml
```diff
@@ -0,0 +1,13 @@
...
+  name: hco-operatorhub
+  namespace: ${TARGET_NAMESPACE}
+spec:
+  source: ${CSC_SOURCE}
+  sourceNamespace: ${MARKETPLACE_NAMESPACE}
+  name: kubevirt-hyperconverged
+  startingCSV: kubevirt-hyperconverged-operator.v${HCO_VERSION}
+  channel: "${HCO_CHANNEL}"
+  installPlanApproval: Automatic
+  ${SUBSCRIPTION_CONFIG}
```
djzager: Along the lines of being more prescriptive, I would think we would define as little as necessary on the Spec of the Subscription. My understanding is the essentials are the name and channel. Including the `installPlanApproval` seems reasonable (even though it is the default) because we want consumers to allow automatic updates.

deploy/kustomize/base/kustomization.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+resources:
```
djzager: :+1: these are what I would think would be essential items for consumers to deploy HCO.

deploy/kustomize/base/subscription.yaml
djzager: Maybe overlays are the way to add these to the spec when appropriate?

orenc1: Thanks  for commenting, David.
In other words, you suggest to move the fields that have high chance to be changed from default values, into overlays with `patchesStrategicMerge` that they'll overwrite the defaults?
e.g. in subscription, that would be name and channel, while the others will be editable on base?

deploy/kustomize/manifests/create_namespace/kustomization.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+bases:
+  - ../marketplace
```
orenc1: Note: the base can be either `../marketplace` or `../image_registry`.
I'll add this in documentation.

deploy/kustomize/README.md
```diff
@@ -0,0 +1,116 @@
...
+## Method 1: Deploy with a kustomization overlay
+`manifests` folder contains pre-configured manifests ready to be deployed using kustomize, for various HCO deployment scenarios.
+The bases are:
+ 1. **marketplace** - this base is used to deploy HCO operator using CatalogSourceConfig, pointing to an OperatorSource, which makes the operator available in OLM marketplace. Thus, the resources in this base are:
+	 - CatalogSourceConfig
+	 - OperatorSource
+2. **image_registry** - this base is used to deploy HCO operator using a direct bundle image, delivered by grpc protocol. The resource this base contains is:
+	- CatalogSource
+
+Applying one of the bases alone results in HCO operator being ready for deployment manually according to [official documentation](https://docs.openshift.com/container-platform/4.3/cnv/cnv_install/installing-container-native-virtualization.html).
```
tiraboschi: Please avoid referring to the downstream documentation in the upstream repo.
I think we should mention this in https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/README.md instead

deploy/kustomize/manifests/create_namespace/kustomization.yaml
tiraboschi: I'd suggest also to add a comment line in the yaml file (#)

deploy/kustomize/manifests/image_registry/catalog_source.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: CatalogSource
+metadata:
+  name: hco-catalogsource
+  namespace: openshift-marketplace
+  imagePullPolicy: Always
+spec:
+  sourceType: grpc
+  image: quay.io/kubevirt/hco-container-registry:latest
+  displayName: KubeVirt HyperConverged
+  publisher: Red Hat
```
tiraboschi: Kubevirt ?

deploy/kustomize/manifests/marketplace/catalog_source_config.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: CatalogSourceConfig
+metadata:
+  name: hco-catalogsource-config
+  namespace: openshift-marketplace
+spec:
+  source: kubevirt-hyperconverged
+  targetNamespace: openshift-marketplace
+  packages: kubevirt-hyperconverged
+  csDisplayName: HCO Operator
+  csPublisher: Red Hat
```
tiraboschi: Kubevirt

deploy/kustomize/template/base/hco_cr.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: hco.kubevirt.io/v1alpha1
+kind: HyperConverged
+metadata:
+  name: kubevirt-hyperconverged
+  namespace: ${TARGET_NAMESPACE}
```
tiraboschi: I think it's worth to add .in suffix to all the yaml files under /template/

deploy/kustomize/manifests/image_registry/catalog_source.yaml
orenc1: In current script the publisher is Red Hat:
https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/deploy/deploy_imageregistry.sh#L59
You suggest to change the publisher to `kubevirt` ?

djzager:
I've been running through this in a cluster and having some issues with OLM. Will continue to do that and link any issues that get created.

djzager:
Alright so I've done my best to catch up on all that Kustomize has to offer and I have a few things that I think need to be done before this can be merged in:

* I think the kustomize manifests need to be generated as part of the `./hack/build-manifests.sh` process. That would mean changes to manifest-templator or another tool like `kustomize_builder`. Specifically, when we build all these manifests downstream, the important names should be in the manifest and not in some overlay.
* I believe that we should follow the layout mentioned in the [kustomize/glossary](https://github.com/kubernetes-sigs/kustomize/blob/ce2e68561903497bbc44a24df85c4f2e7617797e/docs/glossary.md#kustomization-root). That looks something like:

```
├── base
│   ├── hco_cr.yaml
│   ├── kustomization.yaml
│   ├── namespace.yaml
│   ├── operator_group.yaml
│   └── subscription.yaml
├── overlays
│   ├── deploy_private_repo
│   │   ├── kustomization.yaml
│   │   ├── operator_source.patch.yaml
│   │   └── secret.yaml
│   ├── image_registry
│   │   ├── catalog_source.yaml
│   │   └── kustomization.yaml
│   ├── kvm_emulation
│   │   ├── kustomization.yaml
│   │   └── subscription.patch.yaml
│   └── marketplace
│       ├── catalog_source_config.yaml
│       ├── kustomization.yaml
│       └── operator_source.yaml
```
* `bases` has been deprecated and we can just list `resources` in our desired order. That has the benefit of making it so we can specify in the `deploy_private_repo` overlay's kustomization.yaml (I may be missing some specifics but it should give the idea.):

```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

resources:
  - secret.yaml
  - operator_source.yaml
  - ../../base

patchesStrategicMerge:
  - subscription.patch.yaml
```
* My personal preference would be to update `./hack/build-manifests.sh` to properly generate these manifests in `deploy/kustomize` and hold off on `deploy/kustomize/deploy_kustomize.sh` for now. My motivation for that is to reduce what needs to be reviewed in this PR.

**Notes**
When I was playing around with this I was primarily grabbing the `kubevirt-hyperconverged` package from the OperatorHub that was included in the CodeReadyContainers. What we want the **default** scenario to be likely changes what goes in base and what goes in overlays.

WDYT? @yuvalk @orenc1 @tiraboschi

orenc1:
Thanks David for your thorough review.
>     * I think the kustomize manifests need to be generated as part of the `./hack/build-manifests.sh` process. That would mean changes to manifest-templator or another tool like `kustomize_builder`. Specifically, when we build all these manifests downstream, the important names should be in the manifest and not in some overlay.
To my understanding, `build-manifests.sh` is in-charged of creating the CSV and CRDs needed for building the operator bundle image, to be consumed by the OLM. Do you suggest to integrate there the functionality of creating all other manifests which are required for deployment, in kustomize-style, for all deployment scenarios?

>     * I believe that we should follow the layout mentioned in the [kustomize/glossary](https://github.com/kubernetes-sigs/kustomize/blob/ce2e68561903497bbc44a24df85c4f2e7617797e/docs/glossary.md#kustomization-root). That looks something like (rooted at `deploy/kustomize`):
> 
> 
> ```
> deploy/kustomize
> ├── base
> │   ├── hco_cr.yaml
> │   ├── kustomization.yaml
> │   ├── namespace.yaml
> │   ├── operator_group.yaml
> │   └── subscription.yaml
> ├── overlays
> │   ├── deploy_private_repo
> │   │   ├── kustomization.yaml
> │   │   ├── operator_source.patch.yaml
> │   │   └── secret.yaml
> │   ├── image_registry
> │   │   ├── catalog_source.yaml
> │   │   └── kustomization.yaml
> │   ├── kvm_emulation
> │   │   ├── kustomization.yaml
> │   │   └── subscription.patch.yaml
> │   └── marketplace
> │       ├── catalog_source_config.yaml
> │       ├── kustomization.yaml
> │       └── operator_source.yaml
> ```
I can change that and move the bases and overlays to their corresponding directories, although @yuvalk suggested not to use explicitly "base" and "overlay" terms in directories tree.

>     * `bases` has been deprecated and we can just list `resources` in our desired order. That has the benefit of making it so we can specify in the `deploy_private_repo` overlay's kustomization.yaml (I may be missing some specifics but it should give the idea.):
> 
> 
> ```
> apiVersion: kustomize.config.k8s.io/v1beta1
> kind: Kustomization
> 
> resources:
>   - secret.yaml
>   - operator_source.yaml
>   - ../../base
> 
> patchesStrategicMerge:
>   - subscription.patch.yaml
> ```
Acknowledged. Will change that.
>     * My personal preference would be to update `./hack/build-manifests.sh` to properly generate these manifests in `deploy/kustomize` and hold off on `deploy/kustomize/deploy_kustomize.sh` for now. My motivation for that is to reduce what needs to be reviewed in this PR.
See bullet No. 1

orenc1:
Although your suggested tree is the logical and straight-forward way, but it's not possible to use it to deploy in all possible scenarios.
If the user wants to merely prepare the contents, but not to deploy the operator, he can't do it, since `subscription`, `OperatorGroup`, `HCO` CR are in the base.
Unless the user edits `kustomization.yaml` and remove the base, but this require a little bit more manual modification.

tiraboschi:
> > ```
> > * I think the kustomize manifests need to be generated as part of the `./hack/build-manifests.sh` process. That would mean changes to manifest-templator or another tool like `kustomize_builder`. Specifically, when we build all these manifests downstream, the important names should be in the manifest and not in some overlay.
> > ```
> 
> To my understanding, `build-manifests.sh` is in-charged of creating the CSV and CRDs needed for building the operator bundle image, to be consumed by the OLM. Do you suggest to integrate there the functionality of creating all other manifests which are required for deployment, in kustomize-style, for all deployment scenarios?
> 

hack/build-manifests.sh is used, as it is, only upstream while downstream we have a different script that consumes on the fly the partial outputs of an integrated build pipeline.

On my opinion the relevant question is where are going to store the output of this new tool.
If upstream we can simply save it here, where should we store (and then ship) the downstream copy?
On my opinion maintaining downstream just the relevant overlay can be easier.

djzager:
> I can change that and move the bases and overlays to their corresponding directories, although @yuvalk suggested not to use explicitly "base" and "overlay" terms in directories tree.

@yuvalk I would be interested to hear more here. I don't know enough to do anything but ask why? Looking around at their documentation everything is talking about it in terms of [base](https://github.com/kubernetes-sigs/kustomize/blob/master/docs/glossary.md#base) and [overlays](https://github.com/kubernetes-sigs/kustomize/blob/master/docs/glossary.md#overlay).

djzager:
> Although your suggested tree is the logical and straight-forward way, but it's not possible to use it to deploy in all possible scenarios.
> If the user wants to merely prepare the contents, but not to deploy the operator, he can't do it, since `subscription`, `OperatorGroup`, `HCO` CR are in the base.
> Unless the user edits `kustomization.yaml` and remove the base, but this require a little bit more manual modification.

I think this is possible by saying something like:

```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

resources:
  - secret.yaml
  - operator_source.yaml
  - ../../base/namespace.yaml
  - ../../base/operator_group.yaml
```

If you wanted to not include the subscription and resource creation.

djzager:
> Although your suggested tree is the logical and straight-forward way, but it's not possible to use it to deploy in all possible scenarios.
> If the user wants to merely prepare the contents, but not to deploy the operator, he can't do it, since `subscription`, `OperatorGroup`, `HCO` CR are in the base.
> Unless the user edits `kustomization.yaml` and remove the base, but this require a little bit more manual modification.

I am not sure I wholly understand the value of having kustomizations (and a kustomize generator) for all of the different ways that the HyperConverged operator can be made available to the cluster. I personally think that we need to make the HyperConverged operator available upstream in operatorhub so we can make a similar subscription (from a different catalogsource) upstream as we do downstream. In the case the user wants to stray away from what is already available in the cluster, then they need to 1) make our operator available (we can provide scripts to make that easy) and 2) they can take our manifests and create their own overlays to, for example, modify the subscription to be from their custom catalog source.

djzager:
> I can change that and move the bases and overlays to their corresponding directories, although @yuvalk suggested not to use explicitly "base" and "overlay" terms in directories tree.

Seeing what you get from [kubebuilder](https://github.com/jetstack/kubebuilder-sample-controller/tree/master/config) makes me think that would be an appropriate way to structure our kustomize manifests.

djzager:
> Although your suggested tree is the logical and straight-forward way, but it's not possible to use it to deploy in all possible scenarios.
> If the user wants to merely prepare the contents, but not to deploy the operator, he can't do it, since `subscription`, `OperatorGroup`, `HCO` CR are in the base.
> Unless the user edits `kustomization.yaml` and remove the base, but this require a little bit more manual modification.

I do believe that supporting all of these cases is a good goal. I think I may just be struggling our trying to solve all of them in one go. What I think would be a reasonable way to introduce kustomize into this project would be to have:

```
deploy/kustomize
├── base
│   ├── hco_cr.yaml
│   ├── kustomization.yaml
│   ├── namespace.yaml
│   ├── operator_group.yaml
│   └── subscription.yaml
```

And you could use those manifests to replace the appropriate logic in `deploy/deploy_marketplace.sh`. While I am certain that it is a bit more complicated than that. I think that would give us a solid base from which we can build.

My recommendation would be to take the remainder of the work you have already done and put it into a separate pull request that we can discuss and break down further if necessary. With what I have included above you could dynamically copy `deploy/kustomize/base` somewhere and add `deploy/kustomize/kustomization.yaml` that could do as much or as little of:

1. creating the namespace
1. creating the operatorgroup
1. creating the subscription
1. creating the hco cr

Then in your follow-up PR we can really nail down what scenarios we want to explicitly support.

orenc1:
Hi David, thanks for your feedback.
>     1. It isn't okay for us to have `deploy/kustomize/template` and `deploy/kustomize/manifests` (where one is just a wholesale copy of the other is problematic. This kind of stuff is what Kustomize is supposed to be good at: running `envsubst` over every file, deleting lines from `kustomization.yaml` files, and magically adding `private_repo` to the kvm_emulation `kustomization.yaml` (while clever) is a shell script backed by kustomize in name only.
> 
For what I know, kustomize does not offer a mechanism to replace placeholders with environment variables. It can be partially done by using `patchesStrategicMerge` and specify the altered values in specified fields. It cannot change object's name, for example. That's why I needed a shell script to perform the replacements and alterations of `kustomization.yaml` files according to the desired deployment configuration.

>     2. I really believe that you would benefit from separating the "how to make HCO available" problem from the "how to install HCO via OLM" problem. What impact would this have? For one you would be able to get rid of this weird distinction between `base` and `customized` (ie. what you have in `customized` should just be the `base`). Then you could simply `o
> 
The idea is the user could only edit `customized` folder's files and leave the `base` untouched, since it contains the default configuration.
Your last sentence was truncated.
>     3. There are a lot of places that you are still using `bases` and I'm not sure why.
> 
I tried to put the bases into `resources` section (as it was advised it's supported in new kustomize version), but I got an error when specifying a directory there.
>     4. Some of these kustomization setups (`private_repo` stands out) seem to exist just so we can say we have it since you would absolutely need to either 1) modify it with your secret 2) run a script. Why are we bothering with this?
Because the idea is to cover all use cases implemented in the current deploy scripts (`deploy_marketplace.sh` and `deploy_imageregistry.sh`), using a unified, kustomize-based mechanism. Are you suggesting to drop out some setups? I know QE are using the `private_repo`.

djzager:
> For what I know, kustomize does not offer a mechanism to replace placeholders with environment variables. It can be partially done by using `patchesStrategicMerge` and specify the altered values in specified fields. It cannot change object's name, for example. That's why I needed a shell script to perform the replacements and alterations of `kustomization.yaml` files according to the desired deployment configuration.

From kustomize.io:

> Kustomize introduces a template-free way to customize application configuration that simplifies the use of off-the-shelf applications.

I would argue that having a `deploy/kustomize` directory full of templates that require pre-processing in order for them to be useful would defeat the purpose of introducing kustomize. Furthermore, if we can't drop in a `kustomization.yaml` next to our kustomize manifests and run `kubectl apply -k` then how do we expect a customer to do so? From my perspective templates that require a shell script to process is a no-go.

> The idea is the user could only edit `customized` folder's files and leave the `base` untouched, since it contains the default configuration.

My understanding of the use of Kustomize was around the idea that a consumer of these manifests would take them as they are, drop their own `kustomization.yaml`, add overlays if they wish, and `oc apply -k`. Expecting them to edit existing overlays doesn't sound like the right approach.

> Your last sentence was truncated.

:cry:

> I tried to put the bases into `resources` section (as it was advised it's supported in new kustomize version), but I got an error when specifying a directory there.

Looks like the version of kustomize included in `oc` and `kubectl` is pretty old. What you have done is correct.

> Because the idea is to cover all use cases implemented in the current deploy scripts (`deploy_marketplace.sh` and `deploy_imageregistry.sh`), using a unified, kustomize-based mechanism. Are you suggesting to drop out some setups? I know QE are using the `private_repo`.

No. I am not suggesting that we drop any setups. I **do** think it is a good idea to consolidate the `deploy_marketplace.sh` and `deploy_imageregistry.sh` into one thing. It is not clear to me that the `private_repo` is useful, or really even "kustomize-based", since it only works as a template processed by something else (in this case a `deploy_kustomize.sh` script).

I want to attempt again to address the idea that we would benefit from separating the "how to make HCO available" problem from the "how to install HCO via OLM" problem. The first reason that this separation makes sense is that we already have the idea of `CONTENT_ONLY`, which means **only** make HCO available via OLM and **do not** install HCO via OLM. The second reason this makes sense is that the namespaces referenced "making HCO available" differs (`openshift-marketplace || openshift-operator-lifecycle-manager`) from the namespaces referenced when installing HCO via OLM (`kubevirt-hyperconverged`).

What would this look like? Instead of trying to go through all of the use cases, I will try and write out what one case would look like (`CONTENT_ONLY == "false"` and `PRIVATE_REPO == "true"`):

```bash
TMPDIR=$(mktemp -d)
cp -r $SCRIPT_DIR/* $TMPDIR

if [ "$PRIVATE_REPO" = 'true' ]; then
  get_quay_token
  oc create secret generic quay-registry-kubevirt-hyperconverged --from-literal=token=${QUAY_TOKEN}

  cat <<EOF > $TMPDIR/kustomization.yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

bases:
  - private_repo
EOF

  oc apply -k $TMPDIR
  exit 0
fi

# Then handle install
if [ $CONTENT_ONLY = 'true' ]; then
  exit 0
fi

cat <<EOF > $TMPDIR/kustomization.yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

bases:
  - base

resources:
  - namespace.yaml
EOF

oc apply -k $TMPDIR
# do more stuff
```

The important things about this example:

1. It is not complete, it is simply meant to communicate the idea.
1. The content delivery problem is separate from the install problem (ie. 2x calling `oc apply -k`)
1. The `secret.yaml` was removed from `private_repo` in favor of simply creating the secret in the script.
1. This example relies on a `namespace.yaml` (file tree included below for reference).

For reference, the tree looks like:

```
deploy/kustomize
├── base
│   ├── hco_cr.yaml
│   ├── kustomization.yaml
│   ├── operator_group.yaml
│   └── subscription.yaml
├── deploy_kustomize.sh
├── image_registry
│   ├── catalog_source.yaml
│   └── kustomization.yaml
├── kvm_emulation
│   ├── kustomization.yaml
│   └── subscription.patch.yaml
├── marketplace
│   ├── catalog_source_config.yaml
│   ├── kustomization.yaml
│   └── operator_source.yaml
├── namespace.yaml
├── private_repo
│   ├── kustomization.yaml
│   └── operator_source.patch.yaml
└── README.md
```

orenc1:
Hi @djzager, I followed your suggested approach and now the `kustomization.yaml` files are being manufactured and applied along the way of the control flow.
Meaning, each layer's manifests (private repo, marketplace, image_registry) are being applied, and lastly the installation is taking place (if content_only!=true).

"templates"  and "envsubst" are not being used any more.

One thing that is missing from the puzzle, is the way the user can modify some of the given fields, e.g. namespace name, hco version and channel, hco image url (grpc)
Should it be done via the script? (then I'll need to use `sed` or `envsubst`) or should the user edit the yamls himself before applying?

deploy/kustomize/deploy_kustomize.sh
```diff
@@ -0,0 +1,91 @@
...
+  cat <<EOF > $TMPDIR/kustomization.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+bases:
+  - private_repo
+EOF
+  oc apply -k $TMPDIR
+
+  else # not private repo
```
djzager, line 40:
```suggestion
else # not private repo
```

djzager: this whole nest if/else block has some indentation issues that make it hard to read.

deploy/kustomize/deploy_kustomize.sh
```diff
@@ -0,0 +1,91 @@
...
+QUAY_TOKEN=""
+CREATE_NAMESPACE="${CREATE_NAMESPACE:-false}"
+CONTENT_ONLY="${CONTENT_ONLY:-false}"
+KVM_EMULATION="${KVM_EMULATION:-false}"
+OC_TOOL="${OC_TOOL:-oc}"
+
+#####################
+
+SCRIPT_DIR="$(dirname "$0")"
+source $SCRIPT_DIR/get_quay_token.sh
```
djzager, line 22: It is a good idea to just go ahead and define the function here.

deploy/kustomize/deploy_kustomize.sh
```diff
@@ -0,0 +1,91 @@
...
+EOF
+  oc apply -k $TMPDIR
+  fi
+fi
+
+if [ "$CONTENT_ONLY" = 'true' ]; then
+  echo INFO: Content is ready for deployment in OLM.
+  exit 0
+fi
+source $SCRIPT_DIR/retry_loop.sh
```
djzager, line 66: It is a good idea to just go ahead and define the function here. Maybe up with the `get_quay_token`. Don't see a lot of value in putting into a separate script.

deploy/kustomize/deploy_kustomize.sh
orenc1: I moved `get_quay_token` and `retry_loop` functions to separate files mainly for readability, since in bash you must write the function definition before calling it. Unless we'll do something like [this](https://unix.stackexchange.com/questions/449498/call-function-declared-below), and we could put the auxiliary functions at the bottom of main script.

What do you think?

deploy/kustomize/base/operator_group.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: operators.coreos.com/v1
+kind: OperatorGroup
+metadata:
+  name: kubevirt-hyperconverged-group
+  namespace: kubevirt-hyperconverged
+spec:
+  targetNamespaces:
+    - kubevirt-hyperconverged
```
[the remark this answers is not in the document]
orenc1: The Subscription and HCO CR also need to consume the namespace, in their `metadata.namespace` fields.
I'll include a patch for them as well.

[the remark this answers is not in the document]
orenc1: PR is now updated, incl. readme file.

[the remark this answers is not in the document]
djzager: I missed it on the first glance but you should be able to specify `namespace: my-custom-namespace` in the kustomization.yaml that matches the metadata.name in the `namespace.yaml` resource specified and all of the subscription, operator group, and CR would be updated to use the correct namespace.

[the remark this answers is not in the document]
orenc1: I made an update utilizing that feature. I extracted the customized namespace from `namespace.yaml` and use that value to set $TARGET_NAMESPACE (instead of getting from env var) and put that on the base `kustomization.yaml`.

deploy/kustomize/deploy_kustomize.sh
```diff
@@ -0,0 +1,156 @@
...
+QUAY_PASSWORD="${QUAY_PASSWORD:-}"
+CONTENT_ONLY="${CONTENT_ONLY:-false}"
+KVM_EMULATION="${KVM_EMULATION:-false}"
+OC_TOOL="${OC_TOOL:-oc}"
+
+#####################
+
+main() {
+  SCRIPT_DIR="$(dirname "$0")"
+  TARGET_NAMESPACE=$(grep name: $SCRIPT_DIR/namespace.yaml | awk '{print $2}')
```
djzager, line 21: The other option is to take the`TARGET_NAMESPACE` argument (or environment variable) and simply write out the `namespace.yaml`.

deploy/kustomize/README.md
```diff
@@ -0,0 +1,44 @@
+# Intro
```
djzager, line 1: I recommend making the header something like "Kustomize Manifests" and simply describe this directory as manifests for delivering and deploying the HCO via OLM using Kustomize.

deploy/kustomize/README.md
```diff
@@ -0,0 +1,44 @@
...
+The KubeVirt Hyperconverged Cluster Operator (HCO) is delivered and deployed on a running OCP/OKD cluster using the kustomize method. 
+
+# Deploy Modes
+HCO can be delivered and deployed in several configurations, corresponding to several pre-built kustomize overlays, which are explained below.
+## Delivery
```
djzager, line 6: nit. In general I find Markdown easier to read when there are empty lines above and below headings.

deploy/kustomize/README.md
djzager: Instead of having the "Deploy Modes" as you do below. I recommend that here, you mention that there are two necessary steps to install HCO via OLM in a cluster 1) "delivery" - make OLM aware of HCO and 2) "deploy" - use OLM provided APIs to install HCO.

deploy/kustomize/README.md
```diff
@@ -0,0 +1,154 @@
+# Deploy HCO using kustomize
+The KubeVirt Hyperconverged Cluster Operator (HCO) is delivered and deployed on a running OCP/OKD cluster using the kustomize method. 
+
+# Kustomize Manifests
+In order to install HCO on your cluster, two necessary steps to be performed:
+1. **Delivery** - Make HCO recognized and available for the operator-lifecycle-manager (OLM).
+2. **Deployment** - Use OLM provided resources and APIs to deploy HCO on the cluster.
+
+## Directory Tree
```
djzager, line 9: To keep from needing to update this readme anytime we modify this directory, let's just drop the directory tree.

tiraboschi:
ci/prow/hco-e2e-upgrade-aws currently runs on OCP 4.5 and it's going to fail due to a side effect in the fix of https://bugzilla.redhat.com/1825330
It will be addressed on operator-framework/operator-lifecycle-manager#1470 on OLM side.
Temporary ignoring it

djzager:
/test images

Based on https://github.com/kubevirt/hyperconverged-cluster-operator/pull/463#issuecomment-621307110

deploy/kustomize/deploy_kustomize.sh
```diff
@@ -0,0 +1,156 @@
...
+  if [ "$token" == "null" ]; then
+    echo [ERROR] Got invalid Token from Quay. Please check your credentials in QUAY_USERNAME and QUAY_PASSWORD.
+    exit 1
+  else
+    QUAY_TOKEN=\"$token\";
+  fi
+}
+
+# Deploy HCO and OLM Resources with retries
+retry_loop() {
```
fabiand, line 114: @dankenigsberg just FYI - this is a retryloop we need to mitigae the not-auto-create-cr

dankenigsberg: Yes, I'm aware of it. If we auto-create, users with default requirements could avoid this retry_loop. But we cannot avoid in code if we want to support user kustomization of the HCO.cr. We currently have nothing there, but sooner or later, we would.

PR #471, opened 2020-02-24
Set Version and Status for the kubevirt-hyperconverged instance

nunnatsa:
Partial fix for [Bug 1716329](https://bugzilla.redhat.com/show_bug.cgi?id=1716329)

Add version and status in the HCO page in openshift UI for the kubevirt-hyperconverged instance.

Signed-off-by: Nahshon Unna-Tsameret <nunntsa@redhat.com>

### DIFF
--- a/pkg/apis/hco/v1alpha1/hyperconverged_types.go
+++ b/pkg/apis/hco/v1alpha1/hyperconverged_types.go
@@ -1,6 +1,7 @@
 package v1alpha1
 
 import (
+	ownVersion "github.com/kubevirt/hyperconverged-cluster-operator/version"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
@@ -26,6 +27,42 @@ type HyperConvergedSpec struct {
 
 	// LocalStorageClassName the name of the local storage class.
 	LocalStorageClassName string `json:"LocalStorageClassName,omitempty"`
+
+	// Deprecated: Version is the HCO version
+	// +optional
+	Version string `json:"version,omitempty"`
+}
+
+// HyperConvergedDeployStatus defines the HCO deployment phases
+type HyperConvergedDeployStatus string
+
+const (
+	HyperConvergedDeployedStatus  HyperConvergedDeployStatus = "Deployed"
+	HyperConvergedDeployingStatus HyperConvergedDeployStatus = "Deploying"
+	HyperConvergedDeletingStatus  HyperConvergedDeployStatus = "Deleting"
+	HyperConvergedDegradedStatus  HyperConvergedDeployStatus = "Degraded"
+)
+
+const operatorVersionName = "operator"
+
+// operatorVersion implements openshift operator versioning format, e.g.
+//   status:
+//     versions:
+//     - name: operator
+//       version: 1.2.3
+type operatorVersion struct {
+	Name    string `json:"name,omitempty"`
+	Version string `json:"version,omitempty"`
+}
+
+// HCO version
+var operatorVersions = []operatorVersion{
+	{Name: operatorVersionName, Version: ownVersion.Version},
+}
+
+// get HCO version in openshift operator format
+func GetOperatorVersions() []operatorVersion {
+	return operatorVersions
 }
 
 // HyperConvergedStatus defines the observed state of HyperConverged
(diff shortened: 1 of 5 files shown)


pkg/apis/hco/v1alpha1/hyperconverged_types.go
```diff
@@ -23,8 +23,20 @@ type HyperConvergedSpec struct {
...
+	Version string `json:"version,omitempty"`
 }
 
+// HyperConvergedDeployPhase defines the HCO deployment phases
+type HyperConvergedDeployPhase string
+
+const HyperConvergedDeployingPhase = HyperConvergedDeployPhase("Deploying")
+const HyperConvergedDegradedPhase = HyperConvergedDeployPhase("Degraded")
+const HyperConvergedDeployedPhase = HyperConvergedDeployPhase("Deployed")
+const HyperConvergedDeletingPhase = HyperConvergedDeployPhase("Deleting")
```
djzager: nit. Something like:

```
const (
    HyperConvergedDeployingPhase HyperConvergedDeployPhase = "Deploying"
    ...
)
```

Would probably be better.

pkg/apis/hco/v1alpha1/hyperconverged_types.go
```diff
@@ -23,8 +23,20 @@ type HyperConvergedSpec struct {
 
 	// LocalStorageClassName the name of the local storage class.
 	LocalStorageClassName string `json:"LocalStorageClassName,omitempty"`
+
+	// Version is the HCO version
+	// +optional
+	Version string `json:"version,omitempty"`
```
djzager: If the version is not configurable...why would we add it to the spec?

djzager:
Looking at this change. I would be happy if we only introduced a single phase "Deployed" **and** moved the `version` onto the status.

pkg/components/components.go
```diff
@@ -472,6 +473,10 @@ func GetOperatorCR() *hcov1alpha1.HyperConverged {
 		Spec: hcov1alpha1.HyperConvergedSpec{
 			BareMetalPlatform:     false,
 			LocalStorageClassName: "",
+			Version:               ownVersion.Version,
+		},
+		Status: hcov1alpha1.HyperConvergedStatus{
+			Phase: "Deploying",
```
tiraboschi: Better to use the const if we have

nunnatsa:
> I only have a few nitpicky comments. However, I don't really agree with https://bugzilla.redhat.com/show_bug.cgi?id=1716329 as the UI forcing us to use Phase seems wrong.
> 
> Is there any way we can go back to the UI people and figure out an actual pattern for operators to communicate their status in the UI? (ie. some ConditionType + Version on the status of the resource).

I'm not sure how UI works, but I guess that adding logic in front-end tod decide for each custom operator what is the current deployment phase, based on customized information, is a challenging task. A specific operator can understand its deployment phase and declare it. I think it's easier.

pkg/apis/hco/v1alpha1/hyperconverged_types.go
nunnatsa: Not sure I follow - currently, the UI takes it from the spec. Is there anyway else to set it?

nunnatsa:
> > I only have a few nitpicky comments. However, I don't really agree with https://bugzilla.redhat.com/show_bug.cgi?id=1716329 as the UI forcing us to use Phase seems wrong.
> > Is there any way we can go back to the UI people and figure out an actual pattern for operators to communicate their status in the UI? (ie. some ConditionType + Version on the status of the resource).
> 
> I'm not sure how UI works, but I guess that adding logic in front-end to decide for each custom operator what is the current deployment phase, based on customized information, is a challenging task. A specific operator can understand its deployment phase and declare it. I think it's easier.

Got it now - conditions are standard

pkg/apis/hco/v1alpha1/hyperconverged_types.go
djzager: We aren't allowing configuration of this value. So it should not be on the spec.

We could create an issue against the UI to make it so they could read it off the status.

djzager:
> I'm not sure how UI works, but I guess that adding logic in front-end to decide for each custom operator what is the current deployment phase, based on customized information, is a challenging task. A specific operator can understand its deployment phase and declare it. I think it's easier.

The issue with this is that Phase is deprecated, from [api-conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md):

> Some resources in the v1 API contain fields called phase, and associated message, reason, and other status fields. The pattern of using phase is deprecated. Newer API types should use conditions instead. Phase was essentially a state-machine enumeration field, that contradicted system-design principles and hampered evolution, since adding new enum values breaks backward compatibility. Rather than encouraging clients to infer implicit properties from phases, we prefer to explicitly expose the individual conditions that clients need to monitor. Conditions also have the benefit that it is possible to create some conditions with uniform meaning across all resource types, while still exposing others that are unique to specific resource types. See #7856 for more details and discussion.

nunnatsa:
According to this:
https://projects.invisionapp.com/share/7TRU0EYHJNM#/screens/361317782
it is possible to use `status.status` or `status.condition` for the `status` column. This is not working in openshift 4.3

The `version` column is only taken from the `spec.version` field, even in the newer versions.

djzager:
> According to this:
> https://projects.invisionapp.com/share/7TRU0EYHJNM#/screens/361317782
> it is possible to use `status.status` or `status.condition` for the `status` column. This is not working in openshift 4.3

This may be one of those things that you have to set in the CRD itself like [printer columns](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/deploy/crds/hco00.crd.yaml#L7)

nunnatsa:
> > According to this:
> > https://projects.invisionapp.com/share/7TRU0EYHJNM#/screens/361317782
> > it is possible to use `status.status` or `status.condition` for the `status` column. This is not working in openshift 4.3
> 
> This may be one of those things that you have to set in the CRD itself like [printer columns](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/deploy/crds/hco00.crd.yaml#L7)

version - maybe. But this will be a manual task for each version, I think. Also, it doesn't solve the status issue, which is dnamic.

nunnatsa:
@djzager, @tiraboschi - please re-review

nunnatsa:
Continue in #616

PR #473, opened 2020-02-25
Set default value (BlockUninstallIfWorkloadsExist) for UninstallStrategy on CDI

tiraboschi:
Set default value (BlockUninstallIfWorkloadsExist) for UninstallStrategy on CDI

github.com/kubevirt/containerized-data-importer/pull/1118 introduced uninstall
strategies for the CDI-operator.
This PR will:

- set BlockUninstallIfWorkloadsExist as the UninstallStrategy creating CDI CR
- set UninstallStrategy=BlockUninstallIfWorkloadsExist if unset on an existing CDI CR
- test it

### DIFF
--- a/pkg/controller/hyperconverged/hyperconverged_controller.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller.go
@@ -535,11 +535,15 @@ func newCDIForCR(cr *hcov1alpha1.HyperConverged, namespace string) *cdiv1alpha1.
 	labels := map[string]string{
 		"app": cr.Name,
 	}
+	uninstallStrategy := cdiv1alpha1.CDIUninstallStrategyBlockUninstallIfWorkloadsExist
 	return &cdiv1alpha1.CDI{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "cdi-" + cr.Name,
 			Labels:    labels,
 			Namespace: namespace,
+		},
+		Spec: cdiv1alpha1.CDISpec{
+			UninstallStrategy: &uninstallStrategy,
 		},
 	}
 }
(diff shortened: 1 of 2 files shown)


pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -580,6 +584,15 @@ func (r *ReconcileHyperConverged) ensureCDI(instance *hcov1alpha1.HyperConverged
 		}
 	}
 
+	if !reflect.DeepEqual(found.Spec, cdi.Spec) {
+		defaultUninstallStrategy := cdiv1alpha1.CDIUninstallStrategyBlockUninstallIfWorkloadsExist
```
nunnatsa: Maybe move this line into the `if` block?

tiraboschi: Sorry, I don't understand, do you mean inside `if found.Spec.UninstallStrategy == nil {` block?

nunnatsa: Right - sorry for the confusion.

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -580,6 +584,15 @@ func (r *ReconcileHyperConverged) ensureCDI(instance *hcov1alpha1.HyperConverged
 		}
 	}
 
+	if !reflect.DeepEqual(found.Spec, cdi.Spec) {
+		if found.Spec.UninstallStrategy == nil {
```
orenc1: What if `found.Spec.UninstallStrategy` is not nil but some other invalid value that the user inserted manually into the CR spec? Do we have a validator there?

tiraboschi: No, we don't. But I think we can only respect the user provided value that should be eventually detected on CDI operator side.
A nil value is instead different because we know that in the past we didn't set any value and so we want to upgrade it to the current default which is not nil for new deployments.

orenc1: > No, we don't. But I think we can only respect the user provided value that should be eventually detected on CDI operator side.
> A nil value is instead different because we know that in the past we didn't set any value and so we want to upgrade it to the current default which is not nil for new deployments.

So we can't reconcile to default value if, for example, the user had a typo in setting this field in the CR.
Maybe we can check if the existing value is matching one of the permitted values of `UninstallStrategy`, otherwise - reconcile to default.
WDYT?

PR #479, opened 2020-02-28
Hide internal CRDs

tiraboschi:
With github.com/operator-framework/operator-lifecycle-manager/pull/1097
OM supports a mechanism to hide internal CRDs on OLM console.
Start using it.
By default only
- hyperconvergeds.hco.kubevirt.io
- v2vvmwares.kubevirt.io
- hostpathprovisioners.hostpathprovisioner.kubevirt.io
are going to be visible.
The list of visible CRDs can be override as a CLI parameter
to the csv-merger tool.

```release-note
Hide internal CRDs: by default only hyperconvergeds.hco.kubevirt.io, v2vvmwares.kubevirt.io and hostpathprovisioners.hostpathprovisioner.kubevirt.io will be visible in OLM console
```

### DIFF
--- a/tools/csv-merger/csv-merger.go
+++ b/tools/csv-merger/csv-merger.go
@@ -83,7 +83,9 @@ var (
 	namespace           = flag.String("namespace", "kubevirt-hyperconverged", "Namespace")
 	crdDisplay          = flag.String("crd-display", "KubeVirt HyperConverged Cluster", "Label show in OLM UI about the primary CRD")
 	csvOverrides        = flag.String("csv-overrides", "", "CSV like string with punctual changes that will be recursively applied (if possible)")
-	relatedImagesList   = flag.String("related-images-list", "",
+	visibleCRDList      = flag.String("visible-crds-list", "hyperconvergeds.hco.kubevirt.io,v2vvmwares.kubevirt.io,hostpathprovisioners.hostpathprovisioner.kubevirt.io",
+		"Comma separated list of all the CRDs that should be visible in OLM console")
+	relatedImagesList = flag.String("related-images-list", "",
 		"Comma separated list of all the images referred in the CSV (just the image pull URLs or eventually a set of 'image|name' collations)")
 )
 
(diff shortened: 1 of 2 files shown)


tiraboschi:
It fails on:
```
info: Loading sha256:0a7f5552ad858612f864cbb0afc878f87742317d5853c16702583c1fd3c1a00d sig-storage-local-static-provisioner
error: unable to read image registry.svc.ci.openshift.org/ci-op-grzl0qxg/stable@sha256:c3072c3a648455d2820701f1a254efcef84787519e7ba1e6f75c046b219dedcb: received unexpected HTTP status: 504 Gateway Time-out
2020/02/28 17:18:29 Container release in pod release-latest failed, exit code 1, reason Error
2020/02/28 17:18:29 No custom metadata found and prow metadata already exists. Not updating the metadata.
2020/02/28 17:18:31 Ran for 1m6s
```
so unrelated to this PR.
Let's try tomorrow

tiraboschi:
[Failed on OCP 4.3.0](https://pasteboard.co/IWO6tV2.png)

Failed for me on OCP 4.3.0: the annotation is in the CSV but all the CRDs are still visible.

tiraboschi:
While if I try to put the value in a yaml list it complains that it [expects a string](https://pasteboard.co/IWOwczF.png)

tiraboschi:
The relevant feature got merged on Openshift console on Jan 23: 
https://github.com/openshift/console/pull/3877

so this should be just about testing it on a more recent OCP build.

tiraboschi:
Verified on OCP 4.4

tools/csv-merger/csv-merger.go
```diff
@@ -83,7 +83,9 @@ var (
 	namespace           = flag.String("namespace", "kubevirt-hyperconverged", "Namespace")
 	crdDisplay          = flag.String("crd-display", "KubeVirt HyperConverged Cluster", "Label show in OLM UI about the primary CRD")
 	csvOverrides        = flag.String("csv-overrides", "", "CSV like string with punctual changes that will be recursively applied (if possible)")
-	relatedImagesList   = flag.String("related-images-list", "",
+	visibleCRDList      = flag.String("visible-crds-list", "hyperconvergeds.hco.kubevirt.io,v2vvmwares.kubevirt.io,hostpathprovisioners.hostpathprovisioner.kubevirt.io",
```
djzager: my only question is whether or not this needs to be a command line argument. It looks to me like the kind of thing we would always change in the source.

Here is my thought, specifically create vars for visible and hidden CRDs, if we see one that isn't in either `panic` because we missed something. WDYT?

tiraboschi: I added it as CLI argument (with a reasonable default) because I cannot exclude we can have a different configuration between upstream and downstream

PR #481, opened 2020-02-29
Improve upgrade test suite

tiraboschi:
Up to now, the upgrade test lane on CI was testing the upgrade, via OLM,
using a locally built registry image, which already contains a bundle (CSV+CRDs)
from the PR content, as the baseline for the upgrade.
That bundle is pointing to an already available HCO image from quay.io
( quay.io/kubevirt/hyperconverged-cluster-operator:latest ) but everything else comes
from the open PR.
The upgrade suite was going to upgrade from there to a second
locally built registry image which is exactly equal to the first one but it
point to a locally built HCO image with the code from the PR.

This approach was wrong because:
- nothing ensures that the HCO image already available on quay.io
can be successfully used with the bundle (CRDs + CSV) contained in the
PR and so with the locally built pre-upgrade registry image
- nothing tries to upgrade the dependent components from what is tracked
in an already available bundle shipped in the past to the content of the current PR
(we were testing only quay.io/kubevirt/hyperconverged-cluster-operator:latest -> `HCO from current PR`
ignoring other components that can be included with the PR).

A better approach is to start the upgrade from:
- an HCO image built in the past and shipped in quay.io
- a registry image built in the past and shipped in quay.io

In that way we can ensure that we are able to correctly
upgrade from the latest images shipped in the past on quay.io to the content
of the current PR including dependent components.

So, in this PR:
- rename the registry image from hco-registry to hco-container-registry to better describe its content (keeping hco-registry-upgrade for the upgrade registry because it's hardcoded in prow job)
- start the upgrade from quay.io/kubevirt/hco-container-registry:latest as shipped in quay.io
- significatively raise various upgrade timeouts because now we are going to test a real upgrade potentially involving also other dependent components
- force KVM_EMULATION mode when we run on CI env without real virtualization capabilities
- print more logs on upgrade failures

To be addressed in future PRs:
- duplicate the upgrade lane to test a matrix of upgrades, for instance:
0.4.0 -> current PR
1.0.0 -> current PR
1.1.0 -> current PR


tiraboschi:
CI only
/release-note-none

orenc1:
CatalogSource pod is not getting `Ready` within 30 minutes. I suggest to print out its logs to see what's wrong if the retest fails.
/test hco-e2e-upgrade-aws

hack/upgrade-test.sh
```diff
@@ -206,7 +205,7 @@ ${CMD} patch catalogsource hco-catalogsource-example -n ${HCO_CATALOG_NAMESPACE}
 sleep 5
 ./hack/retry.sh 20 30 "${CMD} get pods -n ${HCO_CATALOG_NAMESPACE} | grep hco-catalogsource | grep -v Terminating"
 HCO_CATALOGSOURCE_POD=`${CMD} get pods -n ${HCO_CATALOG_NAMESPACE} | grep hco-catalogsource | grep -v Terminating | head -1 | awk '{ print $1 }'`
-${CMD} wait pod $HCO_CATALOGSOURCE_POD --for condition=Ready -n ${HCO_CATALOG_NAMESPACE} --timeout="120s"
+${CMD} wait pod $HCO_CATALOGSOURCE_POD --for condition=Ready -n ${HCO_CATALOG_NAMESPACE} --timeout="1800s"
```
orenc1:
```suggestion
if [ $? -eq 1 ]
then
  {CMD} describe $HCO_CATALOGSOURCE_POD -n ${HCO_CATALOG_NAMESPACE}
  {CMD} logs $HCO_CATALOGSOURCE_POD -n ${HCO_CATALOG_NAMESPACE}
fi
```

tiraboschi: Done, thanks

tiraboschi:
The patched catalog source pod fails with:
```
catalogsource.operators.coreos.com/hco-catalogsource-example patched
+ sleep 5
+ ./hack/retry.sh 20 30 'oc get pods -n openshift-marketplace | grep hco-catalogsource | grep -v Terminating'
oc get pods -n openshift-marketplace | grep hco-catalogsource | grep -v Terminating
hco-catalogsource-example-7gbqd         0/1     ErrImagePull   0          7s
```
...
```
+ oc wait pod hco-catalogsource-example-7gbqd --for condition=Ready -n openshift-marketplace --timeout=1800s
error: timed out waiting for the condition on pods/hco-catalogsource-example-7gbqd
````

let's try to fetch more logs.

PR #483, opened 2020-03-03
Adding namespace validator for HCO CRD

orenc1:
By adding the 'validation' section to the HCO CRD, the user can now only deploy the HCO CR on the specified namespace.
Otherwise he/she will encounter the following error message from the API server:
`The HyperConverged 'kubevirt-hyperconverged' is invalid: metadata.namespace: Invalid value: '': metadata.namespace in body should match 'kubevirt-hyperconverged`


djzager:
I think I have been taking too long with my changes. That is my fault. Created #484

djzager:
I think it's worth discussing whether or not the additional check on operator startup is necessary.

djzager:
@orenc1 and @nunnatsa any thoughts on #484 and which we should move forward with. I don't actually care but both shouldn't survive :sunglasses:

orenc1:
> @orenc1 and @nunnatsa any thoughts on #484 and which we should move forward with. I don't actually care but both shouldn't survive sunglasses

I might be biased (my PR..), but I believe if both PRs are intending to achieve the same outcome (as described in first comment), mine is much simpler, since it utilizes an existing kubernetes validation feature, directly in the CRD of the object we're trying to deploy.
I've already checked this in an existing cluster and it behaves as expected.
Furthermore, #484 is not passing the CI tests for some reason, and this seems to be passing them (two more, finger crossed..), and we would like to merge this feature before end of sprint.

djzager:
> I might be biased (my PR..), but I believe if both PRs are intending to achieve the same outcome (as described in first comment), mine is much simpler, since it utilizes an existing kubernetes validation feature, directly in the CRD of the object we're trying to deploy.
I've already checked this in an existing cluster and it behaves as expected.
Furthermore, #484 is not passing the CI tests for some reason, and this seems to be passing them (two more, finger crossed..), and we would like to merge this feature before end of sprint.

just so we are clear. The main change that is common between the two is the change to `components.go:GetOperatorCRD` to include a namespace parameter that is used when generating the CustomResourceValidation.

The only real difference between these two PRs is that mine additionally checks that the operator is deployed into the same namespace as what the HyperConverged resource is being restricted to.

How this plays out in your PR is that it is possible for someone to deploy the HCO into the `kubevirt-hyperrrconverged` namespace. You will only be able to create the HyperConverged resource with `types.NamespacedName{Name: 'hyperconverged-cluster', Namespace: 'kubevirt-hyperconverged'}` but the [controller will never act on it](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/pkg/controller/hyperconverged/hyperconverged_controller.go#L165-L175), additionally all of the dependent watches are reconciling to `types.NamespacedName{Name: 'hyperconverged-cluster', Namespace: 'kubevirt-hyperrrconverged'}` that will never exist.

orenc1:
> just so we are clear. The main change that is common between the two is the change to `components.go:GetOperatorCRD` to include a namespace parameter that is used when generating the CustomResourceValidation.
> 
> The only real difference between these two PRs is that mine additionally checks that the operator is deployed into the same namespace as what the HyperConverged resource is being restricted to.
> 
> How this plays out in your PR is that it is possible for someone to deploy the HCO into the `kubevirt-hyperrrconverged` namespace. You will only be able to create the HyperConverged resource with `types.NamespacedName{Name: 'hyperconverged-cluster', Namespace: 'kubevirt-hyperconverged'}` but the [controller will never act on it](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/pkg/controller/hyperconverged/hyperconverged_controller.go#L165-L175), additionally all of the dependent watches are reconciling to `types.NamespacedName{Name: 'hyperconverged-cluster', Namespace: 'kubevirt-hyperrrconverged'}` that will never exist.

IIUC, in this PR, the namespace restriction is not enforced when deploying the operator using OLM subscription, i.e. can be installed at any arbitrary namespace, and only the step of creating the HCO CR is restricted to the specified namespace?
In such case, the user will notice something is wrong when the validator will show the error message.
On the other hand, I tend to agree it's better to restrict the ns for deployment through OLM in first place.

tiraboschi:
handled in https://github.com/kubevirt/hyperconverged-cluster-operator/pull/484

PR #485, opened 2020-03-03
Emit a warning event on failed uninstall

tiraboschi:
```release-note
Emit a warning event on failed uninstall
```

Now Virt and CDI operators implements a validating
webhook that can prevent their CRs from being deleted
when we have existing VMs or DVs.
When the user will try to delete HCO CR, HCO will try
to delete Virt and CDI CRs and so it can fail there.
In that case, HCO CR is flagged for deletion and HCO
will periodically retry with an exponential
backoff mechanism.

Adding here the code to emit a warning event
to make it more evident to the user.

A better approach will be implemented in a future PR:
we are going to implement a validating webhook also here
that will try to delete virt and CDI resources in dry-run
mode and it will reject the deletion of HCO CR with a
clear error message if virt or cdi CRs cannot be deleted.

### DIFF
--- a/pkg/controller/hyperconverged/hyperconverged_controller.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller.go
@@ -1345,12 +1345,17 @@ func ensureKubeVirtDeleted(c client.Client, instance *hcov1alpha1.HyperConverged
 	key, err := client.ObjectKeyFromObject(virt)
 	if err != nil {
 		log.Error(err, "Failed to get object key for KubeVirt")
+		return err
 	}
 
 	found := &kubevirtv1.KubeVirt{}
 	err = c.Get(context.TODO(), key, found)
 
 	if err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Info("KubeVirt resource doesn't exist, there is nothing to remove")
+			return nil
+		}
 		log.Error(err, "Failed to get KubeVirt from kubernetes")
 		return err
 	}
@@ -1364,12 +1369,17 @@ func ensureCDIDeleted(c client.Client, instance *hcov1alpha1.HyperConverged) err
 	key, err := client.ObjectKeyFromObject(cdi)
 	if err != nil {
 		log.Error(err, "Failed to get object key for CDI")
+		return err
 	}
 
 	found := &cdiv1alpha1.CDI{}
 	err = c.Get(context.TODO(), key, found)
 
 	if err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Info("CDI resource doesn't exist, there is nothing to remove")
+			return nil
+		}
 		log.Error(err, "Failed to get CDI from kubernetes")
 		return err
 	}
@@ -1382,12 +1392,17 @@ func ensureNetworkAddonsDeleted(c client.Client, instance *hcov1alpha1.HyperConv
 	key, err := client.ObjectKeyFromObject(networkAddons)
 	if err != nil {
 		log.Error(err, "Failed to get object key for Network Addons")
+		return err
 	}
 
 	found := &networkaddonsv1alpha1.NetworkAddonsConfig{}
 	err = c.Get(context.TODO(), key, found)
 
 	if err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Info("NetworkAddonsConfig doesn't exist, there is nothing to remove")
+			return nil
+		}
 		log.Error(err, "Failed to get NetworkAddonsConfig from kubernetes")
 		return err
 	}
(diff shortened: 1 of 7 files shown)


pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -238,6 +243,11 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
 				err = f(r.client, instance)
 				if err != nil {
 					reqLogger.Error(err, "Failed to manually delete objects")
+					r.recorder.Event(instance, corev1.EventTypeWarning, ErrHCOUninstall, uninstallErrorMsg)
```
djzager: my only concern with this approach is that there are reasons that we could fail that have nothing to do with CDI or KubeVirt's validating webhooks.

tiraboschi: by the way we are neither trying to explicitly delete kubevirt but just CDI

tiraboschi: to keep it simple, let's try to be as generic as possible and consume the error message we get from the component without trying to parse the content

tiraboschi:
The event is visible in pods events on OLM console:
https://pasteboard.co/IXGAgvD.png

still not on CSV events page which can be more evident

tiraboschi:
Now we are correctly associating the event with the CSV so that it's really visible in the UI: https://pasteboard.co/IXQGlGH.png

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -228,14 +237,47 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
...
 				ensureKubeVirtCommonTemplateBundleDeleted,
 			} {
-				err = f(r.client, instance)
+				err = f(r.client, instance, request.Namespace)
 				if err != nil {
 					reqLogger.Error(err, "Failed to manually delete objects")
+
+					errT := ErrHCOUninstall
+					errMsg := uninstallHCOErrorMsg
+					switch i {
```
djzager: I think you should add a TODO to handle the case where a `Delete` fails that isn't because KubeVirt|CDI refused our delete request.

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -228,14 +237,47 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
...
+					case 0:
+						errT = ErrVirtUninstall
+						errMsg = uninstallVirtErrorMsg + err.Error()
+					case 1:
+						errT = ErrCDIUninstall
+						errMsg = uninstallCDIErrorMsg + err.Error()
+					}
+
+					pod, pod_err := hcoutil.GetPod(r.client, reqLogger)
```
djzager: My personal opinion is that this should be a `err := r.emitWarning` or somesuch that has all of this process inside (getting the pod, finding the CSV, and emitting the event).

pkg/controller/hyperconverged/hyperconverged_controller.go
tiraboschi: in that case we are already going to emit a generic error message:
https://github.com/kubevirt/hyperconverged-cluster-operator/pull/485/commits/9d30b92e0a22096e5f16b77572b236ac4a36315e#diff-64e1e2b9251fa5137c91d5a7c8427b88R70

pkg/controller/hyperconverged/hyperconverged_controller.go
tiraboschi: I agree and I think we can reuse that in the future

pkg/controller/hyperconverged/hyperconverged_controller.go
djzager: My issue is that it is possible that our call to `ensureKubeVirtDeleted` or `ensureCDIDeleted` to return an err and it not be because KubeVirt or CDI refused our request.

tiraboschi: now I got your point, thanks

tiraboschi:
With the second commit we can now correctly continue to try deleting CDI CR even if we already deleted kubevirt CR in the past:
https://pasteboard.co/IYr66f3.png

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -1275,54 +1340,93 @@ func componentResourceRemoval(o interface{}, c client.Client, cr *hcov1alpha1.Hy
 	return err
 }
 
-func ensureCDIDeleted(c client.Client, instance *hcov1alpha1.HyperConverged) error {
+func ensureKubeVirtDeleted(c client.Client, instance *hcov1alpha1.HyperConverged, requestNamespace string) error {
+	virt := newKubeVirtForCR(instance, requestNamespace)
```
djzager: isn't this just `instance.Namespace`?

tiraboschi: formally not, but yes, now we are 100% sure that it's the same value by design (being a real singleton).
Let's simplify this one.

tiraboschi:
Ignoring reduced coverage  because we cannot really test delete prevention mechanism without OLM

PR #466, opened 2020-02-21
bump SSP to v1.0.23

tiraboschi:
```release-note
bump SSP to v1.0.23
```

Signed-off-by: Simone Tiraboschi <stirabos@redhat.com>

PR #486, opened 2020-03-04
Revert "bump SSP to v1.0.23"

tiraboschi:
We are still facing an issue on SSP operator that returns
```
  customresourcedefinitions:
    owned:
    - description: Represents a deployment of the predefined VM templates
      displayName: KubeVirt common templates
      kind: KubevirtCommonTemplatesBundle
      name: kubevirtcommontemplatesbundles.kubevirt.io
```

while it should be `name: kubevirtcommontemplatesbundles.ssp.kubevirt.io`

Reverting this now to unblock other changes

Reverts kubevirt/hyperconverged-cluster-operator#466


tiraboschi:
The issue is downstream specific, ignore this

PR #496, opened 2020-03-12
Add pull request template to write release note

maya-r:
Based on CDI's templates.

Hopefully they will remind people to write release notes instead of being pinged to fix it by kubevirt-bot :-)

```release-note
NONE
```
(ironic edit)

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,26 @@
+<!-- This form is for bug reports and feature requests ONLY!
+Also make sure that you visit our User Guide at https://kubevirt.io/user-guide/
+-->
+
+**Is this a BUG REPORT or FEATURE REQUEST?**:
+
+> Uncomment only one, leave it on its own line:
+>
+> /kind bug
+> /kind enhancement
+
+
+**What happened**:
+
+**What you expected to happen**:
+
+**How to reproduce it (as minimally and precisely as possible)**:
+
+**Anything else we need to know?**:
+
+**Environment**:
+- HCO version (use `oc get csv -n kubevirt-hyperconverged`):
+- Kubernetes version (use `kubectl version`):
+- Cloud provider or hardware configuration:
+- Install tools:
+- Others:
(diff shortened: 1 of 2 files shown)


orenc1:
I'm in favor of applying the template to issues but not sure if necessary in pull requests, as it would make it a bit clumsy. Maybe just the release-notes reminder.
/cc @tiraboschi

djzager:
I tend to agree with @orenc1 ...maybe we could compromise and make all of the PR template a collection of comments for reference when making a PR. That way if you do nothing we don't have a bunch of useless text. If you need a reminder on how to add release notes, however, it's there.

orenc1:
maya why did you remove the _issues_ template? I think it's useful.

maya-r:
> maya why did you remove the _issues_ template? I think it's useful.

Misread, it's back.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,27 @@
...
+**What happened**:
+
+**What you expected to happen**:
+
+**How to reproduce it (as minimally and precisely as possible)**:
+
+**Anything else we need to know?**:
+
+**Environment**:
+- CDI version (use `kubectl get deployments hyperconverged-cluster-operator -o yaml`):
```
orenc1: CDI version is irrelevant in our case.
If it was deployed via OLM, it should be:
```suggestion
- HCO version (use `oc get csv -n kubevirt-hyperconverged`):
```
(notice the change from `kubectl` to `oc`, since OLM available in openshift.
And if it was deployed without OLM
In such case, the version is extracted in the following way:
```
HCO_IMAGE=`kubectl get deployment hco-operator -n kubevirt-hyperconverged -o=jsonpath='{.spec.template.spec.containers[0].image}'`
skopeo inspect docker://$HCO_IMAGE | jq '.Labels.version'
```

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,9 @@
+**Release note**:
+<!--  Write your release note:
+1. Enter your extended release note in the below block. If the PR requires additional action from users switching to the new release, include the string "action required".
+2. If no release note is required, just write "NONE".
+-->
+```release-note
+
+```
```
djzager: I think what I had in mind was:

```
<!-- 
**Release note**:
Write your release note:
1. Enter your extended release note in the below block. If the PR requires additional action from users switching to the new release, include the string "action required".
2. If no release note is required, just write "NONE".
..release not block that I can't include in this fence
-->
```

This way if you do nothing on PR submit it won't get in the way.

PR #497, opened 2020-03-12
Let the user set a custom value for smbios at CSV level

tiraboschi:
Let the user set a custom value for smbios in
KubeVirt ConfigMap consuming a new env variable
set in the CSV.

Set it on new deployemnts.
Update to the value in the CSV on upgrades.

### DIFF
--- a/deploy/olm-catalog/kubevirt-hyperconverged/1.1.0/kubevirt-hyperconverged-operator.v1.1.0.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/kubevirt-hyperconverged/1.1.0/kubevirt-hyperconverged-operator.v1.1.0.clusterserviceversion.yaml
@@ -9,7 +9,7 @@ metadata:
     categories: OpenShift Optional
     certified: "false"
     containerImage: quay.io/kubevirt/hyperconverged-cluster-operator:1.1.0
-    createdAt: "2020-03-09 14:34:54"
+    createdAt: "2020-03-13 13:02:31"
     description: |-
       **HyperConverged Cluster Operator** is an Operator pattern for managing multi-operator products.
       Specifcally, the HyperConverged Cluster Operator manages the deployment of KubeVirt,
@@ -1298,6 +1298,12 @@ spec:
                   value: quay.io/kubevirt/kubevirt-v2v-conversion:v2.0.0
                 - name: VMWARE_CONTAINER
                   value: quay.io/kubevirt/kubevirt-vmware:v2.0.0
+                - name: SMBIOS
+                  value: |-
+                    Family: KubeVirt
+                    Manufacturer: KubeVirt
+                    Product: None
+                - name: MACHINETYPE
                 image: quay.io/kubevirt/hyperconverged-cluster-operator:1.1.0
                 imagePullPolicy: IfNotPresent
                 name: hyperconverged-cluster-operator
--- a/pkg/controller/hyperconverged/hyperconverged_controller.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller.go
@@ -32,6 +32,7 @@ import (
 	hcoutil "github.com/kubevirt/hyperconverged-cluster-operator/pkg/util"
 	conditionsv1 "github.com/openshift/custom-resource-status/conditions/v1"
 	objectreferencesv1 "github.com/openshift/custom-resource-status/objectreferences/v1"
+	virtconfig "kubevirt.io/kubevirt/pkg/virt-config"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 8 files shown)


deploy/olm-catalog/kubevirt-hyperconverged/1.1.0/kubevirt-hyperconverged-operator.v1.1.0.clusterserviceversion.yaml
```diff
@@ -9,7 +9,7 @@ metadata:
     categories: OpenShift Optional
     certified: "false"
     containerImage: quay.io/kubevirt/hyperconverged-cluster-operator:1.1.0
-    createdAt: "2020-03-09 14:34:54"
+    createdAt: "2020-03-12 16:41:25"
```
orenc1: Can you drop this change?
It will cause a merge conflict when doing the cherry-pick.

tiraboschi: We are not going to cherry pick this to release-2.3.
But yes, we have to figure out if we can set any special git filter to ignore merge conflicts on the timestamp.

tiraboschi: Maybe this: https://git-scm.com/docs/gitattributes#_defining_a_custom_merge_driver

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -434,6 +434,11 @@ func newKubeVirtConfigForCR(cr *hcov1alpha1.HyperConverged, namespace string) *c
 			"selinuxLauncherType": "spc_t",
 		},
 	}
+	val, ok := os.LookupEnv("SMBIOS")
```
vatsalparekh: Maybe it makes sense to look for all the list of env variables here, for example machine type and smbios both

tiraboschi: currently we are not setting machine type in kubevirt-config, how is it named?

tiraboschi: found: https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-config/config-map.go#L49

test.sh
```diff
@@ -0,0 +1,1799 @@
+/home/stirabos/go/src/github.com/kubevirt/hyperconverged-cluster-operator/tools/csv-merger/csv-merger '--cna-csv=apiVersion: operators.coreos.com/v1alpha1
```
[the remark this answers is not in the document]
tiraboschi: right, sorry

tiraboschi:
> Is there a scenario where we want `smbios` to be empty?

Up to today we weren't passing any value.
The root issue is that we want to use different values upstream and downstream and, due to that, hardcoding it in the source is not a a good idea.

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -434,6 +435,15 @@ func newKubeVirtConfigForCR(cr *hcov1alpha1.HyperConverged, namespace string) *c
...
+	if ok && val != "" {
+		cm.Data[virtconfig.SmbiosConfigKey] = val
+	}
+	val, ok = os.LookupEnv("MACHINETYPE")
+	if ok && val != "" {
+		cm.Data[virtconfig.MachineTypeKey] = val
```
orenc1: 1. IIUC, we are putting nested key-value maps `SMBIOS` and `MACHINETYPE` inside the kubevirt-config CM? It means the consumer must follow that.

2. The jira ticket mentioned also an SKU version (which is the kubevirt version). Where is this referenced? inside SMBIOS internal key-value map?

tiraboschi: > 1. IIUC, we are putting nested key-value maps `SMBIOS` and `MACHINETYPE` inside the kubevirt-config CM? It means the consumer must follow that.

kubevirt expects a json or yaml serialized string, the user should put there what kubevirt expects.
https://github.com/kubevirt/kubevirt/blob/eaa5d58a86523ee4489e369f043eeb2a1ac7ed3b/pkg/virt-config/config-map.go#L291

in hack/build-manifests.sh we are injecting a small piece of yaml

> 2. The jira ticket mentioned also an SKU version (which is the kubevirt version). Where is this referenced? inside SMBIOS internal key-value map?

Yes, the value of Product key in smbios map

PR #498, opened 2020-03-15
Grant authenticated users access to read kubevirt-storage-class-defaults

maya-r:
[Bugzilla bug report](https://bugzilla.redhat.com/show_bug.cgi?id=1784467).

```release-note
Grant authenticated users access to read kubevirt-storage-class-defaults
```

### DIFF
--- a/pkg/controller/hyperconverged/hyperconverged_controller.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller.go
@@ -476,8 +476,6 @@ func (r *ReconcileHyperConverged) ensureHco(req *hcoRequest) error {
 		r.ensureKubeVirtPriorityClass,
 		r.ensureKubeVirtConfig,
 		r.ensureKubeVirtStorageConfig,
-		r.ensureKubeVirtStorageRole,
-		r.ensureKubeVirtStorageRoleBinding,
 		r.ensureKubeVirt,
 		r.ensureCDI,
 		r.ensureNetworkAddons,
@@ -959,6 +957,16 @@ func (r *ReconcileHyperConverged) ensureCDI(req *hcoRequest) (upgradeDone bool, 
 	}
 
 	req.logger.Info("CDI already exists", "CDI.Namespace", found.Namespace, "CDI.Name", found.Name)
+
+	err = r.ensureKubeVirtStorageRole(req)
+	if err != nil {
+		return false, err
+	}
+
+	err = r.ensureKubeVirtStorageRoleBinding(req)
+	if err != nil {
+		return false, err
+	}
 
 	existingOwners := found.GetOwnerReferences()
 
@@ -1530,10 +1538,10 @@ func newKubeVirtStorageConfigForCR(cr *hcov1alpha1.HyperConverged, namespace str
 	}
 }
 
-func (r *ReconcileHyperConverged) ensureKubeVirtStorageRole(req *hcoRequest) (upgradeDone bool, err error) {
+func (r *ReconcileHyperConverged) ensureKubeVirtStorageRole(req *hcoRequest) error {
 	kubevirtStorageRole := newKubeVirtStorageRoleForCR(req.instance, req.Namespace)
 	if err := controllerutil.SetControllerReference(req.instance, kubevirtStorageRole, r.scheme); err != nil {
-		return false, err
+		return err
 	}
 
 	key, err := client.ObjectKeyFromObject(kubevirtStorageRole)
@@ -1545,28 +1553,28 @@ func (r *ReconcileHyperConverged) ensureKubeVirtStorageRole(req *hcoRequest) (up
 	err = r.client.Get(req.ctx, key, found)
 	if err != nil && apierrors.IsNotFound(err) {
 		req.logger.Info("Creating kubevirt storage role")
-		return false, r.client.Create(req.ctx, kubevirtStorageRole)
+		return r.client.Create(req.ctx, kubevirtStorageRole)
 	}
 
 	if err != nil {
-		return false, err
+		return err
 	}
 
 	req.logger.Info("KubeVirt storage role already exists", "KubeVirtConfig.Namespace", found.Namespace, "KubeVirtConfig.Name", found.Name)
 	// Add it to the list of RelatedObjects if found
 	objectRef, err := reference.GetReference(r.scheme, found)
 	if err != nil {
-		return false, err
+		return err
 	}
 	objectreferencesv1.SetObjectReference(&req.instance.Status.RelatedObjects, *objectRef)
 
-	return req.componentUpgradeInProgress, nil
+	return nil
 }
 
-func (r *ReconcileHyperConverged) ensureKubeVirtStorageRoleBinding(req *hcoRequest) (upgradeDone bool, err error) {
+func (r *ReconcileHyperConverged) ensureKubeVirtStorageRoleBinding(req *hcoRequest) error {
 	kubevirtStorageRoleBinding := newKubeVirtStorageRoleBindingForCR(req.instance, req.Namespace)
 	if err := controllerutil.SetControllerReference(req.instance, kubevirtStorageRoleBinding, r.scheme); err != nil {
-		return false, err
+		return err
 	}
 
 	key, err := client.ObjectKeyFromObject(kubevirtStorageRoleBinding)
@@ -1578,22 +1586,22 @@ func (r *ReconcileHyperConverged) ensureKubeVirtStorageRoleBinding(req *hcoReque
 	err = r.client.Get(req.ctx, key, found)
 	if err != nil && apierrors.IsNotFound(err) {
 		req.logger.Info("Creating kubevirt storage rolebinding")
-		return false, r.client.Create(req.ctx, kubevirtStorageRoleBinding)
+		return r.client.Create(req.ctx, kubevirtStorageRoleBinding)
 	}
 
 	if err != nil {
-		return false, err
+		return err
 	}
 
 	req.logger.Info("KubeVirt storage rolebinding already exists", "KubeVirtConfig.Namespace", found.Namespace, "KubeVirtConfig.Name", found.Name)
 	// Add it to the list of RelatedObjects if found
 	objectRef, err := reference.GetReference(r.scheme, found)
 	if err != nil {
-		return false, err
+		return err
 	}
 	objectreferencesv1.SetObjectReference(&req.instance.Status.RelatedObjects, *objectRef)
 
-	return req.componentUpgradeInProgress, nil
+	return nil
 }
 
 func (r *ReconcileHyperConverged) ensureKubeVirtStorageConfig(req *hcoRequest) (upgradeDone bool, err error) {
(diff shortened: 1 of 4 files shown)


deploy/cluster_role.yaml
```diff
@@ -1163,3 +1163,19 @@ rules:
...
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  resourceNames:
+  - kubevirt-storage-class-defaults
+  verbs:
+  - get
+  - watch
```
mhenriks: Should probably add list here as well.  Watchers typically list all items to fill a cache, then watch

deploy/cluster_role.yaml
```diff
@@ -1163,3 +1163,19 @@ rules:
   - '*'
   verbs:
   - '*'
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: hco.kubevirt.io:config-reader
+  namespace: kubevirt-hyperconverged
```
mhenriks: not exactly sure the context in which this manifest is used but the other roles do not specify the namespace.  So maybe also shouldn't here?

deploy/cluster_role.yaml
mhenriks: I also wouldn't be surprised if this file is generated.

mhenriks: Yup, here:  https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/tools/manifest-templator/manifest-templator.go#L72-L74

djzager:
In order to make sure that it gets included in all of the places it should (ie. downstream), we need to make it so that `./hack/build-manifests.sh` properly creates this Role and RoleBinding.

However, even with that change, I don't think the Role will be installed when the HCO is installed via OLM. How exactly is this Role supposed to 1) make it onto a cluster 2) be used?

djzager:
Is this something that the HCO should be creating in it's `Reconcile()`?

deploy/cluster_role.yaml
maya-r: I want to give access to the configmap with namespace kubevirt-hyperconverged and name kubevirt-storage-class-defaults.
if I don't restrict the namespace then I'm granting access to an identically named configmap in anu namespace (if someone happens to use that exact name. unlikely, but I shouldn't do it)

maya-r:
Here's a version creating everything via reconcile. I added a functional test too.
I'm under the impression that we use Reconcile to bridge the gap between upstream vs. downstream namespace.
I might be able to live with a ClusterRoleBinding instead of a RoleBinding, because it does not need to be namespaced. But I don't know where to put it.

maya-r:
oh, this is blocked on #506, and I should adjust it not to focus the unprivileged test.

maya-r:
Rebased to include the latest (unrelated) commits.

maya-r:
ping

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -1005,6 +1008,52 @@ func (r *ReconcileHyperConverged) ensureKubeVirtTemplateValidator(instance *hcov
...
+			Name:      "hco.kubevirt.io:config-reader",
+			Labels:    labels,
+			Namespace: namespace,
+		},
+		Rules: []rbacv1.PolicyRule{
+			{
+				APIGroups: []string{""},
+				Resources: []string{"configmaps"},
+				ResourceNames: []string{"kubevirt-storage-class-defaults"},
+				Verbs: []string{"get", "watch", "list"},
```
tiraboschi, line 1026: can you please run ` go fmt pkg/controller/hyperconverged/hyperconverged_controller.go` and commit back the result?

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -292,6 +293,8 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
 	for _, f := range []func(*hcov1alpha1.HyperConverged, logr.Logger, reconcile.Request) error{
 		r.ensureKubeVirtConfig,
 		r.ensureKubeVirtStorageConfig,
+		r.ensureKubeVirtStorageRole,
+		r.ensureKubeVirtStorageRoleBinding,
```
tiraboschi, line 297: Why have we to do it at runtime via HCO?
Isn't OLM not enough for this?

maya-r: I assume it can. I wasn't sure about how I'm expected to do something like this.
HCO seems to work with kubernetes too, whereas OLM is apparently specific to OpenShift.
Are we expected to add to the OLM files and have the kubernetes bits be generated?

tiraboschi: Yes, something like that:
https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/deploy/cluster_role_binding.yaml and https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/deploy/cluster_role.yaml are used only on plain k8s deployment.

https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/deploy/olm-catalog/kubevirt-hyperconverged/1.1.0/kubevirt-hyperconverged-operator.v1.1.0.clusterserviceversion.yaml is instead the CSV for OLM based deployment.

both the set of files are generated running `hack/build-manifests.sh` according to what's provided by the components operators.
In order to consume a different component you have to patch go.mod and run `go mod vendor` and `hack/config`

maya-r: In this case HCO is adding the ConfigMap, so it feels like the role should be something HCO adds. Is there somewhere that isn't auto-generated for appending to HCO-specific additions?

ovirt-infra:
All tests passed

maya-r:
This looked like the right thing, I edited pkg/components/components.go and it regenerated deploy/*.yaml, but the OLM CSV doesn't look like it picked up any changes.

tiraboschi:
> This looked like the right thing, I edited pkg/components/components.go and build-manifests.sh regenerated deploy/*.yaml, but the OLM CSV doesn't look like it picked up any changes.

You can eventually patch GetInstallStrategyBase functions.
But should this be part of what HCO is going to receive calling the csv-generator tool on CDI operator?

maya-r:
Right now the configmap is generated by HCO in https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/pkg/controller/hyperconverged/hyperconverged_controller.go#L1502-L1524

I don't know why it's here (or why anything that it does is useful), but the web UI looks at this information

maya-r:
So, I'm running into some trouble: OLM only lets me specify a serviceAccountName in a clusterPermissions.
The intention was to do this for the web UI, so maybe I should figure out what serviceAccountName is the web UI instead of trying to allow everyone access to this configMap?

ovirt-infra:
All tests passed

maya-r:
I went back to the version adding the Role and RoleBinding via the reconcile loop, because I don't think it's possible to add them via OLM.

Another option worth considering is granting access to the openshift-console serviceAccountName, that might have the expected effect too.

@orenc1, could you have a look at this pull request? I'm under the impression that @tiraboschi is away on holiday.

orenc1:
> I went back to the version adding the Role and RoleBinding via the reconcile loop, because I don't think it's possible to add them via OLM.
> 
> Another option worth considering is granting access to the openshift-console serviceAccountName, that might have the expected effect too.
> 
> @orenc1, could you have a look at this pull request? I'm under the impression that @tiraboschi is away on holiday.

Hi @maya-r , we are using the OLM to set the required permissions for the service accounts of HCO's dependent operators. If the access for `kubevirt-storage-class-defaults` config map is only required by the console (and not a direct access by the user), I think it would be more "granular" to provide this Role&RoleBinding to the console ServiceAccount. That could be configured by OLM by specifying it in the CSV, instead of having HCO manage these permissions in runtime.

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -469,6 +470,8 @@ func (r *ReconcileHyperConverged) ensureHco(req *hcoRequest) error {
 		r.ensureKubeVirtPriorityClass,
 		r.ensureKubeVirtConfig,
 		r.ensureKubeVirtStorageConfig,
+		r.ensureKubeVirtStorageRole,
```
nunnatsa, line 473: In addition to the general comment about not doing it from reconcile - if there is no other choice, I would call these two methods from within `ensureCDI` and not as part of this loop. This is something that is not monitors by HC as other resources.

orenc1:
> 
> Hi @maya-r , we are using the OLM to set the required permissions for the service accounts of HCO's dependent operators. If the access for `kubevirt-storage-class-defaults` config map is only required by the console (and not a direct access by the user), I think it would be more "granular" to provide this Role&RoleBinding to the console ServiceAccount. That could be configured by OLM by specifying it in the CSV, instead of having HCO manage these permissions in runtime.

I would recommend you to test by manually configuring the required Roles&RoleBindings of this configmap for the console ServiceAccount and see if it solves the issue. If so, add the relevant section in `csv-generator` of CDI so it will end up in HCO's CSV.

maya-r:
I got a response from Tomas Jelinek that the web interface accesses objects in the cluster as the logged in user (I had some trouble experimenting in the web interface).
It isn't possible to add the rights to the serviceAccountName and get the same result.
So this version of the pull request that is adding it via reconcile is the best I can do, I can't achieve the same result using OLM.

orenc1:
OK.
I would suggest to nest `r.ensureKubeVirtStorageConfig` and `r.ensureKubeVirtStorageRole` executions from within `ensureCDI`, as Nahshon said, because these resources are strictly related to Storage.
I would also wait for @tiraboschi who returns on Monday for another review.

pkg/controller/hyperconverged/hyperconverged_controller.go
maya-r: Thanks for the review.
My original motivation to write the code this way is that ensureKubeVirtStorageConfig is currently implemented as its own reconcile function.
If the intention in rotating the code is to "don't test if the rolebinding exists in every reconcile iteration", I don't think that this is a good idea.
It would violate our rule to write reconcile such that we reach eventual consistency. This might negatively affect updates, as one example.
If it's just to make things tidy, I should probably combine it with ensureKubeVirtStorageConfig too.

tiraboschi:
Yes, I also agree with @nunnatsa comment: all the methods called in the first loop in ensureHco method have a specific signature because the code there will use it to check the conditions reported by component operators and eventually the status of an upgrade.
We don't have anything like that for `r.ensureKubeVirtStorageRole` and `r.ensureKubeVirtStorageRoleBinding` so I  agree that they should be called from `r.ensureCDI` so that you can use a simpler signature.
Probably, in a separate PR, we should do the same for `r.ensureKubeVirtStorageConfig` and other Config Maps created there.

orenc1:
> Yes, I also agree with @nunnatsa comment: all the methods called in the first loop in ensureHco method have a specific signature because the code there will use it to check the conditions reported by component operators and eventually the status of an upgrade.
> We don't have anything like that for `r.ensureKubeVirtStorageRole` and `r.ensureKubeVirtStorageRoleBinding` so I agree that they should be called from `r.ensureCDI` so that you can use a simpler signature.
> Probably, in a separate PR, we should do the same for `r.ensureKubeVirtStorageConfig` and other Config Maps created there.

@maya-r ping?

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

PR #499, opened 2020-03-16
BZ 1768751: Add Web UI to rotate-certs.sh

mareklibra:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1768751

**Release note**:
```release-note
Fixes BZ 1768751 by including Web UI within `rotate-certs.sh` script.
```

### DIFF
--- a/tools/rotate-certs.sh
+++ b/tools/rotate-certs.sh
@@ -69,3 +69,18 @@ ${_kubectl} delete pods --namespace "${namespace}" -l kubevirt.io
 echo "# Rotating SSP certificates ..."
 ${_kubectl} delete secrets --ignore-not-found=true --namespace "${namespace}" virt-template-validator-certs
 ${_kubectl} delete pods --namespace "${namespace}" -l kubevirt.io=virt-template-validator
+
+echo "# Rotating Web UI certificates ..."
+oc get crd kwebuis.kubevirt.io >/dev/null
+if [ $? -eq 0 ]; then
+  # valid for 1.4.Z only
+  formerVersion=$(${_kubectl} get kwebui kubevirt-web-ui -o yaml | grep '  version: '|sed 's/^.*: *\(.*\)$/\1/g')
+  echo Detected former Web UI version: ${formerVersion}
+  ${_kubectl} patch kwebui kubevirt-web-ui --patch '{"spec": {"version": ""}}' --type=merge # undeploy
+  while (oc get deployment console -n ${webui_namespace} --no-headers=true 2>/dev/null); do
+    echo "Waiting for Web UI ..." # to undeploy
+    sleep 5
+  done
+  patch=$(echo '{"spec": {"version": "'${formerVersion}'"}}')
+  ${_kubectl} patch kwebui kubevirt-web-ui --patch "${patch}" --type=merge # deploy
+fi
--- a/tools/rotate-certs.sh
+++ b/tools/rotate-certs.sh
@@ -71,13 +71,13 @@ ${_kubectl} delete secrets --ignore-not-found=true --namespace "${namespace}" vi
 ${_kubectl} delete pods --namespace "${namespace}" -l kubevirt.io=virt-template-validator
 
 echo "# Rotating Web UI certificates ..."
-oc get crd kwebuis.kubevirt.io >/dev/null
+${_kubectl} get crd kwebuis.kubevirt.io >/dev/null
 if [ $? -eq 0 ]; then
   # valid for 1.4.Z only
   formerVersion=$(${_kubectl} get kwebui kubevirt-web-ui -o yaml | grep '  version: '|sed 's/^.*: *\(.*\)$/\1/g')
   echo Detected former Web UI version: ${formerVersion}
   ${_kubectl} patch kwebui kubevirt-web-ui --patch '{"spec": {"version": ""}}' --type=merge # undeploy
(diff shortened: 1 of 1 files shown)


mareklibra:
@tiraboschi , can you please suggest branch to backport this PR to?
Fixes BZ 1768751, the Web UI is missing in later releases.

tools/rotate-certs.sh
```diff
@@ -69,3 +69,12 @@ ${_kubectl} delete pods --namespace "${namespace}" -l kubevirt.io
 echo "# Rotating SSP certificates ..."
 ${_kubectl} delete secrets --ignore-not-found=true --namespace "${namespace}" virt-template-validator-certs
 ${_kubectl} delete pods --namespace "${namespace}" -l kubevirt.io=virt-template-validator
+
+echo "# Rotating Web UI certificates ..."
+formerVersion=$(${_kubectl} get kwebui kubevirt-web-ui -o yaml | grep '  version: '|sed 's/^.*: *\(.*\)$/\1/g')
+echo Detected former Web UI version: ${formerVersion}
+${_kubectl} patch kwebui kubevirt-web-ui --patch '{"spec": {"version": ""}}' --type=merge # undeploy
+sleep 15 # give kubevirt-web-ui-operator enough time to react, but it's usually instant
```
mareklibra: The web-ui-operator needs to start reconciliation (means undeployment). We do not need to wait till it's finished.

mareklibra:
> The script should work for 1.4, 2.0, 2.1 and 2.2.

Good point, thank you. This should be valid for 1.4.Z only. I have added check for presence of corresponding CRD.

tools/rotate-certs.sh
```diff
@@ -69,3 +69,16 @@ ${_kubectl} delete pods --namespace "${namespace}" -l kubevirt.io
...
 ${_kubectl} delete pods --namespace "${namespace}" -l kubevirt.io=virt-template-validator
+
+echo "# Rotating Web UI certificates ..."
+oc get crd kwebuis.kubevirt.io >/dev/null
+if [ $? -eq 0 ]; then
+  # valid for 1.4.Z only
+  formerVersion=$(${_kubectl} get kwebui kubevirt-web-ui -o yaml | grep '  version: '|sed 's/^.*: *\(.*\)$/\1/g')
+  echo Detected former Web UI version: ${formerVersion}
+  ${_kubectl} patch kwebui kubevirt-web-ui --patch '{"spec": {"version": ""}}' --type=merge # undeploy
+  sleep 15 # give kubevirt-web-ui-operator enough time to react, but it's usually instant
```
rmohr: An unconditional sleep is not really good. You are probably waiting for the ui-deployment to be scaled down or deleted. Could you either just scale down the deployment (which is waiting), or periodically check that the deployment got deleted?

mareklibra: Actually I already had code here checking presence of pods (they are about to be terminated). But found that this is not important. The operator just needs to get time to initiate the reconciliation - to start undeployment. We do not need to wait for this undeployment to be finished.

If really requested, I can add similar check back, either based on status in the CR or pod presence, but it will make the code here harder to read with no/very limited benefit.

rmohr: What if the undeploy does not happen withing 15 seconds (e.g. temporary error or rbac issue, or whatever)?  You would set the version again after 15 seconds. Wouldn't you then risk that no rotation happened at all?

mareklibra: Ok, wating is added

tools/rotate-certs.sh
```diff
@@ -69,3 +69,18 @@ ${_kubectl} delete pods --namespace "${namespace}" -l kubevirt.io
...
 ${_kubectl} delete pods --namespace "${namespace}" -l kubevirt.io=virt-template-validator
+
+echo "# Rotating Web UI certificates ..."
+oc get crd kwebuis.kubevirt.io >/dev/null
+if [ $? -eq 0 ]; then
+  # valid for 1.4.Z only
+  formerVersion=$(${_kubectl} get kwebui kubevirt-web-ui -o yaml | grep '  version: '|sed 's/^.*: *\(.*\)$/\1/g')
+  echo Detected former Web UI version: ${formerVersion}
+  ${_kubectl} patch kwebui kubevirt-web-ui --patch '{"spec": {"version": ""}}' --type=merge # undeploy
+  while (oc get deployment console -n ${webui_namespace} --no-headers=true 2>/dev/null); do
```
djzager:
```suggestion
  while (${_kubectl} get deployment console -n ${webui_namespace} --no-headers=true 2>/dev/null); do
```

tools/rotate-certs.sh
```diff
@@ -69,3 +69,18 @@ ${_kubectl} delete pods --namespace "${namespace}" -l kubevirt.io
 echo "# Rotating SSP certificates ..."
 ${_kubectl} delete secrets --ignore-not-found=true --namespace "${namespace}" virt-template-validator-certs
 ${_kubectl} delete pods --namespace "${namespace}" -l kubevirt.io=virt-template-validator
+
+echo "# Rotating Web UI certificates ..."
+oc get crd kwebuis.kubevirt.io >/dev/null
```
tiraboschi:
```suggestion
${_kubectl} get crd kwebuis.kubevirt.io >/dev/null
```

PR #502, opened 2020-03-18
updating script to allow for proxy environment usecases

shawn-hurley:
**Release note**:

```release-note
* Addition of `bundle-registry-dest` and `no-mirror` flags to the `mirror_csv_release.sh` script
```

/cc @tiraboschi

### DIFF
--- a/tools/mirror-csv-release/mirror_csv_release.sh
+++ b/tools/mirror-csv-release/mirror_csv_release.sh
@@ -66,6 +66,14 @@ Optional arguments:
     --bundle-registry-tag
         tag of the destination bundle registry image, used only in appregistry mode
         default: 1.0.0
+    
+     --bundle-registry-dest
+         registry location for the bundle registry image
+	 default: DEST_PREFIX
+     
+      --no-mirror
+          skip mirroring the images for CNV to the DEST_PREFIX
+	  default: false
 
 Example:
     ${0##*/}  --version-filter 2.2.0 quay.io/openshift-cnv/container-native-virtualization-hco-bundle-registry:v2.2.0-181 quay.io/tiraboschi/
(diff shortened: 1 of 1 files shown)


ovirt-infra:

tools/mirror-csv-release/mirror_csv_release.sh
```diff
@@ -164,6 +174,14 @@ main() {
                 bundle_registry_tag="$2"
                 shift 2
                 ;;
+            --bundle-registry-dest)
+                bundle_registry_dest="$2"
+                shift 2
+                ;;
+            --no-mirror)
+                no_mirror="$2"
```
tiraboschi: I think it would be better to use just an optional flag without expecting any value as we are doing for instance for --dry-run

shawn-hurley: +1 sorry not that skilled in bash that makes a ton more sense!

tools/mirror-csv-release/mirror_csv_release.sh
```diff
@@ -287,8 +306,9 @@ mirror() {
 build_and_publish_patched_bundle_image() {
     local bundle_image="${1:?}"
     local dest_prefix="${2:?}"
+    local bundle_dest_prefix="${5:-$dest_prefix}"
     local source_registry=${source_image%\/*}/
-    local dest_image=$(get_dest_image "${bundle_image}" "${dest_prefix}")
+    local dest_image=$(get_dest_image "${bundle_image}" "${bundle_dest_prefix}")
```
tiraboschi: You patched the wrong function:
on line 218 we are calling `build_and_publish_patched_bundle_image_appregistry` but this is `build_and_publish_patched_bundle_image`.

I'd suggest to:
- keep this
- patch also `build_and_publish_patched_bundle_image_appregistry` with the same logic
- patch the call on line 220
- remove `, used only in appregistry mode` on line 71

PR #504, opened 2020-03-20
Add forgotten webui_namespace variable

mareklibra:
**Release note**:

```release-note
NONE
```

### DIFF
--- a/tools/rotate-certs.sh
+++ b/tools/rotate-certs.sh
@@ -3,9 +3,10 @@ set -euo pipefail
 
 namespace=openshift-cnv
 cdi_namespace=""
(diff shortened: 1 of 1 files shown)


tools/rotate-certs.sh
```diff
@@ -22,6 +23,10 @@ while true; do
         shift;
         cdi_namespace=$1
         ;;
+    --webui-namespace)
```
tiraboschi: I think you have also to add `,webui-namespace:` on line 9

mareklibra: Added, thank you. I am currently looking for an environment to run this script altogether, not just by snippets.

PR #508, opened 2020-03-23
Fix missing "Available" condition bug

nunnatsa:
Fix a bug where when checking the conditions of an operator, if the "Available" condition is missing,
HCO relate that as Available = True;

Instead, HCO now treat missing Available condition as Available = False

```release-note
NONE
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,8 +14,9 @@ require (
 	github.com/hashicorp/golang-lru v0.5.4 // indirect
 	github.com/imdario/mergo v0.3.8
 	github.com/kubevirt/cluster-network-addons-operator v0.3.1-0.20200226122004-919521cbbe11
-	github.com/onsi/ginkgo v1.10.1
-	github.com/onsi/gomega v1.7.0
+	github.com/mattn/goveralls v0.0.5 // indirect
+	github.com/onsi/ginkgo v1.12.0
+	github.com/onsi/gomega v1.9.0
 	github.com/openshift/custom-resource-status v0.0.0-20190822192428-e62f2f3b79f3
 	github.com/operator-framework/operator-lifecycle-manager v0.0.0-20191220211133-23f5c0292434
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -14,8 +14,9 @@ require (
 	github.com/hashicorp/golang-lru v0.5.4 // indirect
 	github.com/imdario/mergo v0.3.8
 	github.com/kubevirt/cluster-network-addons-operator v0.3.1-0.20200226122004-919521cbbe11
-	github.com/onsi/ginkgo v1.10.1
-	github.com/onsi/gomega v1.7.0
+	github.com/mattn/goveralls v0.0.5 // indirect
+	github.com/onsi/ginkgo v1.12.0
+	github.com/onsi/gomega v1.9.0
```
tiraboschi: I'd suggest to move this to a distinct PR.

djzager:
> I'm just wondering if we can or should (before this PR) refactor `func handleConditionsSSP` to make it able to handle the conditions for all the components and then apply this fix only there.

I would be in favor of that. Since SSP is the only operator in the bunch that is Ansible based (and susceptible to having a "Running" condition but no other).

nunnatsa:
> > I'm just wondering if we can or should (before this PR) refactor `func handleConditionsSSP` to make it able to handle the conditions for all the components and then apply this fix only there.
> 
> I would be in favor of that. Since SSP is the only operator in the bunch that is Ansible based (and susceptible to having a "Running" condition but no other).

Done. The code is identical in the rest of the components. However, the Kubevirt code defines its own condition type, and so we can't directly use it as a parameter to the `handleComponentConditions` method. I added a map function, `translateKubeVirtConds`, to handle this.

tiraboschi:
/override coverage/coveralls
Due to the refactor with `handleComponentConditions` we are implicitly covering less lines: accepting it.

PR #509, opened 2020-03-24
Enable 1.0.0 -> 1.1.0 upgrade path

tiraboschi:
1.1.0 should replace 1.0.0 and not 0.0.3

```release-note
NONE
```


tiraboschi:
> Maybe update version/version.go as well?

yes, thanks!

PR #510, opened 2020-03-25
Enforce a structural schema for HCO CRD

tiraboschi:
https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20190425-structural-openapi.md
defines what's a structural schema for a CRD and currently HCO CRD doens't match.

Try to move back to a fully structural schema.
Please notice that the most relevant change is removing the
namespace restriction (which is explictly forbidden) that
can lead HCO CR to be potentially created in any namespace.

We should eventually consider again about moving to a cluster scoped CR.

**Release note**:
```release-note
Enforce a structural schema for HCO CRD
```

### DIFF
--- a/deploy/crds/hco00.crd.yaml
+++ b/deploy/crds/hco00.crd.yaml
@@ -27,9 +27,8 @@ spec:
             name:
               pattern: kubevirt-hyperconverged
               type: string
-            namespace:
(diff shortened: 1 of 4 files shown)


tiraboschi:
implemented with #610

PR #516, opened 2020-03-30
API Group Overlap Validation Among Operators

orenc1:
A new verification is added during the process of build-manifests.sh, to check whether there are different operators in Kubevirt eco-system which are using the same API groups.
If one or more of such overlaps are detected, an error is shown and build-manifests.sh is exited.

```release-note
API groups overlap among different operators validation.
```

### DIFF
--- a/tools/csv-merger/csv-merger.go
+++ b/tools/csv-merger/csv-merger.go
@@ -24,7 +24,12 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"io/ioutil"
+	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions"
+	"log"
 	"os"
+	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/blang/semver"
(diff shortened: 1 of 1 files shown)


orenc1:
**Note:** Currently the test fails because the API group `kubevirt.io` is being used by these operators: kubevirt, hco, node-maintenance

/cc @tiraboschi

tools/csv-merger/csv-merger.go
```diff
@@ -89,6 +94,7 @@ var (
 		"Comma separated list of all the CRDs that should be visible in OLM console")
 	relatedImagesList = flag.String("related-images-list", "",
 		"Comma separated list of all the images referred in the CSV (just the image pull URLs or eventually a set of 'image|name' collations)")
+	crdDir              = flag.String("crds-dir", "", "the directory containing the CRDs for apigroup validation")
```
tiraboschi: please modify the help message saying that the validation will be performed only if the value is != ""

orenc1: Done, and I removed the triggering of the test from `build-manifests.sh`

PR #517, opened 2020-04-01
bump SSP to v1.0.29

tiraboschi:

**Release note**:
```release-note
bump SSP to v1.0.29
```


tiraboschi:
/hold
test only with an additional rule

PR #528, opened 2020-04-06
Detect Upgrade Mode

nunnatsa:
HCO now find out if it in upgrade mode. This is a preparation for the next phase, where HCO will notify the end-user about doing upgrade and completing it.

On init, HCO add the image Id to the CR's annotations.
On non-init reconciliation, HCO checks if ts image Id match to the image Id in the CR; if not, or if there is no image Id in the CR, HCO is in upgrade mode.
On upgrade mode, when processing is completed and HCO become available, HCO updates the image Id in the CR to the new image Id, and exits upgrade mode.

```release-note
NONE
```

### DIFF
--- a/pkg/controller/hyperconverged/hyperconverged_controller.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller.go
@@ -319,7 +319,7 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
 
 	if !r.upgradeMode && !init && knownHcoVersion != r.ownVersion {
 		r.upgradeMode = true
-		reqLogger.Info(fmt.Sprintf("Upgating from version %s to version %s", knownHcoVersion, r.ownVersion))
+		reqLogger.Info(fmt.Sprintf("Start upgrating from version %s to version %s", knownHcoVersion, r.ownVersion))
 	}
 
 	for _, f := range []func(*hcov1alpha1.HyperConverged, logr.Logger, reconcile.Request) error{
--- a/pkg/controller/hyperconverged/hyperconverged_controller.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller.go
@@ -32,6 +32,7 @@ import (
 	networkaddonsnames "github.com/kubevirt/cluster-network-addons-operator/pkg/names"
 	hcov1alpha1 "github.com/kubevirt/hyperconverged-cluster-operator/pkg/apis/hco/v1alpha1"
 	hcoutil "github.com/kubevirt/hyperconverged-cluster-operator/pkg/util"
+	version "github.com/kubevirt/hyperconverged-cluster-operator/version"
 	conditionsv1 "github.com/openshift/custom-resource-status/conditions/v1"
 	objectreferencesv1 "github.com/openshift/custom-resource-status/objectreferences/v1"
 	virtconfig "kubevirt.io/kubevirt/pkg/virt-config"
@@ -73,8 +74,7 @@ const (
 	ErrHCOUninstall       = "ErrHCOUninstall"
 	uninstallHCOErrorMsg  = "The uninstall request failed on dependent components, please check their logs."
 
-	operatorImageEnv = "OPERATOR_IMAGE"
-	operatorImageCr  = "operatorImage"
+	hcoVersionName = "operator"
 )
 
 // Add creates a new HyperConverged Controller and adds it to the Manager. The Manager will set fields on the Controller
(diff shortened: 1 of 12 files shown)


pkg/controller/hyperconverged/hyperconverged_controller_test.go
```diff
@@ -1581,7 +1581,95 @@ var _ = Describe("HyperconvergedController", func() {
 
 			})
 		})
+
+		Context("Upgrade Mode", func() {
+			expected := getBasicDeployment()
+			okConds := expected.hco.Status.Conditions
+
+			const (
+				OLD_IMAGE = "quay.io/kubuvirt/hyperconverged-cluster-operator:v2.3.0"
```
tiraboschi, line 1590: please use 1.0.0 and 1.1.0

nunnatsa:
@tiraboschi could you please re-review?

nunnatsa:
/block

Freeze for now. We'll try to use image digest instead of version.

nunnatsa:
@djzager, @tiraboschi - please review

pkg/controller/hyperconverged/hyperconverged_controller_test.go
```diff
@@ -1668,6 +1670,97 @@ var _ = Describe("HyperconvergedController", func() {
...
+				os.Setenv("OPERATOR_NAMESPACE", namespace)
+			})
+
+			It("Should update HCO Version Id in the CR on init", func() {
+				os.Setenv(util.HcoKvIoVersionName, OldVersion)
+
+				expected.hco.Status.Conditions = nil
+
+				cl := expected.initClient()
+				foundResource := checkAvailability(expected.hco, cl, true, corev1.ConditionFalse)
```
tiraboschi, line 1695: basically this is the only place where we call checkAvailability with requeue = true; I'm just wondering if we can change the logic here and keep the function simpler

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -364,6 +388,11 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
 			Message: reconcileCompletedMessage,
 		})
 
+		if r.upgradeMode { // update the new image only when upgrade is completed
+			instance.Status.UpdateVersion(hcoVersionName, r.ownVersion)
+			r.upgradeMode = false
```
tiraboschi, line 393: I think we should also log something with reqLogger.Info(...)  explicitly reporting that the upgrade has been completed

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -294,6 +313,11 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
 		}
 	}
 
+	// If the current image is not updated in CR ,then we're updating. This is also works when updating from
+	// an old version, since instance.ObjectMeta.Annotations[operatorImageCr] will be empty.
+	knownHcoVersion, _ := instance.Status.GetVersion(hcoVersionName)
+	r.upgradeMode = !init && knownHcoVersion != r.ownVersion
```
tiraboschi, line 319: please do this with an if so that we can eventually also log that we detected that we are in upgrade mode

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -294,6 +313,15 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
 		}
 	}
 
+	// If the current image is not updated in CR ,then we're updating. This is also works when updating from
+	// an old version, since instance.ObjectMeta.Annotations[operatorImageCr] will be empty.
+	knownHcoVersion, _ := instance.Status.GetVersion(hcoVersionName)
+
+	if !r.upgradeMode && !init && knownHcoVersion != r.ownVersion {
+		r.upgradeMode = true
+		reqLogger.Info(fmt.Sprintf("Upgating from version %s to version %s", knownHcoVersion, r.ownVersion))
```
tiraboschi, line 322: Upgrading

pkg/controller/hyperconverged/hyperconverged_controller_test.go
```diff
@@ -1560,41 +1562,53 @@ var _ = Describe("HyperconvergedController", func() {
...
+				checkAvailability(foundResource, corev1.ConditionTrue)
 
 				// TODO: temporary avoid checking conditions on KubevirtCommonTemplatesBundle because it's currently
 				// broken on k8s. Revert this when we will be able to fix it
 				/*
 					origConds = expected.kvCtb.Status.Conditions
 					expected.kvCtb.Status.Conditions = expected.cdi.Status.Conditions[1:]
 					cl = expected.initClient()
-					checkAvailability(expected.hco, cl, corev1.ConditionFalse)
+					checkAvailability(expected.hco, cl, false, corev1.ConditionFalse)
```
tiraboschi, line 1607: ?

tiraboschi:
ignoring slightly coverage reduction
/override coverage/coveralls

PR #529, opened 2020-04-06
Simplify how we track component versions

tiraboschi:
Currently we are tracking component versions in
different places and different test scripts are consuming
different versions:
- go.mod for source vendoring
- hack/build-manifest.sh for CSV and manifest generators
- hack/defaults
- hack/config

This is definitively error prone, let's simplify this keeping only:
- go.mod for go
- hack/config for all the bash scripts

**Release note**:
```release-note
NONE
```


orenc1:
`hack/common.sh: line 25: CDI_MANIFEST_VERSION: unbound variable`

PR #544, opened 2020-04-22
Fix CI for klog/v2

tiraboschi:
klog officially released v2 on Apr 10th so
/v2 is not anymore a subdir there but the top
level directory.
Implementing an hack to temporary fix it with
a symlink until all the involved modules will
be ready for that.

**Release note**:
```release-note
NONE
```


tiraboschi:
ci/prow/hco-e2e-upgrade-aws currently runs on OCP 4.5 and it's going to fail due to a side effect in the fix of https://bugzilla.redhat.com/1825330
It will be addressed on https://github.com/operator-framework/operator-lifecycle-manager/pull/1470# on OLM side.
Temporary ignoring it

PR #545, opened 2020-04-23
Track operator versions as env variables

tiraboschi:
Track operator versions as env variables
in HCO deployment to enable upgrade status checks
(to be addressed in afuture PR).

**Release note**:
```release-note
Track operator versions as env variables in HCO deployment
```

### DIFF
--- a/pkg/components/components.go
+++ b/pkg/components/components.go
@@ -39,9 +39,20 @@ type StrategyDetailsDeployment struct {
 	ClusterPermissions []StrategyDeploymentPermissions `json:"clusterPermissions,omitempty"`
 }
 
-const hcoName = "hyperconverged-cluster-operator"
+const (
+	hcoName             = "hyperconverged-cluster-operator"
+	kubevirtVersionEnvV = "KUBEVIRT_VERSION"
+	cdiVersionEnvV      = "CDI_VERSION"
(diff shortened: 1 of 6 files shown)


pkg/components/components.go
```diff
@@ -39,9 +39,20 @@ type StrategyDetailsDeployment struct {
...
+	kubevirtVersionEnvV = "KUBEVIRT_VERSION"
+	cdiVersionEnvV      = "CDI_VERSION"
+	cnaoVersionEnvV     = "NETWORK_ADDONS_VERSION"
+	sspVersionEnvV      = "SSP_VERSION"
+	nmoVersionEnvV      = "NMO_VERSION"
+	hppoVersionEnvV     = "HPPO_VERSION"
+)
 
-func GetDeployment(namespace, image, imagePullPolicy, conversionContainer, vmwareContainerString, smbios, machinetype string) appsv1.Deployment {
+func GetDeployment(
```
nunnatsa, line 52: This function takes a lot of parameters. I found it confusing. Can we pass a name-value pairs array instead? we can build it in different method and pass the result.
This approach will also is adding or dropping components.

tiraboschi:
Handled in #584

tiraboschi:
Closing for https://github.com/kubevirt/hyperconverged-cluster-operator/pull/584

PR #548, opened 2020-04-26
WIP refactor(status): recalculate all conditions

djzager:
On every call to Reconcile(), recalculate conditions from scratch.

**Release note**:
```release-note
None
```

### DIFF
--- a/pkg/controller/hyperconverged/hyperconverged_controller.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller.go
@@ -72,6 +72,27 @@ const (
 	uninstallHCOErrorMsg  = "The uninstall request failed on dependent components, please check their logs."
 )
 
+var expectedConditions = []conditionsv1.Condition{
+	{
+		Type:    conditionsv1.ConditionAvailable,
+		Status:  corev1.ConditionTrue,
+		Reason:  reconcileCompletedMessage,
+		Message: reconcileCompleted,
+	},
+	{
+		Type:    conditionsv1.ConditionProgressing,
+		Status:  corev1.ConditionFalse,
+		Reason:  reconcileCompletedMessage,
+		Message: reconcileCompleted,
+	},
+	{
+		Type:    conditionsv1.ConditionDegraded,
+		Status:  corev1.ConditionFalse,
+		Reason:  reconcileCompletedMessage,
+		Message: reconcileCompleted,
+	},
+}
+
 // Add creates a new HyperConverged Controller and adds it to the Manager. The Manager will set fields on the Controller
 // and Start it when the Manager is Started.
 func Add(mgr manager.Manager) error {
@@ -135,10 +156,9 @@ var _ reconcile.Reconciler = &ReconcileHyperConverged{}
 type ReconcileHyperConverged struct {
 	// This client, initialized using mgr.Client() above, is a split client
 	// that reads objects from the cache and writes to the apiserver
-	client     client.Client
-	scheme     *runtime.Scheme
-	recorder   record.EventRecorder
-	conditions []conditionsv1.Condition
+	client   client.Client
+	scheme   *runtime.Scheme
+	recorder record.EventRecorder
 }
 
 // Reconcile reads that state of the cluster for a HyperConverged object and makes changes based on the state read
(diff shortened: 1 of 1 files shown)


pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -695,91 +647,66 @@ func (r *ReconcileHyperConverged) ensureNetworkAddons(instance *hcov1alpha1.Hype
...
+				Status:  corev1.ConditionUnknown,
+				Reason:  fmt.Sprintf("%sMissingCondition", component),
+				Message: fmt.Sprintf("Component %s is not reporting this condition", component),
 			}
 		}
 
-		if !foundAvailableCond {
-			componentNotAvailable(logger, component, `missing "Available" condition`, r)
+		currentStatus := (currentCondition.Status == condition.Status)
+		componentStatus := (componentCondition.Status == condition.Status)
```
nunnatsa, line 699: `Unknown` will override a "negative" condition (e.g. `Available=False` or `Progressing=True`). I guess that if we already know that one operand is unavailable, we better report that than reporting "unknown". So maybe a value comparison is not enough.

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -695,91 +647,66 @@ func (r *ReconcileHyperConverged) ensureNetworkAddons(instance *hcov1alpha1.Hype
...
+
+		if !componentStatus {
+			conditionsv1.SetStatusCondition(&instance.Status.Conditions, *componentCondition)
+		}
+		if currentStatus && componentStatus {
```
nunnatsa, line 704: Personally, I found this more readable. 
``` golang
else if currentStatus {
```

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -320,77 +300,54 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
...
+		Message: reconcileCompletedMessage,
+	})
 
-		// If no operator whose conditions we are watching reports an error, then it is safe
-		// to set readiness.
-		r := ready.NewFileReady()
-		err = r.Set()
+	readyFile := ready.NewFileReady()
+	switch conditionsv1.IsStatusConditionTrue(instance.Status.Conditions, conditionsv1.ConditionUpgradeable) {
+	case true:
```
nunnatsa, line 339: Why switch-case for boolean? I think it's confusing. Please consider if-else.

pkg/controller/hyperconverged/hyperconverged_controller.go
djzager: I've updated it so that the first time we encounter something other than the desired status, we will update it. Subsequently we will only update the status if the component is both 1) not reporting the desired status and 2) it's status is not unknown.

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -72,6 +72,33 @@ const (
 	uninstallHCOErrorMsg  = "The uninstall request failed on dependent components, please check their logs."
 )
 
+var conditions = []conditionsv1.Condition{
```
nunnatsa, line 75: Naming: please consider clearer name for the `conditions` slice. What about `expectedConditions` or something like that?

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -72,6 +72,33 @@ const (
...
+		Message: reconcileCompleted,
+	},
+	{
+		Type:    conditionsv1.ConditionProgressing,
+		Status:  corev1.ConditionFalse,
+		Reason:  reconcileCompletedMessage,
+		Message: reconcileCompleted,
+	},
+	// {
+	// 	Type:    conditionsv1.ConditionDegraded,
```
nunnatsa, line 89: So - don't we monitor components's degraded condition at all?

djzager: fixed

pkg/controller/hyperconverged/hyperconverged_controller.go
djzager: done

PR #560, opened 2020-05-05
Wait a few seconds before checking CR status

tiraboschi:
OCP 4.5 seems more sensitive on this front,
let's try to make it more stable with a small
delay.

**Release note**:
```release-note
NONE
```


fabiand:
No sleep!

orenc1:
It was failed because of a timestamp conflict on the CSV yaml file, I'll make an enhancement which is going to resolve that issue.

tiraboschi:
It seems less relevant now, I'm closing it.

PR #564, opened 2020-05-12
Aggregate Negative Conditions

nunnatsa:
Make sure that when a sub component returns a negative condition (e.g. `Available = "False"` or `Progressing = "True"`), HCO will set all its conditions, and won't have left overs conditions from previous reconciliations.

**Release note**:
```release-note
NONE
```


pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -365,7 +365,7 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
 		return reconcile.Result{Requeue: true}, nil
 	}
 
-	if r.conditions == nil {
+	if r.conditions == nil || r.conditions.empty() {
```
tiraboschi, line 368: I'm still a bit confused about this if.
Are we considering the unknown status when, for any reason, we fail to properly fetch conditions from all the operators?

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -416,6 +416,68 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
 		// One shortcoming is that only one failure of a particular condition can be
 		// captured at one time (ie. if KubeVirt and CDI are both reporting !Available,
 		// you will only see CDI as it updates last).
+
+		if cond, ok := r.conditions[conditionsv1.ConditionDegraded]; ok {
```
tiraboschi, line 420: honestly I'm a bit confused about this one:
r.conditions in the in memory copy of the conditions we are going to write on the hyperconverged resource.
We are going to write them only on line 481.

In precedence we updated r.conditions scanning all the components (but what's going to happen if we don't finish or if we skip one?).
Now we are overriding it.
I'm a bit scared we can loose something.

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -416,6 +416,68 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
...
+			if _, ok = r.conditions[conditionsv1.ConditionProgressing]; !ok {
+				r.setStatusCondition(conditionsv1.Condition{
+					Type:    conditionsv1.ConditionProgressing,
+					Status:  corev1.ConditionTrue,
+					Reason:  cond.Reason,
+					Message: cond.Reason,
```
tiraboschi, line 426: So basically we are replicating here, on the progressing condition, the error message we got from the degraded condition of another operator.
I'm a bit scared that this can be even more confusing at user eyes.

hack/build-charts.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+set -ex
+
+TMP_PLANTUML_DIR=$(mktemp -d)
+wget -O "${TMP_PLANTUML_DIR}/plantuml.jar" http://sourceforge.net/projects/plantuml/files/plantuml.jar/download
+java -jar "${TMP_PLANTUML_DIR}/plantuml.jar" -checkmetadata -tsvg pkg/controller/hyperconverged/hyperconverged_controller.go
```
tiraboschi, line 6: We can also evaluate a dockerized version: https://hub.docker.com/r/plantuml/plantuml-server
but maybe this is still faster

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -335,102 +352,253 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
...
+				Reason:  cond.Reason,
+				Message: cond.Reason,
+			})
+		} // else - Upgradeable is already exists
+		if _, conditionFound = req.conditions[conditionsv1.ConditionAvailable]; !conditionFound { // (#chart 6)
+			req.conditions.setStatusCondition(conditionsv1.Condition{ // (#chart 7)
+				Type:    conditionsv1.ConditionAvailable,
+				Status:  corev1.ConditionFalse,
+				Reason:  cond.Reason,
+				Message: cond.Reason,
```
tiraboschi, line 514: Still not 100% confident about this:
basically if we found a component that reported degraded we are also simply duplicating the error message we got from that component also on available condition on HCO CR.
So when the user tries to check available condition on HCO CR he will find something like "XXX is degraded due to..." that can sound a bit misleading.
Maybe in this cases we can simply use a placeholder message explicitly saying that HCO is not available because we have degraded components.

pkg/controller/hyperconverged/hyperconverged_controller_test.go
```diff
@@ -1796,6 +1814,300 @@ var _ = Describe("HyperconvergedController", func() {
...
+				Expect(cd.Reason).Should(Equal("CDIProgressing"))
+				cd = conditionsv1.FindStatusCondition(conditions, conditionsv1.ConditionDegraded)
+				Expect(cd.Status).Should(BeEquivalentTo("False"))
+				Expect(cd.Reason).Should(Equal(reconcileCompleted))
+				cd = conditionsv1.FindStatusCondition(conditions, conditionsv1.ConditionUpgradeable)
+				Expect(cd.Status).Should(BeEquivalentTo("False"))
+				Expect(cd.Reason).Should(Equal("CDIProgressing"))
+			})
+
+			It("should be not Available when a component is not Available + !", func() {
```
tiraboschi, line 2009: "+ !..."

hack/build-charts.sh
nunnatsa: Their docker is an HTTP server, not a command line. It's much simpler, safer and faster to do that like that.

PR #577, opened 2020-05-19
Reconcile only a few values of KV configmap

tiraboschi:
Reconcile only a few values of KV configmap and only on upgrades

On upgrades only virtconfig.SmbiosConfigKey, virtconfig.MachineTypeKey, virtconfig.SELinuxLauncherTypeKey and "debug.useEmulation"
should be manipulated.
Everything else should keep its value to preserve
user changes.
This is going to happen only during HCO upgrades,
during the lifecycle of the version the user
is still allowed to edit them.

- use const from config-map.go
- set also debug.useEmulation according to KVM_EMULATION

**Release note**:
```release-note
Reconcile only a few values of KV configmap and only on upgrades
```

### DIFF
--- a/pkg/controller/hyperconverged/hyperconverged_controller.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller.go
@@ -679,10 +679,12 @@ func newKubeVirtConfigForCR(cr *hcov1alpha1.HyperConverged, namespace string) *c
 			Labels:    labels,
 			Namespace: namespace,
 		},
+		// only virtconfig.SmbiosConfigKey, virtconfig.MachineTypeKey, virtconfig.SELinuxLauncherTypeKey and
+		// "debug.useEmulation" are going to be manipulated and only on HCO upgrades
 		Data: map[string]string{
-			"feature-gates":       "DataVolumes,SRIOV,LiveMigration,CPUManager,CPUNodeDiscovery,Sidecar",
-			"migrations":          `{"nodeDrainTaintKey" : "node.kubernetes.io/unschedulable"}`,
-			"selinuxLauncherType": "spc_t",
+			virtconfig.FeatureGatesKey:        "DataVolumes,SRIOV,LiveMigration,CPUManager,CPUNodeDiscovery,Sidecar",
+			virtconfig.MigrationsConfigKey:    `{"nodeDrainTaintKey" : "node.kubernetes.io/unschedulable"}`,
+			virtconfig.SELinuxLauncherTypeKey: "spc_t",
 		},
 	}
 	val, ok := os.LookupEnv("SMBIOS")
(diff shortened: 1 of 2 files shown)


pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -723,13 +725,19 @@ func (r *ReconcileHyperConverged) ensureKubeVirtConfig(req *hcoRequest) error {
...
-		req.logger.Info("Updating existing KubeVirt config")
-		found.Data = kubevirtConfig.Data
-		err = r.client.Update(req.ctx, found)
-		if err != nil {
-			req.logger.Error(err, "Failed updating an existing kubevirt config")
-			return err
+	if r.upgradeMode {
+		// only virtconfig.SmbiosConfigKey, virtconfig.MachineTypeKey and selinuxLauncherType are
+		// going to be manipulated and only on HCO upgrades
+		for _, k := range []string{virtconfig.SmbiosConfigKey, virtconfig.MachineTypeKey, "selinuxLauncherType"} {
```
vatsalparekh, line 731:
```suggestion
		for _, k := range []string{virtconfig.SmbiosConfigKey, virtconfig.MachineTypeKey, virtconfig.SELinuxLauncherTypeKey} {
```
why not use `virtconfig.SELinuxLauncherTypeKey` like other two?

pkg/controller/hyperconverged/hyperconverged_controller_test.go
```diff
@@ -207,13 +212,17 @@ var _ = Describe("HyperconvergedController", func() {
...
+				// TODO: add negative test on the value we want to preserve
 				expectedResource := newKubeVirtConfigForCR(hco, namespace)
 				expectedResource.ObjectMeta.SelfLink = fmt.Sprintf("/apis/v1/namespaces/%s/dummies/%s", expectedResource.Namespace, expectedResource.Name)
 				outdatedResource := newKubeVirtConfigForCR(hco, namespace)
 				outdatedResource.ObjectMeta.SelfLink = fmt.Sprintf("/apis/v1/namespaces/%s/dummies/%s", outdatedResource.Namespace, outdatedResource.Name)
 				outdatedResource.Data[virtconfig.SmbiosConfigKey] = "old-smbios-value-that-we-have-to-update"
 				outdatedResource.Data[virtconfig.MachineTypeKey] = "old-machinetype-value-that-we-have-to-update"
+				// force upgrade mode
+				hco.Status.UpdateVersion(hcoVersionName, OldVersion)
+				os.Setenv(util.HcoKvIoVersionName, NewVersion)
```
nunnatsa, line 225: I think we need to add a new test case w/o upgrade. Is that what you meant in the comment on line 216?

tiraboschi:
Waiting for more unit tests
/hold

pkg/controller/hyperconverged/hyperconverged_controller_test.go
tiraboschi: done

PR #578, opened 2020-05-19
Use custom SELinux launcher type for virt-launcher

stu-gott:
```release-note
Use SELinux type defined by KubeVirt project for virt-launcher.
```

### DIFF
--- a/pkg/controller/hyperconverged/hyperconverged_controller.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller.go
@@ -765,7 +765,7 @@ func newKubeVirtConfigForCR(cr *hcov1alpha1.HyperConverged, namespace string) *c
 		Data: map[string]string{
 			virtconfig.FeatureGatesKey:        "DataVolumes,SRIOV,LiveMigration,CPUManager,CPUNodeDiscovery,Sidecar",
 			virtconfig.MigrationsConfigKey:    `{"nodeDrainTaintKey" : "node.kubernetes.io/unschedulable"}`,
-			virtconfig.SELinuxLauncherTypeKey: "spc_t",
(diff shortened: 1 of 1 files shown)


pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -472,7 +472,7 @@ func newKubeVirtConfigForCR(cr *hcov1alpha1.HyperConverged, namespace string) *c
 		Data: map[string]string{
 			"feature-gates":       "DataVolumes,SRIOV,LiveMigration,CPUManager,CPUNodeDiscovery,Sidecar",
 			"migrations":          `{"nodeDrainTaintKey" : "node.kubernetes.io/unschedulable"}`,
-			"selinuxLauncherType": "spc_t",
+			"selinuxLauncherType": "virt_launcher.process",
```
tiraboschi, line 475: Stu, do we also have to update it on existing deployments upgraded to this code?
Please see also: https://github.com/kubevirt/hyperconverged-cluster-operator/pull/577

stu-gott: Yes we would. What does that entail?

tiraboschi: We don't want to reconcile anymore the whole configMap to preserve user set values (on feature-gates for instance) so we have to know which keys should be reconciled on upgrades.

stu-gott:
@tiraboschi This PR has been rebased. It appears that changeset 27167a7af already accounts for the fact that this key needs to be re-evaluated upon update.

Is there anything further to do with this besides the one-line change?

ovirt-infra:
All tests passed

PR #580, opened 2020-05-20
Bump kubevirt to v0.29.2

tiraboschi:
Bump kubevirt to v0.29.2

**Release note**:
```release-note
Bump kubevirt to v0.29.2
```


tiraboschi:
Now it fails the upgrade test with:
```
      KubeVirt is degraded: An error occurred during deployment: unable to create config map &ConfigMap{ObjectMeta:{kubevirt-cpu-plugin-configmap  kubevirt-hyperconverged    0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[app.kubernetes.io/managed-by:kubevirt-operator] map[kubevirt.io/install-strategy-identifier:fcdd3654ec2707d5c36480805c84422f3371314d kubevirt.io/install-strategy-registry:docker.io/kubevirt kubevirt.io/install-strategy-version:v0.29.1] [] []  []},Data:map[string]string{cpu-plugin-configmap.yaml: obsoleteCPUs:
```
who is going to delete that configMap if already created in the past by SSP operator?

tiraboschi:
/unhold
We reconfigured openshift-ci webhook so now we are properly testing this

PR #583, opened 2020-05-21
Avoid sspv1.KubevirtNodeLabellerBundle and cpu-plugin on SSP operator

tiraboschi:
After
https://github.com/MarSik/kubevirt-ssp-operator/pull/167
https://github.com/MarSik/kubevirt-ssp-operator/pull/156
and
https://github.com/kubevirt/kubevirt/pull/3133
node-labeller and cpu plugin are not anymore part of ssp operator
and moved to kubevirt one.
Update HCO according to that.

**Release note**:
```release-note
Avoid sspv1.KubevirtNodeLabellerBundle and cpu-plugin on SSP operator
```

### DIFF
--- a/pkg/controller/hyperconverged/hyperconverged_controller.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller.go
@@ -129,7 +129,6 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		&cdiv1alpha1.CDI{},
 		&networkaddonsv1alpha1.NetworkAddonsConfig{},
 		&sspv1.KubevirtCommonTemplatesBundle{},
-		&sspv1.KubevirtNodeLabellerBundle{},
 		&sspv1.KubevirtTemplateValidator{},
 		&sspv1.KubevirtMetricsAggregation{},
 		&schedulingv1.PriorityClass{},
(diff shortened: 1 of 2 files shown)


ksimon1:
I think these flags can be deleted as well:
```
KVM_INFO_TAG
VIRT_LAUNCHER_TAG
NODE_LABELLER_TAG
CPU_PLUGIN_TAG
``` 
e.g. I found them here: https://github.com/kubevirt/hyperconverged-cluster-operator/blob/348e9e0a7cdd6afd4e54f9352ad786e3a7033494/deploy/operator.yaml#L221

tiraboschi:
> I think these flags can be deleted as well:
> 
> ```
> KVM_INFO_TAG
> VIRT_LAUNCHER_TAG
> NODE_LABELLER_TAG
> CPU_PLUGIN_TAG
> ```

They comes from the partial CSV we get running the csv-generator tool on the binary image of ssp-operator.
So we need a new release of the ssp-operator, then change hack/config to point there and run hack/build-manifest.sh

We should also probably integrate this with https://github.com/kubevirt/hyperconverged-cluster-operator/pull/580 to consume the new version of ssp-operator and the new version of virt oeprator at the same time.

The we also need to understand how to handle the upgrade for who deployed in the past when KubevirtNodeLabellerBundle was part of ssp operator.

tiraboschi:
ignore slightly reduced test coverage
/override coverage/coveralls

PR #584, opened 2020-05-21
Upgrade - check components

nunnatsa:
**Release note**:
```release-note
HCO, during upgrade, checks the upgrade status of the underline components.
```

### DIFF
--- a/pkg/controller/hyperconverged/hyperconverged_controller.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller.go
@@ -168,10 +168,11 @@ type ReconcileHyperConverged struct {
 
 type hcoRequest struct {
 	reconcile.Request
-	logger     logr.Logger
-	conditions hcoConditions
-	ctx        context.Context
-	instance   *hcov1alpha1.HyperConverged
+	logger                     logr.Logger
+	conditions                 hcoConditions
+	ctx                        context.Context
+	instance                   *hcov1alpha1.HyperConverged
+	componentUpgradeInProgress bool
 }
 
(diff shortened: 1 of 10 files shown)


nunnatsa:
/test hco-e2e-aws

/hold 
Waiting for a new CDI release with a fix to the ["Missing ObservedVersion field in Some Cases" issue](https://github.com/kubevirt/containerized-data-importer/issues/1212)

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -234,6 +236,7 @@ func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcil
 
 	if !r.upgradeMode && !init && knownHcoVersion != r.ownVersion {
 		r.upgradeMode = true
+		req.componentUpgradeDone = true
```
tiraboschi, line 239: if we detect the upgrade mode we track it with:
r.upgradeMode = true
but then
req.componentUpgradeDone = true

componentUpgradeDone should't we have the opposite on componentUpgradeDone because we know that somoething can also happen to other components?

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -433,6 +437,7 @@ func (r *ReconcileHyperConverged) ensureHco(req *hcoRequest) error {
 			})
 			return err
 		}
+		req.componentUpgradeDone = req.componentUpgradeDone && upgradeDone
```
tiraboschi, line 440: not really sure about this:
it was initialized to true, we set it to false on the first component that is still not upgrade.
When do we set it back to true?

PR #596, opened 2020-05-27
Add VM Import Operator to bump-bot + compare versions fix

orenc1:
1. Adding the new VM Import Operator to be sampled and bumped by the bot.
2. Fixing an issue where the most recent component version is not tagged as "latest" - in such case, the greater version number will win.

**Release note**:

```release-note
Add VM Import Operator to Bump-Bot
```


orenc1:
/hold
need to refactor to support releases with non-numerical tags (e.g. "rc")

orenc1:
> /hold
> need to refactor to support releases with non-numerical tags (e.g. "rc")
Semvers with letters are now supported.

PR #601, opened 2020-05-28
network: enable KubeMacPool again

phoracek:
It should be stabilized and good to go.

Signed-off-by: Petr Horacek <phoracek@redhat.com>

**Release note**:

```release-note
KubeMacPool is part of the installation again
```

### DIFF
--- a/pkg/controller/hyperconverged/hyperconverged_controller.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller.go
@@ -1003,6 +1003,7 @@ func newNetworkAddonsForCR(cr *hcov1alpha1.HyperConverged, namespace string) *ne
 			LinuxBridge: &networkaddonsv1alpha1.LinuxBridge{},
 			Ovs:         &networkaddonsv1alpha1.Ovs{},
 			NMState:     &networkaddonsv1alpha1.NMState{},
+			KubeMacPool: &networkaddonsv1alpha1.KubeMacPool{},
 		},
 	}
 }
--- a/pkg/controller/hyperconverged/hyperconverged_controller.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller.go
@@ -1040,7 +1040,7 @@ func (r *ReconcileHyperConverged) ensureNetworkAddons(req *hcoRequest) (upgradeD
 		}
 	}
 
-	if !reflect.DeepEqual(found.Spec, networkAddons.Spec) {
+	if !reflect.DeepEqual(found.Spec, networkAddons.Spec) && !r.upgradeMode {
 		req.logger.Info("Updating existing Network Addons")
 		found.Spec = networkAddons.Spec
 		return false, r.client.Update(req.ctx, found)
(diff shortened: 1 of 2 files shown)


pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -1003,6 +1003,7 @@ func newNetworkAddonsForCR(cr *hcov1alpha1.HyperConverged, namespace string) *ne
 			LinuxBridge: &networkaddonsv1alpha1.LinuxBridge{},
 			Ovs:         &networkaddonsv1alpha1.Ovs{},
 			NMState:     &networkaddonsv1alpha1.NMState{},
+			KubeMacPool: &networkaddonsv1alpha1.KubeMacPool{},
```
tiraboschi, line 1006: This is probably more related to line 1044; now we have a flag (`r.upgradeMode`) to signal that we are in the middle of an upgrade.
Do we have to reconcile this config map continuously or should we coordinate this with the upgrade process?

phoracek: When we add this attribute, it will render the status of CNAO "Progressing". Would that ensure that we signal the upgrade?

tiraboschi: It's detected by HCO comparing the value of a new environmental variable read from the CSV with what was stored on HCO CR in the previous run.
We are not going to reset it until all the observed components reports Available=true, Degraded=false, Progressing=False.
With https://github.com/kubevirt/hyperconverged-cluster-operator/pull/584 we also started explicitly checking the version reported by each single operator (still not all of them).

While with `r.upgradeMode=true`, HCO is not going to report true with the readiness probe so that OLM, and other users, should see it less flappy.

phoracek: I'm still not sure I follow. Is there any extra action needed to support this PR?

tiraboschi: OLM is not transactional and doesn't have any rollback mechanism.
The question is basically if directly changing this config map, regardless of the upgrade status of CNAO, is safe enough or if it can cause more troubles to an older version of CNAO that for, any reason, is still not completely progressed to a newer version.

phoracek: I see. It should be able to progress to a healthy state. However, we don't want to consume KubeMacPool from an old version of CNAO. Can we avoid changing this if we see the upgrade is not completed yet?

tiraboschi: yes, this is exactly what I'm suggesting.
HCO will set `r.upgradeMode=true` as soon as it detects that it's running with a version newer/different from what we tracked in the CSV.
Then we check the upgrade status of CNAO on line 1060; we can consume this there.

phoracek: I think I get it now. I'm not sure what would be the best solution, but saying "not reconfigure while upgrading" sounds like a safe bet.

PR #618, opened 2020-06-03
Avoid deploying NodeLabellerBundle when running with KVM_EMULATION

tiraboschi:
Avoid deploying NodeLabellerBundle when running
with KVM_EMULATION because it will be completely useless

**Release note**:
```release-note
Avoid deploying NodeLabellerBundle when running with KVM_EMULATION
```

### DIFF
--- a/hack/deploy.sh
+++ b/hack/deploy.sh
@@ -28,8 +28,8 @@ HCO_RESOURCE_NAME="kubevirt-hyperconverged"
 HCO_CRD_NAME="hyperconvergeds.hco.kubevirt.io"
 
 CI=""
-if [ "$1" == "CI" ]; then
-	echo "deploying on CI"
+if [ -n "$KUBEVIRT_PROVIDER" ]; then
+	echo "deploying on KubevirtCI ${KUBEVIRT_PROVIDER}"
 	CI="true"
 elif [ "$HOSTNAME" == "hco-e2e-aws" ] || [ "$HOSTNAME" == "e2e-aws-cnv" ]; then
 	echo "deploying on AWS CI"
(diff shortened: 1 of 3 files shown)


ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

dankenigsberg:
@tiraboschi  would you explain the motivation for this PR? What are its downsides?

tiraboschi:
> @tiraboschi would you explain the motivation for this PR? What are its downsides?

We are already configuring KubevirtNodeLabellerBundle according to KVM_EMULATION env variable with a parameter in the spec section of its CR, see:
https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/pkg/controller/hyperconverged/hyperconverged_controller.go#L1276

But I understood from @ksimon1 that this is just a waste of resources and deployment time because even in that case KubevirtNodeLabeller will be instantiated by SSP operator although completely useless.

ovirt-infra:
All tests passed

ksimon1:
> > @tiraboschi would you explain the motivation for this PR? What are its downsides?
> 
> We are already configuring KubevirtNodeLabellerBundle according to KVM_EMULATION env variable with a parameter in the spec section of its CR, see:
> https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/pkg/controller/hyperconverged/hyperconverged_controller.go#L1276
> 
> But I understood from @ksimon1 that this is just a waste of resources and deployment time because even in that case KubevirtNodeLabeller will be instantiated by SSP operator although completely useless.

Yes, that is correct. If labeller doesn't have access to kvm, it is useless, because it can report falsely usable cpu model / feature.

tiraboschi:
ignoring sliglty decreased code coverage because we cannot really rely on real KVM support on upstream CI environments
/override coverage/coveralls

tiraboschi:
obsolete

PR #634, opened 2020-06-09
Remove old SSP CRDs

nunnatsa:
**Release note**:

```release-note
Remove old, unused CRDs from SSP
```

PR #636, opened 2020-06-10
re-enable SSP components conditions checks

orenc1:
SSP have been implemented the standard conditions HCO expects to reconcile on.
Those checks are now restored.

**Release note**:

```release-note
NONE
```


orenc1:
closing in favour of #634 - which removes the old SSP CRDs as well after upgrade completion.
/close

PR #639, opened 2020-06-16
Intercept delete in validating webhook

yuvalturg:
Previously, when trying to delete the HCO resource while VMs exist in
the cluster, the HCO would stay in Terminating state.  This patch
implements a webhook that intercepts delete requests for the HCO.
When validating a delete request, the webhook will try to remove the
KubeVirt and CDI resources in dry run mode, and reject the request in
case the removal of those resources fail.

**Release note**:

```release-note
Intercept delete requests in validating webhook
```


ovirt-infra:

yuvalturg:
> I still think HyperConverged is the wrong place for the New resource methods. But if this is the only option, then all of the sub resource creation methods should be there, not only few of them.

True, they should all be there, but I only moved those that were on the way towards sharing the deletion code.  We can address them in a different patch.
As for the location, like we discussed previously, I really dont mind moving it to a HyperConvergedFactory, but it should be part of v1alpha1, otherwise we go into cyclic imports etc

PR #644, opened 2020-06-19
pin HCO to operator-sdk v0.17.0

phoracek:
cluster-network-addons-operator pulls in dependency on operator-sdk
v0.18.0. This depedency is not exercised on context of HCO. However,
since HCO is pinned to Kubernetes v1.16, it has to stay on operator-sdk
v0.17.0 to keep compatibility. This patch makes sure we pin the
dependency version.

Signed-off-by: Petr Horacek <phoracek@redhat.com>

**Release note**:

```release-note
NONE
```


ovirt-infra:
Some tests failed

PR #669, opened 2020-06-29
Create a PriorityClass for KubeVirt on startup

yuvalturg:
Until now, the PriorityClass for KubeVirt was created during
reconciliation (after creating the CR for the HCO).  This caused the
virt-operator to exit with a FailedCreate status, and the CSV to stay in
a Failed status.
This patch creates the PriorityClass when the HCO starts, allowing the
virt-operator to be installed properly and the CSV's status to change to
Succeeded.

**Release note**:

```release-note
Create KubeVirt's PriorityClass when HCO starts (BZ#1851856)
```

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -5,14 +5,16 @@ import (
 	"fmt"
 
 	"errors"
+	"os"
+
 	"github.com/go-logr/logr"
 	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
-	"os"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	csvv1alpha1 "github.com/operator-framework/operator-lifecycle-manager/pkg/api/apis/operators/v1alpha1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	schedulingv1 "k8s.io/api/scheduling/v1"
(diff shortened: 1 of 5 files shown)


ovirt-infra:

pkg/util/util.go
```diff
@@ -96,3 +98,20 @@ func GetCSVfromPod(pod *corev1.Pod, c client.Client, logger logr.Logger) (*csvv1
...
 }
+
+func NewKubeVirtPriorityClass() *schedulingv1.PriorityClass {
+	return &schedulingv1.PriorityClass{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "scheduling.k8s.io/v1",
+			Kind:       "PriorityClass",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "kubevirt-cluster-critical",
```
tiraboschi, line 109: I think that we should also enforce Labels as we are doing for other resources, see:
https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/pkg/controller/hyperconverged/hyperconverged_controller.go#L760

tiraboschi: Rethinking about this: we should label the priority class for tracking purposes but we should not try to delete it when the user deletes HCO CR, I'll address this in a separate PR

tiraboschi:
/override coverage/coveralls
ignoring slightly reduced coverage

PR #670, opened 2020-06-30
Avoid deleting KubeVirt Priority Class

tiraboschi:
Avoid deleting KubeVirt Priority Class

KV Priority Class is needed to start virt-operator
but OLM is currently not able to create it so HCO
is trying to create it ASAP.
With #669
we are trying to create it as soon as HCO starts.
But, for the same reason, we should not trying to
delete it it when the user deletes HCO CR because
virt-operator should continue running even if HCO
CR got deleted.
In the past we were trying to delete the priority
class but we was failing on that just because we forgot
to set app label on that and componentResourceRemoval is
explictly checking for that before really deleting.

**Release note**:
```release-note
Avoid deleting KV Priority Class when the user deletes HCO CR
```

### DIFF
--- a/cmd/hyperconverged-cluster-operator/main.go
+++ b/cmd/hyperconverged-cluster-operator/main.go
@@ -233,8 +233,13 @@ func main() {
 	}
 }
 
+// KubeVirtPriorityClass is needed by virt-operator but OLM is not able to
+// create it so we have to create it ASAP.
+// When the user deletes HCO CR virt-operator should continue running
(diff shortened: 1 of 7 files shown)


ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

PR #690, opened 2020-07-07
Simplify ensureKubeVirtStorageConfig

maya-r:
As discussed in #498, this function only has upgrade logic because it's called by reconcile.
By calling it from ensureCDI we can drop the upgrade logic.

Catch up with this signature in tests.

**Release note**:
```release-note
NONE
```


ovirt-infra:
All tests passed

PR #698, opened 2020-07-13
Make the default binding mechanism to  be Masquerade

alonSadan:
Make the default binding mechanism to be `masquerade`

Since HCO's focus are legacy workloads where live migration is a common practice,
It is required to encourage  the use of `masquerade` binding over bridge.
However, the `bridge` binding is still available and it is even the default,
which sometimes makes the usage confusing for users.
That's why we would like `masquerade` to be our default binding mechanism.

This PR injects `masquerade` value to the default network interface field in kubevirt configMap.
Then this value will be used when installing kubevirt, thus making masquerade as the
default binding mechanism.

another thing changed in this PR is to make the controller to
to vendor the configMap keys and not hardcode them.
Signed-off-by: alonSadan <asadan@redhat.com>

**Release note**:

```release-note
Expose kubevirt default network interface in HCO-CR, give it a value of `masquerade` and pass it to kubevirt configMap 
```

### DIFF
--- a/pkg/controller/hyperconverged/hyperconverged_controller.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller.go
@@ -79,6 +79,8 @@ const (
 	metricsAggregationOldCrdName    = "kubevirtmetricsaggregations.kubevirt.io"
 	nodeLabellerBundlesOldCrdName   = "kubevirtnodelabellerbundles.kubevirt.io"
 	templateValidatorsOldCrdName    = "kubevirttemplatevalidators.kubevirt.io"
+
+	kubevirtDefaultNetworkInterfaceValue = "masquerade"
 )
 
 // Add creates a new HyperConverged Controller and adds it to the Manager. The Manager will set fields on the Controller
@@ -764,11 +766,12 @@ func newKubeVirtConfigForCR(cr *hcov1beta1.HyperConverged, namespace string) *co
 			Namespace: namespace,
 		},
 		// only virtconfig.SmbiosConfigKey, virtconfig.MachineTypeKey, virtconfig.SELinuxLauncherTypeKey and
-		// "debug.useEmulation" are going to be manipulated and only on HCO upgrades
+		// virtconfig.UseEmulationKey are going to be manipulated and only on HCO upgrades
 		Data: map[string]string{
 			virtconfig.FeatureGatesKey:        "DataVolumes,SRIOV,LiveMigration,CPUManager,CPUNodeDiscovery,Sidecar",
 			virtconfig.MigrationsConfigKey:    `{"nodeDrainTaintKey" : "node.kubernetes.io/unschedulable"}`,
 			virtconfig.SELinuxLauncherTypeKey: "virt_launcher.process",
+			virtconfig.NetworkInterfaceKey:    kubevirtDefaultNetworkInterfaceValue,
 		},
 	}
 	val, ok := os.LookupEnv("SMBIOS")
@@ -781,7 +784,7 @@ func newKubeVirtConfigForCR(cr *hcov1beta1.HyperConverged, namespace string) *co
 	}
 	val, ok = os.LookupEnv("KVM_EMULATION")
 	if ok && val != "" {
-		cm.Data["debug.useEmulation"] = val
+		cm.Data[virtconfig.UseEmulationKey] = val
 	}
 	return cm
 }
@@ -818,12 +821,12 @@ func (r *ReconcileHyperConverged) ensureKubeVirtConfig(req *hcoRequest) (upgrade
 
 	if r.upgradeMode {
 		// only virtconfig.SmbiosConfigKey, virtconfig.MachineTypeKey, virtconfig.SELinuxLauncherTypeKey and
-		// "debug.useEmulation" are going to be manipulated and only on HCO upgrades
+		// virtconfig.UseEmulationKey are going to be manipulated and only on HCO upgrades
 		for _, k := range []string{
 			virtconfig.SmbiosConfigKey,
 			virtconfig.MachineTypeKey,
 			virtconfig.SELinuxLauncherTypeKey,
-			"debug.useEmulation",
+			virtconfig.UseEmulationKey,
 		} {
 			if found.Data[k] != kubevirtConfig.Data[k] {
 				req.logger.Info(fmt.Sprintf("Updating %s on existing KubeVirt config", k))
--- a/pkg/controller/hyperconverged/hyperconverged_controller_components_test.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller_components_test.go
@@ -124,7 +124,7 @@ var _ = Describe("HyperConverged Components", func() {
 		var req *hcoRequest
 
 		updatableKeys := [...]string{virtconfig.SmbiosConfigKey, virtconfig.MachineTypeKey, virtconfig.SELinuxLauncherTypeKey}
-		unupdatableKeys := [...]string{virtconfig.FeatureGatesKey, virtconfig.MigrationsConfigKey}
+		unupdatableKeys := [...]string{virtconfig.FeatureGatesKey, virtconfig.MigrationsConfigKey, virtconfig.NetworkInterfaceKey}
 
 		BeforeEach(func() {
 			hco = newHco()
@@ -182,6 +182,7 @@ var _ = Describe("HyperConverged Components", func() {
 			// values we should preserve
 			outdatedResource.Data[virtconfig.FeatureGatesKey] = "old-featuregates-value-that-we-should-preserve"
 			outdatedResource.Data[virtconfig.MigrationsConfigKey] = "old-migrationsconfig-value-that-we-should-preserve"
+			outdatedResource.Data[virtconfig.NetworkInterfaceKey] = "old-defaultnetworkinterface-value-that-we-should-preserve"
 
 			cl := initClient([]runtime.Object{hco, outdatedResource})
 			r := initReconciler(cl)
@@ -221,6 +222,7 @@ var _ = Describe("HyperConverged Components", func() {
 			// values we should preserve
 			outdatedResource.Data[virtconfig.FeatureGatesKey] = "old-featuregates-value-that-we-should-preserve"
 			outdatedResource.Data[virtconfig.MigrationsConfigKey] = "old-migrationsconfig-value-that-we-should-preserve"
+			outdatedResource.Data[virtconfig.DefaultNetworkInterface] = "old-defaultnetworkinterface-value-that-we-should-preserve"
 
 			cl := initClient([]runtime.Object{hco, outdatedResource})
 			r := initReconciler(cl)


ovirt-infra:

pkg/apis/hco/v1alpha1/hyperconverged_types.go
```diff
@@ -29,6 +29,9 @@ type HyperConvergedSpec struct {
 
 	// operator version
 	Version string `json:"version,omitempty"`
+
+	//configurations to be passed to kubevirt config-map
+	KubevirtConfiguration kubevirtv1.KubeVirtConfiguration `json:"kubevirtConfiguration,omitempty"`
```
tiraboschi, line 34: We are going to move to v1beta1 with https://github.com/kubevirt/hyperconverged-cluster-operator/pull/695

I'd suggest to do this only there.
We also have to properly think about the upgrade path for who already deployed in the past.

alonSadan: Thank you @tiraboschi I'll try and sync with #695. 
About the second point, can you elaborate?
We are changing the default, but anyone who want's can change it back, it also stated for a long time the we 
recommend to use masquerade and not bridge. so at least in that case we are "coverd". aren't we?
@phoracek  what is your opinion about that second point?

alonSadan:
> Please sync with #695

Thanks. will do

alonSadan:
> We think that this is too much,
> can we limit it to the scope described in the title of the PR (Make the default binding mechanism to be Masquerade) ?

Thanks @tiraboschi. 
I will the required changes

nunnatsa:
> > We think that this is too much,
> > can we limit it to the scope described in the title of the PR (Make the default binding mechanism to be Masquerade) ?
> 
> Thanks @tiraboschi.
> I will make the required changes

If you can also not do it with KV API, to avoid strong coupling, and future integration error.

Also,

```yaml
spec:
  configuration:
     ...
```

That leave no place for other components. Let's do something like:
```yaml
spec:
  configuration:
    kubevirt:
      ...
```

alonSadan:
Hey @tiraboschi  @nunnatsa  @orenc1
I made the changes we talked about.
now only the default binding is exposed in it is given a value of `masquerade`.
would much appreciate your reviews.

another thing: how do I get rid of this [`make sanity` error](https://travis-ci.org/github/kubevirt/hyperconverged-cluster-operator/builds/708733245#L6286)?? in my laptop it doesn't error about difs... 

Thanks:)

tools/operator-courier/Dockerfile
```diff
@@ -1,5 +1,6 @@
 FROM registry.access.redhat.com/ubi8/python-36
 
+RUN pip3 install operator-courier
```
nunnatsa, line 3: Done in line 4. Please remove

deploy/hco.cr.yaml
```diff
@@ -3,4 +3,7 @@ apiVersion: hco.kubevirt.io/v1alpha1
 kind: HyperConverged
 metadata:
   name: kubevirt-hyperconverged
-spec: {}
+spec:
+  configuration:
+    kubevirt:
+      default-network-interface: masquerade
```
phoracek, line 9: If we want this to be used when people install HCO through OLM, it needs to be set in CSV as the default example (@tiraboschi keep me honest here). We may also need to deliver this change to all the automation scripts we have where HCO CR is stored without this setting.

Would it make sense to not expose it at all and just hardcode it as the default for everybody?

orenc1: this will have to be in the CSV's `metadata.annotations.alm-examples` do be visible for the user who creating the CR from OLM UI upon installation, yes.

orenc1: unless we hardcoding the defaults, and the CR should contain only changes over the default values.
btw, we should definitely have a proper validation on the schema.

phoracek: I think we should change the default internally and at most allow people to override it. Otherwise, I have doubts it will reach to all

alonSadan: > I think we should change the default internally and at most allow people to override it. Otherwise, I have doubts it will reach to all
hmmm I think it makes sense. Though a bit enforcing, we want the users to get the preferred configuration when they deploy. The freedom remains in the form of applying a config map, after kubevirt is deployed, and that is  the user's/admin's responsibility at that stage.
when I wrote the PR I thought that exposing the configuration is the right way to go, but now I am not so sure. So, if there are no objections(@tiraboschi  @nunnatsa  @orenc1 ) then I think I will hard-code the default to be masquerade.

tools/operator-courier/Dockerfile
alonSadan: Done

pkg/apis/hco/v1alpha1/hyperconverged_types.go
```diff
@@ -14,6 +14,16 @@ import (
 // HyperConvergedName is the name of the HyperConverged resource that will be reconciled
 const HyperConvergedName = "kubevirt-hyperconverged"
 
+// KubevirtConfiguration are the default configurations to be passed to kubevirt-configMap
+type KubevirtConfiguration struct {
+	DefaultNetworkInterface string `json:"default-network-interface,omitempty"`
```
nunnatsa, line 19: JSON naming should be camelCase. Please change to:

```golang
DefaultNetworkInterface string `json:"defaultNetworkInterface,omitempty"`
```

alonSadan: @nunnatsa 
I understand it is a strong convention. I made it this way (with the `-`), so that it will match the keys in kubevirt-configMap. 
https://github.com/kubevirt/kubevirt/blob/master/pkg/virt-config/config-map.go#L52
I am not sure it is a good enough reason.
what do you think?

nunnatsa: > what do you think?

it should be json complaint

alonSadan: Done.

nunnatsa:
/hold

wait for v1beta1

nunnatsa:
v1beta1 merged (#695). Please move your v1alpha1 code to v1beta1.

alonSadan:
> v1beta1 merged (#695). Please move your v1alpha1 code to v1beta1.

Great. Will do

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -764,11 +766,12 @@ func newKubeVirtConfigForCR(cr *hcov1beta1.HyperConverged, namespace string) *co
 			Namespace: namespace,
 		},
 		// only virtconfig.SmbiosConfigKey, virtconfig.MachineTypeKey, virtconfig.SELinuxLauncherTypeKey and
-		// "debug.useEmulation" are going to be manipulated and only on HCO upgrades
+		// "debug.useEmulation" are going to be manipulated and only on HCO upgrades;
 		Data: map[string]string{
 			virtconfig.FeatureGatesKey:        "DataVolumes,SRIOV,LiveMigration,CPUManager,CPUNodeDiscovery,Sidecar",
 			virtconfig.MigrationsConfigKey:    `{"nodeDrainTaintKey" : "node.kubernetes.io/unschedulable"}`,
 			virtconfig.SELinuxLauncherTypeKey: "virt_launcher.process",
+			virtconfig.NetworkInterfaceKey:    "masquerade",
```
nunnatsa, line 774: Don't you want to use the constant you defined in line 83?

alonSadan: correct!
Done.

tools/operator-courier/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM registry.access.redhat.com/ubi8/python-36
 
-RUN pip3 install operator-courier
+RUN pip3 install --upgrade operator-courier==2.1.9
```
nunnatsa, line 3: this is not required anymore. It was a temporary workaround

alonSadan: Tnx. removed it

pkg/controller/hyperconverged/hyperconverged_controller_components_test.go
```diff
@@ -182,6 +182,7 @@ var _ = Describe("HyperConverged Components", func() {
 			// values we should preserve
 			outdatedResource.Data[virtconfig.FeatureGatesKey] = "old-featuregates-value-that-we-should-preserve"
 			outdatedResource.Data[virtconfig.MigrationsConfigKey] = "old-migrationsconfig-value-that-we-should-preserve"
+			outdatedResource.Data[virtconfig.DefaultNetworkInterface] = "old-defaultnetworkinterface-value-that-we-should-preserve"
```
RamLavi, line 185: not sure I understand what this does. can you elaborate?

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -79,6 +79,8 @@ const (
 	metricsAggregationOldCrdName    = "kubevirtmetricsaggregations.kubevirt.io"
 	nodeLabellerBundlesOldCrdName   = "kubevirtnodelabellerbundles.kubevirt.io"
 	templateValidatorsOldCrdName    = "kubevirttemplatevalidators.kubevirt.io"
+
```
RamLavi, line 82: please correct me if I'm wrong, but according to what I [see](https://github.com/kubevirt/kubevirt/blob/ddce8abb40142ccc229feb77593719965e6ff431/pkg/virt-config/config-map.go#L351) - masquarade is already default in kubeivrt.

alonSadan: please note that in the link you sent, it parses the values in the configMap, and it accepts `Bridge`, `Slirp` and `Masquerade` as proper values, but it will not determine the default interface, only parsing. 
The value is determined [here](https://github.com/kubevirt/kubevirt/blob/ddce8abb40142ccc229feb77593719965e6ff431/pkg/virt-config/config-map.go#L162) and it is being set [here](https://github.com/kubevirt/kubevirt/blob/ddce8abb40142ccc229feb77593719965e6ff431/pkg/virt-config/virt-config.go)

that default configuration will take place unless who ever creates the configMap states otherwise. in this case the HCO is creating the configMap, and if we will inject `masquerade` it will not use kubevirt-default-value which is bridge.

note that if someone will deploy kubevirt without HCO it will still get bridge as default.

pkg/controller/hyperconverged/hyperconverged_controller_components_test.go
alonSadan: Test that when we upgrade kubevirt,  the default network configuration will stay intact.
but thanks to you  I saw I forgot to add it's key [here](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/2d1bf98db4d20d4f8f82248f91254ad0ac078152/pkg/controller/hyperconverged/hyperconverged_controller_components_test.go#L127)

Will change that.

pkg/controller/hyperconverged/hyperconverged_controller.go
RamLavi: ack thanks!

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -781,7 +784,7 @@ func newKubeVirtConfigForCR(cr *hcov1beta1.HyperConverged, namespace string) *co
 	}
 	val, ok = os.LookupEnv("KVM_EMULATION")
 	if ok && val != "" {
-		cm.Data["debug.useEmulation"] = val
+		cm.Data[virtconfig.UseEmulationKey] = val
```
orenc1, line 787: Please also update the comments in lines 769 and 824 for consistency.

alonSadan: Done

orenc1:
tests passed on aws, overriding azure (hit public IP allocation limit).
/override ci/prow/hco-e2e-azure
/override ci/prow/hco-e2e-upgrade-azure
/override ci/prow/hco-e2e-upgrade-prev-azure

alonSadan:
Good Morning.
@orenc1 @nunnatsa  can you please approve so it will be merged? 
Thanks.

PR #747, opened 2020-08-13
Dirty hack to unlock CI

tiraboschi:
This is just a dirty hack to unlock CI harcoding (again) the name of the registry used by openshift-ci.
We should find the proper way to consume at bundle build time, but with this at least we can unlock CI and gain more time.

Please be aware that this is going  to fail if our jobs get migrated from build01 to build02 and so on.

**Release note**:
```release-note
NONE
```


ovirt-infra:
All tests passed

tiraboschi:
/unhold
it's really a dirty hack, but at least we can continue working on other stuff while we properly address this

PR #784, opened 2020-09-07
Adapt to latest origin-operator-registry image

yuvalturg:
/registry was removed from the latest origin-operator-registry image,
and the workdir was changed as well, this breaks the hco's bundle image
build.

**Release note**:

```release-note
None
```

### DIFF
--- a/deploy/Dockerfile
+++ b/deploy/Dockerfile
@@ -1,9 +1,17 @@
 FROM quay.io/openshift/origin-operator-registry:latest
 
-COPY deploy/olm-catalog /registry
+WORKDIR /registry
+
+COPY deploy/olm-catalog .
+
+USER root
 
 # Initialize the database
-RUN initializer --manifests /registry/kubevirt-hyperconverged --output bundles.db
+RUN ["initializer", "--manifests", "/registry/kubevirt-hyperconverged", "--output", "bundles.db"]
+
+RUN ["chown", "1001", "bundles.db"]
+
+USER 1001
 
 # There are multiple binaries in the origin-operator-registry
 # We want the registry-server


ovirt-infra:

deploy/Dockerfile
```diff
@@ -1,9 +1,17 @@
 FROM quay.io/openshift/origin-operator-registry:latest
 
-COPY deploy/olm-catalog /registry
+WORKDIR /registry
+
+COPY deploy/olm-catalog .
+
+USER root
```
orenc1, line 7: Could you please elaborate on why is that necessary?

yuvalturg: Sure, tl;dr - https://github.com/operator-framework/operator-registry/pull/430 introduced a change that breaks our image.  Previously /registry was created in the operator-registry image, and had g+rwx permissions, so we could create and use /registry/bundles.db.  This is not the case anymore, so when running initializer, the bundles.db can't be written to /registry.  In addition to that, there is not /bin/sh, so RUN should be used with exec and not with `sh -c` - they are working on a fix in https://github.com/operator-framework/operator-registry/pull/435 but with this change, we won't depend on /bin/sh

deploy/Dockerfile
```diff
@@ -1,9 +1,17 @@
 FROM quay.io/openshift/origin-operator-registry:latest
 
-COPY deploy/olm-catalog /registry
+WORKDIR /registry
+
+COPY deploy/olm-catalog .
```
orenc1, line 5: Isn't it the same result? Why did you split `COPY deploy/olm-catalog /registry` into two dockerfile commands?

yuvalturg: Yes and no... yes because the end result will be the same - having /registry/kubevirt-hyperconverged, and no because we need WORKDIR to be set for the RUN and ENTRYPOINT (it was set previously on the operator-registry as well), but it was removed.  The alternative would be to keep the COPY command as it is, and move WORKDIR after, does it matter ?

orenc1: /lgtm

tiraboschi (approved):

PR #791, opened 2020-09-10
Bump VM_IMPORT to v0.2.1

pkliczewski:
Bump VM_IMPORT to v0.2.1

**Release note**:
```release-note
Bump VM_IMPORT to v0.2.1
```

Signed-off-by: Piotr Kliczewski <piotr.kliczewski@gmail.com>


ovirt-infra:
All tests passed

deploy/olm-catalog/kubevirt-hyperconverged/1.2.0/kubevirt-hyperconverged-operator.v1.2.0.clusterserviceversion.yaml
```diff
@@ -1793,15 +1793,15 @@ spec:
                 - name: CONTROLLER_IMAGE
                   value: docker.io/kubevirt/cdi-controller@sha256:89937377c6e851ee9ba85d861b87d49b87084ef2b7e05257ceed779529552d30
                 - name: IMPORTER_IMAGE
-                  value: docker.io/kubevirt/cdi-importer@sha256:a5d9743e30a097969d5905ea1bcb5eefb0b7e22b43313a30cc51dc5d43316a91
+                  value: docker.io/kubevirt/cdi-importer@
```
pkliczewski, line 1796: @tiraboschi I see that `make build-manifests` removed hashes but `make sanity` fails on it. Is this expected?

pkliczewski: Interesting. I run `make build-manifests` and the hashes are back. I consider it to be a feature :smile:

ovirt-infra:
All tests passed

deploy/olm-catalog/kubevirt-hyperconverged/1.2.0/kubevirt-hyperconverged-operator.v1.2.0.clusterserviceversion.yaml
tiraboschi: We use `skopeo inspect` to fetch the digest of the image starting from the tag, "sometimes" it can fail due to network errors and so the empty digest but yes, than we have `make sanity` to double check (and we run it also on Travis).
@nunnatsa maybe we should try to fail earlier on `make build-manifests`

nunnatsa: Yes. But I prefer some kind of caching - We're reading the same value over and over again.

e.g.  have a file with all the digests, that will only built if the hack/config file was changed.

ovirt-infra:
All tests passed

nunnatsa (approved):
/approve

PR #799, opened 2020-09-14
Bump CDI to v1.23.0

tiraboschi:
Bump CDI to v1.23.0

Signed-off-by: Simone Tiraboschi <stirabos@redhat.com>

**Release note**:
```release-note
Bump CDI to v1.23.0
```

PR #800, opened 2020-09-15
Build manifests outside of docker.

maya-r:
Needed by CDI v1.23.0 which dynamically generates the CDI CRD,
which requires Go and source code.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/build-manifests.sh
+++ b/hack/build-manifests.sh
@@ -66,10 +66,15 @@ function gen_csv() {
   local crds="${operatorName}.crds.yaml"
 
   # TODO: Use oc to run if cluster is available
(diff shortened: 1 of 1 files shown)


maya-r:
Related to #799

ovirt-infra:
All tests passed

tiraboschi:
We run `make sanity` on Travis to validate the PR,
I don't think we can do this on Travis.

nunnatsa (changes requested):
This solution does not seems like a long time solution, but a temporary fix of another component (CDI) I would prefer new fixed version of CDI instead of this fix.

awels:
https://github.com/kubevirt/containerized-data-importer/pull/1372 fixes it for master, I have the cherry pick already lined up.

awels:
I have release CDI 1.23.1 which includes a fix for this issue.

PR #812, opened 2020-09-18
[CI] Fix the address of openshift-ci internal registry for 4.6

tiraboschi:
[CI] Fix the address of openshift-ci internal registry for 4.6

openshift-ci is currently configured to use two different
addresses for its internal registry depending from the OCP version:
- OCP 4.5: registry.svc.ci.openshift.org
- OCP 4.6: default-route-openshift-image-registry.apps.build01.ci.devcluster.openshift.com

Fix the upgrade test lanes for OCP 4.6 now that we
are running again on OCP 4.6.

This is going to be solved only once we will move to openshift-ci generated
index images also for the upgrade tests.

**Release note**:
```release-note
NONE
```


ovirt-infra:
All tests passed

PR #823, opened 2020-09-22
Disable VMI migration upon K8s unschedulable taint

danielBelenky:
Currently, the default behaviour is to migrate the VMI when a node is
tainted with the node.kubernetes.io/unschedulable key. This behaviour is
wrong and can be un-expected from the user's perspective because the
unschedulable taint means "do not schedule new workloads on this node"
and not "evict existing workloads from this node".

kubevirt/kubevirt#4012 adds support for the
eviction API so we can properly handle node drains and specific
evictions on VMI pods.

This patch avoid setting (and explicitly removes during upgrades if
set in the past) the default node drain key so we won't migrate VMIs
when we're not expected to do so.

Fixes: https://bugzilla.redhat.com/1881676

Signed-off-by: Daniel Belenky <dbelenky@redhat.com>

**Release note**:

```release-note
VMIs will no longer migrate when node is tainted with node.kubernetes.io/unschedulable by default. Users can now use the proper node drain API to evacuate multiple VMIs from a node.
```

### DIFF
--- a/pkg/controller/hyperconverged/hyperconverged_controller.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller.go
@@ -771,11 +771,11 @@ func newKubeVirtConfigForCR(cr *hcov1beta1.HyperConverged, namespace string) *co
 		// only virtconfig.SmbiosConfigKey, virtconfig.MachineTypeKey, virtconfig.SELinuxLauncherTypeKey,
 		// virtconfig.FeatureGatesKey and virtconfig.UseEmulationKey are going to be manipulated
 		// and only on HCO upgrades.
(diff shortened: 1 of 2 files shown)


danielBelenky:
/hold

Waiting for https://github.com/kubevirt/kubevirt/pull/4012

ovirt-infra:
All tests passed

tiraboschi (changes requested):
Please notice that virtconfig.MigrationsConfigKey is not reconciled so this is enough for new deployment but not for upgrades

stu-gott (approved):
/lgtm

ovirt-infra:
All tests passed

tiraboschi (approved):

nunnatsa (approved):
/approve

PR #832, opened 2020-09-24
Reconcile IMS ConfigMap

tiraboschi:
IMS ConfigMap should be reconciled to HCO opinionated values.

Fixes: https://bugzilla.redhat.com/1879993

Signed-off-by: Simone Tiraboschi <stirabos@redhat.com>

**Release note**:
```release-note
Reconcile IMS ConfigMap
```

PR #854, opened 2020-10-02
Partially reconcile IMS ConfigMap

tiraboschi:
v2v-vmware is created and managed by HCO operator;
with #832 we started properly reconciling it.
Unfortunately, due to a bad design, v2v-vmware is
also abused by other components that injects other
values there.
A proper design consist in splitting it into two distinct
config-maps. As a quick but dirty fix let's start
reconciling only a few values from HCO.

Fixes: https://bugzilla.redhat.com/1884538

**Release note**:
```release-note
Partially reconcile IMS ConfigMap
```


ovirt-infra:
All tests passed

fdupont-redhat (approved):
LGTM

PR #893, opened 2020-10-21
Split operator and webhook into two different pods

tiraboschi:
Split operator and webhook into two different pods

Currently we are abusing the pod readiness to signal
to OLM that HCO is not ready for an upgrade.
This has a lot of side effects, one of this is the
validating webhook being not able to receive traffic
when exposed by a pod that is not reporting ready=true.
This can cause a lot of side effects if not deadlocks
when the system reach a status where, for any possible
reason, HCO pod cannot be ready and so HCO pod cannot
validate any further update or delete request on HCO CR.
A proper solution is properly use the readiness probe
only to report the pod readiness and communicate
status to OLM via conditions once OLM will be ready for:
https://github.com/operator-framework/enhancements/blob/master/enhancements/operator-conditions.md
in the meanwhile a quick (but dirty!) solution is to
expose the same hco binary on two distinct pods:
the first one will run only the controller and
the second one (almost always ready) just the validating
webhook one.
A deeper code refactor to produce two distinct binaries
is not worth now because we are going to use OLM operator
conditions soon.

Fixes: https://bugzilla.redhat.com/1889401

**Release note**:
```release-note
split operator and webhook into two different pods
```

### DIFF
--- a/cmd/hyperconverged-cluster-operator/main.go
+++ b/cmd/hyperconverged-cluster-operator/main.go
@@ -110,6 +110,19 @@ func main() {
 		os.Exit(1)
 	}
 
+	// TODO: remove this once we will move to OLM operator conditions
+	// Get the webhook mode
+	operatorWebhookMode, err := hcoutil.GetWebhookModeFromEnv()
+	if err != nil {
+		log.Error(err, "Failed to get operator webhook mode from the environment")
+		os.Exit(1)
+	}
+	if operatorWebhookMode {
+		log.Info("operatorWebhookMode: running only the validating webhook")
+	} else {
+		log.Info("operatorWebhookMode: running only the real operator")
+	}
+
 	if runInLocal {
 		depOperatorNs = operatorNsEnv
 	}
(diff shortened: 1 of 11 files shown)


ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

cmd/hyperconverged-cluster-operator/main.go
```diff
@@ -142,8 +155,13 @@ func main() {
 
 	ctx := context.TODO()
 
+	lockName := "hyperconverged-cluster-operator-lock"
+	if operatorWebhookMode {
```
zcahana, line 159: Is leader election actually necessary for webhook mode?

tiraboschi: You are right, it's not: OLM it's going to expose it over a service that can have more than one endpoint.
I'm going to fix it.

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

nunnatsa (approved):
/approve

PR #918, opened 2020-11-05
Signal unsupported mode with a TaintedConfiguration condition

zcahana:
This PR adds a new condition, `TaintedConfiguration: True` to the HCO CR, when it's annotated with a hidden/unofficial configuration annotation.

**Release note**:
```release-note
Added a new condition to signal when HCO runs with an unsupported configuration.
```

### DIFF
--- a/pkg/apis/hco/v1beta1/hyperconverged_types.go
+++ b/pkg/apis/hco/v1beta1/hyperconverged_types.go
@@ -114,9 +114,17 @@ func (vs *Versions) getVersion(name string) (string, bool) {
 	return "", false
 }
 
-// ConditionReconcileComplete communicates the status of the HyperConverged resource's
-// reconcile functionality. Basically, is the Reconcile function running to completion.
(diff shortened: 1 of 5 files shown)


ovirt-infra:

nunnatsa (commented):
This PR is totally do what it should (except for updating the new condition on init). I wander if we need to use the condition map for it. The condition map is a helper for aggregating the operand conditions. But the new condition is not related to the operands conditions. 

Actually, the map is not even used until there are operand CRs - i.e. during the first reconcilation. (init == true).

Please consider to change `detectTaintedConfiguration` to update (add/remove the new condition) the `status.conditions` directly (or to update a condition array function parameter). Then there are few options to use it: calling it at the end of `updateConditions` *and* at the end of `setInitialConditions`; or call it only once after calling to `res, err := r.doReconcile(req)`.

zcahana:
Made some updates:
- Per @nunnatsa's suggestion, we now directly write the conditions to `status.conditions` instead of buffering.
- Moved the JSONPatch annotation constants to a common package, to avoid future cyclic package dependencies.

PR #919, opened 2020-11-05
Distinguish reconciliation requests triggered by the primary resource from ones triggered by others

tiraboschi:
Distinguish reconciliation requests triggered by the primary resource from ones triggered by others

Previous all the reconciliation resources were always
enqueued using the the same HyperConverged object
and this was not allowing us to distinguish reconciliation
requests triggered by a change on HCO CR (the primary resource)
from reconciliation requests triggered the other way around
(a change on one of the CRs created and controlled by HCO).

The watch on the primary resource is enqueuing with
InstrumentedEnqueueRequestForObject, while the watches on secondary
resources enque with EnqueueRequestsFromMapFunc so, in this second case,
we can easily set there a placeholder reconcile.Request just to be able
to distinguish the two flows.

With this, each time we are going to reconcile a CR controlled by HCO,
we can distinguish the case where we are updating it due a change
started from HCO CR from the case where we are reconciling it to
overwrite a change directly introduced (by the user?)
on the controlled resource reverting it to an opinionated state.

Let's also add an additional flag (Overwritten) to EnsureResult considering
it as a subset of Updated to be able to report it back to handlers.

Fixing then the log messages and adding relevant unit tests.

**Release note**:
```release-note
Distinguish reconciliation requests triggered by the primary resource from ones triggered by others
```


ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

tiraboschi:
Ignoring slightly reduced coverage
/override coverage/coveralls

nunnatsa (approved):
/approve

PR #927, opened 2020-11-09
Hide nodePlacement APIs from the UI

tiraboschi:
Hide nodePlacement APIs from the UI

**Release note**:
```release-note
Hide nodePlacement APIs from the UI
```


ovirt-infra:
All tests passed

nunnatsa (commented):
Don't we only want it in release-1.2.0?

tiraboschi:
> Don't we only want it in release-1.2.0?

yes, right: we can do it only there

PR #933, opened 2020-11-10
Validate a CR update with all the operators

tiraboschi:
Before really propagating an update from HCO CR
to all the managed operators,
try to validate it with all the operators in
dry-run mode. If just one of them refuses
the update, refuse as well and report it to the user
with a clear error message.

**Release note**:
```release-note
Validate a CR update with all the operators
```

### DIFF
--- a/pkg/webhooks/webhooks.go
+++ b/pkg/webhooks/webhooks.go
@@ -4,9 +4,12 @@ import (
 	"context"
 	"fmt"
 	"github.com/go-logr/logr"
+	networkaddonsv1 "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/v1"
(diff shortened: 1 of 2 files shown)


ovirt-infra:
All tests passed

ovirt-infra:
All tests passed

pkg/webhooks/webhooks.go
```diff
@@ -74,14 +82,44 @@ func (wh WebhookHandler) updateOperatorCr(ctx context.Context, hc *v1beta1.Hyper
 
 	switch obj := exists.(type) {
 	case *kubevirtv1.KubeVirt:
-		existingKv := obj
+		existing := obj
```
nunnatsa, line 85: Actually, we don't need this variable. But I wouln't hold the PR for that. (same note for all the cases)

nunnatsa (approved):
/approve

PR #949, opened 2020-11-19
Add a mutating webhook to protect hco namespace

tiraboschi:
Add a mutating webhook to protect hco namespace

Introducing a mutating admission webhooks, deployed by
the OLM, bound to the namespace where HCO is running
to protect it from uneducated deletion the can
potentially made it stuck.
The webhook scope is limited to that specific namespace
by the use of an objectSelector.

In this first step the mutating webhook is simply refusing,
with a clear error message,to admit the deletion of the
namespace if the Hyperconverged CR is still there.
A future enhancement can try to delete the Hyperconverged CR
as a side effect (that's why a mutating webhook with side effects)
and, only when it and all its descendants are correctly
and safely deleted, admit the deletion of the namespace.
This requires other ancillary work to make the deletion safe
avoiding race conditions.
So let's simply refuse to delete the namespace as a first
and safer step.

Test it in CI at the end of upgrade test lanes:
the test will try to delete the namespace when the Hyperconverged CR
is still there looking for a specific error message,
the test then will delete the hyperconverged CR and finally
successfully delete the namespace.

Fixes: https://bugzilla.redhat.com/1881874

**Release note**:
```release-note
Add a mutating webhook to protect hco namespace
```

### DIFF
--- a/pkg/controller/hyperconverged/hyperconverged_controller.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller.go
@@ -296,21 +296,22 @@ func (r *ReconcileHyperConverged) doReconcile(req *common.HcoRequest) (reconcile
 	// negative conditions (!Available, Degraded, Progressing)
 	req.Conditions = common.NewHcoConditions()
 
+	fin_dropped := false
 	// Handle finalizers
+	if contains(req.Instance.ObjectMeta.Finalizers, badFinalizerName) {
+		req.Logger.Info("removing a finalizer set in the past (without a fully qualified name)")
+		req.Instance.ObjectMeta.Finalizers, fin_dropped = drop(req.Instance.ObjectMeta.Finalizers, badFinalizerName)
+		req.Dirty = req.Dirty || fin_dropped
+	}
 	if req.Instance.ObjectMeta.DeletionTimestamp.IsZero() {
 		// Add the finalizer if it's not there
 		if !contains(req.Instance.ObjectMeta.Finalizers, FinalizerName) {
+			req.Logger.Info("setting a finalizer (with fully qualified name)")
 			req.Instance.ObjectMeta.Finalizers = append(req.Instance.ObjectMeta.Finalizers, FinalizerName)
-			req.Dirty = true
-		}
-		if contains(req.Instance.ObjectMeta.Finalizers, badFinalizerName) {
-			req.Instance.ObjectMeta.Finalizers = drop(req.Instance.ObjectMeta.Finalizers, badFinalizerName)
-			req.Dirty = true
+			req.Dirty = req.Dirty || fin_dropped
 		}
 	} else {
-		if contains(req.Instance.ObjectMeta.Finalizers, FinalizerName) {
-			return r.ensureHcoDeleted(req)
-		}
+		return r.ensureHcoDeleted(req)
 	}
 
 	// If the current version is not updated in CR ,then we're updating. This is also works when updating from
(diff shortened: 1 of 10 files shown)


pkg/webhooks/webhooks.go
```diff
@@ -181,3 +183,55 @@ func (wh WebhookHandler) ValidateDelete(hc *v1beta1.HyperConverged) error {
...
 }
+
+func (wh WebhookHandler) HandleMutatingNsDelete(ns *corev1.Namespace, dryRun bool) error {
+	wh.logger.Info("validating namespace deletion", "name", ns.Name)
+
+	operatorNsEnv, err := hcoutil.GetOperatorNamespaceFromEnv()
+	if err != nil {
+		wh.logger.Error(err, "failed to get operator namespace from the environment")
+		return err
+	}
```
nunnatsa, line 194: This is not going to changed. Maybe do it when creating the WebhookHandler object.

pkg/webhooks/webhooks.go
```diff
@@ -181,3 +183,55 @@ func (wh WebhookHandler) ValidateDelete(hc *v1beta1.HyperConverged) error {
...
+
+	return &BlockDeletionWithHyperConvergedError{}
+}
+
+type BlockDeletionWithHyperConvergedError struct{}
+
+func (m *BlockDeletionWithHyperConvergedError) Error() string {
+	return "HyperConverged CR is still present, please remove it before deleting the containing namespace"
+}
```
nunnatsa, line 237: Why do we need this type? if it just returning a string, we can just do:

```golang

	return BlockDeletionWithHyperConvergedError
}

var BlockDeletionWithHyperConvergedError = errors.New("HyperConverged CR is still present, please remove it before deleting the containing namespace")
```

tiraboschi: done

pkg/webhooks/webhooks.go
tiraboschi: done

nunnatsa (approved):
/approve

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -430,6 +430,7 @@ func (r *ReconcileHyperConverged) ensureHcoDeleted(req *common.HcoRequest) (reco
 
 	// Remove the finalizer
 	req.Instance.ObjectMeta.Finalizers = drop(req.Instance.ObjectMeta.Finalizers, FinalizerName)
+	req.Instance.ObjectMeta.Finalizers = drop(req.Instance.ObjectMeta.Finalizers, badFinalizerName)
 	req.Dirty = true
```
nunnatsa, line 434: I think we maybe need to requeue only if the finilizer were actually removed. Somethinkg like:
```golang
        if contains(req.Instance.ObjectMeta.Finalizers, FinalizerName) {
	    req.Instance.ObjectMeta.Finalizers = drop(req.Instance.ObjectMeta.Finalizers, FinalizerName)
            req.Dirty = true
        }
	if contains(req.Instance.ObjectMeta.Finalizers, badFinalizerName) {
	    req.Instance.ObjectMeta.Finalizers = drop(req.Instance.ObjectMeta.Finalizers, badFinalizerName)
            req.Dirty = true
        }
```

Or maybe change the `drop` function to also return boolean `dropped`, and use this result instead of checking again as suggested above.

tiraboschi:
Ignoring slightly reduced coverage
/override coverage/coveralls

PR #966, opened 2020-11-26
Switch to use the new SSP operator 

zcahana:
This PR updates HCO operator to use the new [Golang-based SSP operator](https://github.com/kubevirt/ssp-operator), instead of the old [Ansible-based SSP operator](https://github.com/kubevirt/kubevirt-ssp-operator).

~Note 1: This isn't mergeable yet, since the new SSP operator doesn't have a released image yet.~
~Note 2: I plan to add additional unit tests to cover the upgrade logic (removal of old CRDs, etc).~

```release-note
Switch to use the new SSP operator.
```


pkg/controller/operands/ssp.go
```diff
@@ -3,289 +3,125 @@ package operands
 import (
```
zcahana, line 3: Github's diff view of this file is terrible.
In essence, this file has been rewritten from scratch, so it better be viewed as a whole file, and compared to the old file as a whole.

cmd/hyperconverged-cluster-operator/main.go
```diff
@@ -32,6 +31,7 @@ import (
 	apiruntime "k8s.io/apimachinery/pkg/runtime"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	cdiv1beta1 "kubevirt.io/containerized-data-importer/pkg/apis/core/v1beta1"
+	sspv1alpha1 "kubevirt.io/ssp-operator/api/v1alpha1"
```
tiraboschi, line 34: I fear we can accept it only if >= v1beta1

zcahana: What is the criteria to move from v1alpha1 to v1beta1? Or this is merely a syntactical requirement?

nunnatsa (commented):
Starting the review. Currently - two comments.

pkg/controller/operands/ssp.go
```diff
@@ -3,289 +3,125 @@ package operands
...
-			req.Logger.Info("Reconciling an externally updated KubeVirt Node Labeller Bundle's Spec to its opinionated values")
-		}
-		kvNLB.Spec.DeepCopyInto(&found.Spec)
-		err := Client.Update(req.Ctx, found)
-		if err != nil {
-			return false, false, err
-		}
 
-		return true, !req.HCOTriggered, nil
+		crdsToRemove: append([]string{}, oldSSPCrds...),
```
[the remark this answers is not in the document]
zcahana: I too don't really like that global var and the appends. These are basically Go's poorman's constant slice and slice copy.
As you said this isn't much of a big deal, so I'll just inline the CRD names...

pkg/controller/operands/ssp.go
```diff
@@ -3,289 +3,125 @@ package operands
...
+	// Attempt to remove old CRDs
+	if len(handler.crdsToRemove) > 0 && (!req.UpgradeMode || res.Updated) {
+		unremovedCRDs := make([]string, 0, len(handler.crdsToRemove))
 
-	if cr.Spec.Workloads.NodePlacement != nil {
-		if cr.Spec.Workloads.NodePlacement.Affinity != nil {
-			cr.Spec.Workloads.NodePlacement.Affinity.DeepCopyInto(&spec.Affinity)
-		}
+		for _, crdName := range handler.crdsToRemove {
+			removed := removeCrd(handler.Client, req, crdName)
```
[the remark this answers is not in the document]
zcahana: Yes this cries out for concurrency. I think this is simple enough so better be implemented now rather than postponed to sometimes between [unknown, never].

cmd/hyperconverged-cluster-operator/main.go
tiraboschi: The criteria is:

Since it is a "beta" suffixed, it should work with subsequent releases. Indeed, the meaning of api versions is as following:

alpha
API versions with ‘alpha’ in their name are early candidates for new functionality coming into Kubernetes. These may contain bugs and are not guaranteed to work in the future.

beta
‘beta’ in the API version name means that testing has progressed past alpha level, and that the feature will eventually be included in Kubernetes. Although the way it works might change, and the way objects are defined may change completely, the feature itself is highly likely to make it into Kubernetes in some form.

stable
These do not contain ‘alpha’ or ‘beta’ in their name. They are safe to use.

an operator is required to have it's CRD at least to v1beta1 to be considered GA as our previous version was.

pkg/controller/operands/ssp.go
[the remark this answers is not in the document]
zcahana: Done.

pkg/controller/operands/ssp.go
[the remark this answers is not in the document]
zcahana: Done.

pkg/controller/operands/ssp.go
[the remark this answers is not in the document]
zcahana: By the way, the way I implemented this assumes that the controller-runtime's Client is thread-safe. I'm not sure whether this assumption is correct or not, but [asked](https://groups.google.com/u/3/g/kubebuilder/c/CT0wdZykPwc).

zcahana:
Rebased + several updates.

pkg/controller/operands/ssp.go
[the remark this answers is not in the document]
zcahana: Here's the answer I got from one of the controller-runtime devs (for some reason it doesn't appear on the google group thread):

_Yes, the delegating client is safe to use concurrently. The delegating client itself doesn't have any moving parts after its creation, and the underlying clients (cached, and live) should be already safe._

cmd/hyperconverged-cluster-operator/main.go
akrejcir: The API version of `ssp-operator` is now `v1beta1`.

zcahana: @akrejcir thanks, will update this PR to consume `v1beta1`.

zcahana:
/hold

We need to revert efba5d2 before merging this PR - and only after the fix is applied and released in [kubevirt/ssp-operator](https://github.com/kubevirt/ssp-operator).

zcahana:
@tiraboschi per our discussion, should we go ahead an /override the two upgrade tests?

pkg/controller/operands/ssp_test.go
```diff
@@ -70,822 +79,549 @@ var _ = Describe("SSP Operands", func() {
...
-					Reason:  "KubevirtCommonTemplatesBundleProgressing",
-					Message: "KubevirtCommonTemplatesBundle is progressing: Bar",
-				})))
-				Expect(req.Conditions[]).To(ContainElement(testlib.RepresentCondition(conditionsv1.Condition{
-					Type:    conditionsv1.ConditionDegraded,
-					Status:  corev1.ConditionTrue,
-					Reason:  "KubevirtCommonTemplatesBundleDegraded",
-					Message: "KubevirtCommonTemplatesBundle is degraded: Bar",
-				})))
+		Context("NodeLabeller node placement", func() {
```
nunnatsa, line 113: This is a bit misleading to me. This Context and the one in line 248 uses the old names and I find it confusing. But I undersand the logic. Your call...

Also, since workloads and infra are not depeneded in each other, do we reaally need this split in the contexts. Maybe we do it inside the same one?

pkg/controller/operands/ssp_test.go
```diff
@@ -70,822 +79,549 @@ var _ = Describe("SSP Operands", func() {
...
+				Expect(foundResource.Spec.TemplateValidator.Placement.NodeSelector["key1"]).Should(Equal("value1"))
+
+				Expect(req.Conditions).To(BeEmpty())
+			})
+		})
 
-			Expect(existingResource.Spec.Affinity.NodeAffinity).ToNot(BeNil())
-			Expect(existingResource.Spec.Tolerations).To(HaveLen(2))
-			Expect(existingResource.Spec.NodeSelector["key1"]).Should(Equal("value1"))
+		Context("SSP Upgrade", func() {
```
nunnatsa, line 383: Maybe it worth to add a test case(s) where the SSP CR already exist (for future upgrades). Although this is a very unlikely to be relevant.

pkg/controller/operands/ssp_test.go
zcahana: The use of `NodeLabeller` and `TemplateValidator` don't refer to the old CRDs, but rather to the components (which still exist) and their corresponding fields within `SSP.Spec`.

About merging both under the same context... I was considering this too while implementing this, but thought that it would be good to test for them being and staying independent when the code change. But rethinking this now, I think it'd be better to shrink the test code in ~140 lines vs. future-proofing from unlikely bugs. So, will merge the contexts.

zcahana: Done.

pkg/controller/operands/ssp_test.go
zcahana: Not sure what is the test case to be added? Meaning, what logic to we want to test?

nunnatsa: Just make sure that the HyperConverged CR 's node placement related data is copied to the SSP CR

zcahana: This behavior is already covered in various tests under the "Node placement" context.

zcahana:
Waiting for merge of #1011 so that I can rebase on top of that (instead of cherry-picking its commit into this PR).

zcahana:
@nunnatsa no need to rerun the `upgrade-` tests, these are expected to fail here and it was concluded to /override them.

zcahana:
/test hco-e2e-upgrade-prev-azure 
/test hco-e2e-image-index-azure

flakytest.

nunnatsa:
> @nunnatsa no need to rerun the `upgrade-` tests, these are expected to fail here and it was concluded to /override them.

Right. What about prev ? they must pass. But the same is for all azure test that currently not working

zcahana:
> Right. What about prev ? they must pass, I guess. But the same is for all azure test that currently not working

Yes, prev must pass. All failures seems to be due to CI flakyness.

nunnatsa:
test hco-e2e-upgrade-aws
/test hco-e2e-upgrade-azure

zcahana:
@tiraboschi / @nunnatsa can we already `/override` the upgrade tests, so that we don't accidentally `/retest` their failures?

nunnatsa:
> /test hco-e2e-upgrade-aws

This will fail for sure. It can't pas until we'll update the image in quay.io after merging this one.

hack/check_old_ssp_removed.sh
```diff
@@ -0,0 +1,83 @@
...
+)
+
+OLD_SSP_KINDS=(
+	"KubevirtTemplateValidator"
+	"KubevirtCommonTemplatesBundle"
+	"KubevirtNodeLabellerBundle"
+	"KubevirtMetricsAggregation"
+)
+
+#echo "Check that all old SSP CRDs were removed"
```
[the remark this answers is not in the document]
zcahana: This block shouldn't be commented - just a remainder from my local testing of it.
Thanks for spotting this!

pkg/controller/commonTestUtils/testUtils.go
```diff
@@ -98,6 +99,38 @@ func NewHyperConvergedConfig() *sdkapi.NodePlacement {
 	}
 }
 
+func NewHyperConvergedConfig2() *sdkapi.NodePlacement {
```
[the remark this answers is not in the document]
zcahana: The dup's purpose is to create `NodePlacement` instance which is different than the one created by the already existing `NewHyperConvergedConfig()` function. The naming is not great, I agree. I wanted to (re)name them `NewNodePlacement()` and `NewOtherNodePlacement()`, but then preferred to avoid touching lots of test files that are already referencing `NewHyperConvergedConfig()` (as this PR is already quite too large). I could use `NewOtherHyperConvergedConfig()`, which is also not great since this isn't creating a HyperConvergedConfig, but at least is still consistent with the original function.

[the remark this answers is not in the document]
zcahana: Eventually renamed both to `NewNodePlacement()` and `NewOtherNodePlacement()` - took the toll of touching 7-8 additional "unrelated" files.

hack/check_old_ssp_removed.sh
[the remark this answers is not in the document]
zcahana: Fixed.

nunnatsa (approved):
/approve

nunnatsa:
/override ci/prow/hco-e2e-upgrade-azure

Failure is expected

/override ci/prow/hco-e2e-image-index-gcp
hco-e2e-image-index-azure hco-e2e-image-index-aws passed

nunnatsa (approved):
/approve

PR #969, opened 2020-11-26
Provide default values for ceph RBD.

maya-r:
**Release note**:
```release-note
Have the storage suggestion configmap suggest Block/RWX for ceph RBD
```

### DIFF
--- a/pkg/controller/operands/cdi.go
+++ b/pkg/controller/operands/cdi.go
@@ -293,6 +293,8 @@ func NewKubeVirtStorageConfigForCR(cr *hcov1beta1.HyperConverged, namespace stri
 		localSC = *(&cr.Spec.LocalStorageClassName)
 	}
 
+	ocsRBD := "ocs-storagecluster-ceph-rbd"
+
 	labels := map[string]string{
 		hcoutil.AppLabel: cr.Name,
 	}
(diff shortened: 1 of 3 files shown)


tiraboschi (changes requested):
Can you please also add a unit test for this ensuring that HCO is properly covering the upgrade scenario (the ConfigMap was already there from a previous version and HCO should reconcile it adding the missing values)?

I tend to think that HCO is not going to reconcile that object.

maya-r:
Not sure how to run the upgrade tests.
I'm surprised they don't invoke the regular functional tests. Should I change them to do this?

tiraboschi:
> Not sure how to run the upgrade tests.
> I'm surprised they don't invoke the regular functional tests. Should I change them to do this?

I was simply asking for a unit test to ensure that HCO is really going to add the two new keys to that config map if the config map has been created in the past.
I added it in https://github.com/kubevirt/hyperconverged-cluster-operator/pull/969/commits/dea37daa9f0f37dc85c71c22ba741995dfcfbeb9 , please take a look.

The issue is that, as I was suspecting, HCO is not going to add the two new keys because HCO is currently not reconciling that object but just the CR for cdi-operator.

tiraboschi (approved):

PR #974, opened 2020-11-29
Use the new webhook image in deployment/CI/CSV

nunnatsa:

**Release note**:

```release-note
NONE
```


deploy/operator.yaml
```diff
@@ -112,7 +112,7 @@ spec:
         - name: APP
```
zcahana, line 112: Is the `WEBHOOK_MODE` env var still needed?

zcahana (commented):
Just a tiny comment, otherwise lgtm.

nunnatsa:
> Just a tiny comment, otherwise lgtm.

Yes, because of the upgrade tests, that uses the same csv, but witb previous image.

zcahana:
> Yes, because of the upgrade tests, that uses the same csv, but witb previous image.

ok then,
/lgtm

PR #976, opened 2020-11-30
Use OPERATOR_NAME/OPERATOR_NAMESPACE envvars in build-manifests.sh

zcahana:
Modify `build-manifests.sh` to use the `OPERATOR_NAME` and `OPERATOR_NAMESPACE` environment variables to override these values, instead of `NAME` and `NAMESPACE`. (`NAME` is used in WSL to point at the Windows host, so causing errors when running this script).

```release-note
NONE
```


orenc1 (approved):
/approve

PR #992, opened 2020-12-08
Handling OVS opt-in annotation in HCO CR

orenc1:
HCO will reconcile CNAO CR spec stanza with 'ovs:{}' field, if and only if HCO CR has an annotation of `deployOVS=true`, or if it was upgraded from previous version.
`alm-examples` now contains HCO annotation of `deployOVS=false` by default

```release-note
Handling OVS opt-in annotation in HCO CR
```


pkg/controller/operands/networkAddons_test.go
```diff
@@ -362,6 +362,81 @@ var _ = Describe("CNA Operand", func() {
...
+			hcoOVSConfig := commonTestUtils.NewHco()
+			hcoOVSConfig.Annotations = map[string]string{}
+			hcoOVSConfig.Annotations["deployOVS"] = "false"
+
+			existingResource := NewNetworkAddons(hcoOVSConfig)
+
+			cl := commonTestUtils.InitClient([]runtime.Object{hco, existingResource})
+			handler := (*genericOperand)(newCnaHandler(cl, commonTestUtils.GetScheme()))
+			res := handler.ensure(req)
+			Expect(res.UpgradeDone).To(BeFalse())
```
RamLavi, line 401: why does the first test have `Expect(res.Updated).To(BeTrue())` and the other 2 don't? is it intentional?

pkg/controller/operands/networkAddons_test.go
```diff
@@ -362,6 +362,81 @@ var _ = Describe("CNA Operand", func() {
 			Expect(req.Conditions).To(BeEmpty())
 		})
 
+		It("should have ovs if deployOVS annotation is set to true", func() {
```
RamLavi, line 365: you can move these 3 tests to 1 test table and save some code:
```
import "github.com/onsi/ginkgo/extensions/table"

		type ovsAnnotationParams struct {
			annotationExists    bool
			annotationValue     string
			ovsDeployExpected   bool
		}
		DescribeTable("when reconciling ovs-cni", func( o ovsAnnotationParams) {
			hcoOVSConfig := commonTestUtils.NewHco()
			hcoOVSConfig.Annotations = map[string]string{}

			if o.annotationExists {
				hcoOVSConfig.Annotations["deployOVS"] = o.annotationValue
			}

			existingResource := NewNetworkAddons(hcoOVSConfig)

			cl := commonTestUtils.InitClient([]runtime.Object{hco, existingResource})
			handler := (*genericOperand)(newCnaHandler(cl, commonTestUtils.GetScheme()))
			res := handler.ensure(req)
			Expect(res.UpgradeDone).To(BeFalse())
			Expect(res.Updated).To(BeTrue())
			Expect(res.Err).To(BeNil())

			foundResource := &networkaddonsv1.NetworkAddonsConfig{}
			Expect(
				cl.Get(context.TODO(),
					types.NamespacedName{Name: existingResource.Name, Namespace: existingResource.Namespace},
					foundResource),
			).To(BeNil())

			if o.ovsDeployExpected {
				Expect(existingResource.Spec.Ovs).ToNot(BeNil(), "Ovs spec should be added")
			} else {
				Expect(existingResource.Spec.Ovs).To(BeNil(), "Ovs spec should not be added")
			}
			Expect(req.Conditions).To(BeEmpty())

		},
			Entry("should have ovs if deployOVS annotation is set to true", ovsAnnotationParams{
				annotationExists:  true,
				annotationValue:   "true",
				ovsDeployExpected: true,
			}),
			Entry("should not have ovs if deployOVS annotation is not set to true", ovsAnnotationParams{
				annotationExists:  true,
				annotationValue:   "false",
				ovsDeployExpected: false,
			}),
			Entry("should not have ovs if deployOVS annotation does not exist", ovsAnnotationParams{
				annotationExists:  false,
				annotationValue:   "",
				ovsDeployExpected: false,
			})
		)
```

RamLavi (changes requested):
added comments

deploy/index-image/kubevirt-hyperconverged/1.3.0/kubevirt-hyperconverged-operator.v1.3.0.clusterserviceversion.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: operators.coreos.com/v1alpha1
...
...spec":{"templateValidatorReplicas":2}},{"apiVersion":"ssp.kubevirt.io/v1","kind":"KubevirtCommonTemplatesBundle","metadata":{"name":"kubevirt-common-template-bundle","namespace":"kubevirt"}},{"apiVersion":"ssp.kubevirt.io/v1","kind":"KubevirtMetricsAggregation","metadata":{"name":"kubevirt-metrics-aggregation","namespace":"kubevirt"}},{"apiVersion":"ssp.kubevirt.io/v1","kind":"KubevirtNodeLabellerBundle","metadata":{"name":"kubevirt-node-labeller-bundle","namespace":"kubevirt"}},{"apiVersion":"cdi.kubevirt.io/v1beta1","kind":"CDI","metadata":{"name":"cdi","namespace":"cdi"},"spec":{"imagePullPolicy":"IfNotPresent"}},{"apiVersion":"nodemaintenance.kubevirt.io/v1beta1","kind":"NodeMaintenance","metadata":{"name":"nodemaintenance-example"},"spec":{"nodeName":"node02","reason":"Test node maintenance"}},{"apiVersion":"hostpathprovisioner.kubevirt.io/v1beta1","kind":"HostPathProvisioner","metadata":{"name":"hostpath-provisioner"},"spec":{"imagePullPolicy":"IfNotPresent","pathConfig":{"path":"/var/hpvolumes","useNamingPrefix":false}}},{"apiVersion":"v2v.kubevirt.io/v1beta1","kind":"VMImportConfig","metadata":{"name":"vm-import-operator-config"},"spec":{"imagePullPolicy":"IfNotPresent"}}]'
```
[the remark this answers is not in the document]
orenc1: this is generated change.
the actual code change is in `components.go`. i'll split the commits.

pkg/controller/operands/networkAddons.go
```diff
@@ -61,12 +61,24 @@ func (h *cnaHooks) updateCr(req *common.HcoRequest, Client client.Client, exists
 		return false, false, errors.New("can't convert to CNA")
 	}
 
+	// If deployOVS annotation doesn't exists prior the upgrade - set this annotation to true;
+	// Otherwise - remain the value as it is.
+	if req.UpgradeMode {
```
dankenigsberg, line 66: let me know if the code already does this, but we should *not* do this on each and every upgrade. We should add the annotation only if we upgrade from an "old" cnao, e.g. < v0.43.0.

pkg/controller/operands/networkAddons.go
```diff
@@ -61,12 +61,24 @@ func (h *cnaHooks) updateCr(req *common.HcoRequest, Client client.Client, exists
...
+		_, exists := req.Instance.Annotations["deployOVS"]
+		if !exists {
+			req.Instance.Annotations = map[string]string{}
```
nunnatsa, line 69: The current code removed all the existing annotations in case the `deployOVS` annotation is missing. Is this on purpose?

Also, please set `req.Dirty = true`. That will make sure the change is written back to the cluster.

nunnatsa (changes requested):

nunnatsa:
I think this PR depends in #993; @orenc1, please check that.

pkg/controller/operands/networkAddons.go
RamLavi: This code adds the annotation if `UpgradeMode==true` AND the annotation is not already there.
This is agnostic of the version, but notice that after the first upgrade, the annotation will be there so this code will not change the annotation.

pkg/controller/operands/networkAddons.go
orenc1: Thanks, fixed.

pkg/controller/operands/networkAddons.go
dankenigsberg: @RamLavi assume someone has a fresh install of cnao v0.43.0 (without ovs). Nothing adds any annotation to HCO.
What would happen when they upgrade to cnao v0.44.0 in 3 months?

orenc1:
> I think this PR depends in #993; @orenc1, please check that.

It doesn't depend on it. The functionality of deploying / undeploying OVS daemonset according to existence / in-existence of `ovs: {}` field in CNAO CR's spec, is already implemented in the current CNAO version.

tests/go.mod
```diff
@@ -4,25 +4,19 @@ go 1.15
...
 	github.com/ghodss/yaml v1.0.1-0.20190212211648-25d852aebe32
 	github.com/go-openapi/spec v0.19.7 // indirect
 	github.com/golang/groupcache v0.0.0-20200121045136-8c9f03a8e57e // indirect
 	github.com/golang/mock v1.4.0 // indirect
-	github.com/google/go-github/v32 v32.1.0 // indirect
 	github.com/hashicorp/golang-lru v0.5.3 // indirect
-	github.com/kr/text v0.2.0 // indirect
-	github.com/niemeyer/pretty v0.0.0-20200227124842-a10e7caefd8e // indirect
-	github.com/openshift/api v3.9.1-0.20190924102528-32369d4db2ad+incompatible // indirect
+	github.com/kubevirt/cluster-network-addons-operator v0.43.0
```
nunnatsa, line 17: This is done as part of #993, but #993 didn't pass tests yet.

pkg/controller/operands/networkAddons_test.go
```diff
@@ -362,6 +363,59 @@ var _ = Describe("CNA Operand", func() {
...
+					types.NamespacedName{Name: existingResource.Name, Namespace: existingResource.Namespace},
+					foundResource),
+			).To(BeNil())
+
+			if o.ovsDeployExpected {
+				Expect(existingResource.Spec.Ovs).ToNot(BeNil(), "Ovs spec should be added")
+			} else {
+				Expect(existingResource.Spec.Ovs).To(BeNil(), "Ovs spec should not be added")
+			}
+			//Expect(req.Conditions).To(BeEmpty())
```
orenc1, line 399: @RamLavi @nunnatsa 
I changed these 3 ovs-opt-in tests to table, as advised.
Looks like the conditions are not always empty, and for the negative tests (ovs should not be deployed), I'm getting that HCO conditions are not empty (see below). Is that expected and I can safely delete `Expect(req.Conditions).To(BeEmpty())` ?

```
• Failure [0.001 seconds]
CNA Operand
/home/ocohen/Repositories/fork/hyperconverged-cluster-operator/pkg/controller/operands/networkAddons_test.go:27
  NetworkAddonsConfig
  /home/ocohen/Repositories/fork/hyperconverged-cluster-operator/pkg/controller/operands/networkAddons_test.go:29
    when reconciling ovs-cni
    /home/ocohen/Repositories/fork/hyperconverged-cluster-operator/pkg/controller/operands/networkAddons_test.go:371
      should not have ovs if deployOVS annotation is not set to true [It]
      /home/ocohen/Repositories/fork/hyperconverged-cluster-operator/pkg/controller/operands/networkAddons_test.go:408

      Expected
          <common.HcoConditions | len:3>: {
              "Available": {
                  Type: "Available",
                  Status: "False",
                  Reason: "NetworkAddonsConfigConditions",
                  Message: "NetworkAddonsConfig resource has no conditions",
                  LastHeartbeatTime: {
                      Time: 0001-01-01T00:00:00Z,
                  },
                  LastTransitionTime: {
                      Time: 0001-01-01T00:00:00Z,
                  },
              },
              "Progressing": {
                  Type: "Progressing",
                  Status: "True",
                  Reason: "NetworkAddonsConfigConditions",
                  Message: "NetworkAddonsConfig resource has no conditions",
                  LastHeartbeatTime: {
                      Time: 0001-01-01T00:00:00Z,
                  },
                  LastTransitionTime: {
                      Time: 0001-01-01T00:00:00Z,
                  },
              },
              "Upgradeable": {
                  Type: "Upgradeable",
                  Status: "False",
                  Reason: "NetworkAddonsConfigConditions",
                  Message: "NetworkAddonsConfig resource has no conditions",
                  LastHeartbeatTime: {
                      Time: 0001-01-01T00:00:00Z,
                  },
                  LastTransitionTime: {
                      Time: 0001-01-01T00:00:00Z,
                  },
              },
          }
      to be empty

```

pkg/controller/operands/networkAddons_test.go
orenc1: Because the first test case is updating the CR (adding the OVS field), while the two other are not changing it (OVS is not in the CNAO CR).

tests/go.mod
orenc1: Probably my IDE fetched the most recent module (`tests/go.mod` did not contain CNAO module before that). I changed to current version - 0.42.4

pkg/controller/operands/networkAddons.go
RamLavi: CNAO is not used standalone in CNV context, but controlled by HCO. Let's talk in HCO terms.
- when someone installs a fresh install of HCO, he will use the alm-example and gets the `deployOvs:false` annotation that will cause CNAO to not deploy ovs. when he upgrades in 3 months, the ovs will not be deployed, due to this annotation.
- when someone installs a fresh install of HCO with `deployOvs:true`, ovs will be deployed and this will persist to future upgrades.
- (your case) when someone installs a fresh install of HCO without adding the annotation, ovs will not e deployed, but on upgrade it'll be added `deployOvs:true` and ovs will be deployed. Is this a valid and supported scenario though?

hack/upgrade-test.sh
```diff
@@ -254,7 +259,9 @@ ${CMD} get pod $HCO_CATALOGSOURCE_POD -n ${HCO_CATALOG_NAMESPACE} -o yaml | grep
 
 dump_sccs_after
 
-Msg "brutally delete HCO removing the namespace where it's running"
+Msg "Check that OVS is deployed or not deployed according to deployOVS annotation in HCO CR."
```
RamLavi, line 262: I think you forgot to add the call to `hack/check_ovs.sh`

hack/check_ovs.sh
```diff
@@ -0,0 +1,44 @@
...
+  # if the annotation existed in the previous version and was not "true" - OVS should not be deployed.
+    echo "check that OVS does not exist in CNAO CR Spec"
+    [[ $(${CMD} get networkaddonsconfigs cluster -o jsonpath='{.spec.ovs}') == '' ]]
+
+    echo "check that OVS DaemonSet does not exist"
+    [[ $(${CMD} get ds ovs-cni-amd64 --no-headers --ignore-not-found | wc -l) == '0' ]]
+  fi
+fi
+
+echo "OVS Opt-in test completed successfully."
```
RamLavi: please add a new line

tests/go.mod
```diff
@@ -14,7 +14,7 @@ require (
 	github.com/golang/groupcache v0.0.0-20200121045136-8c9f03a8e57e // indirect
 	github.com/golang/mock v1.4.0 // indirect
 	github.com/hashicorp/golang-lru v0.5.3 // indirect
-	github.com/kubevirt/cluster-network-addons-operator v0.43.0
+	github.com/kubevirt/cluster-network-addons-operator v0.42.4
```
RamLavi, line 17: This is only temporary right?

pkg/cmd/oren/main.go
```diff
@@ -0,0 +1 @@
+package oren
```
RamLavi, line 1: can you change the package name `oren` and file path `pkg/cmd/oren/main.go`?

hack/upgrade-test.sh
```diff
@@ -205,6 +205,11 @@ echo "----- Images before upgrade"
 ${CMD} get deployments -n ${HCO_NAMESPACE} -o yaml | grep image | grep -v imagePullPolicy
 ${CMD} get pod $HCO_CATALOGSOURCE_POD -n ${HCO_CATALOG_NAMESPACE} -o yaml | grep image | grep -v imagePullPolicy
 
+
+echo "----- HCO deployOVS annotation before the upgrade"
+PREVIOUS_OVS_ANNOTATION=$(${CMD} get HyperConverged kubevirt-hyperconverged -o jsonpath='{.metadata.annotations.deployOVS}')
```
RamLavi, line 210: question: is this test performed on various versions? meaning, do we check upgrade from to all supported versions?

RamLavi (changes requested):
added initial comments and questions

tests/go.mod
orenc1: yes, #993 hasn't been merged yet.

hack/upgrade-test.sh
orenc1: We have several upgrade test lanes:

- ci/prow/hco-e2e-upgrade[aws|azure]  - tests an upgrade from current version to the code in the current PR.
- ci/prow/hco-e2e-upgrade-prev[aws|azure] - tests an upgrade from the previous minor version (Y-1) directly to the content of the PR.

`PREVIOUS_OVS_ANNOTATION` is the annotation in the cluster, prior to the upgrade.

pkg/controller/operands/networkAddons.go
dankenigsberg: > he will use the alm-example and gets the `deployOvs:false` annotation

This is the bit that I was missing.

> (your case) when someone installs a fresh install of HCO without adding the annotation, ovs will not e deployed, but on upgrade it'll be added `deployOvs:true` and ovs will be deployed. Is this a valid and supported scenario though?

Yes, it is supported and [documented](https://docs.openshift.com/container-platform/4.6/virt/install/installing-virt-cli.html#virt-subscribing-cli_installing-virt-cli). We must not assume that we are installed via GUI.

orenc1: I updated the PR.
On upgrade, I'm checking whether OVS exists on CNAO CR if the `deployOVS` annotation does not exist.

- In case ovs exists - the annotation is being set to true (and therefore CNAO CR will be reconciled with `ovs: {}`).
- In case ovs doesn't exist - the annotation is being set to false (and therefore CNAO CR will be reconciled _without_ the ovs field).

That way we preserve the existence / in-existence of OVS between upgrades.

RamLavi: reviewing the HLD, I'm afraid I wasn't accurate in my last case. let me clear it up:
- (your case) when someone installs a fresh install of HCO without adding the annotation, ovs will not be deployed, **and the `deployOvs:false` annotation will be added** . When upgrading, the annotation will keep the ovs from being further deployed.

@dankenigsberg sorry for the mistake.
@orenc1 I don't think we actually need to check the CNAO-CR. in previous versions, the `deployOVS` annotation does not exist, and ovs is always deployed. on the new version, the annotation is always added (as mentioned before - if not exist, will be reconciled to add `deployOvs:false` as this is the default behavior unless we're on upgrade mode)

orenc1: done.

pkg/controller/operands/networkAddons_test.go
nunnatsa: It's not so important. We just need to make sure that upgrade is working with and without the annotation

pkg/controller/operands/networkAddons.go
```diff
@@ -61,12 +61,31 @@ func (h *cnaHooks) updateCr(req *common.HcoRequest, Client client.Client, exists
 		return false, false, errors.New("can't convert to CNA")
 	}
 
+	// If deployOVS annotation doesn't exists:
+	// * if on upgrade - set it to true
+	// * if not on upgrade (fresh install / manually removing the annotation) - set it to false.
+	_, annotation_exists := req.Instance.Annotations["deployOVS"]
```
RamLavi, line 67: I think that if `req.Instance.Annotations` is nil then this line will crash, better rearrange it like so:
```
	// If deployOVS annotation doesn't exists:
	// * if on upgrade - set it to true
	// * if not on upgrade (fresh install / manually removing the annotation) - set it to false.
	annotationExists := false
	if req.Instance.GetAnnotations() == nil {
		req.Instance.Annotations = map[string]string{}
	} else {
		_, annotationExists = req.Instance.GetAnnotations()["deployOVS"]
	}
	if !annotationExists {
		if req.UpgradeMode {
			req.Instance.Annotations["deployOVS"] = "true"
			req.Logger.Info("deployOVS annotation did not exist in previous version and set to true in new version.")
		} else {
			req.Instance.Annotations["deployOVS"] = "false"
			req.Logger.Info("deployOVS annotation does not exist and it is being set to false.")
		}
		req.Dirty = true
	}
```

RamLavi: and perhaps we should move it to a separate function `updateDeployOvsAnnotation` or something (I am bad with namings)

pkg/controller/operands/networkAddons_test.go
```diff
@@ -362,6 +363,59 @@ var _ = Describe("CNA Operand", func() {
...
+
+			if o.annotationExists {
+				hcoOVSConfig.Annotations["deployOVS"] = o.annotationValue
+			}
+
+			existingResource := NewNetworkAddons(hcoOVSConfig)
+
+			cl := commonTestUtils.InitClient([]runtime.Object{hco, existingResource})
+			handler := (*genericOperand)(newCnaHandler(cl, commonTestUtils.GetScheme()))
+			res := handler.ensure(req)
```
RamLavi, line 383: is this the normal flow? `NewNetworkAddons()` and only then `handler.ensure(req)` is called?
IIUC, `handler.ensure()` is calling the `updateCr()` function, which updates the annotation if it does not exist.
If that is so, then should the annotation not exist - it'll get added too late when `updateCr()` is called. if this is true, we may need to find a more suitable place the update the annotation if it does not exist.

tests/func-tests/virtual_machines_test.go
```diff
@@ -15,6 +18,9 @@ import (
 	k8smetav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kubevirtv1 "kubevirt.io/client-go/api/v1"
 	"kubevirt.io/client-go/kubecli"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
```
RamLavi, line 22: please remove extra new lines

tests/func-tests/virtual_machines_test.go
```diff
@@ -3,7 +3,10 @@ package tests_test
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/kubevirt/cluster-network-addons-operator/pkg/apis"
```
RamLavi, line 6: the import sections are mixed here. we should have 4 groups:
1. standard golang library imports.
2. 3rd party librarys
3. test libraries
4. (optional) organization libraries (kubevirt)
5. self imports

like so: 
```
import (
	"encoding/json"
	"fmt"
	"time"

	corev1 "k8s.io/api/core/v1"
	k8smetav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1" (notice a duplication here)
	"k8s.io/client-go/kubernetes/scheme"

	. "github.com/onsi/ginkgo"
	. "github.com/onsi/gomega"
	testscore "kubevirt.io/kubevirt/tests"

	"github.com/kubevirt/cluster-network-addons-operator/pkg/apis"
	networkaddonsv1 "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/v1"
	"kubevirt.io/kubevirt/tests/flags"
	kubevirtv1 "kubevirt.io/client-go/api/v1"
	"kubevirt.io/client-go/kubecli"

	tests "github.com/kubevirt/hyperconverged-cluster-operator/tests/func-tests"
)
```
you may need to fmt this.

RamLavi (changes requested):
added comments. Also, can you split the alm-example auto-generated to a separate commit as agreed?

pkg/controller/operands/networkAddons.go
orenc1: if `req.Instance.Annotations == nil`, the line doesn't crash, it returns `false` to the second variable.

RamLavi (changes requested):
comments regarding the e2e tests

pkg/controller/operands/networkAddons.go
RamLavi: are you sure? I think that if req.Instance.Annotations==nil then nil["deployOVS"] should fail.

hack/check_ovs.sh
```diff
@@ -0,0 +1,46 @@
...
+    [[ $(${CMD} get HyperConverged kubevirt-hyperconverged -n kubevirt-hyperconverged -o jsonpath='{.metadata.annotations.deployOVS}') == 'true' ]]
+
+    echo "check that OVS exists in CNAO CR Spec"
+    [[ $(${CMD} get networkaddonsconfigs cluster -o jsonpath='{.spec.ovs}') == '{}' ]]
+
+    echo "check that OVS DaemonSet exists"
+    [[ $(${CMD} get ds ovs-cni-amd64 -n kubevirt-hyperconverged --no-headers --ignore-not-found | wc -l) == '1' ]]
+
+  else
+    if [[ -n ${PREVIOUS_OVS_ANNOTATION} ]] && [[ ${PREVIOUS_OVS_ANNOTATION} != 'true' ]]; then
```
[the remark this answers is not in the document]
orenc1: changed to `elif`.

pkg/controller/operands/networkAddons.go
RamLavi: @dankenigsberg FYI we updated the [design](https://docs.google.com/document/d/1-6hdXgaWF3RNZM-6dw5mPn4r78SmtTiTnfffldHr-1Q/edit)

tests/go.mod
RamLavi: #993 is merged now, reminder to remove it :)

hack/check_upgrade_ovs.sh
```diff
@@ -0,0 +1,65 @@
...
+    echo "check that OVS annotation in HCO CR in new version is set to true"
+    [[ $(${CMD} get HyperConverged kubevirt-hyperconverged -n kubevirt-hyperconverged -o jsonpath='{.metadata.annotations.deployOVS}') == 'true' ]]
+
+    echo "check that OVS exists in CNAO CR Spec"
+    [[ $(${CMD} get networkaddonsconfigs cluster -o jsonpath='{.spec.ovs}') == '{}' ]]
+
+    echo "check that OVS DaemonSet exists"
+    [[ $(${CMD} get ds ovs-cni-amd64 -n kubevirt-hyperconverged --no-headers --ignore-not-found | wc -l) == '1' ]]
+
+  elif [[ -n ${PREVIOUS_OVS_ANNOTATION} ]] && [[ ${PREVIOUS_OVS_ANNOTATION} != 'true' ]]; then
```
RamLavi, line 35: continuing the discussion from [here](https://github.com/kubevirt/hyperconverged-cluster-operator/pull/992#discussion_r542387935), I meant you can remove this entire condition since it'll always happen.
```suggestion
  elif
```

hack/check_upgrade_ovs.sh
```diff
@@ -0,0 +1,65 @@
...
+    # if the annotation existed in the previous version and was not "true" - OVS should not be deployed.
+      echo "check that OVS does not exist in CNAO CR Spec"
+      [[ $(${CMD} get networkaddonsconfigs cluster -o jsonpath='{.spec.ovs}') == '' ]]
+
+      echo "check that OVS DaemonSet does not exist"
+      [[ $(${CMD} get ds ovs-cni-amd64 -n kubevirt-hyperconverged --no-headers --ignore-not-found | wc -l) == '0' ]]
+  fi
+
+elif [[ ${RELEASE_DELTA} == 1 ]]; then
```
RamLavi, line 44: I don't understand why there is a difference if we are upgrading from previous minor version or from same version, can you elaborate?

RamLavi (changes requested):
added comments

hack/check_upgrade_ovs.sh
orenc1: if we are upgrading from previous minor version, the annotation does not exist in the previous version, and upon upgrade to the current version (CNV 2.6), we'll have to deploy ovs and add the true annotation.
if we are upgrading from the same version, we already have this opt-in machanism, thus: annotation not exists == no OVS is deployed.
that's why we have a distinction between the two cases.

hack/check_upgrade_ovs.sh
orenc1: you mean just `else`?
you're right that the two conditions are complimentary.

orenc1:
- **hco-e2e-upgrade-aws**: Step 7/13: Verify the hyperconverged-cluster deployment is using the new image  - image not contain `ci-op-s1mmhvvc/stable`
- **hco-e2e-upgrade-azure**: cluster bootstrap error.
- **hco-e2e-upgrade-prev-aws**: step 13/13:
```
DELETE_ERROR_TEXT='Error from server (InternalError): Internal error occurred: failed calling webhook "mutate-ns-hco.kubevirt.io": Post "https://hco-webhook-service.kubevirt-hyperconverged.svc:4343/mutate-ns-hco-kubevirt-io?timeout=30s": x509: certificate signed by unknown authority (possibly because of "x509: ECDSA verification failure" while trying to verify candidate authority certificate "Red Hat, Inc.")'
+ echo 'Error from server (InternalError): Internal error occurred: failed calling webhook "mutate-ns-hco.kubevirt.io": Post "https://hco-webhook-service.kubevirt-hyperconverged.svc:4343/mutate-ns-hco-kubevirt-io?timeout=30s": x509: certificate signed by unknown authority (possibly because of "x509: ECDSA verification failure" while trying to verify candidate authority certificate "Red Hat, Inc.")'
+ grep 'denied the request: HyperConverged CR is still present, please remove it before deleting the containing namespace'
```
- **hco-e2e-upgrade-prev-azure**: cluster bootstrap error.

/retest

hack/check_upgrade_ovs.sh
RamLavi: this logic is different from what HCO actually does, which is to check if ovs was previously deployed (in absence of the annotation during upgrade) and add the annotation accordingly.
The distinction between the types of RELEASE_DELTA is not a close fit to what we do in the code logic, and might break. it might also do a disservice as documentation of what HCO actually does.

I suggest checking if OVS is deployed in CNAO CR before the upgrade (similar to how we pre-save the PREVIOUS_OVS_ANNOTATION) and act upon it here

pkg/controller/operands/networkAddons_test.go
RamLavi: After @orenc1 explained that indeed the flow where the annotation will be added in a certain reconcile and handled in the next one is by design, I'm ok with this.

RamLavi:
/lgtm
Thanks Oren!

tiraboschi (approved):

PR #994, opened 2020-12-09
Add recommended labels into operands&OLM workloads

erkanerol:
With this PR, we are adding recommended labels[1] into all operands we manage with hco-operator.

````
app.kubernetes.io/managed-by=hco-operator
app.kubernetes.io/version=<version-of-hco-operator>
app.kubernetes.io/component=<component-of-operand-like-compute-or-storage>
app.kubernetes.io/part-of=hyperconverged-cluster
````

We are keeping this as it is in order to not break any existing mechanism.
```
app=kubevirt-hyperconverged
```

***OLM PART***
This PR adds the labels into the workloads we deploy with OLM as well.  See the changes in
- deploy/index-image/kubevirt-hyperconverged/1.4.0/kubevirt-hyperconverged-operator.v1.4.0.clusterserviceversion.yaml
- deploy/olm-catalog/kubevirt-hyperconverged/1.4.0/kubevirt-hyperconverged-operator.v1.4.0.clusterserviceversion.yaml

***operator.yaml***
This PR adds the labels into the workloads we deploy with operator.yaml. Since we don't know who is going to deploy with this file, we didn't use "managed-by" label in this manifest file. 

***Testing***

- Unit tests in hyperconverged_controller_test.go and e2e tests in check_labels.sh: checks all objects under .status.relatedObjects of HCO and verifies that all of them have these labels.

**Release note**:
```release-note
Adds some of the recommended labels* into workloads

* https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/ 
```

1. https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

### DIFF
--- a/pkg/apis/hco/v1alpha1/hyperconverged_resources.go
+++ b/pkg/apis/hco/v1alpha1/hyperconverged_resources.go
@@ -19,7 +19,7 @@ func (r *HyperConverged) getNamespace(defaultNamespace string, opts []string) st
 	return defaultNamespace
 }
 
-func (r *HyperConverged) getLabels() map[string]string {
+func (r *HyperConverged) getLabels(component hcoutil.AppComponent) map[string]string {
 	hcoName := hcoutil.HyperConvergedName
 
 	if r.Name != "" {
@@ -27,7 +27,11 @@ func (r *HyperConverged) getLabels() map[string]string {
 	}
 
 	return map[string]string{
-		hcoutil.AppLabel: hcoName,
+		hcoutil.AppLabel:          hcoName,
+		hcoutil.AppLabelManagedBy: hcoutil.OperatorName,
+		hcoutil.AppLabelVersion:   hcoutil.GetHcoKvIoVersion(),
+		hcoutil.AppLabelPartOf:    hcoutil.HyperConvergedCluster,
+		hcoutil.AppLabelComponent: string(component),
 	}
 }
 
@@ -35,7 +39,7 @@ func (r *HyperConverged) NewKubeVirt(opts ...string) *kubevirtv1.KubeVirt {
 	return &kubevirtv1.KubeVirt{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "kubevirt-" + r.Name,
-			Labels:    r.getLabels(),
+			Labels:    r.getLabels(hcoutil.AppComponentCompute),
 			Namespace: r.getNamespace(r.Namespace, opts),
 		},
 		Spec: kubevirtv1.KubeVirtSpec{
@@ -49,7 +53,7 @@ func (r *HyperConverged) NewCDI(opts ...string) *cdiv1alpha1.CDI {
 	return &cdiv1alpha1.CDI{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "cdi-" + r.Name,
-			Labels:    r.getLabels(),
+			Labels:    r.getLabels(hcoutil.AppComponentStorage),
 			Namespace: r.getNamespace(hcoutil.UndefinedNamespace, opts),
 		},
 		Spec: cdiv1alpha1.CDISpec{
@@ -62,7 +66,7 @@ func (r *HyperConverged) NewNetworkAddons(opts ...string) *networkaddonsv1alpha1
 	return &networkaddonsv1alpha1.NetworkAddonsConfig{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      networkaddonsnames.OPERATOR_CONFIG,
-			Labels:    r.getLabels(),
+			Labels:    r.getLabels(hcoutil.AppComponentNetwork),
 			Namespace: r.getNamespace(hcoutil.UndefinedNamespace, opts),
 		},
 		Spec: networkaddonsshared.NetworkAddonsConfigSpec{
@@ -79,7 +83,7 @@ func (r *HyperConverged) NewKubeVirtCommonTemplateBundle(opts ...string) *sspv1.
 	return &sspv1.KubevirtCommonTemplatesBundle{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "common-templates-" + r.Name,
-			Labels:    r.getLabels(),
+			Labels:    r.getLabels(hcoutil.AppComponentSchedule),
 			Namespace: r.getNamespace(hcoutil.OpenshiftNamespace, opts),
 		},
 	}
@@ -93,7 +97,7 @@ func (r *HyperConverged) NewKubeVirtPriorityClass() *schedulingv1.PriorityClass 
 		},
 		ObjectMeta: metav1.ObjectMeta{
 			Name:   "kubevirt-cluster-critical",
-			Labels: r.getLabels(),
+			Labels: r.getLabels(hcoutil.AppComponentSchedule),
 		},
 		// 1 billion is the highest value we can set
 		// https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/#priorityclass
--- a/pkg/components/components.go
+++ b/pkg/components/components.go
@@ -547,6 +547,7 @@ func GetClusterPermissions() []rbacv1.PolicyRule {
 				"watch",
 				"create",
 				"delete",
+				"update",
 			},
 		},
 		{
--- a/pkg/components/components.go
+++ b/pkg/components/components.go
@@ -57,7 +57,7 @@ func GetDeploymentOperator(namespace, image, imagePullPolicy, conversionContaine
 	}
 }
 
-func GetDeploymentWebhook(namespace, image, imagePullPolicy string, env []corev1.EnvVar) appsv1.Deployment {
+func GetDeploymentWebhook(namespace, image, imagePullPolicy, hcoKvIoVersion string, env []corev1.EnvVar) appsv1.Deployment {
 	deploy := appsv1.Deployment{
 		TypeMeta: metav1.TypeMeta{
 			APIVersion: "apps/v1",
@@ -69,7 +69,7 @@ func GetDeploymentWebhook(namespace, image, imagePullPolicy string, env []corev1
 				"name": hcoNameWebhook,
 			},
 		},
-		Spec: GetDeploymentSpecWebhook(namespace, image, imagePullPolicy, env),
+		Spec: GetDeploymentSpecWebhook(namespace, image, imagePullPolicy, hcoKvIoVersion, env),
 	}
 
 	InjectVolumesForWebHookCerts(&deploy)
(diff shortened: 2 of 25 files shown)


nunnatsa (commented):
Not sure I like the long `if` conditions in the `updateCr` methods. 

Each `if` checks different field in, and then, within the if block, but there is no relation between the field.

so instead:

```golang
if !reflect.DeepEqual(require.A, exists.A) || !reflect.DeepEqual(require.B, exists.B) {
    handle A
    handle B
}
```

I think it will be clrearer to do:
```golang
if !reflect.DeepEqual(require.A, exists.A) {
    handle A
}

if !reflect.DeepEqual(require.B, exists.B) {
    handle B
}
```

pkg/components/components.go
```diff
@@ -116,7 +116,10 @@ func GetDeploymentSpecOperator(namespace, image, imagePullPolicy, conversionCont
 		Template: corev1.PodTemplateSpec{
 			ObjectMeta: metav1.ObjectMeta{
 				Labels: map[string]string{
-					"name": hcoName,
+					"name":                    hcoName,
+					hcoutil.AppVersionLabel:   hcoKvIoVersion,
```
zcahana, line 120: Here you add the labels directly to the podSpec - shouldn't this be added to the Deployment's labels too?
Same comment applies both to the operator and webhook deployments.

pkg/util/consts.go
```diff
@@ -38,3 +50,15 @@ const (
...
 )
+
+type AppComponent string
+
+const (
+	ComputeComponent    AppComponent = "compute"
+	StorageComponent                 = "storage"
+	ImportComponent                  = "import"
+	NetworkComponent                 = "network"
+	MonitorComponent                 = "monitor"
```
zcahana, line 61: I think "monitoring" or maybe "metrics" could be a better name than "monitor".

pkg/util/consts.go
```diff
@@ -38,3 +50,15 @@ const (
 	HCONSWebhookPath            = "/mutate-ns-hco-kubevirt-io"
 	WebhookPort                 = 4343
 )
+
+type AppComponent string
+
+const (
+	ComputeComponent    AppComponent = "compute"
+	StorageComponent                 = "storage"
+	ImportComponent                  = "import"
```
zcahana, line 59: Don't know if "import" merits a component name of its own (i.e., if its significant enough for the user?). Maybe this can fit as part of "compute" component? Or maybe "migration" is a better name?

pkg/util/labels.go
```diff
@@ -0,0 +1,12 @@
+package util
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+func DeepCopyLabels(src, tgt *metav1.ObjectMeta) {
+	if src.Labels != nil {
+		tgt.Labels = make(map[string]string, len(src.Labels))
```
zcahana, line 7: It would be safer to "append" to the target labels instead of overwriting - i.e. check if already exists before creating...

pkg/util/labels.go
```diff
@@ -0,0 +1,12 @@
+package util
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+func DeepCopyLabels(src, tgt *metav1.ObjectMeta) {
+	if src.Labels != nil {
```
zcahana, line 6: I think it's nicer to:
```go
if src.Labels == nil {
	return
}
// rest of code, unindented
```

tools/csv-merger/csv-merger.go
```diff
@@ -456,6 +499,20 @@ func main() {
 
 }
 
+func overwriteWithStandardLabels(labels map[string]string, version string, component hcoutil.AppComponent) {
+	appendedLabels := map[string]string{
```
zcahana, line 503: What's the use of the extra map + for loop? (except that it indents a bit more nicely)
Why not just this 4-liner:
```go
labels[hcoutil.AppManagedByLabel] = "olm"
labels[hcoutil.AppVersionLabel] = version
labels[hcoutil.AppPartOfLabel] = hcoutil.HyperConvergedCluster
labels[hcoutil.AppComponentLabel] = string(component)
```

pkg/apis/hco/v1alpha1/hyperconverged_resources.go
```diff
@@ -19,23 +21,27 @@ func (r *HyperConverged) getNamespace(defaultNamespace string, opts []string) st
...
 
 	if r.Name != "" {
 		hcoName = r.Name
 	}
 
 	return map[string]string{
-		hcoutil.AppLabel: hcoName,
+		hcoutil.AppLabel:          hcoName,
+		hcoutil.AppManagedByLabel: hcoutil.OperatorName,
+		hcoutil.AppVersionLabel:   os.Getenv(hcoutil.HcoKvIoVersionName),
```
zcahana, line 34: Also, here you're setting the `HCO_KV_IO_VERSION` version to all operands, although in the CSV you're setting the operand specific version (e.g., CNAO 0.44.0) - need to decide on a consistent approach. If the operand specific version is used, then probably the application name label should reflect that (e.g., `hcoutil.AppLabel: "cnao"`).

pkg/apis/hco/v1alpha1/hyperconverged_resources.go
```diff
@@ -19,23 +21,27 @@ func (r *HyperConverged) getNamespace(defaultNamespace string, opts []string) st
...
+func (r *HyperConverged) getLabels(component hcoutil.AppComponent) map[string]string {
 	hcoName := hcoutil.HyperConvergedName
 
 	if r.Name != "" {
 		hcoName = r.Name
 	}
 
 	return map[string]string{
-		hcoutil.AppLabel: hcoName,
+		hcoutil.AppLabel:          hcoName,
```
zcahana, line 32: What about the `app.kubernetes.io/instance` instance label, should we be adding that as well?

zcahana:
A few more general comments:
1. I think this PR can have a release note.
2. I know you already tested this, but would it be easy to add an automated test that verifies that all `relatedObjects` have these labels?
3. Note regarding reconciliation of labels - if we want to ensure that the labels are always reconciled, even when changed by the user, then we need to modify the [predicate](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/7098156c11a82a8ff96b79621f9daf01b664cce2/pkg/controller/hyperconverged/hyperconverged_controller.go#L114) we're using such that it'll also accept label changes (currently it only admits changes to spec and/or annotations).

pkg/components/components.go
erkanerol: I discussed this with @tiraboschi  and explained that OLM doesn't allow us to set labels for deployments. It takes only DeploymentSpec.

pkg/util/consts.go
```diff
@@ -25,6 +25,18 @@ const (
 	CurrentAPIVersion      = APIVersionBeta
 	APIVersionGroup        = "hco.kubevirt.io"
 	APIVersion             = APIVersionGroup + "/" + APIVersionBeta
+	// Recommended labels by Kubernetes. See
+	// https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/
+	AppLabelPrefix    = "app.kubernetes.io"
```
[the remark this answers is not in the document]
erkanerol: IT makes sense!! I just updated as you recommended.

tools/csv-merger/csv-merger.go
erkanerol: Good catch!

pkg/apis/hco/v1alpha1/hyperconverged_resources.go
erkanerol: We discussed which labels we need to implement and possible key/value pairs. We didn't invest on this "instance" label at the moment.

pkg/util/consts.go
erkanerol: After the discussions in chat, I updated this part as "monitoring". Also changed the names of enums like the ServiceType naming of k8s https://github.com/kubernetes/kubernetes/blob/ddf3eb5a1877338da806c2be15b53f456b6342c9/pkg/apis/core/types.go#L3446

pkg/util/consts.go
erkanerol: It was suggested by @fabiand  at the beginning of the discussions.

pkg/util/labels.go
erkanerol: It sounds better! Just fixed.

pkg/util/labels.go
erkanerol: We are checking the deep equality for labels since we are so opinionated about the labels of our operands. This was a result of a discussion with @tiraboschi

pkg/apis/hco/v1alpha1/hyperconverged_resources.go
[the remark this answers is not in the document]
zcahana: Correcting myself as per @yuvalturg comment, there aren't syscalls involved in reading the Environment variable - these are served by the runtime from memory, and hence the cost of `Getenv()` multiple times is reasonable. Nevertheless this might be somewhat cleaner to read them once, store, and serve from a state variable.

pkg/components/components.go
zcahana: Seems like this capability was recently added in [operator-framework/api#46](https://github.com/operator-framework/api/pull/46) (although [its parent issue](https://github.com/operator-framework/operator-lifecycle-manager/issues/1522) is still open).

pkg/util/labels.go
erkanerol: we talked this with @zcahana  and we are ok with it.

pkg/apis/hco/v1alpha1/hyperconverged_resources.go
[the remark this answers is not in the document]
erkanerol: fixed.

pkg/components/components.go
erkanerol: I added the labels into the StrategyDeploymentSpec objects as well. OLM is not currently using them but my understanding is that it is going to use.

erkanerol:
> A few more general comments:
> 
>     1. I think this PR can have a release note.
> 
>     2. I know you already tested this, but would it be easy to add an automated test that verifies that all `relatedObjects` have these labels?
> 
>     3. Note regarding reconciliation of labels - if we want to ensure that the labels are always reconciled, even when changed by the user, then we need to modify the [predicate](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/7098156c11a82a8ff96b79621f9daf01b664cce2/pkg/controller/hyperconverged/hyperconverged_controller.go#L114) we're using such that it'll also accept label changes (currently it only admits changes to spec and/or annotations).

1. I added a release note.
2. I added both unit tests and e2e tests. 
3. After checking the code base, we noticed that we don't use predicate for secondary resources so there is no need a change.

zcahana:
Looks good @erkanerol!
/lgtm

zcahana:
@erkanerol no need to `/test hco-e2e-image-index-gcp` since `hco-e2e-image-index-aws` and `hco-e2e-image-index-azure` have passed, so our friend `kubevirt-bot` will make sure to `/override hco-e2e-image-index-gcp`.

tiraboschi (approved):

PR #998, opened 2020-12-14
Patch csv-generator to workaround an OLM bug

tiraboschi:
**Release note**:
```release-note
patch csv-generator to workaround an OLM bug
```

### DIFF
--- a/tools/csv-merger/csv-merger.go
+++ b/tools/csv-merger/csv-merger.go
@@ -333,6 +333,13 @@ func main() {
 
 			strategySpec := csvStruct.Spec.InstallStrategy.StrategySpec
 
+			// temporary workaround for https://bugzilla.redhat.com/1907381
+			// a custom .spec.template.annotations["description"] it's causing a failure on OLM
+			// TODO: remove this once fixed on OLM side
+			for _, deployment := range strategySpec.DeploymentSpecs {
+				delete(deployment.Spec.Template.Annotations, "description")
+			}
+
 			installStrategyBase.DeploymentSpecs = append(installStrategyBase.DeploymentSpecs, strategySpec.DeploymentSpecs...)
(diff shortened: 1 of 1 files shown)


tools/csv-merger/csv-merger.go
```diff
@@ -333,6 +333,13 @@ func main() {
 
 			strategySpec := csvStruct.Spec.InstallStrategy.StrategySpec
 
+			// temporary workarounf for RHBZ#....
```
zcahana, line 336: workarounf --> workaround

zcahana, line 336: Is there actually a bugzilla on this? If so, it's worthwhile to link it so a future maintainer will have the context for why it was added (and when can be removed).

tools/csv-merger/csv-merger.go
```diff
@@ -333,6 +333,13 @@ func main() {
...
+			// a custom .spec.template.annotations["description"] it's causing a failure on OLM
+			// TODO: remove this once fixed on OLM side
+			for _, deployment := range strategySpec.DeploymentSpecs {
+				delete(deployment.Spec.Template.Annotations, "description")
```
zcahana, line 340: This handles a very specific annotation ("description").
Should we instead mimic the [OLM logic](https://github.com/operator-framework/operator-lifecycle-manager/blob/4b66803055a8ab611447c33ed86e755ad39cb313/pkg/controller/install/deployment.go#L263-L269) and just mirror the annotations from the CSV?

tools/csv-merger/csv-merger.go
tiraboschi: done

tools/csv-merger/csv-merger.go
tiraboschi: we have more than one deployment in the CSV,
replicating the same long description more than once it's just going to make it unnecessarily long without any real value

zcahana: True, there's no value in replicating the same description over and over.

PR #1001, opened 2020-12-16
Add cert-manager into deploy.sh for webhooks

erkanerol:
When OLM is not available in the cluster, our webhooks don't work because of missing certificates. 

This PR adds cert-manager deployment into deploy.sh and adds self-signed issuer. 

For each webhook, we need to create a Certificate CR and add annotations into webhooks so that cert-manager can inject caBundle into them.

Also note that this PR replaces "kubectl create" commands with "kubectl apply" for idempotency. 

**Release note**:
```release-note
NONE
```


deploy/index-image/kubevirt-hyperconverged/1.3.0/kubevirt-hyperconverged-operator.v1.3.0.clusterserviceversion.yaml
```diff
@@ -1827,7 +1827,20 @@ spec:
                   initialDelaySeconds: 5
                   periodSeconds: 5
                 resources: {}
+                volumeMounts:
```
tiraboschi, line 1830: This should not touch the CSV

erkanerol: Resolved!

hack/deploy.sh
```diff
@@ -130,12 +133,19 @@ then
...
-	"${CMD}" create -f _out/operator.yaml
+	"${CMD}" apply -f _out/operator.yaml
 else
 	sed -E 's|^(\s*)- name: KVM_EMULATION$|\1- name: KVM_EMULATION\n\1  value: "true"|' < _out/operator.yaml > _out/operator-ci.yaml
 	cat _out/operator-ci.yaml
-	"${CMD}" create -f _out/operator-ci.yaml
+	"${CMD}" apply -f _out/operator-ci.yaml
 fi
 
 # Wait for the HCO to be ready
```
tiraboschi, line 151: Maybe we should also start explicitly checking the status of other webhooks here

erkanerol: Hi @tiraboschi . Thanks for checking the PR. Could you please elaborate these two words in your comment?

1) "status" -> did you mean the status of deployments or status of validatingwebhookconfiguration CRs ?
2) "other" -> other than validating? other than hco? other than hco + node-maintenance?

tiraboschi: yes, sorry.
As on line 144/148 but for hco webhooks and node-maintenance deployments

erkanerol: Check for node-maintenance-operator is available in the for loop here https://github.com/kubevirt/hyperconverged-cluster-operator/pull/1001/commits/74a9f3a6e33f643a2926bdf09977d256bb92c7ba#diff-012508938ec4e97ecf905766cea58011d7b8a1da6b7f82d362ffa10220b33184L158

I just added a wait for hco-webhook as well.

nunnatsa (approved):
/approve

PR #1015, opened 2020-12-21
test: force RollingUpgrade on all the deployments

tiraboschi:
test only

**Release note**:

```release-note
NONE
```

### DIFF
--- a/deploy/index-image/kubevirt-hyperconverged/1.3.0/kubevirt-hyperconverged-operator.v1.3.0.clusterserviceversion.yaml
+++ b/deploy/index-image/kubevirt-hyperconverged/1.3.0/kubevirt-hyperconverged-operator.v1.3.0.clusterserviceversion.yaml
@@ -1835,7 +1835,7 @@ spec:
             matchLabels:
               name: cluster-network-addons-operator
           strategy:
-            type: Recreate
+            type: RollingUpdate
           template:
             metadata:
               labels:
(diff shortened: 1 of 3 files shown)


zcahana:
@tiraboschi These are pretty good results for the first round of tests - the two `e2e-upgrade-` failures don't seem related to the CNAO issue we've seen. Should we /retest everything to see if this continues to pass?

tiraboschi:
yes,
let's run again the failed ones
/retest

tiraboschi:
still
```
Command: oc get clusterserviceversions -n kubevirt-hyperconverged
NAME                                        DISPLAY                                    VERSION   REPLACES                                  PHASE
kubevirt-hyperconverged-operator.v1.3.0     KubeVirt HyperConverged Cluster Operator   1.3.0     kubevirt-hyperconverged-operator.v1.2.0   Pending
kubevirt-hyperconverged-operator.v100.0.0   KubeVirt HyperConverged Cluster Operator   100.0.0   kubevirt-hyperconverged-operator.v1.3.0   Pending
```
probably because this is patching only the CSV we want to upgrade to, but the issue is with the CSV we start from.

nunnatsa:
> probably because this is patching only the CSV we want to upgrade to, but the issue is with the CSV we start from.

No, in this test we're using the same CSV and only replacing the images

tiraboschi:
Not working

PR #1025, opened 2020-12-27
Fix digester

nunnatsa:
Sometimes split the long line at the end of the deploy/images.env

This fix prints this line in two parts

**Release note**:

```release-note
NONE
```


orenc1 (approved):
I hope splitting to half would be enough to mitigate that writer issue.

/approve

PR #1027, opened 2020-12-29
Publish updated images on every merge

orenc1:
Every merge to master will execute a workflow for keeping HCO images up-to-date.
The workflow is executing the following steps:

1. Build the application binaries
2. Build the application images (`hyperconverged-cluster-operator`, `hyperconverged-cluster-webhook`)
3. Push the images to quay.io/kubevirt
4. Update the CSV with the new images' digests
5. Create the bundle and index images, containing the new HCO images and a unique semver in the CSV.
6. Push the bundle image (`hyperconverged-cluster-bundle`) and index image (`hyperconverged-cluster-index`) to quay.io.

**Release note**:

```release-note
Continuously update HCO images on quay.io/kubevirt, with unique semver in CSV.
```


nunnatsa:
I don't understand this PR. What are we trying to acheave here? Are we going to rename the CSV in each commit? how it's going to work?

But mostly I don't understand why taking the CSV without dealing with the application images - it will do nothing if the CSV will keep pointing to old images. 

If we're trying to build some kind of CI/CD, I think that this is not the way to go.

.github/workflows/build-push-images.yaml
```diff
@@ -0,0 +1,36 @@
+name: Build and Push Images
+
+on:
+  push:
+    branches:
+      - master
```
nunnatsa: maybe we want to add release-* ?

We can also do that at the release branch itself.

hack/upgrade-create-new-version.sh
```diff
@@ -23,6 +23,7 @@ LATEST_VERSION=$(ls -d ${PACKAGE_DIR}/*/ | sort -rV | head -1 | cut -d '/' -f 5)
 OPERATOR_NAME="kubevirt-hyperconverged-operator"
 LATEST_CSV_DIR="${PACKAGE_DIR}/${LATEST_VERSION}"
 LATEST_CSV_NAME="${OPERATOR_NAME}.v${LATEST_VERSION}.clusterserviceversion.yaml"
+LATEST_CSV_NAME="${OPERATOR_NAME}.v${LATEST_VERSION}.clusterserviceversion.yaml"
```
nunnatsa, line 26: Duplecation

tools/csv-merger/csv-merger.go
```diff
@@ -124,6 +124,7 @@ var (
 	hppoVersion                   = flag.String("hppo-version", "", "HPP operator version")
 	vmImportVersion               = flag.String("vm-import-version", "", "VM-Import operator version")
 	apiSources                    = flag.String("api-sources", cwd+"/...", "Project sources")
+	enableUniqueSemver            = flag.String("enable-unique-version", "", "Insert a skipRange annotation to support unique semver in the CSV")
```
nunnatsa, line 127: for boolean value, use `flag.Bool` or `flag.BoolVar`

tools/csv-merger/csv-merger.go
```diff
@@ -306,6 +307,11 @@ func main() {
 
 		// This is the basic CSV without an InstallStrategy defined
 		csvBase := components.GetCSVBase(csvParams)
+
+		if *enableUniqueSemver == "true" {
```
nunnatsa, line 311: See above. Use bool for boolean values

.github/workflows/build-push-images.yaml
```diff
@@ -0,0 +1,36 @@
...
+      QUAY_USERNAME: ${{ secrets.QUAY_USERNAME }}
+      QUAY_PASSWORD: ${{ secrets.QUAY_PASSWORD }}
+      REGISTRY_NAMESPACE: kubevirt
+    steps:
+      - name: Checkout the latest code
+        uses: actions/checkout@v2
+        with:
+          ref: master
+      - name: Build Applications Images
+        run: make container-build
```
nunnatsa, line 22: I maybe miss something here. As fa as I understand, the default image tag is "latest". I don't think we want that. `latest` should be the latest *released* image. 

I would tag each image twice - a uniqe image tag - and that maybe not needed, and another floating tag like `latest-unstable` or anything else but `latest`.

The unique tag is not mandatory, but it will allow rollbacks.

deploy/olm-catalog/kubevirt-hyperconverged/1.4.0/kubevirt-hyperconverged-operator.v1.4.0.clusterserviceversion.yaml
```diff
@@ -9,7 +9,7 @@ metadata:
     categories: OpenShift Optional
     certified: "false"
     containerImage: quay.io/kubevirt/hyperconverged-cluster-operator@sha256:a1b248457d1d6296390e7aa8b92e3db2a7786f32fe483ad9f44ca40ad1d69280
-    createdAt: "2021-01-20 12:20:36"
+    createdAt: "2021-01-21 09:41:24"
```
nunnatsa, line 12: if this is the only change in this file, please cancle this change to avoid conflicts.

nunnatsa (changes requested):
Added some comments, but in general - looks good!

.github/workflows/build-push-images.yaml
orenc1: does wildcard work on branch names?

.github/workflows/build-push-images.yaml
orenc1: What if we tag it with the latest CSV version (currently 1.4.0) for continuous updates throughout the release?

.github/workflows/build-push-images.yaml
orenc1: ok, it [does](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-including-branches-and-tags)

Makefile
```diff
@@ -5,8 +5,8 @@ SOURCES            := $(shell find . -name '*.go' -not -path "*/vendor/*")
 SHA                := $(shell git describe --no-match  --always --abbrev=40 --dirty)
 IMAGE_REGISTRY     ?= quay.io
 IMAGE_TAG          ?= latest
-OPERATOR_IMAGE     ?= kubevirt/hyperconverged-cluster-operator
-WEBHOOK_IMAGE      ?= kubevirt/hyperconverged-cluster-webhook
+OPERATOR_IMAGE     ?= orenc/hyperconverged-cluster-operator
```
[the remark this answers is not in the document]
orenc1: probably.

nunnatsa:
wait to #1081

orenc1:
Execution example on my fork:
https://github.com/orenc1/hyperconverged-cluster-operator/runs/1765836987?check_suite_focus=true

nunnatsa (approved):
/approve

nunnatsa (approved):

nunnatsa (approved):

PR #1029, opened 2020-12-30
Adds how to run webhook from local environment

erkanerol:
This PR

- adds an environment variable to the webhook so that we can configure tls secrets directory with an environment variable
- adds a short script (local_webhook.sh) to fetch tls secrets (created by cert-manager) from cluster to local environment
- adds a section into the developer document about how to use telepresence for opening a tunnel between local environments and k8s clusters. 

**Release note**:
```release-note
NONE
```


zcahana:
This basically looks good to me, but as I'm not entirely sure how developers are using `make local`, I'd like to verify that we are still able to run `hco-operator` locally, and the webhook remotely, without fiddling with telepresence, etc. Is that still supported?

nunnatsa:
> This basically looks good to me, but as I'm not entirely sure how developers are using `make local`, I'd like to verify that we are still able to run `hco-operator` locally, and the webhook remotely, without fiddling with telepresence, etc. Is that still supported?

We never ran the webhoo before, at least not sucesfully.

zcahana:
I chatted with @erkanerol and we concluded that it's still possible for a developer to just run the operator locally without carrying about/setting up the webhook.
Anyway, from my perspective this is good to go with.
/lgtm
/approve

pkg/components/components.go
```diff
@@ -1242,7 +1242,7 @@ func InjectVolumesForWebHookCerts(deploy *appsv1.Deployment) {
 		deploy.Spec.Template.Spec.Containers[index].VolumeMounts = append(container.VolumeMounts,
 			corev1.VolumeMount{
 				Name:      "apiservice-cert",
-				MountPath: hcov1beta1.WebhookCertDir,
+				MountPath: hcov1beta1.GetWebhookCertDir(),
```
nunnatsa, line 1245: I think this one is not needed.  Here we're creating the mount for the container, so we need the constant here.

hack/make_local.py
```diff
@@ -18,7 +18,7 @@ def gen_local_deployments(outdir):
         csv = yaml.safe_load(csv_file)
 
     with open(f'{outdir}/local.yaml', 'w') as out:
-        csv = [x for x in csv['spec']['install']['spec']['deployments'] if x['name'] != 'hco-operator']
+        csv = [x for x in csv['spec']['install']['spec']['deployments'] if x['name'] != 'hco-operator' and x['name'] != 'hco-webhook' ]
```
nunnatsa, line 21: I'm not sure about this line.

originally, we wanted to debug the operator. Now we have two applications to debug, but usually we'll want to debug only one of them.

I guess most of the cases, the developer will want to debug the operator, but will still need the webhook in place. And if the developer wants to debug the webhook, it still required a running operator.

So I think we need to have two flags (e.g. using envarinmet variables) to independently run the operatrot and/or the webhook. e.g. `DEBUG_OPERATOR` and `DEBUG_WEBHOOK`.

docs/run-locally.md
```diff
@@ -96,4 +96,25 @@ environment.
 Now it is possible to run HCO from VS Code.
 
 ![](../images/run_local_from_vscode.png)
- 
\ No newline at end of file
+
```
nunnatsa, line 99: General commment about this readme - since we remove the HC CR from the deployment, please add instruction how to deploy it.

nunnatsa (changes requested):

pkg/components/components.go
erkanerol: fixed.

docs/run-locally.md
erkanerol: done

hack/make_local.py
erkanerol: added DEBUG_OPERATOR and DEBUG_WEBHOOK variables into the script for this purpose

hack/config
```diff
@@ -10,6 +10,7 @@ HPP_VERSION="v0.7.0"
 CONVERSION_CONTAINER_VERSION="v2.0.0"
 VMWARE_CONTAINER_VERSION="v2.0.0-9"
 VM_IMPORT_VERSION="v0.2.5"
+HCO_KV_IO_VERSION="1.4.0"
```
nunnatsa, line 13: Please remove this one. Use `source ./hack/config`, then you can use the `CSV_VERSION` env. But please do not constant it here.

pkg/apis/hco/v1beta1/hyperconverged_webhook.go
```diff
@@ -51,7 +62,7 @@ func (r *HyperConverged) SetupWebhookWithManager(ctx context.Context, mgr ctrl.M
 	whHandler = handler
 	whHandler.Init(hcolog, mgr.GetClient(), operatorNsEnv, isOpenshift)
 
-	certs := []string{filepath.Join(WebhookCertDir, WebhookCertName), filepath.Join(WebhookCertDir, WebhookKeyName)}
+	certs := []string{filepath.Join(GetWebhookCertDir(), WebhookCertName), filepath.Join(GetWebhookCertDir(), WebhookKeyName)}
```
nunnatsa, line 65: Maybe you can call `GetWebhookCertDir` only once, either here or maybe in an `init()` method to a global var?

nunnatsa (changes requested):

hack/config
erkanerol: Done.

pkg/apis/hco/v1beta1/hyperconverged_webhook.go
erkanerol: Done

nunnatsa (approved):

PR #1034, opened 2021-01-05
Allow setting the 'HotplugVolumes' KV feature gate from HC CR

nunnatsa:
Add The `FeatureGates` field to the HyperConverged Spec.

Modify the the `HotplugVolumes` feature gate in the KC CR, and in KV config map according to new `FeatureGates` field in the HyperConverged CR.

**Release note**:

```release-note
Functional Changes:
Modify the the `HotplugVolumes` feature gate in the KC CR, and in KV config map according to new `FeatureGates` field in the HyperConverged CR.

API Changes:
Add The `FeatureGates` optional field to the HyperConverged Spec, with format of a map of string => boolean.
```


nunnatsa:
Missing DeepCopy for the new `FeatureGaes` field. Will add.

pkg/controller/operands/kubevirt_test.go
```diff
@@ -21,11 +21,21 @@ import (
 	kubevirtv1 "kubevirt.io/client-go/api/v1"
 	"os"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"strings"
 
 	virtconfig "kubevirt.io/kubevirt/pkg/virt-config"
 )
 
 var _ = Describe("KubeVirt Operand", func() {
+
```
enp0s3, line 30: Maybe it will be better to move this const block under `Context("Feature Gates", func() {` ?

pkg/controller/operands/kubevirt_test.go
```diff
@@ -247,6 +257,205 @@ var _ = Describe("KubeVirt Operand", func() {
...
+			})
+
+			It("should not add the feature gates if they are not set in HyperConverged CR", func() {
+				existingResource := NewKubeVirtConfigForCR(hco, commonTestUtils.Namespace)
+				By("KV CR should contain the fgEnabled feature gate", func() {
+					Expect(existingResource.Data[virtconfig.FeatureGatesKey]).Should(Equal(constFeatureGates))
+				})
+			})
+
+			It("should add the feature gates if they are set in HyperConverged CR", func() {
```
enp0s3, line 284: maybe for clarity we can rephrase to `if they exist and enabled..` just when we say `set` it can be also set to false and then it won't appear in the list

pkg/controller/operands/kubevirt_test.go
```diff
@@ -247,6 +257,205 @@ var _ = Describe("KubeVirt Operand", func() {
...
+				Expect(res.Updated).To(BeTrue())
+				Expect(res.Overwritten).To(BeFalse())
+				Expect(res.Err).To(BeNil())
+
+				foundResource := &corev1.ConfigMap{}
+				Expect(
+					cl.Get(context.TODO(),
+						types.NamespacedName{Name: existingResource.Name, Namespace: existingResource.Namespace},
+						foundResource),
+				).To(BeNil())
```
enp0s3, line 320: This code block has been reused in the next 3 tests, ca we move it to a separate helper function under the whole context?

tiraboschi (approved):

pkg/controller/operands/kubevirt_test.go
```diff
@@ -247,6 +257,205 @@ var _ = Describe("KubeVirt Operand", func() {
...
+
+				By("KV CR should not contain the WithHostPassthroughCPU feature gate", func() {
```
enp0s3, line 380: I think the description is a little bit misleading since I don't see any reference to `WithHostPassthroughCPU` in this test case. Maybe the aim was to check that only the default const feature gate list of KV exists, and everything else that is not managed by the HCO CR will be deleted from the list

pkg/controller/operands/kubevirt_test.go
```diff
@@ -499,6 +708,147 @@ var _ = Describe("KubeVirt Operand", func() {
...
+				Expect(res.Err).To(BeNil())
+
+				foundResource := &kubevirtv1.KubeVirt{}
+				Expect(
+					cl.Get(context.TODO(),
+						types.NamespacedName{Name: existingResource.Name, Namespace: existingResource.Namespace},
+						foundResource),
+				).To(BeNil())
+
+				By("KV CR should contain the WithHostPassthroughCPU feature gate", func() {
```
enp0s3, line 766: I think we are missing a dedicated `Expect` to locate the `WithHostPassthroughCPU`

enp0s3 (changes requested):
Thank you! looks good. please see my comments below

pkg/controller/operands/kubevirt_test.go
```diff
@@ -247,6 +257,205 @@ var _ = Describe("KubeVirt Operand", func() {
 			Expect(foundResource.Data).To(Equal(outdatedResource.Data))
 			Expect(foundResource.Data).To(Not(Equal(expectedResource.Data)))
 		})
+
+		Context("Feature Gates", func() {
```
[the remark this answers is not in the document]
nunnatsa: It's not. Please read again. I guess it's just a github diff consusion

pkg/controller/operands/kubevirt_test.go
nunnatsa: It reused in the CM tests

pkg/controller/operands/kubevirt_test.go
nunnatsa: typo. will fix

pkg/controller/operands/kubevirt.go
```diff
@@ -300,6 +354,13 @@ func translateKubeVirtConds(orig []kubevirtv1.KubeVirtCondition) []conditionsv1.
 }
 
 func NewKubeVirtConfigForCR(cr *hcov1beta1.HyperConverged, namespace string) *corev1.ConfigMap {
+	const constFeatureGates = "DataVolumes,SRIOV,LiveMigration,CPUManager,CPUNodeDiscovery,Sidecar,Snapshot"
```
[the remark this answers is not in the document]
nunnatsa: You are right. However, since we didn't remove the CM yet, this logic keep the old FGs with the old behaviour (for backward compatiblity), and currenly only the `HotplugVolumes` is controlled by HCO.

erkanerol:
/lgtm

As far as I see, @tiraboschi  has already approved the PR. The last changes after it looks good to me.

nunnatsa:
Thanks!

PR #1035, opened 2021-01-05
Remove unused local.yaml

erkanerol:
Removes creation of local.yaml file, which is unused after #1029

```release-note
NONE
```

### DIFF
--- a/hack/make_local.py
+++ b/hack/make_local.py
@@ -13,25 +13,6 @@ def get_env(line):
     return f'{env}={environ.get(env)}'
 
 
-def gen_local_deployments(outdir):
-    with open(f'deploy/olm-catalog/kubevirt-hyperconverged/{CSV_VERSION}/kubevirt-hyperconverged-operator.v{CSV_VERSION}.clusterserviceversion.yaml') as csv_file:
-        csv = yaml.safe_load(csv_file)
-
-    with open(f'{outdir}/local.yaml', 'w') as out:
-        csv = [x for x in csv['spec']['install']['spec']['deployments']]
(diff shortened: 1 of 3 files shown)


hack/make_local.py
```diff
@@ -13,25 +13,6 @@ def get_env(line):
     return f'{env}={environ.get(env)}'
 
 
-def gen_local_deployments(outdir):
```
nunnatsa, line 16: Please remove the `import yaml` line

nunnatsa (commented):
Just a thought: should we rename `hack/make_local.py`, as it's now only generating env vars? Not sure about it, though.

hack/make_local.sh
```diff
@@ -23,7 +23,6 @@ chmod +x _local/deploy.sh
 
 kubectl config set-context --current --namespace=${hco_namespace}
 _local/deploy.sh
-# kubectl apply -f _local/local.yaml
```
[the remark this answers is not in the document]
erkanerol: It was commented with previous PR. The comment is being deleted with this PR.

hack/make_local.py
erkanerol: Since I was not using linter for Python, I didn't catch it. Now, I installed a plugin. Thanks for this catch. Fixed.

erkanerol:
> Just a thought: should we rename `hack/make_local.py`, as it's now only generating env vars? Not sure about it, though.

I just renamed as "generate_local_env.py".

nunnatsa (approved):
/approve

PR #1039, opened 2021-01-07
test: tolerate conditions with Unknown status

tiraboschi:
Due to other errors, the reconciliation
loop can fail and HCO conditions can temporary
assume Unknown status which is semantically acceptable.
Let's tolerate it.

Test code only.

**Release note**:
```release-note
NONE
```


zcahana:
Looks reasonable.
/lgtm
/approve

zcahana (approved):

PR #1047, opened 2021-01-11
Add ENABLE_SSP option to hack/deploy.sh script

zcahana:
This PR adds a ENABLE_SSP environment variable option to `hack/deploy.sh` script, that allows to determine whether ssp-operator and associated resources will be deployed.

The default value is `false`, and is aimed to remove SSP deployment from plain k8s environments.
On OpenShift, this script can be run as `ENABLE_SSP=true hack/deploy.sh` to include SSP.

Note:
The SSP CRD is nevertheless deployed, also when SSP deployment is disabled - to support this we may need a bit larger change in how the CRD files are generated. I think this is harmless enough for now.

```release-note
Add ENABLE_SSP option to hack/deploy.sh script
```

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -642,6 +642,8 @@ spec:
 apiVersion: v1
 kind: Service
 metadata:
+  labels:
+    name: ssp-operator
   name: ssp-operator-service
 spec:
   ports:
@@ -656,6 +658,8 @@ spec:
 apiVersion: v1
 kind: Service
 metadata:
+  labels:
+    name: node-maintenance-operator
   name: node-maintenance-operator-service
 spec:
   ports:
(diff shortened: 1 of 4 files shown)


tiraboschi:
this is probably too much because we have two distinct use cases here:
1. deploy on plain K8s
2. deploy on Openshift bypassing OLM

and with this we can properly address 1. but this is going to cause troubles on 2.

erkanerol:
@zcahana  What do you think about setting the environment variable automatically by checking the cluster version like here? 
https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/pkg/util/cluster.go#L33

Would it be overkill?

zcahana:
> @zcahana What do you think about setting the environment variable automatically by checking the cluster version like here?
> https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/pkg/util/cluster.go#L33
> 
> Would it be overkill?

Good idea - will add!

nunnatsa (approved):
/approve

PR #1049, opened 2021-01-12
WIP: KubeVirt feature gates may only be modified from HC CR

nunnatsa:
Before this PR, KubeVirt feature gates are modified by changing the feature-gate list in KubeVirt configmap. This changes were not controlled by HCO, but HCO overrode them to default list during upgrades. In #1034, the `HotplugVolumes` feature gate was added, but it can only be set/disabled/deleted by modifying the new `featureGates` list in HC CR.

This PR makes all the feature gates to behave like `HotplugVolumes`, meaning they all can only be modified in the HC CR, but they are also copied to the config map, since KubeVirt still reads the feature gates from there. But any change in the config map will be reconciled according to the list in HC CR.

**Release note**:

```release-note
KubeVirt feature gates may only be modified from HC CR

The ConfigMap feature gate is read-only.
```

### DIFF
--- a/pkg/controller/hyperconverged/hyperconverged_controller.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller.go
@@ -297,6 +297,12 @@ func (r *ReconcileHyperConverged) doReconcile(req *common.HcoRequest) (reconcile
 			req.Logger.Error(err, "Failed to add conditions to status")
 			return reconcile.Result{}, err
 		}
+
+	}
+
+	// Set default feature gates, if not pre-defined in the CR
+	if applyDefaultFeatureGates(req.Instance, r.operandHandler.GetDefaultFeatureGates()) {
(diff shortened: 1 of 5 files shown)


nunnatsa:
/hold

API still under disccussion.

dankenigsberg:
> /hold
> 
> API still under disccussion.

Can you list (in the PR text, or even in an .md file) which are the `featureGates` that HCO becomes familiar with? What are their default value? What is their meaning?

Would you be kind to add a example HCO yaml showing select features enabled?

As I user of HCO I would like to know all the options I have to tweak it.

fabiand:
What is the status of tihs PR, is this work needed?

nunnatsa:
This PR is outdated. The feature gates implementation was changed from map of [string]bool to struct of named boolean fields.

This PR is no longer relevant.

PR #1052, opened 2021-01-13
Fix webhook configuration of ssp

erkanerol:

```release-note
NONE
```

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -650,7 +650,7 @@ spec:
     protocol: TCP
     targetPort: 9443
   selector:
-    name: ssp-operator
+    control-plane: ssp-operator
   type: ClusterIP
 ---
 apiVersion: v1
--- a/deploy/webhooks.yaml
+++ b/deploy/webhooks.yaml
@@ -124,7 +124,7 @@ webhooks:
       name: ssp-operator-service
       namespace: kubevirt-hyperconverged
       path: /validate-ssp-kubevirt-io-v1beta1-ssp
-      port: 8443
+      port: 9443
   failurePolicy: Fail
   matchPolicy: Equivalent
   name: vssp.kb.io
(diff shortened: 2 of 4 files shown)


deploy/webhooks.yaml
```diff
@@ -124,7 +124,7 @@ webhooks:
       name: ssp-operator-service
       namespace: kubevirt-hyperconverged
       path: /validate-ssp-kubevirt-io-v1beta1-ssp
-      port: 8443
+      port: 9443
```
zcahana, line 127: Good catch. And +1 for auto-generating this file sometime in the future :-)

deploy/webhooks.yaml
```diff
@@ -139,5 +139,6 @@ webhooks:
     - UPDATE
     resources:
     - ssps
+    scope: Cluster
```
zcahana, line 142: The SSP CRD isn't cluster-scoped, see here: https://github.com/kubevirt/hyperconverged-cluster-operator/blob/20df7b1db67c06d4d7edc3fe362d2be8e09e27fe/deploy/crds/scheduling-scale-performance00.crd.yaml#L15

deploy/operator.yaml
```diff
@@ -650,7 +650,7 @@ spec:
     protocol: TCP
     targetPort: 9443
   selector:
-    name: ssp-operator
+    control-plane: ssp-operator
```
zcahana, line 653: This file is autogenerated, so fixing it inline won't do - probably need to fix this at the manifest-templator.

zcahana (changes requested):

deploy/operator.yaml
erkanerol: done

deploy/webhooks.yaml
erkanerol: reverted

zcahana:
/lgtm

Since we don't have any non-OLM lane that exercises the SSP webhook (created in OpenShift only), I guess we can make do for now with manual verification that this webhook is now usable.
@tiraboschi should we be adding a non-OLM lane that uses `deploy.sh` on OpenShift?

PR #1063, opened 2021-01-18
Add Two New Feature Gates

nunnatsa:
This PR add support in two new feature gates:
* `withHostModelCPU`: Support migration for VMs with host-model CPU mode
* `withHostPassthroughCPU`: Allow migrating a virtual machine with CPU host-passthrough mode.

This two new feature gate may be added to the HyperConverged CR:
```yaml
kind: HyperConverged
metadata:
  name: kubevirt-hyperconverged
spec:
  infra: {}
  workloads: {}
  featureGates:
    hotplugVolumes: true
    withHostPassthroughCPU: true
    withHostModelCPU: true
```

**Release note**:
```release-note
Added two new feature gates:
* `withHostModelCPU`: Support migration for VMs with host-model CPU mode
* `withHostPassthroughCPU`: Allow migrating a virtual machine with CPU host-passthrough mode.
```

### DIFF
--- a/deploy/crds/hco00.crd.yaml
+++ b/deploy/crds/hco00.crd.yaml
@@ -156,6 +156,12 @@ spec:
                   hotplugVolumes:
                     description: Allow attaching a data volume to a running VMI
                     type: boolean
+                  withHostModelCPU:
+                    description: Support migration for VMs with host-model CPU mode
+                    type: boolean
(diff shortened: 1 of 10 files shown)


tiraboschi:
if we consider it still not a stable API/feature, I think this is the best we can do right now
/lgtm

PR #1064, opened 2021-01-18
Upgrade controller-runtime version to v0.7.0

andreyod:
controller-runtime v0.7.0 introduces various breaking ang high impact changes.

For full list of changes see:
https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.7.0

Relevant changes:
* Use client.Object and client.ObjectList
* Add a context to Reconciler interface
* client.ObjectKeyFromObject now uses client.Object
* Upgrade v1beta1 admission types to v1
* Add predicate for annotations change on update event
* Simplify usage of EnqueueRequestsFromMapFunc

Kubevirt related change:
Stop using kubevirt.io/kubevirt/pkg/virt-config package as it eventually uses incompatible k8s package.

**Release note**:

```release-note
Upgrade controller-runtime to v0.7.0
```


pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -182,7 +183,7 @@ type ReconcileHyperConverged struct {
 // Note:
 // The Controller will requeue the Request to be processed again if the returned error is non-nil or
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
-func (r *ReconcileHyperConverged) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+func (r *ReconcileHyperConverged) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
```
nunnatsa, line 186: OK, this is new... But if we finally receiving the context from the SDK, we need to stop use ours. Please use it in `common.NewHcoRequest` (will require to change its signature, to add `ctx context.Context` as its first parameter).

nunnatsa (changes requested):
Just one comment about the new context parameter in `Reconcile`

pkg/controller/hyperconverged/hyperconverged_controller.go
andreyod: Done

PR #1072, opened 2021-01-19
[WIP] Add func-tests into operator image

erkanerol:

**Release note**:
```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ sanity:
 	./hack/build-manifests.sh
 	git difftool -y --trust-exit-code --extcmd=./hack/diff-csv.sh
 
-build: build-operator build-csv-merger build-webhook
+build: build-operator build-csv-merger build-webhook build-functest
 
 build-operator: $(SOURCES) ## Build binary from source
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -25,7 +25,7 @@ sanity:
 	./hack/build-manifests.sh
 	git difftool -y --trust-exit-code --extcmd=./hack/diff-csv.sh
 
-build: build-operator build-csv-merger build-webhook
+build: build-operator build-csv-merger build-webhook build-functest
```
nunnatsa, line 28: Please remove `build-webhook`. Not in use here.

erkanerol: Since we call `make build` in sanity checks as well, I reverted this part and call the targets in Dockerfile selectively.

erkanerol:
Question: Why don't we vendor files for func-tests? While adding the test binary into the operator image, should not we vendor dependencies of tests as well?  @tiraboschi  @nunnatsa

erkanerol:
We decided to discuss this approach again and to find another way. Closing...

PR #1082, opened 2021-01-21
WIP: Report Upgradable OperatorCondition to OLM

thetechnick:
This PR introduces reporting of HCO Conditions back to OLM via the `OperatorCondition` CRD.
Setting the `Upgradable` condition to `False` will allow us to block OLM from disturbing a running upgrade process within HCO.

Note: 
I am not using operator-lib (https://github.com/operator-framework/operator-lib/blob/v0.4.0/conditions/conditions.go), because it is hardcoded to discover the `OperatorCondition` Namespace from the filesystem, making it necessary to wrap our tests in a container or mess around in the developers local filesystem.

Tests will follow.

Ref: https://olm.operatorframework.io/docs/advanced-tasks/communicating-operator-conditions-to-olm/

**Release note**:

```release-note
OLMs OperatorCondition is now used to block OLM from upgrading HCO, while HCO is upgrading it's child components.
```


pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -336,6 +340,18 @@ func (r *ReconcileHyperConverged) doReconcile(req *common.HcoRequest) (reconcile
 	knownHcoVersion, _ := req.Instance.Status.GetVersion(hcoVersionName)
 
 	if !r.upgradeMode && !init && knownHcoVersion != r.ownVersion {
+		if hcoutil.GetClusterInfo().IsManagedByOLM() {
```
nunnatsa, line 343: Sicnse most of the values are constant, lets add it all to the `setOperatorCondition` method and get a cleaner code.

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -675,6 +691,18 @@ func (r *ReconcileHyperConverged) completeReconciliation(req *common.HcoRequest)
 
 		// if in upgrade mode, and all the components are upgraded - upgrade is completed
 		if r.upgradeMode && req.ComponentUpgradeInProgress {
+			if hcoutil.GetClusterInfo().IsManagedByOLM() {
```
nunnatsa, line 694: same

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -336,6 +340,18 @@ func (r *ReconcileHyperConverged) doReconcile(req *common.HcoRequest) (reconcile
 	knownHcoVersion, _ := req.Instance.Status.GetVersion(hcoVersionName)
 
 	if !r.upgradeMode && !init && knownHcoVersion != r.ownVersion {
+		if hcoutil.GetClusterInfo().IsManagedByOLM() {
+			// Ensure OLM is not interfering while we are upgrading components.
+			if err := setOperatorCondition(req.Ctx, r.client, metav1.Condition{
```
nunnatsa, line 345: is the setOperatorCondition will be a method in the Reconciler type, there no need to pass the client and the version as parameters.

pkg/controller/hyperconverged/hyperconverged_controller.go
nunnatsa, line 343: let's also add the condition into the `setOperatorCondition` method

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -336,6 +340,18 @@ func (r *ReconcileHyperConverged) doReconcile(req *common.HcoRequest) (reconcile
...
+				Type:    "Upgradable",
```
nunnatsa, line 346: please use constants

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -845,6 +873,28 @@ func drop(slice []string, s string) ([]string, bool) {
 	return newSlice, dropped
 }
 
+// sets a Condition on OLMs OperatorCondition to report status information about this Operator back to OLM.
+func setOperatorCondition(ctx context.Context, c client.Client, cond metav1.Condition) error {
```
nunnatsa, line 877: Suggestion:
```golang
func (r *ReconcileHyperConverged) setOperatorCondition(ctx context.Context, cond metav1.ConditionStatus) 
```

nunnatsa (commented):
Preliminary Review

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -819,6 +806,47 @@ func (r *ReconcileHyperConverged) detectTaintedConfiguration(req *common.HcoRequ
 	}
 }
 
+func (r *ReconcileHyperConverged) setOperatorConditionUpgradable(ctx context.Context, status metav1.ConditionStatus) error {
+	cond := metav1.Condition{
+		Type:    olmUpgradableConditionType,
+		Status:  metav1.ConditionTrue,
+		Reason:  reconcileCompleted,
+		Message: reconcileCompletedMessage,
```
nunnatsa, line 816: Not sure we need any Reason and mesaage here. I would keep it empty. Anyway, "reconcile completed" is not the right message.

thetechnick:
Closing this until we can move forward with:
https://bugzilla.redhat.com/show_bug.cgi?id=1927340

PR #1089, opened 2021-01-24
Try to fix the covaeralls issue

nunnatsa:
Since HCO moved to run sanity in github action, the coveralls report is wrong - relatives to
the current branch insted of to the master branch, so the coverage check does not show increasing
or decreasing.

This PR moves to use the shogo82148/actions-goveralls@v1 action in order to solve this issue.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.github/workflows/pr-sanity.yaml
+++ b/.github/workflows/pr-sanity.yaml
@@ -4,6 +4,8 @@ name: Sanity Checks
 
 # Controls when the action will run.
 on:
+  push:
+    branches: [ main ]
   pull_request:
     branches: [ master ]
 
--- a/.github/workflows/pr-sanity.yaml
+++ b/.github/workflows/pr-sanity.yaml
@@ -35,15 +35,9 @@ jobs:
       - name: Unit tests
         run: make test
 
-      - name: Install goveralls
-        env:
-          GO111MODULE: off
-        run: go get github.com/mattn/goveralls
-
-      - name: Send coverage
-        env:
-          COVERALLS_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: goveralls -coverprofile=./coverprofiles/cover.coverprofile -service=github
+      - uses: shogo82148/actions-goveralls@v1
+        with:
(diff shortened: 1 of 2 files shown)


nunnatsa:
didn't work

.github/workflows/pr-sanity.yaml
```diff
@@ -4,6 +4,8 @@ name: Sanity Checks
 
 # Controls when the action will run.
 on:
+  push:
+    branches: [ main ]
```
erkanerol, line 8: @nunnatsa should not it be "master" ?

nunnatsa: fixed

orenc1: i would prefer to use native yaml format, but not critical for now.
```
on:
  push:
    branches:
    - master
```

hack/build-tests.sh
```diff
@@ -14,7 +14,7 @@ if [ "${JOB_TYPE}" == "travis" ]; then
     go mod vendor
     PACKAGE_PATH="pkg/"
     mkdir -p coverprofiles
-    KUBEVIRT_CLIENT_GO_SCHEME_REGISTRATION_VERSION=v1 ginkgo -r -cover -outputdir=./coverprofiles -coverprofile=cover.coverprofile ${PACKAGE_PATH}
+    KUBEVIRT_CLIENT_GO_SCHEME_REGISTRATION_VERSION=v1 ginkgo -r -covermode atomic -outputdir=./coverprofiles -coverprofile=cover.coverprofile ${PACKAGE_PATH}
```
orenc1, line 17: are you sure you don't need to still specify the `cover` flag to run the coverage tests?
```
  -cover
    	Run tests with coverage analysis, will generate coverage profiles with the package name in the current directory.
```

nunnatsa: It's not needed if we are using the `-covermod` flag

orenc1: /lgtm

PR #1094, opened 2021-01-25
add default for HypervStrictCheck featureGate

kwiesmueller:
I'm not sure if this is the right way to do this. This will set the `HypervStrictCheck` FeatureGate to true if it (or any feature gates at all) are not provided.
It can be set to false by specifying it in the HCO config.

**Release note**:

```release-note
The kubevirt `HypervStrictCheck` FeatureGate will default to true.
```

/cc @ksimon1 @omeryahud

### DIFF
--- a/pkg/apis/hco/v1beta1/hyperconverged_types.go
+++ b/pkg/apis/hco/v1beta1/hyperconverged_types.go
@@ -40,7 +40,7 @@ type HyperConvergedSpec struct {
 	// the feature. Setting `false` or removing the feature gate, disables the feature.
 	// +optional
 	// +TODO: Always keep the default FeatureGates in sync with the default field values in HyperConvergedFeatureGates //NOSONAR
-	// +kubebuilder:default={withHostModelCPU: true, withHostPassthroughCPU: false}
+	// +kubebuilder:default={withHostModelCPU: true, withHostPassthroughCPU: false, hypervStrictCheck: true}
 	FeatureGates *HyperConvergedFeatureGates `json:"featureGates,omitempty"`
 
 	// operator version
@@ -58,6 +58,7 @@ type HyperConvergedConfig struct {
 // by default yet.
 // +optional
 // +k8s:openapi-gen=true
+// +kubebuilder:default={}
 type HyperConvergedFeatureGates struct {
 	// Allow migrating a virtual machine with SRIOV interfaces.
 	// When enabled virt-launcher pods of virtual machines with SRIOV
@@ -80,6 +81,12 @@ type HyperConvergedFeatureGates struct {
 	// +optional
 	// +kubebuilder:default=true
 	WithHostModelCPU *bool `json:"withHostModelCPU,omitempty"`
+
+	// Enable HyperV strict host checking for HyperV enlightenments
+	// Defaults to true, even when HyperConvergedFeatureGates is empty
+	// +optional
+	// +kubebuilder:default=true
+	HypervStrictCheck *bool `json:"hypervStrictCheck,omitempty"`
 }
 
 func (fgs *HyperConvergedFeatureGates) IsHotplugVolumesEnabled() bool {
@@ -96,6 +103,10 @@ func (fgs *HyperConvergedFeatureGates) IsWithHostPassthroughCPUEnabled() bool {
 
 func (fgs *HyperConvergedFeatureGates) IsWithHostModelCPUEnabled() bool {
 	return (fgs != nil) && (fgs.WithHostModelCPU != nil) && (*fgs.WithHostModelCPU)
+}
+
+func (fgs *HyperConvergedFeatureGates) IsHypervStrictCheckEnabled() bool {
+	return (fgs != nil) && (fgs.HypervStrictCheck != nil) && (*fgs.HypervStrictCheck)
 }
 
 // HyperConvergedStatus defines the observed state of HyperConverged
--- a/pkg/controller/operands/kubevirt.go
+++ b/pkg/controller/operands/kubevirt.go
@@ -295,38 +295,37 @@ func (h *kvConfigHooks) addEnabledFeatureGates(req *common.HcoRequest, foundFgSp
 
 func filterOutDisabledFeatureGates(req *common.HcoRequest, foundFgSplit []string, resultFg []string) (bool, []string) {
 	fgChanged := false
+	featureGateChecks := getFeatureGateChecks(req.Instance.Spec.FeatureGates)
 	for _, fg := range foundFgSplit {
-		// Remove if not in HC CR
-		switch fg {
-		case HotplugVolumesGate:
-			if !req.Instance.Spec.FeatureGates.IsHotplugVolumesEnabled() {
-				fgChanged = true
-				continue
-			}
-		case kvWithHostPassthroughCPU:
-			if !req.Instance.Spec.FeatureGates.IsWithHostPassthroughCPUEnabled() {
-				fgChanged = true
-				continue
-			}
-		case kvWithHostModelCPU:
-			if !req.Instance.Spec.FeatureGates.IsWithHostModelCPUEnabled() {
-				fgChanged = true
-				continue
-			}
-		case SRIOVLiveMigrationGate:
-			if !req.Instance.Spec.FeatureGates.IsSRIOVLiveMigrationEnabled() {
-				fgChanged = true
-				continue
-			}
-		case kvHypervStrictCheck:
-			if !req.Instance.Spec.FeatureGates.IsHypervStrictCheckEnabled() {
-				fgChanged = true
-				continue
-			}
+		if isFeatureGateMissingFrom(featureGateChecks, fg) {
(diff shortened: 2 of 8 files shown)


kwiesmueller:
This is probably missing code generation etc.
Just want to clarify the approach first.

omeryahud:
The direction looks good to me
@orenc1 @tiraboschi  Can you take a look please?

tiraboschi:
can you please also run `make build-manifests` to regenerate the deploy manifests?

pkg/apis/hco/v1beta1/hyperconverged_types.go
```diff
@@ -72,6 +72,11 @@ type HyperConvergedFeatureGates struct {
 	// Support migration for VMs with host-model CPU mode
 	// +optional
 	WithHostModelCPU *bool `json:"withHostModelCPU,omitempty"`
+
+	// Enable HyperV strict host checking for HyperV enlightenments
+	// Defaults to true, even when HyperConvergedFeatureGates is empty
```
nunnatsa, line 77: I find it confusing. What do you think about setting this during reconciling? Somethink like this:

```golang
	if req.Instance.Spec.FeatureGates == nil {
		trueVal := true
		req.Instance.Spec.FeatureGates = &hcov1beta1.HyperConvergedFeatureGates{
			WithHostModelCPU: &trueVal,
		}
	}
```

pkg/controller/operands/kubevirt_test.go
```diff
@@ -1385,21 +1386,29 @@ var _ = Describe("KubeVirt Operand", func() {
 
 		enabled := true
 		disabled := false
+		numOfDefaultGates := 1
```
nunnatsa, line 1389: Should be const, rather than var

nunnatsa (commented):
Added online questions about how to enable features by default. @tiraboschi - what do you think?

pkg/apis/hco/v1beta1/hyperconverged_types.go
tiraboschi: proper defaulting on read?
https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#defaulting

kwiesmueller: Yeah, I think this should be the way to go: https://book.kubebuilder.io/cronjob-tutorial/webhook-implementation.html

deploy/olm-catalog/kubevirt-hyperconverged/1.4.0/kubevirt-hyperconverged-operator.v1.4.0.clusterserviceversion.yaml
```diff
@@ -9,7 +9,7 @@ metadata:
     categories: OpenShift Optional
     certified: "false"
     containerImage: quay.io/kubevirt/hyperconverged-cluster-operator@sha256:a137fd1ee55ae8806cbdd66912c7931c7f567315f0a927b3a12090112e96edb1
-    createdAt: "2021-01-28 18:56:06"
+    createdAt: "2021-01-31 03:41:38"
```
nunnatsa, line 12: Please undo this change if it's the only change in this file. It just cause conflicts.

nunnatsa (changes requested):
Please implement as in #1109 and #1113

nunnatsa (changes requested):
Missing handling in `filterOutDisabledFeatureGates`.

Missing unit test for `IsHypervStrictCheckEnabled`.

kwiesmueller:
Thanks for spotting those @nunnatsa !

nunnatsa (approved):
/approve

nunnatsa:
I approve it even though it caused a new sonarCloud warning abut the complecity of `filterOutDisabledFeatureGates`...

kwiesmueller:
@nunnatsa have a look at the new commit. Does this change make sense as a cleanup?
I'm not sure if I like the map much more than the switch but it should reduce complexity.

pkg/controller/operands/kubevirt.go
```diff
@@ -294,35 +295,36 @@ func (h *kvConfigHooks) addEnabledFeatureGates(req *common.HcoRequest, foundFgSp
...
 		resultFg = append(resultFg, fg)
 	}
 	return fgChanged, resultFg
 }
 
+func isFeatureGateMissingFrom(checks featureGateChecks, featureGate string) bool {
+	if check, isKnown := checks[featureGate]; isKnown {
+		return !check()
+	}
+	return false
```
nunnatsa, line 313: I think it should be `true` here. If we get here, the FG **is** missing. Do you have a test to cover this (It's a bit confusing).

pkg/controller/operands/kubevirt.go
```diff
@@ -294,35 +295,36 @@ func (h *kvConfigHooks) addEnabledFeatureGates(req *common.HcoRequest, foundFgSp
...
+}
+
+type featureGateChecks map[string]func() bool
+
+func getFeatureGateChecks(featureGates *hcov1beta1.HyperConvergedFeatureGates) featureGateChecks {
```
nunnatsa, line 318: I don't have strong opinion about this solution. I'm not sure it's not "too much". 

Regardless, maybe we can add a check for `featureGates == nil` before creating the map.

pkg/controller/operands/kubevirt.go
```diff
@@ -528,5 +530,9 @@ func getKvFeatureGateList(fgs *hcov1beta1.HyperConvergedFeatureGates) []string {
 		res = append(res, SRIOVLiveMigrationGate)
 	}
 
+	if fgs.IsHypervStrictCheckEnabled() {
```
nunnatsa, line 533: Maybe we can use the `filterOutDisabledFeatureGates` method here as well?

nunnatsa (changes requested):
Added 3 commnets. 2 of them may be ignored, but this one - https://github.com/kubevirt/hyperconverged-cluster-operator/pull/1094/files#r572610672 - I think it's a bug. Please re-check.

pkg/controller/operands/kubevirt.go
kwiesmueller: I agree, switched it over while testing.
Will add a test.

pkg/controller/operands/kubevirt.go
kwiesmueller: It will cause more allocations than the switch. Not sure if it's a concern though.

nunnatsa: Yes, and it called twice each time.

pkg/controller/operands/kubevirt_test.go
```diff
@@ -1569,15 +1582,107 @@ var _ = Describe("KubeVirt Operand", func() {
...
+			}
+
+			changed, result := filterOutDisabledFeatureGates(
+				requestWithFeatureGates(fgs),
+				allGates,
+				make([]string, 0, len(allGates)))
+			Expect(changed).To(BeFalse())
+			Expect(result).To(Equal(allGates))
+		})
+		It("should remove gates that don't have a matching field", func() {
```
kwiesmueller, line 1665: @nunnatsa I'm not sure if that assumption is correct.
If my brain is not playing tricks on me the previous code did not remove gates that don't exist while it should have imo.
What do you think?

kwiesmueller:
@nunnatsa could you please take another look?

nunnatsa:
> @nunnatsa could you please take another look?

@kwiesmueller - sanity test failed. Details are here: https://github.com/kubevirt/hyperconverged-cluster-operator/pull/1094/checks?check_run_id=1865547440

```
• Failure [0.000 seconds]
HyperconvergedController Reconcile HyperConverged Update Conflict Error [It] Should requeue in case of update conflict 
/home/runner/work/hyperconverged-cluster-operator/hyperconverged-cluster-operator/pkg/controller/hyperconverged/hyperconverged_controller_test.go:1457

  Expected
      <nil>: nil
  not to be nil

  /home/runner/work/hyperconverged-cluster-operator/hyperconverged-cluster-operator/pkg/controller/hyperconverged/hyperconverged_controller_test.go:1475
```

That usually mean that an unexpected error is returned somewhere in `doReconcile`

nunnatsa:
Both `hack/ginkgo.sh pkg/controller/hyperconverged/...` and `hack/ginkgo.sh pkg/controller/operands/...` are failing

pkg/controller/operands/kubevirt_test.go
kwiesmueller: I removed this test for now as it collided with existing behavior.

kwiesmueller:
@nunnatsa the problem should be fixed.
It was a mix of the nil check and the behavior change I discussed in the other comment.

nunnatsa:
/override coverage/coveralls - not less tests, but some refactoring cause changs in the coverage precentage.

/lgtm
/approve

PR #1100, opened 2021-01-27
Reduce SonarClaud number of warnings

nunnatsa:
1. In SSP operator handler, use a constant instead of "ssp.kubevirt.io"
2. In SSP operator handler, use a constant instead of "kubevirt.io"
3. refactor `func getQuickStartHandlers()` in pkg/controller/operands/quickStart.go to get cleaner code. Reduce the number of code branches by splitting to smaller functions.
4. refactor `func ComponentResourceRemoval()` in pkg/util/util.go to get cleaner code. Reduce the number of code branches by splitting to smaller functions.
5. in `pkg/controller/commonTestUtils/eventEmitterMock.go`, add comments to explain non implemented mock methods.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/operands/quickStart.go
+++ b/pkg/controller/operands/quickStart.go
@@ -3,6 +3,7 @@ package operands
 import (
 	"context"
 	"errors"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
@@ -106,7 +107,7 @@ func (h qsHooks) updateCr(req *common.HcoRequest, Client client.Client, exists r
 	return false, false, nil
 }
 
-func checkCrdExists(ctx context.Context, Client client.Client, logger log.Logger) (bool, error) {
+func checkCrdExists(ctx context.Context, Client client.Client, logger log.Logger) error {
 	qsCrd := &extv1.CustomResourceDefinition{
 		TypeMeta: metav1.TypeMeta{
 			Kind: customResourceDefinitionName,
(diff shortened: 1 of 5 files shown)


pkg/controller/operands/quickStart.go
```diff
@@ -129,57 +129,78 @@ func checkCrdExists(ctx context.Context, Client client.Client, logger log.Logger
 
 func getQuickStartHandlers(logger log.Logger, Client client.Client, Scheme *runtime.Scheme, hc *hcov1beta1.HyperConverged) ([]Operand, error) {
 	crdExists, err := checkCrdExists(context.TODO(), Client, logger)
```
erkanerol, line 131: We should handle the error here.

nunnatsa: We do. it returned.

checkCrdExists returns false if error or if no CRD found. If the method returns true, error is nil.

pkg/controller/operands/quickStart.go
```diff
@@ -129,57 +129,78 @@ func checkCrdExists(ctx context.Context, Client client.Client, logger log.Logger
...
 
-	if err != nil {
-		return nil, err
-	}
+func processQuickstartFile(path string, info os.FileInfo, logger log.Logger, hc *hcov1beta1.HyperConverged, Client client.Client, Scheme *runtime.Scheme) Operand {
+	if !info.IsDir() && strings.HasSuffix(info.Name(), ".yaml") {
+		file, err := os.Open(path)
+		if err != nil {
+			logger.Error(err, "Can't open the quickStart yaml file", "file name", path)
+			return nil
```
erkanerol, line 192: why do we return nil here?

nunnatsa: I don't want to fail the holw thing over quickstart. This is not a logical change, just moving things around.

pkg/util/util.go
```diff
@@ -215,22 +215,52 @@ func ComponentResourceRemoval(ctx context.Context, c client.Client, obj interfac
...
+		return err
+	}
+
+	if !shouldDeleteResource(resource, hcoName, logger) {
+		return nil
+	}
+
+	opts := getDeletionOption(dryRun, wait)
+
+	if deleted, err := doDeleteResource(ctx, c, resource, opts); !deleted {
```
erkanerol, line 229: I would implement doDeleteResource such that it only returns error. If it is not nil, just return it. Otherwise, continue.

pkg/controller/operands/quickStart.go
erkanerol: As far as I see from the old logic, we were returning the error for this case. We are changing the logic.

pkg/controller/operands/quickStart.go
erkanerol: Yes. Actually we return now.  This is not a big deal and we can keep it as it is. 

let me share my opinion:

my understanding from the Go world, the returned error has a kind of priority. It says whether something went wrong or not. If it is nil, then you can check the other returned values. Now, we have an assumption about the implementation of checkCrdExists function. We know it won't return  (true, err) if the error is not nil. However, it may change in the future and we cannot blame the person who changes it in this way. He/She can say "Hey. I am returning an error. Why don't you respect it.".  Therefore, this implementation puts another responsibility into shoulders of the maintainers of checkCrdExists function, which doesn't seem so good to me.

PR #1104, opened 2021-01-28
Fix code smells reported by SonarCloud

zcahana:
1. Extract duplicate magic strings as constants
2. Rename variables to adhere to naming conventions
3. Refactor methods with high cognitive complexity

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/controller/hyperconverged/hyperconverged_controller.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller.go
@@ -184,33 +184,26 @@ type ReconcileHyperConverged struct {
 // The Controller will requeue the Request to be processed again if the returned error is non-nil or
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
 func (r *ReconcileHyperConverged) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
-
-	secCRPlaceholder, err := getSecondaryCRPlaceholder()
+	hcoTriggered, err := isTriggeredByHyperConverged(request)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
 
-	hcoTriggered := true
-	pRequest := request
-	if request.NamespacedName == secCRPlaceholder {
-		hcoTriggered = false
-		hco, err := getHyperconverged()
+	reconcileRequest := request
+	if hcoTriggered {
+		log.Info("Reconciling HyperConverged operator")
+
+		r.operandHandler.Reset()
+	} else {
+		log.Info("The reconciliation got triggered by a secondary CR object")
+
+		reconcileRequest, err = getHyperconvergedReconcileRequest()
 		if err != nil {
 			return reconcile.Result{}, err
 		}
-		pRequest = reconcile.Request{
-			NamespacedName: hco,
-		}
-	} else {
-		r.operandHandler.Reset()
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -361,6 +304,30 @@ func (r *ReconcileHyperConverged) doReconcile(req *common.HcoRequest) (reconcile
 	return reconcile.Result{}, nil
 }
 
+func (r *ReconcileHyperConverged) constructHcoRequest(ctx context.Context, request reconcile.Request) (*common.HcoRequest, error) {
```
nunnatsa, line 307: This function does two things. It check if the source of the request is a change in HC or in a secondery watched resource, *and* creates the hcoRequest. The latest already implemented in common.NewHcoRequest.

I think this method need a new name and the last line should be placed in the original place, out of this method.

zcahana: Done.

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -615,52 +565,48 @@ func (r *ReconcileHyperConverged) aggregateComponentConditions(req *common.HcoRe
...
 
-			if _, conditionFound = req.Conditions[conditionsv1.ConditionAvailable]; !conditionFound { // (#chart 17) {
-				req.Conditions.SetStatusCondition(conditionsv1.Condition{ // (#chart 18)
-					Type:    conditionsv1.ConditionAvailable,
-					Status:  corev1.ConditionTrue,
-					Reason:  reconcileCompleted,
-					Message: reconcileCompletedMessage,
-				})
-			}
 		}
```
nunnatsa, line 607: Worth a new function?

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -790,6 +736,55 @@ func (r *ReconcileHyperConverged) detectTaintedConfiguration(req *common.HcoRequ
 	}
 }
 
+func (r *ReconcileHyperConverged) updateHCO(req *common.HcoRequest) error {
```
nunnatsa, line 739: we're not updating HCO - HC operator, but HC CR. Pleasse consider renaming.

nunnatsa (commented):
Added some low priority comments. maybe for next time? Not blocking changes.

pkg/controller/hyperconverged/hyperconverged_controller.go
zcahana: Good point. Will rename.

pkg/controller/hyperconverged/hyperconverged_controller.go
zcahana: Considered this (regarding these 3 comments) - I think it'll cause the overall flow to become less clear than is now.

zcahana:
Some suspicious failures in the different lanes here - need to investigate what's causing them.

zcahana:
Some lanes are failing due to timeouts after long hours of test execution - not clear where it's stuck.
Temporarily reverting 9f58a27 while trying to locate the offending code.

zcahana:
Tests pass after reverting 38b94c0. Hence, this is the offending commit. Reason is not clear yet - debugging...

nunnatsa (approved):
/approve

PR #1112, opened 2021-01-31
Fix CSV upgrade workflow

orenc1:
Currently, the "unstable" bundle image specified in the "unstable" index image at _quay.io/kubevirt_, stays with the same tag upon updates, which causing the cluster not to pull the updated bundle image when an upgrade is available, because the same tag already exists on the cluster's nodes.
This PR adds the bundle image's digest to the index, instead of unchanging tag, ensuring that a cluster subscribed to `quay.io/kubevirt/hyperconverged-cluster-index:1.4.0-unstable` will be able to be upgraded automatically when the index image is polled.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/build-index-image.sh
+++ b/hack/build-index-image.sh
@@ -40,6 +40,11 @@ function create_index_image() {
 
   docker build -t "${BUNDLE_IMAGE_NAME}" -f bundle.Dockerfile --build-arg "VERSION=${CURRENT_VERSION}" .
   docker push "${BUNDLE_IMAGE_NAME}"
+
+  # Extract the digest of the bundle image, to be added to the index image
+  DIGEST=$("${PROJECT_ROOT}/tools/digester/digester" --image "${BUNDLE_IMAGE_NAME}" -d)
+  BUNDLE_IMAGE_NAME="${IMAGE_REGISTRY}/${REGISTRY_NAMESPACE}/${BUNDLE_REGISTRY_IMAGE_NAME}@sha256:${DIGEST}"
+
   # shellcheck disable=SC2086
   ${OPM} index add --bundles "${BUNDLE_IMAGE_NAME}" ${INDEX_IMAGE_PARAM} --tag "${INDEX_IMAGE_NAME}" -u docker
   docker push "${INDEX_IMAGE_NAME}"


hack/build-index-image.sh
```diff
@@ -40,6 +40,11 @@ function create_index_image() {
 
   docker build -t "${BUNDLE_IMAGE_NAME}" -f bundle.Dockerfile --build-arg "VERSION=${CURRENT_VERSION}" .
   docker push "${BUNDLE_IMAGE_NAME}"
+
+  # Extract the digest of the bundle image, to be added to the index image
+  DIGEST=$(skopeo inspect docker://${BUNDLE_IMAGE_NAME} | jq .Digest | tr -d '"')
```
nunnatsa, line 45: A silly comment: first, the digester can do that, using the `--image` flag. But also, the `docker push` command also output the digest, so maybe it's possibel to capture it nin line 42?

orenc1: Thanks for the input.
I've changed it to use the digester tool. Prefer it over trying to capture the digest in `docker push` output.

hack/build-index-image.sh
```diff
@@ -40,6 +40,11 @@ function create_index_image() {
 
   docker build -t "${BUNDLE_IMAGE_NAME}" -f bundle.Dockerfile --build-arg "VERSION=${CURRENT_VERSION}" .
   docker push "${BUNDLE_IMAGE_NAME}"
+
+  # Extract the digest of the bundle image, to be added to the index image
+  DIGEST=$("${PROJECT_ROOT}/tools/digester/digester" --image "${BUNDLE_IMAGE_NAME}" -d)
+  BUNDLE_IMAGE_NAME="${IMAGE_REGISTRY}/${REGISTRY_NAMESPACE}/${BUNDLE_REGISTRY_IMAGE_NAME}@sha256:${DIGEST}"
```
nunnatsa, line 46: Isn't it clearer like this?

```bash
DIGEST=$("${PROJECT_ROOT}/tools/digester/digester" --image "${BUNDLE_IMAGE_NAME}")
BUNDLE_IMAGE_NAME="${IMAGE_REGISTRY}/${REGISTRY_NAMESPACE}/${BUNDLE_REGISTRY_IMAGE_NAME}${DIGEST}"
```

nunnatsa (approved):
Tiny comment fornext time.

/approve

hack/build-index-image.sh
orenc1: It returns the full pullspec of the image, so that would be enough:
```
BUNDLE_IMAGE_NAME=$("${PROJECT_ROOT}/tools/digester/digester" --image "${BUNDLE_IMAGE_NAME}")
```

PR #1114, opened 2021-02-01
Refactor ReconcileHyperConverged.Reconcile()

zcahana:
Fixes a code smell (high cognitive complexity) reported by SonarCloud.

This is a another attempt to refactor this method after c953851 in #1104 resulted in unexplained test failures (timeouts in deleting secondary objects). This time I'll apply the refactoring gradually until the test failures are reproduced.

FOOTNOTE:
Yes, these refactoring iterations are supposed to be done locally accompanied by frequent unit testing executions. However, these particular failures aren't caught by the unit tests - only by the heavyweight CI tests.

**Release note**:
-->
```release-note
NONE
```

### DIFF
--- a/pkg/controller/hyperconverged/hyperconverged_controller.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller.go
@@ -191,46 +191,40 @@ func (r *ReconcileHyperConverged) Reconcile(ctx context.Context, request reconci
 		return reconcile.Result{}, err
 	}
 
-	resolvedRequest := request
+	resolvedRequest, err := resolveReconcileRequest(request, hcoTriggered)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	hcoRequest := common.NewHcoRequest(ctx, resolvedRequest, log, r.upgradeMode, hcoTriggered)
+
 	if hcoTriggered {
 		logger.Info("Reconciling HyperConverged operator")
-
 		r.operandHandler.Reset()
 	} else {
 		logger.Info("The reconciliation got triggered by a secondary CR object")
-
-		hco, err := getHyperconverged()
-		if err != nil {
-			return reconcile.Result{}, err
-		}
-		resolvedRequest = reconcile.Request{
-			NamespacedName: hco,
-		}
 	}
 
-	req := common.NewHcoRequest(ctx, resolvedRequest, log, r.upgradeMode, hcoTriggered)
-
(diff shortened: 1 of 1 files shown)


pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -285,6 +283,16 @@ func (r *ReconcileHyperConverged) Reconcile(ctx context.Context, request reconci
...
 }
 
+func isTriggeredByHyperConverged(request reconcile.Request) (bool, error) {
+	placeholder, err := getSecondaryCRPlaceholder()
+	if err != nil {
+		return false, err
+	}
+
+	isSecondary := request.NamespacedName == placeholder
+	return !isSecondary, nil
```
nunnatsa, line 293: Can we return the actual boolean value instead of its opposite value? I find it a bit confusing. Something like:

```golang
        isHyperConvergedChange := request.NamespacedName != placeholder
	return isHyperConvergedChange, nil
```

zcahana: We could, it's just a trade of where we put the negation. I thought it's clearer as-is, but don't mind changing.

zcahana: Changed.

zcahana:
For what it's worth, the problem in c953851 from #1104, was that the `err` returned from [`r.doReconcile()`](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/c9538514855f07943c0197ff4536b2d586987f32/pkg/controller/hyperconverged/hyperconverged_controller.go#L222) was [overwritten](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/c9538514855f07943c0197ff4536b2d586987f32/pkg/controller/hyperconverged/hyperconverged_controller.go#L231) with `nil`, therefore the [`if apierrors.IsConflict(err)`](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/c9538514855f07943c0197ff4536b2d586987f32/pkg/controller/hyperconverged/hyperconverged_controller.go#L233) check later yielded false and therefore the request was not requeued.

This is now fixed in this PR, by checking the `err` right after the return from `doReconcile()`.

@nunnatsa this is ready for review now.

zcahana:
@override-bot

nunnatsa (approved):
/approve

PR #1120, opened 2021-02-02
Move cmds' common code to a new package

nunnatsa:
In order to reduce code duplication, added the new `cmdcommon` package that contains code that was common to the operator and the webhook commands.

`cmdcommon.HcCmdHelper` is the type that holds the common methods.

**Release note**:

```release-note
NONE
```


orenc1 (approved):
/approve

PR #1125, opened 2021-02-04
Refactor controller registration code to reduce boilerplating

thetechnick:
When adding a new dependency to the HyperConverged Reconciler, you need to add that dependency in 4(!) places.
- `newReconciler()`
- `Add()`
- `AddToManager()`
- `AddToManagerFuncs` type definition

Because these functions are spread across 2 different packages, it prevents defining interfaces on the consumer side (or you have to define them twice).
Also if we ever have another controller in our operator it will get the same dependencies whether it actually uses it or not.

This is leftover boilerplating from operator-sdk (or kubebuilder 1.x) style bootstraping, something that was revamped completely for kubebuilder 2.x.

Signed-off-by: Nico Schieder <nschieder@redhat.com>

**Release note**:

```release-note
NONE
```


erkanerol:
I am familiar with the existent AddToManager approach. It allows you to add new controllers without touching the consumer side. It allows you to ship a couple of controllers as a bundle. I would keep it as it is since it doesn't annoy me :D 

However, I have not a strong opinion. @tiraboschi  or @nunnatsa may have.

thetechnick:
@erkanerol It's definitely nothing big, I just think it complicates our code for no reason.
I am not sure if I would ship the controllers separately. IMHO that belongs into an`/internal` package.

Only thing that should be imported into other projects from this one is `pkg/apis`, although I don't think we have any consumers of our repo atm.

If we add more controllers or have any benefit of this more dynamic registration process via `init()`, just ping me and I am happy to add it back in. :)

Here is an example of somewhat recent kubebuilder boilerplate (for reference):

`// main.go` Kubebuilder
```golang
	if err = (&controllers.ZoneReconciler{
		Client: mgr.GetClient(),
		Log:    ctrl.Log.WithName("controllers").WithName("Zone"),
	}).SetupWithManager(mgr); err != nil {
		setupLog.Error(err, "unable to create controller", "controller", "Zone")
		os.Exit(1)
	}
```

erkanerol:
@thetechnick As I said, I hadn't a strong opinion. Now, you convinced me :) I am putting LGTM and waiting for others' responses.

/lgtm

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -79,6 +79,11 @@ func Add(mgr manager.Manager, ci hcoutil.ClusterInfo) error {
 	return add(mgr, newReconciler(mgr, ci), ci)
 }
 
+// RegisterReconciler creates a new HyperConverged Reconciler and registers it into manager.
+func RegisterReconciler(mgr manager.Manager, ci hcoutil.ClusterInfo) error {
```
nunnatsa, line 83: This is exactly the same as `Add(mgr manager.Manager, ci hcoutil.ClusterInfo) error` - same signature and same implementation. Why do we need both?

nunnatsa: Beside that, I never understood the original code, so this is simpler and clearer :+1:

thetechnick: Mistake from me, `Add` can be removed completely.

We can also think about:
- constructor function vs. direct instantiation e.g. what we have now with `RegisterReconciler` vs. `(&ReconcileHyperConverged{...}).SetupWithManager` (kubebuilder style)
- flatten the package structure in `pkg/controller`

nunnatsa (approved):
/approve

PR #1130, opened 2021-02-07
Bump CDI to version v1.30.0

nunnatsa:
This PR bumps the CDI version to v1.30.0.

It also changes the CDI images location from docker.io to quay.io

In addition, this PR fixes a bug in the `digester`, where from time to time, it prints additional charachters at the end of the file. The problem was that the file was created on top of an existing file, but wasn't truncated before writing. In cases where the new file content is shorter than the existing file, there were some leftovers. The fix was to add the `os.O_TRUNC` flag when calling to the os.Open function, to truncate the file.

**Release note**:

```release-note
* Bupm CDI to version v1.30.0
*  changes the CDI images location from docker.io to quay.io
```


orenc1 (approved):
/override-bot
/lgtm

PR #1136, opened 2021-02-11
Remove BareMetalPlatform from spec

arnongilboa:
Signed-off-by: Arnon Gilboa <agilboa@redhat.com>

Currently this [spec field](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/docs/kni.md#config-baremetal-platform) only controls the kubevirt-storage-class-defaults configmap, and intentionally does NOT reconcile it (to let customers modify the configmap after installation). The name of the field is also opaque and unclear.

**Release note**:
```release-note
Deprecate BareMetalPlatform from spec
```


deploy/kustomize/base/kustomization.yaml
```diff
@@ -4,4 +4,3 @@ kind: Kustomization
 resources:
   - operator_group.yaml
   - subscription.yaml
-  - hco_cr.yaml
```
orenc1, line 7: the HCO CR is still needed to complete the deployment, also if it has an empty spec.

orenc1 (changes requested):

deploy/kustomize/base/kustomization.yaml
arnongilboa: fixed

orenc1:
kubevirt-ci failed because of cert-manager unknown certificate authority issue:
```
Error from server (InternalError): error when creating "_out/webhooks.yaml": Internal error occurred: failed calling webhook "webhook.cert-manager.io": Post https://cert-manager-webhook.cert-manager.svc:443/mutate?timeout=10s: x509: certificate signed by unknown authority
```
Let's see if that error is consistent:
/test pull-hyperconverged-cluster-operator-e2e-k8s-1.17

orenc1:
/lgtm
on my side.
the only downside for removing completely the `bareMetalPlatform` from HCO's CRD, is when people will use old HCO CR manifest containing that field, using `kubectl`or `oc` with client-side validation enabled (`--validate=true`) [default for `oc` is without client side schema validation], which in that case they'll get:
```
error validating data: ValidationError(HyperConverged.spec): unknown field "bareMetalPlatform" in io.kubevirt.hco.v1beta1.HyperConverged.spec; if you choose to ignore these errors, turn validation off with --validate=false
```

in any case, unknown fields got pruned before persisting the CR to etcd.
https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#field-pruning

/cc @tiraboschi

tiraboschi (approved):

orenc1 (approved):

PR #1138, opened 2021-02-16
Fix a bug where HCO failed to trigger some events

nunnatsa:
This PR was originally about increasing the unit test coverage, but I found and fix a bug diring this work.

The bug was that the HCO failed to trigger delete events for the removed operand resources.

This PR also improves the test utilities:

## Change the fake client to produce error using functions, in order to allow checking specific errors.

The fake client now support the following types and methods:
```golang
type FakeWriteErrorGenerator func(obj client.Object) error
type FakeReadErrorGenerator func(key client.ObjectKey) error

InitiateDeleteErrors(f FakeWriteErrorGenerator)
InitiateUpdateErrors(f FakeWriteErrorGenerator)
InitiateGetErrors(f FakeReadErrorGenerator)
InitiateCreateErrors(f FakeWriteErrorGenerator)
```

Example usage:
```golang
cl := expected.initClient()
cl.InitiateUpdateErrors(func(obj client.Object) error {
	if _, ok := obj.(*kubevirtv1.KubeVirt); ok {
		return errors.New("fake update error")
	}
	return nil
})
```

## Add the ability to check if events are emitted:
This PR introduces the `commonTestUtils.MockEvent` type.

Use the new commonTestUtils.NewEventEmitterMock method to create a fake event emitter. This will initiate a list of events.

Clear the received events to prepare to a new test case by calling to the new `EventEmitterMock.Reset()` method.

To validate if the expected events were actually emitted, use the new `EventEmitterMock.CheckEvents(expectedEvents []MockEvent) bool` method.

## Example
```golang
eventEmitter := commonTestUtils.NewEventEmitterMock()
handler := NewOperandHandler(cli, commonTestUtils.GetScheme(), true, eventEmitter)

...

fakeError := fmt.Errorf("fake KV deletion error")
cli.InitiateDeleteErrors(func(obj client.Object) error {
	if unstructed, ok := obj.(runtime.Unstructured); ok {
		kind := unstructed.GetObjectKind()
		if kind.GroupVersionKind().Kind == "KubeVirt" {
			return fakeError
		}
	}
	return nil
})

eventEmitter.Reset()

err = handler.EnsureDeleted(req)
Expect(err).Should(Equal(fakeError))

By("Check that event was emitted", func() {
	expectedEvents := []commonTestUtils.MockEvent{
		{
			EventType: corev1.EventTypeWarning,
			Reason:    ErrVirtUninstall,
			Msg:       uninstallVirtErrorMsg + fakeError.Error(),
		},
	},
	Expect(eventEmitter.CheckEvents(expectedEvents)).To(BeTrue())
})

```

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/commonTestUtils/eventEmitterMock.go
+++ b/pkg/controller/commonTestUtils/eventEmitterMock.go
@@ -7,18 +7,57 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sync"
 )
 
-type EventEmitterMock struct{}
+type MockEvent struct {
+	EventType string
+	Reason    string
+	Msg       string
+}
+
+type EventEmitterMock struct {
+	expectedEvents *sync.Map
+}
+
+func NewEventEmitterMock(expectedEvents ...MockEvent) *EventEmitterMock {
+	eem := &EventEmitterMock{}
+	eem.SetExpectedEvents(expectedEvents...)
+
+	return eem
+}
+
+func (eem *EventEmitterMock) SetExpectedEvents(expectedEvents ...MockEvent) {
+	events := &sync.Map{}
+	for _, event := range expectedEvents {
+		events.Store(event, false)
+	}
+
+	eem.expectedEvents = events
+}
 
 func (EventEmitterMock) Init(_ context.Context, _ manager.Manager, _ hcoutil.ClusterInfo, _ logr.Logger) {
 	/* not implemented; mock only */
 }
 
-func (EventEmitterMock) EmitEvent(_ runtime.Object, _, _, _ string) {
-	/* not implemented; mock only */
+func (eem *EventEmitterMock) EmitEvent(_ runtime.Object, eventType, reason, msg string) {
+	eem.expectedEvents.Store(MockEvent{
+		EventType: eventType,
+		Reason:    reason,
+		Msg:       msg,
+	}, true)
 }
 
 func (EventEmitterMock) UpdateClient(_ context.Context, _ client.Reader, _ logr.Logger) {
 	/* not implemented; mock only */
+}
+
+func (eem EventEmitterMock) GotExpectedEvent() bool {
+	gotAllEvents := true
+	eem.expectedEvents.Range(func(key, value interface{}) bool {
+		gotAllEvents = gotAllEvents && value.(bool)
+		return value.(bool)
+	})
+
+	return gotAllEvents
 }
--- a/pkg/controller/commonTestUtils/eventEmitterMock.go
+++ b/pkg/controller/commonTestUtils/eventEmitterMock.go
@@ -17,23 +17,22 @@ type MockEvent struct {
 }
 
 type EventEmitterMock struct {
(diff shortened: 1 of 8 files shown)


pkg/controller/commonTestUtils/eventEmitterMock.go
```diff
@@ -7,18 +7,57 @@ import (
...
+		Reason:    reason,
+		Msg:       msg,
+	}, true)
 }
 
 func (EventEmitterMock) UpdateClient(_ context.Context, _ client.Reader, _ logr.Logger) {
 	/* not implemented; mock only */
 }
+
+func (eem EventEmitterMock) GotExpectedEvent() bool {
```
erkanerol, line 55: ---This comment is not critical. Just an improvement idea. Current form is already OK. ---

- Instead of `expectedEvents` map, we can initialize it with an empty map named `eventStore`.
- Instead of setting map value in `EmitEvent`, we can insert the event into the `eventStore`
- Instead of `GotExpectedEvent`, we can implement two methods
   - ContainEvents (events []MockEvent) bool -> Checks the events exist in the `eventStore` or not. It can be useful when you want to check specific events but you don't care other emitted events in your test. It can be useful for negative checks as well like you don't expect having a specific event in the emitter.
   - EqualsTo(events []MockEvent) -> Does  equality check. More events in the emitter is not accepted. It can be useful when you want to check all events. No more no less.

pkg/controller/operands/operandHandler.go
```diff
@@ -178,7 +179,7 @@ func (h OperandHandler) EnsureDeleted(req *common.HcoRequest) error {
 				errorCh <- err
 			} else {
 				key := client.ObjectKeyFromObject(o)
-				if err := h.client.Get(tCtx, key, o); err == nil {
+				if err := h.client.Get(tCtx, key, o); apierrors.IsNotFound(err) {
```
erkanerol, line 182: Adding a test case with `FakeReadErrorGenerator` + IsNotFound error may be useful.

nunnatsa: I tend to thik that this check is wrong. This will work with the fake client, but in the eventual consistance system like K8s, we can't prommise that the resource was already removed. I'll remove this check.. @tiraboschi - what do you think?

tiraboschi: hcoutil.EnsureDeleted implements a wait parameter and on line 164 we are calling it with wait=true.
In that case, we are also waiting for the resource to be really disappeared and not simply flagged for deletion so I think that the check is valid.

nunnatsa: This is checked in `util.validateDeletion` that called from `util.ComponentResourceRemoval`. At this place, the resource is deleted for sure

tiraboschi: yes, right, and if util.validateDeletion failed to delete withing the allowed time, it should return an error that is propagated and then returned from hcoutil.EnsureDeleted so for sure we are not in this branch

nunnatsa: Removed. No need to test it.

pkg/controller/commonTestUtils/eventEmitterMock.go
nunnatsa: Done

PR #1139, opened 2021-02-18
Fix  --csv-overrides flag

erkanerol:

While refactoring the codes for Sonar, we broke this part. Here is the fix. 

**Release note**:
```release-note
NONE
```

### DIFF
--- a/hack/build-manifests.sh
+++ b/hack/build-manifests.sh
@@ -249,21 +249,14 @@ hppCsv="${TEMPDIR}/${hhpFile}.${CSV_EXT}"
 vmImportFile=$(create_vm_import_csv)
 importCsv="${TEMPDIR}/${vmImportFile}.${CSV_EXT}"
 csvOverrides="${TEMPDIR}/csv_overrides.${CSV_EXT}"
-
+keywords="  keywords:
+  - KubeVirt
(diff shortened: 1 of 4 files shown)


tiraboschi (commented):
Can we please also add some kind of test/check in hack/build-manifests.sh to prevent future regressions?

tiraboschi (approved):

hack/build-manifests.sh
```diff
@@ -390,6 +383,11 @@ ${PROJECT_ROOT}/tools/csv-merger/csv-merger \
   --operator-image-name="${HCO_OPERATOR_IMAGE}" \
   --webhook-image-name="${HCO_WEBHOOK_IMAGE}" > "${CSV_DIR}/${OPERATOR_NAME}.v${CSV_VERSION}.${CSV_EXT}"
 
+rendered_csv="$(cat "${CSV_DIR}/${OPERATOR_NAME}.v${CSV_VERSION}.${CSV_EXT}")"
+rendered_keywords="$(echo "$rendered_csv" |grep 'keywords' -A 3)"
```
orenc1, line 387: I think `-A 3` argument on grep is not robust enough, as it will break if the number of keywords will be changed.

PR #1143, opened 2021-02-18
Add OvS Opt-In and jsonpatch annotations documentation

orenc1:
Documenting how users can consume the ovs opt-in and jsonpatch annotations feature.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/docs/cluster-configuration.md
+++ b/docs/cluster-configuration.md
@@ -152,14 +152,15 @@ spec:
 ```
 
 ## Configurations via Annotations
-In addition to `featureGates` field in HCO CR's spec, the user can set annotations in the HCO CR to unfold more configuration options.
+In addition to `featureGates` field in HyperConverged CR's spec, the user can set annotations in the HyperConverged CR to unfold more configuration options.  
+**Warning:** Annotations are less formal means of cluster configuration and may be dropped without the same deprecation process of a regular API, such as in the `spec` section.
 
 ### OvS Opt-In Annotation
 Starting from HCO version 1.3.0, OvS CNI support is disabled by default on new installations.  
-In order to enable the deployment of OvS CNI DaemonSet on all _workload_ nodes, an annotation of `deployOVS: true` must be set on HCO CR.  
-It can be set while creating the HCO custom resource during the initial deployment, or during run time.
-* To enable OvS CNI on the cluster, the HCO CR should be similar to:  
-```
+In order to enable the deployment of OvS CNI DaemonSet on all _workload_ nodes, an annotation of `deployOVS: true` must be set on HyperConverged CR.  
+It can be set while creating the HyperConverged custom resource during the initial deployment, or during run time.
+* To enable OvS CNI on the cluster, the HyperConverged CR should be similar to:  
+```yaml
 apiVersion: hco.kubevirt.io/v1beta1
 kind: HyperConverged
 metadata:
@@ -172,11 +173,11 @@ metadata:
 kubectl annotate HyperConverged kubevirt-hyperconverged -n kubevirt-hyperconverged deployOVS=true --overwrite
 ```
 
-If a HCO was upgraded to 1.3.0 from a previous version, the annotation will be added as `true` and OvS will be deployed.  
+If HCO was upgraded to 1.3.0 from a previous version, the annotation will be added as `true` and OvS will be deployed.  
 Subsequent upgrades to newer versions will preserve the state from previous version, i.e. OvS will be deployed in the upgraded version if and only if it was deployed in the previous one.
 
 ### jsonpatch Annotations
-HCO enables users to modify the operand CRs directly using jsonpatch annotations in HCO CR.  
+HCO enables users to modify the operand CRs directly using jsonpatch annotations in HyperConverged CR.  
 Modifications done to CRs using jsonpatch annotations won't be reconciled back by HCO to the opinionated defaults.  
 The following annotations are supported in the HyperConverged CR:
 * `kubevirt.kubevirt.io/jsonpatch` - for KubeVirt configurations
@@ -188,7 +189,7 @@ The patch’s path is relative to the `spec` field in each CR.
 
 #### Examples
 * The user wants to set the KubeVirt CR’s `spec.configuration.migrations.allowPostCopy` field to `true`. In order to do that, the following annotation should be added to the HyperConverged CR:
-```
+```yaml
 metadata:
   annotations:
     kubevirt.kubevirt.io/jsonpatch: |-
@@ -201,7 +202,7 @@ metadata:
       ]
 ```
 * The user wants to override the default URL used when uploading to a DataVolume, by setting the CDI CR's `spec.config.uploadProxyURLOverride` to `myproxy.example.com`. In order to do that, the following annotation should be added to the HyperConverged CR:
-```
+```yaml
 metadata:
   annotations:
     containerizeddataimporter.kubevirt.io/jsonpatch: |-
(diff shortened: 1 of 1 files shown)


docs/cluster-configuration.md
```diff
@@ -150,3 +150,70 @@ spec:
     withHostModelCPU: true
     withHostPassthroughCPU: true
 ```
+
+## Configurations via Annotations
+In addition to `featureGates` field in HCO CR's spec, the user can set annotations in the HCO CR to unfold more configuration options.
```
dankenigsberg, line 155: You should warn the reader that annotations are less formal and may be dropped without the same deprecation process of normal API

docs/cluster-configuration.md
```diff
@@ -150,3 +150,70 @@ spec:
...
+        {
+          "op": "add",
+          "path": "/config/uploadProxyURLOverride",
+          "value": "myproxy.example.com"
+        }
+      ]
+```
+
+### WARNING
+Using the jsonpatch annotation feature incorrectly might lead to unexpected results and could potentially render Kubevirt-Hyperconverged system unstable.
```
dankenigsberg, line 218: Please warn that this is particularly dangerous on upgrade, when the structure or semantics of the underlying cr changes

docs/cluster-configuration.md
```diff
@@ -150,3 +150,70 @@ spec:
...
+If a HCO was upgraded to 1.3.0 from a previous version, the annotation will be added as `true` and OvS will be deployed.  
+Subsequent upgrades to newer versions will preserve the state from previous version, i.e. OvS will be deployed in the upgraded version if and only if it was deployed in the previous one.
+
+### jsonpatch Annotations
+HCO enables users to modify the operand CRs directly using jsonpatch annotations in HCO CR.  
+Modifications done to CRs using jsonpatch annotations won't be reconciled back by HCO to the opinionated defaults.  
+The following annotations are supported in the HyperConverged CR:
+* `kubevirt.kubevirt.io/jsonpatch` - for KubeVirt configurations
+* `containerizeddataimporter.kubevirt.io/jsonpatch` - for CDI configurations
+* `networkaddonsconfig.kubevirt.io/jsonpatch` - for CNAO configurations
```
dankenigsberg, line 184: Can you add links to where the cnao/kv/cdi crs are defined?

dankenigsberg (changes requested):

docs/cluster-configuration.md
orenc1: By "defined" do you mean where their schemas are defined?
If so, the types are in their respective repos HCO is vendoring.
[kubevirt](https://github.com/kubevirt/kubevirt/blob/master/staging/src/kubevirt.io/client-go/api/v1/types.go#L1264)
[CDI](https://github.com/kubevirt/containerized-data-importer/blob/main/pkg/apis/core/v1beta1/types.go#L323)
[CNAO](https://github.com/kubevirt/cluster-network-addons-operator/blob/master/pkg/apis/networkaddonsoperator/v1/networkaddonsconfig_types.go#L15)

Is that what you meant?

dankenigsberg: I'm looking for human-readable explanation/sample for each of the three RCs.

Lacking that, please file an issue on the relevant components, and link to this machine-readable schema.

tiraboschi: The CRDs for components operators are rendered also inside this repo and they should already contain up to date openAPIv3 documentation (see https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/deploy/crds/cluster-network-addons00.crd.yaml as an example).

I think that the best option for us is using some to tool to extract the documentation of each component CR from its openAPIv3 schema in its CRD and convert it to markdown to get something human readable.

tiraboschi: The operator SDK currently doesn't provide this feature (see: https://github.com/operator-framework/operator-sdk/issues/2804 ), waiting to get it implemented in controller-runtime ( https://github.com/kubernetes-sigs/controller-tools/issues/240 ).
The most promising option are https://github.com/prometheus-operator/prometheus-operator/tree/master/cmd/po-docgen or https://github.com/kubernetes/kubernetes/blob/master/cmd/genkubedocs/gen_kube_docs.go

orenc1: The schemas of the CRDs of these components are subject for changes for time to time.
To avoid having continuously maintaining these changes, we can suggest the user to use `kubectl explain <resource name>` to explore what configuration options are available for a given CR, which a human readable description and a type. That way we are agnostic of changes and even which version was deployed on the cluster.
e.g.
```
$ oc explain kv.spec
KIND:     KubeVirt
VERSION:  kubevirt.io/v1

RESOURCE: spec <Object>

DESCRIPTION:
     <empty>

FIELDS:
   certificateRotateStrategy	<Object>

   configuration	<Object>
     holds kubevirt configurations. same as the virt-configMap

   customizeComponents	<Object>

   imagePullPolicy	<string>
     The ImagePullPolicy to use.

   imageRegistry	<string>
     The image registry to pull the container images from Defaults to the same
     registry the operator's container image is pulled from.

   imageTag	<string>
     The image tag to use for the continer images installed. Defaults to the
     same tag as the operator's container image.

   infra	<Object>
     selectors and tolerations that should apply to KubeVirt infrastructure
     components

   monitorAccount	<string>
     The name of the Prometheus service account that needs read-access to
     KubeVirt endpoints Defaults to prometheus-k8s

   monitorNamespace	<string>
     The namespace Prometheus is deployed in Defaults to openshift-monitor

   productName	<string>
     Designate the apps.kubevirt.io/part-of label for KubeVirt components.
     Useful if KubeVirt is included as part of a product. If ProductName is not
     specified, the part-of label will be omitted.

   productVersion	<string>
     Designate the apps.kubevirt.io/version label for KubeVirt components.
     Useful if KubeVirt is included as part of a product. If ProductVersion is
     not specified, KubeVirt's version will be used.

   uninstallStrategy	<string>
     Specifies if kubevirt can be deleted if workloads are still present. This
     is mainly a precaution to avoid accidental data loss

   workloads	<Object>
     selectors and tolerations that should apply to KubeVirt workloads
```

dankenigsberg: In my humble opinion, each rc should have a web-accessible human-readable explanation with examples, which this doc should point to.
`kubectl explain kv.spec` is not exactly that.

However this sorts not block this pr. I'd ask the relevant projects to document themselves, and maybe this.

docs/cluster-configuration.md
nunnatsa, line 155: Please use the term "HyperConverged CR" instead of "HCO CR"

docs/cluster-configuration.md
```diff
@@ -150,3 +150,70 @@ spec:
     withHostModelCPU: true
     withHostPassthroughCPU: true
 ```
+
+## Configurations via Annotations
+In addition to `featureGates` field in HCO CR's spec, the user can set annotations in the HCO CR to unfold more configuration options.
+
+### OvS Opt-In Annotation
+Starting from HCO version 1.3.0, OvS CNI support is disabled by default on new installations.  
+In order to enable the deployment of OvS CNI DaemonSet on all _workload_ nodes, an annotation of `deployOVS: true` must be set on HCO CR.
```
nunnatsa, line 159: here too

docs/cluster-configuration.md
```diff
@@ -150,3 +150,70 @@ spec:
...
+* OvS CNI can also be enabled during run time of HCO, by annotating its CR:
+```
+kubectl annotate HyperConverged kubevirt-hyperconverged -n kubevirt-hyperconverged deployOVS=true --overwrite
+```
+
+If a HCO was upgraded to 1.3.0 from a previous version, the annotation will be added as `true` and OvS will be deployed.  
+Subsequent upgrades to newer versions will preserve the state from previous version, i.e. OvS will be deployed in the upgraded version if and only if it was deployed in the previous one.
+
+### jsonpatch Annotations
+HCO enables users to modify the operand CRs directly using jsonpatch annotations in HCO CR.
```
nunnatsa, line 179: Here too

nunnatsa (commented):
Semantic comment:
Please use the term "HyperConverged CR" instead of "HCO CR"

docs/cluster-configuration.md
```diff
@@ -150,3 +150,70 @@ spec:
...
 ```
+
+## Configurations via Annotations
+In addition to `featureGates` field in HCO CR's spec, the user can set annotations in the HCO CR to unfold more configuration options.
+
+### OvS Opt-In Annotation
+Starting from HCO version 1.3.0, OvS CNI support is disabled by default on new installations.  
+In order to enable the deployment of OvS CNI DaemonSet on all _workload_ nodes, an annotation of `deployOVS: true` must be set on HCO CR.  
+It can be set while creating the HCO custom resource during the initial deployment, or during run time.
+* To enable OvS CNI on the cluster, the HCO CR should be similar to:
```
nunnatsa, line 161: Here too

docs/cluster-configuration.md
orenc1: added warning.

docs/cluster-configuration.md
orenc1: added, also advised to remove any jsonpatch annotations prior upgrading HCO.

docs/cluster-configuration.md
orenc1: Please see the addition.
It may be not perfect, but at least the users can have a way of exploring the configuration options they have for each CR.

nunnatsa (approved):
/approve

PR #1149, opened 2021-02-23
Add unit tests to check the event emitter

nunnatsa:

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -48,7 +48,10 @@ var ErrNoNamespace = fmt.Errorf("namespace not found for current environment")
 
 // ErrRunLocal indicates that the operator is set to run in local mode (this error
 // is returned by functions that only work on operators running in cluster mode)
-var ErrRunLocal = fmt.Errorf("operator run mode forced to local")
+var (
+	ErrRunLocal          = fmt.Errorf("operator run mode forced to local")
+	GetOperatorNamespace = getOperatorNamespace
+)
 
 func GetOperatorNamespaceFromEnv() (string, error) {
 	if namespace, ok := os.LookupEnv(OperatorNamespaceEnv); ok {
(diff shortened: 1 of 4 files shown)


pkg/util/util.go
```diff
@@ -48,7 +48,10 @@ var ErrNoNamespace = fmt.Errorf("namespace not found for current environment")
 
 // ErrRunLocal indicates that the operator is set to run in local mode (this error
 // is returned by functions that only work on operators running in cluster mode)
-var ErrRunLocal = fmt.Errorf("operator run mode forced to local")
+var (
+	ErrRunLocal          = fmt.Errorf("operator run mode forced to local")
+	GetOperatorNamespace = getOperatorNamespace
```
erkanerol, line 53: A clever way to mock functions :+1:

pkg/util/util.go
```diff
@@ -124,6 +127,9 @@ func GetPod(ctx context.Context, c client.Reader, logger logr.Logger, ci Cluster
 }
 
 func GetCSVfromPod(pod *corev1.Pod, c client.Reader, logger logr.Logger) (*csvv1alpha1.ClusterServiceVersion, error) {
+	if pod == nil {
+		return nil, errors.New("pod is null")
```
erkanerol, line 131: null -> nil

erkanerol:
While reviewing this PR, I noticed the renaming `event_emmiter.go` → `eventEmitter.go `. Then I did some research about file naming conventions in Go. The old one seems better to me after checking these:

https://stackoverflow.com/questions/25161774/what-are-conventions-for-filenames-in-go

Check go files under https://github.com/kubernetes/kubernetes/tree/master/pkg/controller

Check go files under https://github.com/openshift/oc/tree/master/pkg/cli

PR #1151, opened 2021-02-24
Populate all mandatory KV configurations

nunnatsa:
1. Force the SELinuxLauncherType, NetworkInterface, smbios and the kvmEmulation values in KV CR, to a hard-coded value, or an environment variable value.
2. Set the old feature gates from the config map as a hard coded values in KubrVirt CR. These feature gates are not exposed by the API, and HCO does not allow to modified them. The featue gates are: `DataVolumes`, `SRIOV`, `LiveMigration`, `CPUManager`, `CPUNodeDiscovery`, `Sidecar` and `Snapshot`
3. Automatic compute the enabled feature gates using reflection

Still WIP; TODO:
- [x] Add few more tests to check that the reconcile works (there are already tests in the `operands` package).

**Release note**:

```release-note
Will add later
```

### DIFF
--- a/deploy/crds/hco00.crd.yaml
+++ b/deploy/crds/hco00.crd.yaml
@@ -47,10 +47,17 @@ spec:
             properties:
               featureGates:
                 default:
+                  cpuManager: true
+                  cpuNodeDiscovery: true
+                  dataVolumes: true
                   gpu: false
                   hostDevices: false
                   hotplugVolumes: false
                   hypervStrictCheck: true
+                  liveMigration: true
+                  sidecar: true
+                  snapshot: true
+                  sriov: true
                   sriovLiveMigration: false
                   withHostModelCPU: true
                   withHostPassthroughCPU: false
@@ -60,12 +67,22 @@ spec:
                 properties:
                   cpuManager:
                     default: true
+                    description: TODO add description
                     type: boolean
                   cpuNodeDiscovery:
                     default: true
+                    description: TODO add description
                     type: boolean
                   dataVolumes:
                     default: true
+                    description: indicates that we support turning on DataVolume workflows.
+                      This means using DataVolumes in the VM and VMI definitions.
+                      There was a period of time where this was in alpha and needed
+                      to be explicility enabled. It also means that someone is using
+                      KubeVirt with CDI. So by not enabling this feature gate, someone
+                      can safely use kubevirt without CDI and know that users of kubevirt
+                      will not be able to post VM/VMIs that use CDI workflows that
+                      aren't available to them
                     type: boolean
                   gpu:
                     default: false
@@ -86,15 +103,22 @@ spec:
                     type: boolean
                   liveMigration:
                     default: true
+                    description: Enables VMIs to be live migrated. Without this, migrations
+                      are not possible and will be blocked
                     type: boolean
                   sidecar:
                     default: true
+                    description: Enables using our sidecar hooks for injecting custom
+                      logic into the VMI startup flow. This is a very advanced feature
+                      that has security implications, which is why it is opt-in only
                     type: boolean
                   snapshot:
                     default: true
+                    description: Enables the alpha offline snapshot functionality
                     type: boolean
                   sriov:
                     default: true
+                    description: TODO add description
                     type: boolean
                   sriovLiveMigration:
                     default: false
--- a/deploy/crds/hco00.crd.yaml
+++ b/deploy/crds/hco00.crd.yaml
@@ -58,6 +58,15 @@ spec:
                   a flag to `true` will enable the feature. Setting `false` or removing
                   the feature gate, disables the feature.
                 properties:
+                  cpuManager:
+                    default: true
+                    type: boolean
+                  cpuNodeDiscovery:
+                    default: true
+                    type: boolean
+                  dataVolumes:
+                    default: true
+                    type: boolean
                   gpu:
                     default: false
                     description: Allow assigning GPU and vGPU devices to virtual machines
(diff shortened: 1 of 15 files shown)


pkg/apis/hco/v1beta1/hyperconverged_types.go
```diff
@@ -54,79 +55,109 @@ type HyperConvergedConfig struct {
...
 
+type FeatureGate *bool
+
 // HyperConvergedFeatureGates is a set of optional feature gates to enable or disable new features that are not enabled
 // by default yet.
 // +optional
 // +k8s:openapi-gen=true
 // +kubebuilder:default={}
 type HyperConvergedFeatureGates struct {
+	// +optional
```
tiraboschi, line 66: should we also change `+kubebuilder:default={sriovLiveMigration: false, hotplugVolumes: false, gpu: false, hostDevices: false, withHostPassthroughCPU: false, withHostModelCPU: true, hypervStrictCheck: true}` on line 44?

nunnatsa: you mean `true` for all of them, I guess

tiraboschi: yes, right, true for all the new ones

deploy/crds/hco00.crd.yaml
```diff
@@ -75,6 +101,25 @@ spec:
...
+                  liveMigration:
+                    default: true
+                    description: Enables VMIs to be live migrated. Without this, migrations
+                      are not possible and will be blocked
+                    type: boolean
+                  sidecar:
+                    default: true
+                    description: Enables using our sidecar hooks for injecting custom
+                      logic into the VMI startup flow. This is a very advanced feature
+                      that has security implications, which is why it is opt-in only
```
dankenigsberg, line 113: you say that it is opt-in, but the default is set to `true`. It does not make sense to me.
Can the description add a URL to an example of who this can be used?

deploy/crds/hco00.crd.yaml
```diff
@@ -75,6 +101,25 @@ spec:
...
+                    type: boolean
+                  snapshot:
+                    default: true
+                    description: Enables the alpha offline snapshot functionality
```
dankenigsberg, line 117: why do we want to expose this? Can a developer harm anything but their own data if they use this feature?

dankenigsberg (changes requested):
(partial review)

deploy/index-image/community-kubevirt-hyperconverged/1.4.0/hco00.crd.yaml
```diff
@@ -75,6 +101,25 @@ spec:
                     description: Enable HyperV strict host checking for HyperV enlightenments
                       Defaults to true, even when HyperConvergedFeatureGates is empty
                     type: boolean
+                  liveMigration:
+                    default: true
+                    description: Enables VMIs to be live migrated. Without this, migrations
+                      are not possible and will be blocked
```
dankenigsberg, line 107: Why would I, as sysadmin, like to disable `liveMigration`? I believe that this should be hard-coded to `true`.

dankenigsberg (changes requested):

deploy/crds/hco00.crd.yaml
tiraboschi: This one was explicitly enabled as for https://bugzilla.redhat.com/1745328 since 2.1.0.
I think that in this PR we should keep all the values exactly as they were in the past, then we should use a separate PR with a clear explanation each time we want to change a value and so the expected behaviour of the product.

dankenigsberg: 1. this PR should not be self-contradictory. It says that `sidecar` is opt-in, but does the opposite. This does not stand.
2. the BZ says nothing about the correct default value for this.
3. HCO never provided this API. Since it is so dangerous, I'm not at all convinced that it needs to be exposed at all. The `kubevirtconfg` annotation may be enough for this.
4. I'm OK with hard-coding this to `true`, but I prefer to have a follow-up commit fixing it to `false` included in this very PR. We don't want to expose our users to this any further moment.

docs/cluster-configuration.md
```diff
@@ -150,7 +152,7 @@ kind: HyperConverged
 metadata:
   name: kubevirt-hyperconverged
 spec:
-  infra: {}
+  infra: { }
```
erkanerol, line 155: Is this on purpose?

pkg/apis/hco/v1beta1/hyperconverged_types.go
```diff
@@ -66,67 +69,73 @@ type HyperConvergedFeatureGates struct {
...
-	return (fgs != nil) && (fgs.HotplugVolumes != nil) && (*fgs.HotplugVolumes)
+	// +optional
+	// +k8s:openapi-gen=false
+	enabledGateList map[string]bool `json:"-"`
 }
 
-func (fgs *HyperConvergedFeatureGates) IsGPUAssignmentEnabled() bool {
-	return (fgs != nil) && (fgs.GPU != nil) && (*fgs.GPU)
-}
+func (fgs *HyperConvergedFeatureGates) RebuildEnabledGateMap() map[string]bool {
```
erkanerol, line 111: Here we are caching values of other public fields in a private field. As far as I see, we have to call this method in some places to invalidate the cache. I think we should not do it. Cache invalidation is one of the hardest things in software development :) 

Also, I didn't get what we gain with this change. If code duplication is the matter, I think we can improve it with some helper functions.

pkg/apis/hco/v1beta1/hyperconverged_types_test.go
```diff
@@ -301,200 +301,89 @@ var _ = Describe("HyperconvergedTypes", func() {
...
+			It("Should return nil if HyperConvergedFeatureGates is nil", func() {
 				var fgs *HyperConvergedFeatureGates = nil
-				Expect(fgs.IsHotplugVolumesEnabled()).To(BeFalse())
+				Expect(fgs.RebuildEnabledGateMap()).To(BeNil())
 			})
 
 			It("Should return false if HotplugVolumes does not exist", func() {
 				fgs := &HyperConvergedFeatureGates{}
-				Expect(fgs.IsHotplugVolumesEnabled()).To(BeFalse())
+				m := fgs.RebuildEnabledGateMap()
```
erkanerol, line 315: You know these are not equal in two ways. We are over-checking here. I think we should go back to the old way.

nunnatsa:
Close in favor of #1156

PR #1152, opened 2021-02-24
Force go 1.15.8 on github action for consistency

tiraboschi:
github action started using go 1.16 by default introducing
inconsistencies in go.sum between go 1.15.z and 1.16.

Forcing go 1.15.8 on github action until we are ready to
completely move to 1.16.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/.github/workflows/pr-sanity.yaml
+++ b/.github/workflows/pr-sanity.yaml
@@ -26,7 +26,10 @@ jobs:
 
       - uses: actions/setup-go@v2
         with:
-          go-version: '^1.15' # The Go version to download (if necessary) and use.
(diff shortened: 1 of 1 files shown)


zcahana:
@tiraboschi while this is a temporary setting, we don't know for how long. So it would be useful to add a oneliner comment to the yaml, to specify why we're forcing this to 1.15.8.

(yes I should have made that remark before /approving, but we still have some time before the tests pass :-) )

tiraboschi:
> @tiraboschi while this is a temporary setting, we don't know for how long. So it would be useful to add a oneliner comment to the yaml, to specify why we're forcing this to 1.15.8.

+1

PR #1158, opened 2021-03-01
Fix csplit delimiter in build-manifests.sh

orenc1:
CRDs might contain the '---' string inside their body, and they will be splitted on that line, resulting in malformed yaml.
This fix will split the CRDs only if a line contain only '---'.

**Release note**:

```release-note

```


nunnatsa (approved):
/approve

PR #1164, opened 2021-03-04
Add Live Migration Configuration

nunnatsa:
* Add a new `LiveMigrationConfig` struct to the `HyperConverged`'s `Spec` field.
* Propagate the live migration configuration to the KubeVirt CR.
* Change the `FeatureGates` field to be a struct instead of a pointer to struct, in order to to get the default feature gate values without write them twice.
* Fix bugs in webhook unit test when we go false positive results - some of the test passed but didn't work; e.g. got error, but not the right one

### Live Migration Configurations
Set the live migration configurations by modifying the fields in the `liveMigrationConfig` under the `spec` field

#### bandwidthPerMigration
Bandwidth limit of each migration, in MiB/s. The format is a number and with the `Mi` suffix, e.g. `64Mi`.

**default**: 64Mi

#### completionTimeoutPerGiB
The migration will be canceled if it has not completed in this time, in seconds per GiB of memory. For example, a virtual
machine instance with 6GiB memory will timeout if it has not completed migration in 4800 seconds. If the Migration Method
is BlockMigration, the size of the migrating disks is included in the calculation. The format is a number.

**default**: 800

#### parallelMigrationsPerCluster
Number of migrations running in parallel in the cluster. The format is a number.

**default**: 5

#### parallelOutboundMigrationsPerNode
Maximum number of outbound migrations per node. The format is a number.

**default**: 2

#### progressTimeout:
The migration will be canceled if memory copy fails to make progress in this time, in seconds. The format is a number.

**default**: 150

#### Example
```yaml
kind: HyperConverged
metadata:
  name: kubevirt-hyperconverged
spec:
  liveMigrationConfig:
    bandwidthPerMigration: 64Mi
    completionTimeoutPerGiB: 800
    parallelMigrationsPerCluster: 5
    parallelOutboundMigrationsPerNode: 2
    progressTimeout: 150
```

**Release note**:

```release-note
Add a new LiveMigrationConfig struct to the HyperConverged's Spec field. see doc/api.md
Propagate the live migration configuration to the KubeVirt CR.
Change the FeatureGates field to be a struct instead of a pointer to struct, in order to to get the default feature gate values without write them twice.
```


docs/cluster-configuration.md
```diff
@@ -162,15 +162,75 @@ spec:
     withHostPassthroughCPU: true
 ```
 
+## Live Migration Configurations
```
dankenigsberg, line 165: I'm happy with the text (I did not review the code)

pkg/apis/hco/v1beta1/hyperconverged_types.go
```diff
@@ -39,9 +39,12 @@ type HyperConvergedSpec struct {
 	// featureGates is a map of feature gate flags. Setting a flag to `true` will enable
 	// the feature. Setting `false` or removing the feature gate, disables the feature.
 	// +optional
-	// +TODO: Always keep the default FeatureGates in sync with the default field values in HyperConvergedFeatureGates //NOSONAR
-	// +kubebuilder:default={sriovLiveMigration: false, hotplugVolumes: false, gpu: false, hostDevices: false, withHostPassthroughCPU: false, withHostModelCPU: true, hypervStrictCheck: true}
-	FeatureGates *HyperConvergedFeatureGates `json:"featureGates,omitempty"`
+	FeatureGates HyperConvergedFeatureGates `json:"featureGates,omitempty"`
```
tiraboschi, line 42: Great achievement, I always hated this!

tiraboschi (approved):

tiraboschi:
The reduced coverage is only due to auto generated code in pkg/apis/hco/v1beta1/zz_generated.deepcopy.go
we can ignore it
/override coverage/coveralls

nunnatsa:
Coverage droped by 2% because of addition of a generation code.

/override coverage/coveralls

PR #1165, opened 2021-03-04
Wait to set CDI featuregates until "config authority" annotation is set

mhenriks:
Have to wait for CDI operator to "upgrade" the CDI resource before setting featuregates.

Signed-off-by: Michael Henriksen <mhenriks@redhat.com>

**Release note**:

```release-note
BugFix:  Wait to set CDI featuregates until "config authority" annotation is set
```


pkg/controller/operands/cdi.go
```diff
@@ -81,11 +82,14 @@ func (h *cdiHooks) updateCr(req *common.HcoRequest, Client client.Client, exists
 		return false, false, errors.New("can't convert to CDI")
 	}
 
-	// HCO reconciles the CR for CDI excluding the `spec.CDIConfig`,
```
tiraboschi, line 84: can we please keep this comment until we will properly reconcile the whole spec stanza?

tiraboschi (approved):

tiraboschi:
sonarcloud is reporting a 3.0% Duplicated Lines (%) in the new unit test, we can tolerate it
/approve

pkg/controller/operands/cdi.go
```diff
@@ -167,9 +172,10 @@ func NewCDI(hc *hcov1beta1.HyperConverged, opts ...string) (*cdiv1beta1.CDI, err
 func NewCDIWithNameOnly(hc *hcov1beta1.HyperConverged, opts ...string) *cdiv1beta1.CDI {
 	return &cdiv1beta1.CDI{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      "cdi-" + hc.Name,
-			Labels:    getLabels(hc, hcoutil.AppComponentStorage),
-			Namespace: getNamespace(hcoutil.UndefinedNamespace, opts),
+			Name:        "cdi-" + hc.Name,
+			Labels:      getLabels(hc, hcoutil.AppComponentStorage),
+			Namespace:   getNamespace(hcoutil.UndefinedNamespace, opts),
+			Annotations: map[string]string{cdiConfigAuthorityAnnotation: ""},
```
nunnatsa, line 178: The `NewCDIWithNameOnly` function provides a very limited instace of CDI, just for searching (and also reused in the `NewCDI` function).

@mhenriks - is there a reason why the setting of the annotation is done here and not as part of the `NewCDI` function?

tiraboschi: I approved it because the Annotations are part of the metadata stanza as the name and the labels so I found more coherent setting them all here.
@nunnatsa do you see any potential risk or is it just a matter of consistency?

nunnatsa: Not a risk. 

Just I don't want this method to grow. Moving this code to `NewCDI` will keep the same logic, but when we only need to search for CDI (e.g. for deletion, for update) we don't need anything else. I want to keep the `NewCDIWithNameOnly` as small as possible.

mhenriks: Just seeing this @nunnatsa basically, it would be very bad if a new instance of CDI (created by HCO) does not have this annotation.  So it seemed to fit here IMO as I think the spirit of this function is "minimally required" values

nunnatsa: > Just seeing this @nunnatsa basically, it would be very bad if a new instance of CDI (created by HCO) does not have this annotation. So it seemed to fit here IMO as I think the spirit of this function is "minimally required" values

This method **does not** creates a new instance of CDI that will be used later to create an actual CDI instance, this is a helper function. It also called from the `newCDI` function itself, but from other places too. The place to set any other values beside name and namespace is in `newCDI`.

PR #1167, opened 2021-03-07
Update feature gate APIs

nunnatsa:
1. Remove the `sidecar` feature gate, because it allows VM owners un arbitrary code in virt-launcher context.
2. Remove the `SRIOVLiveMigration` feature gate from the API. This feature is not mature and HCO does not allow enabling it.
3. Remove the `HotplugVolumes`, `GPU` and `HostDevices` featue gates from the API. Instead, HCO always adds them to KubeVirt CR.
4. Remove the `WithHostModelCPU` and `HypervStrictCheck` feature gates fom the API. Insted, HCO adds them if SSP is deployed.

## API Changes
Remove the foloowing Feature Gates from the `spec.featureGates` field:
* `SRIOVLiveMigration`
* `HotplugVolumes`
* `GPU`
* `HostDevices`
* `WithHostModelCPU`
* `HypervStrictCheck`

**Release note**:

```release-note
1. Remove the sidecar feature gate, because it allows VM owners un arbitrary code in virt-launcher context.
2. Remove the SRIOVLiveMigration feature gate from the API. This feature is not mature and HCO does not allow enabling it.
3. Remove the HotplugVolumes, GPU and HostDevices featue gates from the API. Instead, HCO always adds them to KubeVirt CR.
4. Remove the WithHostModelCPU and HypervStrictCheck feature gates fom the API. Insted, HCO adds them if SSP is 
```


dankenigsberg:
>     1. Remove the `sidecar` feature gate, because enabling this feature gate may cause a security hole.

I'd say: Remove the `sidecar` feature gate, because it allows VM owners un arbitrary code in virt-launcher context. "security hole" sounds more frightening than it is

> 
>     2. Remove the `SRIOVLiveMigration` feature gate from the API. This feature is not mature and HCO does not allow enabling it.

I think that we'd have to expose this anyway (it's a cool feature, but it requires extra privs for virt-launcher). But ack, exposing `SRIOVLiveMigration` can certainly wait for another PR.

docs/cluster-configuration.md
```diff
@@ -117,37 +117,17 @@ or new features that are not enabled by default.
...
-### withHostModelCPU Feature Gate
-
-Set the `withHostModelCPU` feature gate in order to enable support migration for VMs with host-model CPU mode
-
-Additional information: [LibvirtXMLCPUModel](https://wiki.openstack.org/wiki/LibvirtXMLCPUModel)
-
 ### withHostPassthroughCPU Feature Gate
-
 Set the `withHostPassthroughCPU` feature gate in order to allow migrating a virtual machine with CPU host-passthrough
 mode.
```
dankenigsberg, line 122:
```suggestion
mode. This can provide slightly better CPU performance, but should be enabled only when the Cluster is homogeneous from CPU HW perspective.
```

dankenigsberg (changes requested):

docs/cluster-configuration.md
dankenigsberg: The features gates we keep should be properly documented: each should have an adequate explanation for why an admin should use them, and when.

nunnatsa: This sentence is already there, a few lines below.

tiraboschi (approved):

PR #1173, opened 2021-03-09
Remove KubeVirt ConfigMap

nunnatsa:
KubeVirt is about to drop the configMap support, and will only be configurable by modifiying the KubeVirt custom resource.

This PR removes the KubeVirt confgMap. It does not create the configMap and does not reconcile it.

Upon upgrading from previous version, HCO now removes the KubeVirt configMap if exists. Before deleting the configMap, HCO adopt the relevant configurations into the HyperConverged custome resource. The only field that actually abopted is the `migration` field that is adopted into the HyperConverged CR's `LifeMigrationConfig` field. In addition, HCO copies the configMap to a new one, with the name of "kubevirt-config-backup".

**Release note**:

```release-note
Remove the kubevirt-config ConfigMap
```


erkanerol:
@nunnatsa  Could you please check this part? https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/tests/func-tests/main_test.go#L54 It may require some changes as well.

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -843,6 +852,145 @@ func (r *ReconcileHyperConverged) recoverHCOVersion(request *common.HcoRequest)
...
+			Namespace: req.Namespace,
+		},
+	}
+
+	if err := hcoutil.GetRuntimeObject(req.Ctx, r.client, cm, req.Logger); err != nil {
+		if apierrors.IsNotFound(err) {
+			req.Logger.Info("KubeVirt configmap already removed")
+			return false, nil
+		}
+		req.Logger.Info("failed to create KubeVirt configmap", "error", err.Error())
```
erkanerol, line 885: "failed to get"

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -843,6 +852,145 @@ func (r *ReconcileHyperConverged) recoverHCOVersion(request *common.HcoRequest)
...
+	hcoLiveMigrationConfig := hcov1beta1.LiveMigrationConfigurations{}
+	err := yaml.NewYAMLOrJSONDecoder(strings.NewReader(kvLiveMigrationConfig), 1024).Decode(&hcoLiveMigrationConfig)
+	if err != nil {
+		req.Logger.Error(err, "Failed to read the KubeVirt ConfigMap, and its content was ignored. This ConfigMap will be deleted. The backup ConfigMap called "+backupKvCmName)
+		return false
+	}
+
+	if !reflect.DeepEqual(req.Instance.Spec.LiveMigrationConfig, hcoLiveMigrationConfig) {
+		req.Logger.Info("updating the HyperConverged CR from the KeubeVirt configMap")
+		kvConfigMapToHyperConvergedCr(req, hcoLiveMigrationConfig)
```
erkanerol, line 967: Why do we need this? Why don't we directly assign
`req.Instance.Spec.LiveMigrationConfig = hcoLiveMigrationConfig`

I am sure I am missing something but I don't know what it is :D

nunnatsa: That allow keeping HCO defaults if the configmap does not include all the keys.

erkanerol: After reading the test cases carefully, I understood it. Thanks!

nunnatsa:
> @nunnatsa Could you please check this part? https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/tests/func-tests/main_test.go#L54 It may require some changes as well.

Removed

pkg/controller/hyperconverged/testUtils_test.go
```diff
@@ -58,9 +58,9 @@ func initReconciler(client client.Client) *ReconcileHyperConverged {
 }
 
 type BasicExpected struct {
-	hco                  *hcov1beta1.HyperConverged
-	pc                   *schedulingv1.PriorityClass
-	kvConfig             *corev1.ConfigMap
+	hco *hcov1beta1.HyperConverged
+	pc  *schedulingv1.PriorityClass
+	//	kvConfig             *corev1.ConfigMap
```
tiraboschi, line 63: can we delete this?

pkg/controller/hyperconverged/testUtils_test.go
```diff
@@ -79,7 +79,7 @@ func (be BasicExpected) toArray() []runtime.Object {
 	return []runtime.Object{
 		be.hco,
 		be.pc,
-		be.kvConfig,
+		//		be.kvConfig,
```
tiraboschi, line 82: same

tiraboschi (approved):

tiraboschi (approved):

PR #1179, opened 2021-03-11
Support Permitted Host Devices

nunnatsa:
Administrators can control which host devices are exposed and permitted to be used in the cluster. Permitted host devices in the cluster will need to be allowlisted in KubeVirt CR by its vendor:product selector for PCI devices or mediated device names.

This PR adds the `permittedHostDevices` field to the HyperConverged `spec` field. HCO propagates this field to the KubeVirt custome resource.

## API Changes
Use the `permittedHostDevices` field in order to manage the permitted host devices.

The `permittedHostDevices` field is an optional field under the HyperConverged `spec` field.

The `permittedHostDevices` field contains two optional arrays: the `pciHostDevices` and the `mediatedDevices` array.

HCO propagates these arrays as is to the KubeVirt custom resource; i.e. no merge is done, but a replacement.

The `pciHostDevices` array is an array of `PciHostDevice` objects. The fields of this object are:
* `pciVendorSelector` - a combination of a **`vendor_id:product_id`** required to identify a PCI device on a host.

   This identifier 10de:1eb8 can be found using `lspci`; for example:
   ```shell
   lspci -nnv | grep -i nvidia
   ```

* `resourceName` - name by which a device is advertised and being requested.
* `externalResourceProvider` - indicates that this resource is being provided by an external device plugin.

  KubeVirt in this case will only permit the usage of this device in the cluster but will leave the allocation and
  monitoring to an external device plugin.

  **default**: `false`

The `mediatedDevices` array is an array of `MediatedDevice` objects. The fields of this object are:
* `mdevNameSelector` - name of a mediated device type required to identify a mediated device on a host.

   For example: mdev type nvidia-226 represents GRID T4-2A.

   The selector is matched against the content of `/sys/class/mdev_bus/$mdevUUID/mdev_type/name`.
* `resourceName` - name by which a device is advertised and being requested.
* `externalResourceProvider` - indicates that this resource is being provided by an external device plugin.

  KubeVirt in this case will only permit the usage of this device in the cluster but will leave the allocation and
  monitoring to an external device plugin.

  **default**: `false`

### Permitted Host Devies Example
```yaml
kind: HyperConverged
metadata:
  name: kubevirt-hyperconverged
spec:
  permittedHostDevices:
    pciHostDevices:
      - pciVendorSelector: "10DE:1EB8"
        resourceName: "nvidia.com/TU104GL_Tesla_T4"
        externalResourceProvider: true
      - pciVendorSelector: "8086:6F54"
        resourceName: "intel.com/qat"
    mediatedDevices:
      - mdevNameSelector: "GRID T4-1Q"
        resourceName: "nvidia.com/GRID_T4-1Q"
```

**Release note**:

```release-note
Support Permitted Host Devices
```


pkg/controller/operands/kubevirt.go
```diff
@@ -254,6 +255,49 @@ func getKVConfig(hc *hcov1beta1.HyperConverged) (*kubevirtv1.KubeVirtConfigurati
 	return config, nil
 }
 
+func toKvPermittedHostDevices(permittedDevices *hcov1beta1.PermittedHostDevices) *kubevirtv1.PermittedHostDevices {
```
thetechnick, line 258: I we just pass through this setting transparently to kubevirt by definition, do we really want to mirror the API type here?

We could also use `kubevirtv1.PermittedHostDevices/MediatedHostDevice` directly, so we just have to bump our kubevirt dependency to add updated documentation/settings from kubevirt instead of having to change code.

If there are ever settings introduced in KubeVirt that we don't want to expose via HCO, we can always replace the kubevirt reference with our own struct as long as the json annotations stay compatible.
Obviously we have to ensure that we don't accidentally release HCO with new options here, but that should be easy to spot in code reviews when we update kubevirt.

thetechnick (changes requested):

pkg/controller/operands/kubevirt.go
nunnatsa: We thought about it, but we don't want to be affected by other compponent when it comes to the API itself, in order to keep HCO API in HCO control; e.g for documentation an so on.

nunnatsa: @tiraboschi  ^^^

thetechnick: nvm. Makes sense, just wanted to check :+1:

thetechnick (approved):

nunnatsa:
New auto generated code is not covered, and that droped the coverage by 3.3%

/override coverage/coveralls

PR #1180, opened 2021-03-11
Add pprof instrumentation for optional cpu and memory profiling

thetechnick:
Adding pprof for optional cpu and memory profiling to debug memory spikes and inform us about needed requests/limits for our operator.

This PR starts a new HTTP server if `HCO_PPROF_ADDR` env var is set. From this server memory allocation, cpu profiles and heap information can be downloaded while the operator is deployed on a cluster [1].

Basic commands to do something
```shell
# port forward to the operator pod (blocking)
$ kubectl port-forward pod/<hco-pod-name-goes-here> 8065:<port from HCO_PPROF_ADDR>

# download and inspect cpu profile
$ go tool pprof http://localhost:8065

# get and inspect memory allocation profile
$ curl -sK -v http://localhost:8065/debug/pprof/allocs > allocs.out
$ go tool pprof allocs.out

# get and inspect heap profile
$ curl -sK -v http://localhost:8065/debug/pprof/heap > heap.out
$ go tool pprof heap.out
```

use `web` in the pprof console to start a new browser tab with a svg diagram.

Ref:
[1] https://golang.org/pkg/net/http/pprof/
https://blog.golang.org/pprof
https://www.freecodecamp.org/news/how-i-investigated-memory-leaks-in-go-using-pprof-on-a-large-codebase-4bec4325e192/

Signed-off-by: Nico Schieder <nschieder@redhat.com>

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cmd/cmdcommon/cmdcommon.go
+++ b/cmd/cmdcommon/cmdcommon.go
@@ -1,8 +1,11 @@
 package cmdcommon
 
 import (
+	"context"
 	"flag"
 	"fmt"
+	"net/http"
+	"net/http/pprof"
 	"os"
 	"runtime"
 
@@ -44,6 +47,48 @@ func (h HcCmdHelper) InitiateCommand() {
 	h.printVersion()
 
 	h.checkNameSpace()
+}
+
+const pprofAddrEnvVar = "HCO_PPROF_ADDR"
+
+// Registers a pprof server for cpu and memory profiling the running operator.
+func (h HcCmdHelper) RegisterPPROFServer(mgr manager.Manager) error {
+	pprofAddr := os.Getenv(pprofAddrEnvVar)
+	if len(pprofAddr) == 0 {
+		return nil
+	}
+
+	h.Logger.Info("Registering pprof server.")
+
+	mux := http.NewServeMux()
+	mux.HandleFunc("/debug/pprof/", pprof.Index)
+	mux.HandleFunc("/debug/pprof/cmdline", pprof.Cmdline)
+	mux.HandleFunc("/debug/pprof/profile", pprof.Profile)
+	mux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
+	mux.HandleFunc("/debug/pprof/trace", pprof.Trace)
+
+	s := &http.Server{Addr: pprofAddr, Handler: mux}
+	return mgr.Add(manager.RunnableFunc(func(ctx context.Context) error {
+		errCh := make(chan error)
+		defer func() {
+			for range errCh {
+			} // drain errCh for GC
+		}()
+
+		go func() {
+			// start http Server
+			defer close(errCh)
+			errCh <- s.ListenAndServe()
+		}()
+
+		select {
+		case err := <-errCh:
+			return err
+		case <-ctx.Done():
+			s.Close()
+			return nil
+		}
+	}))
 }
 
 func (h HcCmdHelper) GetWatchNS() string {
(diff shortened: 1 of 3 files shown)


tiraboschi:
I set an hold label, just because I don't think we should merge it in production code but check the result on upstream dev environments/CI lanes

tiraboschi:
According to https://golang.org/doc/diagnostics#profiling profiling production services is considered safe and anyway we are going to expose pprof only if an environmental value is explicitly set and only an administrator can alter it.
I think we can keep this enabled also in production code.
Removing my hold label.

thetechnick:
Just leaving this here for future reference:
> Can I profile my production services?
> Yes. It is safe to profile programs in production, but enabling some profiles (e.g. the CPU profile) adds cost. You should expect to see performance downgrade. The performance penalty can be estimated by measuring the overhead of the profiler before turning it on in production.

https://golang.org/doc/diagnostics

As far as I could gather, only impact will be while profiling - e.g. calling the `/debug/pprof/profile` endpoint:
https://golang.org/src/net/http/pprof/pprof.go?s=3632:3684#L110

erkanerol:
I think we should add these into this PR:
 
1) A permanent doc about 
  - How to enable it? (kubectl patch deployment/subscription)
  - How to expose it? (port-forward or k8s service)
  - How to get the data? (curl etc.)
  - How to analyze it? (go tool pprof etc.)
2) A simple test which exposes this endpoint and sends some request to that endpoint

cmd/cmdcommon/cmdcommon.go
```diff
@@ -46,6 +49,48 @@ func (h HcCmdHelper) InitiateCommand() {
...
+	h.Logger.Info("Registering pprof server.")
+
+	mux := http.NewServeMux()
+	mux.HandleFunc("/debug/pprof/", pprof.Index)
+	mux.HandleFunc("/debug/pprof/cmdline", pprof.Cmdline)
+	mux.HandleFunc("/debug/pprof/profile", pprof.Profile)
+	mux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
+	mux.HandleFunc("/debug/pprof/trace", pprof.Trace)
+
+	s := &http.Server{Addr: pprofAddr, Handler: mux}
```
erkanerol, line 70: I would use the env variable as just a switch like "PPROF_ENABLE=true" since the address must be like `localhost:<port>` in the cluster and users may set it with wrong values with this format. Using a fixed port doesn't seem dangerous to me as well.

thetechnick: I don't think normal users would interact much if at all with this flag, so ease of use is not really a concern to me.

Hardcoded ports/addresses is a no go for me, as you can't start the operator locally if there is a port conflict.
Having to reconfigure half of your local stack, just because `:8080` is super popular is really annoying :D

Same with `localhost` in there, if you are in a pod, you probably want to bind to 0.0.0.0/:: while locally you prop want to include localhost.

orenc1:
looks like the okd cluster provided by openshift-ci is not stable enough.
`ci/prow/okd-hco-e2e-image-index-aws` is failing because 2 nodes in the cluster are not ready, and pods are stuck in pending state.
```
"message": "0/6 nodes are available: 1 node(s) didn't match Pod's node affinity, 2 node(s) had taint
{node.kubernetes.io/not-ready: }, that the pod didn't tolerate, 3 node(s) had taint 
{node-role.kubernetes.io/master: }, that the pod didn't tolerate.", 
```

orenc1:
openshift-ci okd lanes are not stable at the moment. overriding to unblock the merge.
/override ci/prow/okd-hco-e2e-image-index-aws
/override ci/prow/okd-hco-e2e-image-index-azure

PR #1181, opened 2021-03-11
Keep forcing the absence of the KubeVirt ConfigMap after upgrade

nunnatsa:
This PR add code to the reconcile logic, to keep search the KubeVirt ConfigMap, and if it created, HCO will remove it.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/operands/kubevirtConfigMap.go
+++ b/pkg/controller/operands/kubevirtConfigMap.go
@@ -56,7 +56,7 @@ func (handler kubeVirtCmHandler) ensure(req *common.HcoRequest) *EnsureResult {
 
 	unstructuredCm, err := hcoutils.ToUnstructured(cm)
 	if err != nil {
-		return res.Error(fmt.Errorf("failed to get Unstructured object form the %s ConfigMap; %w", kvCmName, err))
+		return res.Error(fmt.Errorf("failed to get Unstructured object from the %s ConfigMap; %w", kvCmName, err))
 	}
 
 	policy := metav1.DeletePropagationForeground
--- a/pkg/controller/operands/kubevirtConfigMap.go
+++ b/pkg/controller/operands/kubevirtConfigMap.go
@@ -1,0 +1,77 @@
+package operands
+
+import (
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/kubevirt/hyperconverged-cluster-operator/pkg/controller/common"
+	hcoutils "github.com/kubevirt/hyperconverged-cluster-operator/pkg/util"
+)
+
+const kvCmName = "kubevirt-config"
+
+// Make sure that the kubevirt-config configMap does not exist
+type kubeVirtCmHandler struct {
+	client  client.Client
+	emitter hcoutils.EventEmitter
+}
+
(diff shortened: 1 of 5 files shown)


nunnatsa:
okd-hco-e2e-image-index-aws lane succeeded.
/override ci/prow/okd-hco-e2e-image-index-azure

pkg/controller/operands/kubevirtConfigMap.go
```diff
@@ -0,0 +1,77 @@
...
+
+	if err != nil {
+		return res.Error(fmt.Errorf("failed to read the %s ConfigMap; %w", kvCmName, err))
+	}
+
+	req.Logger.Info(fmt.Sprintf("removing the unused %s ConfigMap", kvCmName))
+
+	unstructuredCm, err := hcoutils.ToUnstructured(cm)
+	if err != nil {
+		return res.Error(fmt.Errorf("failed to get Unstructured object form the %s ConfigMap; %w", kvCmName, err))
```
orenc1, line 59: s/form/from

pkg/controller/operands/kubevirtConfigMap_test.go
```diff
@@ -0,0 +1,212 @@
...
+		fakeError := fmt.Errorf("fake delete error")
+		c.InitiateDeleteErrors(func(obj client.Object) error {
+			if unstructed, ok := obj.(runtime.Unstructured); ok {
+				kind := unstructed.GetObjectKind()
+				if kind.GroupVersionKind().Kind == "ConfigMap" && obj.GetName() == kvCmName {
+					return fakeError
+				}
+			}
+			return nil
+		})
```
orenc1, line 188: general question: why is `HcoTestClient.deleteError` has a receiver function of `FakeWriteErrorGenerator` that gets `client.Object` and not `client.ObjectKey`?
i.e. why does the delete function needs the whole object, instead just its key?

pkg/controller/operands/kubevirtConfigMap_test.go
```diff
@@ -0,0 +1,212 @@
...
+	It("should remove the configMap if it exists", func() {
+		cm := &corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      kvCmName,
+				Namespace: commonTestUtils.Namespace,
+			},
+			Data: map[string]string{
+				"fakeKey": "fakeValue",
+			},
+		}
```
orenc1, line 67: The block of the config-map definition is duplicated 4 times in this file, for almost every test case (4 out of 5).
Why not defining it in the top, under `var` of the `Describe` block?

orenc1 (commented):
Please see my comments.
2 typos, 2 questions (one general, one specific).
Thanks

pkg/controller/operands/kubevirtConfigMap_test.go
nunnatsa: The client.Client interface uses the Writer interface. The Delete function defined like this:

```go
// Delete deletes the given obj from Kubernetes cluster.
Delete(ctx context.Context, obj Object, opts ...DeleteOption) error
```

So this is what the Delete method get. Now, the fake error function should be flixible, to allow different errors, so it get what the Delete function get, so it can play with it.

pkg/controller/operands/kubevirtConfigMap_test.go
nunnatsa: Common var is not an option because some cases are desructive (e.g. the client.Get function).

I can use a NewKvConfig function instead.

orenc1:
`nmstate-handler` crashloops again in the okd lane...
/retest

tiraboschi (approved):

PR #1189, opened 2021-03-16
Run KV smoke tests on openshift-ci

tiraboschi:
Run KV smoke tests on openshift-ci

Enable a new test line to run KV smoke tests on openshift-ci for each PR on this repo.
Another PR on https://github.com/openshift/release will configure a test lane on azure and on GCP
were we have nested virtualization support. We cannot run it on AWS where we need KVM_EMULATION
which is by far too slow to be reliable.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/hack/kv-smoke-tests.sh
+++ b/hack/kv-smoke-tests.sh
@@ -1,0 +1,234 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+
+INSTALLED_NAMESPACE=${INSTALLED_NAMESPACE:-"kubevirt-hyperconverged"}
+
+source hack/common.sh
+source cluster/kubevirtci.sh
+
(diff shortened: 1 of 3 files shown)


hack/kv-smoke-tests.sh
```diff
@@ -0,0 +1,248 @@
...
+EOF
+
+cat <<EOF | ${CMD} create -f -
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: kubevirt-test-config
+  namespace: ${INSTALLED_NAMESPACE}
+data:
+  test-config.json: |
```
erkanerol, line 218: if the content of this has to be same as `kubevirt-testing-configuration.json `, I would create it from file. `kubectl create configmap --from-file=...`

PR #1192, opened 2021-03-17
[release-1.3] Update Images Digests

orenc1:

**Release note**:

```release-note
update image digests before release
```


tiraboschi (approved):

nunnatsa:
@tiraboschi - this should be copied to `master` 1.3 directory, because it will break the upgrade-prev lanes.

tiraboschi:
> @tiraboschi - this should be copied to `master` 1.3 directory, because it will break the upgrade-prev lanes.

Yes, I'll manually do it.
Then we should find a mechanism to automate this.

orenc1:
> > @tiraboschi - this should be copied to `master` 1.3 directory, because it will break the upgrade-prev lanes.
> 
> Yes, I'll manually do it.
> Then we should find a mechanism to automate this.

We can have a workflow that mirrors the contents of the manifests version folder of a release branch, to it's respective version folder in master, which can be triggered by a special keyword we define in the PR (for example).
Anyway it must go through a PR to master...

PR #1203, opened 2021-03-24
Reconcile CDI configurations

nunnatsa:
HCO now forces the CDI configurations during reconciliation.

The HyperConverged custom resource contains two new fields:
* `scratchSpaceStorageClass` - Override the storage class to used for scratch space during transfer operations. The scratch space storage class is determined in the following order:

  value of scratchSpaceStorageClass, if that doesn't exist, use the default storage class, if there is no default storage class, use the storage class of the DataVolume, if no storage class specified, use no storage class for scratch space.

  HCO will propagate this field to CDI's `spec.config.scratchSpaceStorageClass`
* `resourceRequirements` that contains the `storageWorkloads` field. This field content is a standard kubernetes resourceConfiguration type.

   HCO will propagate this field to CDI's `spec.config.podResourceConfigurations` filed.

**Release note**:

```release-note
HCO now forces the CDI configurations during reconciliation.

API Changes:
The HyperConverged's spec field now contains two new fields:
* scratchSpaceStorageClass
* resourceRequirements.storageWorkloads
```


deploy/crds/hco00.crd.yaml
```diff
@@ -887,6 +887,47 @@ spec:
...
+                          resources required. If Requests is omitted for a container,
+                          it defaults to Limits if that is explicitly specified, otherwise
+                          to an implementation-defined value. More info: https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/'
+                        type: object
+                    type: object
+                type: object
+              scratchSpaceStorageClass:
+                description: "Override the storage class to used for scratch space
+                  during transfer operations. The scratch space storage class is determined
+                  in the following order: \n value of scratchSpaceStorageClass, if
```
tiraboschi, line 925: can we avoid a visible \n?

docs/cluster-configuration.md
```diff
@@ -252,6 +252,54 @@ spec:
...
 ```
+
+## Storage Class for Scratch Space
+
+Administrators can Override the storage class to used for scratch space during transfer operations by setting the
+`scratchSpaceStorageClass` field under the HyperConverged `spec` field.
+
+The scratch space storage class is determined in the following order:
+
+value of scratchSpaceStorageClass, if that doesn't exist, use the default storage class, if there is no default storage
```
tiraboschi, line 263: to we have any validation?
what is going to happen if the user set a not existing storage class?

pkg/apis/hco/v1beta1/hyperconverged_types.go
```diff
@@ -167,6 +180,15 @@ type MediatedHostDevice struct {
 	ExternalResourceProvider bool   `json:"externalResourceProvider,omitempty"`
 }
 
+// ResourceRequirements is a list of resource requirements for the operand workloads pods
+// +k8s:openapi-gen=true
+type OperandResourceRequirements struct {
+	// StorageWorkloads defines the resources requirements for storage workloads. It will propagate to the CDI custom
+	// resource
+	// +optional
+	StorageWorkloads *corev1.ResourceRequirements `json:"storageWorkloads,omitempty"`
```
tiraboschi, line 189: is it worth to set this under the workload stanza?

pkg/apis/hco/v1beta1/hyperconverged_types.go
```diff
@@ -54,6 +54,19 @@ type HyperConvergedSpec struct {
 	// +optional
 	CertConfig *HyperConvergedCertConfig `json:"certConfig,omitempty"`
 
+	// ResourceRequirements describes the resource requirements for the operand workloads.
+	// +optional
+	ResourceRequirements *OperandResourceRequirements `json:"resourceRequirements,omitempty"`
+
+	// Override the storage class to used for scratch space during transfer operations. The scratch space storage class
```
tiraboschi, line 61: to use

pkg/apis/hco/v1beta1/hyperconverged_types.go
```diff
@@ -54,6 +54,19 @@ type HyperConvergedSpec struct {
...
+	// is determined in the following order:
+	//
```
tiraboschi, line 63: maybe this adds a not so nice \n

docs/cluster-configuration.md
nunnatsa: As with anything else: the webhook will send it to the CDI API server. It its job to validate.

pkg/apis/hco/v1beta1/hyperconverged_types.go
nunnatsa: I don't think we can set the same values for all the operand's workloads. It is also only supported in CDI.

pkg/apis/hco/v1beta1/hyperconverged_types.go
nunnatsa: This is copy&paste from CDI...

maybe "to be used"?

docs/cluster-configuration.md
tiraboschi: ok

pkg/apis/hco/v1beta1/hyperconverged_types.go
tiraboschi: "Override the storage class used for scratch space during transfer operations."

pkg/apis/hco/v1beta1/hyperconverged_types.go
tiraboschi: agrred

tiraboschi (approved):

PR #1210, opened 2021-03-25
Migrate IMS config

tiraboschi:
migrate IMS config to HCO CR

- add vddkInitImage parameter to HCO CR
- make HCO the source of truth for it
- during upgrade, copy existing values from IMS CM to HCO CR onlyif still not set on HCO CR
- reconcile the whole CM for IMS
- add unit tests

**Release note**:
```release-note
let the user configure vddkInitImage from HCO CR
```

### DIFF
--- a/pkg/controller/hyperconverged/hyperconverged_controller.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller.go
@@ -864,7 +864,9 @@ func (r *ReconcileHyperConverged) recoverHCOVersion(request *common.HcoRequest) 
 const (
 	kvCmName         = "kubevirt-config"
 	backupKvCmName   = kvCmName + "-backup"
+	imsCmName        = "v2v-vmware"
 	liveMigrationKey = "migrations"
+	vddkInitImakeKey = "vddk-init-image"
 )
 
 func (r *ReconcileHyperConverged) migrateBeforeUpgrade(req *common.HcoRequest) (bool, error) {
@@ -873,12 +875,17 @@ func (r *ReconcileHyperConverged) migrateBeforeUpgrade(req *common.HcoRequest) (
 		return false, err
 	}
 
-	cdiConfigMmodified, err := r.migrateCdiConfigurations(req)
+	cdiConfigModified, err := r.migrateCdiConfigurations(req)
 	if err != nil {
 		return false, err
 	}
 
-	return kvConfigMmodified || cdiConfigMmodified, nil
+	imsConfigModified, err := r.migrateImsConfigurations(req)
+	if err != nil {
+		return false, err
+	}
+
+	return kvConfigMmodified || cdiConfigModified || imsConfigModified, nil
 }
 
 func (r ReconcileHyperConverged) migrateKvConfigurations(req *common.HcoRequest) (bool, error) {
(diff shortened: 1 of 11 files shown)


pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -937,6 +944,45 @@ func (r ReconcileHyperConverged) migrateCdiConfigurations(req *common.HcoRequest
...
+		}
+		req.Logger.Info("failed to get IMS configmap", "error", err.Error())
+		return false, err
+	}
+
+	modified := false
+	vddkInitImage, ok := cm.Data[vddkInitImakeKey]
+	if !ok {
+		// if the user never set the value we should simply continue
+		req.Dirty = false || req.Dirty
```
nunnatsa, line 969: This line will never change `req.Dirty` - please drop it.

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -937,6 +944,45 @@ func (r ReconcileHyperConverged) migrateCdiConfigurations(req *common.HcoRequest
...
+		modified = false
```
nunnatsa, line 970: `modified` is already `false`

So I think we can drop this whole if branch, and only check `if ok`

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -937,6 +944,45 @@ func (r ReconcileHyperConverged) migrateCdiConfigurations(req *common.HcoRequest
...
+	} else {
+		if *req.Instance.Spec.VddkInitImage != vddkInitImage {
```
nunnatsa, line 972: This is something I'm not sure about. If the HyperConverged CR already contains the VDDK image, so someone populated it on purpose. I don't think we need to adopt the old configurations in this case. So I would change the if to:
```golang
if req.Instance.Spec.VddkInitImage == nil && *req.Instance.Spec.VddkInitImage != vddkInitImage {
```

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -937,6 +944,45 @@ func (r ReconcileHyperConverged) migrateCdiConfigurations(req *common.HcoRequest
...
+			req.Logger.Info("updating the HyperConverged CR from the IMS configMap")
+			req.Instance.Spec.VddkInitImage = &vddkInitImage
+			req.Dirty = true
+			modified = true
+		} else {
```
nunnatsa, line 977: There is no effet for this else - please drop it

pkg/controller/hyperconverged/hyperconverged_controller.go
tiraboschi: done

pkg/controller/hyperconverged/hyperconverged_controller.go
tiraboschi: done

pkg/controller/hyperconverged/hyperconverged_controller.go
tiraboschi: done

pkg/controller/hyperconverged/hyperconverged_controller.go
tiraboschi: done

nunnatsa (approved):
/approve

nunnatsa:
Added auto generated code

/override coverage/coveralls

PR #1213, opened 2021-03-29
Fix the go version in all go-based github actions to 1.15.10

nunnatsa:
The previous setting in some actions was `^1.15` so it became 1.16

We don't use 1.16 yet, so we need to fix it to the latests 1.15.x

**Release note**:

```release-note
NONE
```


tiraboschi (approved):

nunnatsa:
We configured the smoke tests not to run in these cases, but it will only work for new PRs. Overriding.

PR #1222, opened 2021-03-31
Expose an environmental value to virtio win image

tiraboschi:
Expose an environmental value to consume at runtime
a virtio win image that should be propagated to
a config map consumed by the UI component
when the user tries to boot/import a Windows VM.

In the past the UI was using an hardcoded value
but due to that it wasn't properly working on
disconnected environments.

Fixes: https://bugzilla.redhat.com/1942839

**Release note**:
```release-note
Expose an environmental value to virtio win image
```


deploy/images.csv
```diff
@@ -1,9 +1,11 @@
 image_var,name,tag,digest
+<<<<<<< HEAD
```
yaacov, line 2: typo ?

tiraboschi: yes, thanks!

PR #1239, opened 2021-04-11
Move ConsoleCLIDownload controller to operandHandler

erkanerol:

While fixing linter issues, we noticed that we are ignoring the error coming from cliDownloadHandler. Then we questioned why we behave ConsoleCliDownload objects differently and we decided to move it into existing operandHandler. 

This PR is supposed to not change anything in terms of operator behaviour. It should only change the implementation. 

***Note for reviwers***
Old implementation didn't have any openshift check and it was trying to create the object in all environments. It was checking the errors and it was ignoring `meta.NoResourceMatchError` errors.

New implementation assumes that `ConsoleCliDownload` CRD exists in Openshift clusters. The controller is simply disabled for vanilla Kubernetes. 

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/controller/operands/cliDownload.go
+++ b/pkg/controller/operands/cliDownload.go
@@ -16,50 +16,50 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-type cLIDownloadHandler genericOperand
+type cliDownloadHandler genericOperand
 
-func newCLIDownloadHandler(Client client.Client, Scheme *runtime.Scheme) *cLIDownloadHandler {
-	return &cLIDownloadHandler{
+func newCLIDownloadHandler(Client client.Client, Scheme *runtime.Scheme) *cliDownloadHandler {
+	return &cliDownloadHandler{
 		Client:                 Client,
 		Scheme:                 Scheme,
 		crType:                 "ConsoleCLIDownload",
 		removeExistingOwner:    false,
 		setControllerReference: false,
-		hooks:                  &cLIDownloadHooks{},
+		hooks:                  &cliDownloadHooks{},
 	}
 }
 
-type cLIDownloadHooks struct{}
+type cliDownloadHooks struct{}
 
-func (h cLIDownloadHooks) getFullCr(hc *hcov1beta1.HyperConverged) (client.Object, error) {
+func (h cliDownloadHooks) getFullCr(hc *hcov1beta1.HyperConverged) (client.Object, error) {
 	return NewConsoleCLIDownload(hc), nil
 }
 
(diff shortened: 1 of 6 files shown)


pkg/controller/operands/cliDownload.go
```diff
@@ -1,57 +1,72 @@
...
 	hcoutil "github.com/kubevirt/hyperconverged-cluster-operator/pkg/util"
 	consolev1 "github.com/openshift/api/console/v1"
-	objectreferencesv1 "github.com/openshift/custom-resource-status/objectreferences/v1"
-	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/tools/reference"
 )
 
-type CLIDownloadHandler genericOperand
+type cLIDownloadHandler genericOperand
```
nunnatsa, line 19: Please fix casing: `cliDownloadHandler`

pkg/controller/operands/cliDownload.go
```diff
@@ -1,57 +1,72 @@
...
+		return false, false, errors.New("can't convert to ConsoleCLIDownload")
+	}
+	if !reflect.DeepEqual(found.Spec, desired.Spec) ||
+		!reflect.DeepEqual(found.Labels, desired.Labels) {
+		if req.HCOTriggered {
+			req.Logger.Info("Updating existing ConsoleCLIDownload's Spec to new opinionated values")
+		} else {
+			req.Logger.Info("Reconciling an externally updated ConsoleCLIDownload's Spec to its opinionated values")
 		}
-		err = objectreferencesv1.SetObjectReference(&req.Instance.Status.RelatedObjects, *objectRef)
```
nunnatsa, line 40: How do you handle this?

pkg/controller/operands/cliDownload.go
erkanerol: Fixed

pkg/controller/operands/cliDownload.go
erkanerol: I thought: We have the similar functionality here: https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/pkg/controller/operands/operand.go#L145 and it is a part of operandHandler. Therefore, we don't need to have such  a logic here. Am I wrong?

nunnatsa: No, you are right - I was confused with https://github.com/kubevirt/hyperconverged-cluster-operator/blob/master/pkg/controller/operands/operand.go#L160

erkanerol: Thanks!

nunnatsa (approved):
/approve

PR #1240, opened 2021-04-11
Use CI-generated index image for HCO upgrade tests

orenc1:
Take advantage of Openshift-CI 'openshift-operators-ci' workflows to build ci-generated index image and then subscribe to the initial version of HCO (same y-stream and previous y-stream).
Then, using a custom logic, upgrade it to the version generated by the code in the open PR (called 100.0.0) and run validation tests.

The upgrade test configuration using ci-index is made in https://github.com/openshift/release/pull/17442.
This PR should be merged before the one in openshift/release could pass the new test lanes.

**Release note**:

```release-note
switch to ci-index for HCO upgrade tests
```

### DIFF
--- a/hack/upgrade-test-index-image.sh
+++ b/hack/upgrade-test-index-image.sh
@@ -1,0 +1,212 @@
+#!/bin/bash -e
+#
+# This file is part of the KubeVirt project
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Copyright 2021 Red Hat, Inc.
+#
+# Usage:
+# make upgrade-test-index-image
+#
+# Use Openshift-CI "optional-operators-ci-*" workflow to:
+# - Build an internal index image based off of the index image at:
+#   quay.io/kubevirt/hyperconverged-cluster-index
+#   with the appropriate tag of the version
+# - Add to that index a new bundle, named 100.0.0 with the contents
+#   of the open PR (this can include new dependent images, new CRDs...).
+# - Subscribe to the initial channel, using the "optional-operators-ci-subscribe"
+#   step.
+# This script is then upgrading HCO to 100.0.0 version, by patching the subscripiton,
+# and performs various validations against the upgraded version.
+
+
+MAX_STEPS=15
+CUR_STEP=1
+RELEASE_DELTA="${RELEASE_DELTA:-1}"
+HCO_DEPLOYMENT_NAME=hco-operator
+HCO_WH_DEPLOYMENT_NAME=hco-webhook
+HCO_NAMESPACE="kubevirt-hyperconverged"
+HCO_KIND="hyperconvergeds"
+HCO_RESOURCE_NAME="kubevirt-hyperconverged"
+PACKAGE_DIR="./deploy/olm-catalog/community-kubevirt-hyperconverged"
+INITIAL_CHANNEL=$(ls -d ${PACKAGE_DIR}/*/ | sort -rV | awk "NR==${RELEASE_DELTA}" | cut -d '/' -f 5)
+TARGET_VERSION=100.0.0
+TARGET_CHANNEL=${TARGET_VERSION}
+echo "INITIAL_CHANNEL: $INITIAL_CHANNEL"
+
+function Msg {
+    { set +x; } 2>/dev/null
+    echo "--"
+    for a in "$@"; do
+        echo "Upgrade Step ${CUR_STEP}/${MAX_STEPS}: $a"
+    done
+    echo "--"
+    ((CUR_STEP += 1))
+    set -x
+}
+
+source ./hack/upgrade-openshiftci-config
+
+function cleanup() {
+    rv=$?
+    if [ "x$rv" != "x0" ]; then
+        echo "Error during upgrade: exit status: $rv"
+        make dump-state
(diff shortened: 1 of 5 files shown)


deploy/index-image/bundle.Dockerfile.ci-index-image-upgrade
```diff
@@ -0,0 +1,28 @@
+FROM quay.io/fedora/fedora:33-x86_64 AS builder
+ARG INITIAL_VERSION=1.4.0
+ARG TARGET_VERSION=100.0.0
```
nunnatsa, line 3: Not sure why we need a builder image here, just to sedding two files. I think it can be done within one Dockerfile, or from the outside before calling to docker build; e.g. copy all the files to a temporary directory, do the change and then call to docker build.

nunnatsa: Also, I find the file name a bit confusing: I think it should start with `Dockerfile` and the rest will be its postfix. Using `Dockerfile` as postfix will also work (I not sure I like it better),

deploy/olm-catalog/bundle.Dockerfile.ci-index-image-upgrade
```diff
@@ -0,0 +1,28 @@
+FROM quay.io/fedora/fedora:33-x86_64 AS builder
```
nunnatsa, line 1: Is this a copy of `deploy/index-image/bundle.Dockerfile.ci-index-image-upgrade`? why do we need both?

hack/upgrade-test-index-image.sh
```diff
@@ -0,0 +1,212 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Copyright 2021 Red Hat, Inc.
+#
+# Usage:
+# make upgrade-test-index-image
+#
+# Use Openshift-CI "optional-operators-ci-*" workflow to:
+# - Build an internal index image based off the index image at:
```
nunnatsa, line 23: based off => based of

deploy/index-image/bundle.Dockerfile.ci-index-image-upgrade
nunnatsa: > Not sure why we need a builder image here, just to sedding two files. I think it can be done within one Dockerfile, or from the outside before calling to docker build; e.g. copy all the files to a temporary directory, do the change and then call to docker build.

After reading openshift/release#17442, I understand why it should be done within a Dockerfile, but I still think that the builder pattern is not needed here. However, I would join all the `RUN` commands into one, in order to create less layers; e.g.
```
COPY ${PACKAGE_NAME}/${INITIAL_VERSION}/manifests/ /manifests/
COPY ${PACKAGE_NAME}/${INITIAL_VERSION}/metadata /metadata/
RUN mv /manifests/kubevirt-hyperconverged-operator.v${INITIAL_VERSION}.clusterserviceversion.yaml \
    /manifests/kubevirt-hyperconverged-operator.v${TARGET_VERSION}.clusterserviceversion.yaml && \
  sed -i 's/${INITIAL_VERSION}/${TARGET_VERSION}/g' /manifests/kubevirt-hyperconverged-operator.v${TARGET_VERSION}.clusterserviceversion.yaml && \
  sed -i 's/${INITIAL_VERSION}/${TARGET_VERSION}/g' /metadata/annotations.yaml
```

orenc1: > Not sure why we need a builder image here, just to sedding two files. I think it can be done within one Dockerfile, or from the outside before calling to docker build; e.g. copy all the files to a temporary directory, do the change and then call to docker build.

There is no way to do the modification on the ci-operator config, and I wanted to avoid having 100.0.0 manifests at the directory. Also, I didn't find a way to resolve the `INITIAL_VERSION` (currently 1.4.0) in the dockerfile context, that's why I used an additional ARG.



> Also, I find the file name a bit confusing: I think it should start with `Dockerfile` and the rest will be its postfix. Using `Dockerfile` as postfix will also work (I not sure I like it better),

I'll update the dockerfile name.



> > Not sure why we need a builder image here, just to sedding two files. I think it can be done within one Dockerfile, or from the outside before calling to docker build; e.g. copy all the files to a temporary directory, do the change and then call to docker build.
> 
> After reading [openshift/release#17442](https://github.com/openshift/release/pull/17442), I understand why it should be done within a Dockerfile, but I still think that the builder pattern is not needed here. However, I would join all the `RUN` commands into one, in order to create less layers; e.g.
> 
> ```
> COPY ${PACKAGE_NAME}/${INITIAL_VERSION}/manifests/ /manifests/
> COPY ${PACKAGE_NAME}/${INITIAL_VERSION}/metadata /metadata/
> RUN mv /manifests/kubevirt-hyperconverged-operator.v${INITIAL_VERSION}.clusterserviceversion.yaml \
>     /manifests/kubevirt-hyperconverged-operator.v${TARGET_VERSION}.clusterserviceversion.yaml && \
>   sed -i 's/${INITIAL_VERSION}/${TARGET_VERSION}/g' /manifests/kubevirt-hyperconverged-operator.v${TARGET_VERSION}.clusterserviceversion.yaml && \
>   sed -i 's/${INITIAL_VERSION}/${TARGET_VERSION}/g' /metadata/annotations.yaml
> ```

will do.

deploy/olm-catalog/bundle.Dockerfile.ci-index-image-upgrade
orenc1: Because all the contents under `deploy/index-image` are being deleted and recreated in `build-manifests.sh`

hack/upgrade-test-index-image.sh
orenc1: It should be either _based off (US) of_ or _based on_ (UK)
https://english.stackexchange.com/questions/15081/based-on-instead-of-based-off-of

orenc1:
the override-bot is not overriding lanes that were not failed. `ci/prow/hco-e2e-kv-smoke-gcp` seems to be stuck.

ci/prow/hco-e2e-kv-smoke-azure succeeded

PR #1245, opened 2021-04-13
Avoid using custom vendoring for functional tests

tiraboschi:
Now that also the kubevirt client-go has been
updated to depend on a more recent set of
kubernetes go libraries we can finally avoid
forcing a custom vendoring just for the
functional tests.

Bump k8s.io/api v0.20.2 to align with kubevirt and CDI.

**Release note**:
```release-note
NONE
```


erkanerol:
I am happy to see that we are using a single `go.mod` file for prod+tests and we are vendoring test dependencies as well, which makes git environment cleaner while implementing functional tests. 

/approve
/lgtm

PR #1250, opened 2021-04-14
Workaround for SSP stuck upgrades

tiraboschi:
SSP operator occasionally can get stuck on upgrades and
it will stay in that state forever.
As a workaround we can periodically (every 12 minutes)
refresh an annotation on KubeVirtCommonTemplateBundle CR just
to trigger another reconciliation loop on the ansible operator.
The annotation will be removed at the end of the upgrade process.

Fixes: https://bugzilla.redhat.com/1909975

**Release note**:
```release-note
Introduce a workaround for SSP stuck upgrades
```

### DIFF
--- a/pkg/controller/hyperconverged/hyperconverged_controller_components_test.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller_components_test.go
@@ -20,6 +20,7 @@ import (
 	virtconfig "kubevirt.io/kubevirt/pkg/virt-config"
 	"os"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"time"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/ginkgo/extensions/table"
@@ -30,6 +31,8 @@ import (
 
 	"k8s.io/client-go/tools/reference"
 )
+
+const hcoForceUpdate = "hcoForceUpdate"
 
 var _ = Describe("HyperConverged Components", func() {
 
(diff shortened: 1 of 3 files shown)


tiraboschi:
we already removed out travis integration, this PR passes sanity tests locally.
/override continuous-integration/travis-ci/pr

tiraboschi:
/hold
waiting to understand if this workaround is really needed

pkg/controller/hyperconverged/hyperconverged_controller_components_test.go
```diff
@@ -1164,69 +1167,361 @@ var _ = Describe("HyperConverged Components", func() {
...
+			res := r.ensureKubeVirtCommonTemplateBundle(req)
+
+			foundResource := &sspv1.KubevirtCommonTemplatesBundle{}
+			Expect(
+				cl.Get(context.TODO(),
+					types.NamespacedName{Name: expectedResource.Name, Namespace: expectedResource.Namespace},
+					foundResource),
+			).To(BeNil())
+
+			Expect(foundResource.ObjectMeta.Annotations).To(HaveKey("hcoForceUpdate"))
```
erkanerol, line 1415: @tiraboschi  Should not we check that the value is updated?

tiraboschi: The exact value is time dependent so we cannot really hard-code the expected value in a test (at least not easily without mocking time.Now()); checking against a variable unknown value is not that different that simply checking that at least we have a new annotation there.
Then we also have another test that explicitly checks if we are updating outdated values when needed.

tiraboschi:
Ignoring travis-ci since we already deconfigured it.
Trusting /approve from Erkan even if he wasn't part of the team at release-1.2 time.

pkg/controller/hyperconverged/hyperconverged_controller_components_test.go
tiraboschi: No sorry, you were right. I'll address this is an additional PR.

tiraboschi: Done, plese see:
https://github.com/kubevirt/hyperconverged-cluster-operator/pull/1256

erkanerol: Thanks!

PR #1255, opened 2021-04-15
Change KubevirtHyperconvergedClusterOperatorCRModification alert rule

erkanerol:

`kubevirt_hco_out_of_band_modifications_count` metric is a counter which keeps the number of changes on operands that trigger overwrites. 

Current implementation checks whether it is more than zero or not. If it is more than zero, it waits 10 minutes and triggers the alert.  Alert won't resolve without restarting the operator+counter.

This implementation checks whether there is an increment in the last 10 minutes. If there is, it triggers the alert immediately. It resolves if there is no change in the last 10 minutes.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/hack/prom-rule-ci/prom-rules-tests.yaml
+++ b/hack/prom-rule-ci/prom-rules-tests.yaml
@@ -9,7 +9,7 @@ tests:
   - interval: 1m
     input_series:
       - series: 'kubevirt_hco_out_of_band_modifications_count{component_name="kubevirt"}'
-        values: "0 0 0 1 1 1 1 1 1 1 1 1 1"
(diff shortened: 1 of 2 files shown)


erkanerol:
I talked with @fabiand  about the expected behavior of this alert and agreed on this one. Basically,
- The alert will be immediately fired if someone changes an operand.
- It will last 10 minutes and resolve if there is another increment on the counter.

We think that users have some integrations in production environments (SMS, email, slack etc. ) and the alert will be seen in those environments even when the alert is resolved after a while.

tiraboschi (approved):

PR #1257, opened 2021-04-16
Hardening the defaulting mechanism

tiraboschi:
Hardening the defaulting mechanism

The defaulting mechanism for certConfig is currently
not working when the user completely omits the whole
spec stanza.
Fixing it and adding tests to avoid regressions:
- forcing defaults to be coherent at all the levels
- using values instead of pointers since we don't need nil when
  we know that we can have for sure at least a default value
- updating contributing documentation

See:
- https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required
- https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#defaulting
as a reference.

BugFixes: https://bugzilla.redhat.com/1949795

**Release note**:
```release-note
hardening the defaulting mechanism
```


hack/check_defaults.sh
```diff
@@ -0,0 +1,86 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Copyright 2021 Red Hat, Inc.
+#
+# This script checks that old SSP operator resources are removed from the cluster
```
dollierp, line 19: This comment should be fixed.

tiraboschi: done, thanks

tiraboschi:
Currently failing on upgrades from 1.3.0:
```
INFO[2021-04-21T15:59:39Z] time="2021-04-21T15:26:46Z" level=info msg="error transitioning InstallPlan" id=YCGyi ip=install-kbpcf namespace=kubevirt-hyperconverged phase=Installing syncError="error validating existing CRs against new CRD's schema: hyperconvergeds.hco.kubevirt.io: error validating custom resource against new schema &apiextensions.CustomResourceValidation{OpenAPIV3Schema:(*apiextensions.JSONSchemaProps)(0xc013042e00)}: [[].spec.certConfig: Required value, [].spec.featureGates: Required value, [].spec.liveMigrationConfig: Required value]" updateError="Operation cannot be fulfilled on installplans.operators.coreos.com \"install-kbpcf\": the object has been modified; please apply your changes to the latest version and try again" 
```

PR #1259, opened 2021-04-16
Fix cluster-sync for dual stack env

tiraboschi:
Fix cluster-sync for dual stack env

**Release note**:

```release-note
NONE
```


nunnatsa (approved):
/approve

PR #1261, opened 2021-04-18
fix subscription patch command in index image upgrade

orenc1:
when using `oc get <resource_type> -o name`, the response is in the form of `resource/name`, and an error is thrown when specifying the resource type twice.
```
error: there is no need to specify a resource type as a separate argument when passing arguments in resource/name form (e.g. 'oc get resource/<resource_name>' instead of 'oc get resource resource/<resource_name>' 
```

**Release note**:

```release-note
NONE
```


nunnatsa (approved):
/approve

orenc1:
irrelevant for this PR:
/override ci/prow/okd-hco-e2e-upgrade-aws

PR #1264, opened 2021-04-18
fix sed command in upgrade-index-ci Dockerfile

orenc1:
When using single quotes in the dockerfile, the variables are not evaluated in the sed command, thus changing to double quotes.
Also, fixing unstable publish workflow - 'env' was missing in UNSTABLE env var.

**Release note**:

```release-note
NONE
```


nunnatsa (approved):
/approve

PR #1273, opened 2021-04-20
Adjustments to the test upgrade index-image script

orenc1:
When the subscription is created with `"installPlanApproval":"Manual"`, patching it to `"Automatic"` doesn't auto-approve the install plan of the new version (100.0.0 in this case), thus we have to approve it.

Also, when publishing the unstable bundle image, use floating tag instead of sha digest.

**Release note**:

```release-note

```


nunnatsa (approved):
/approve

PR #1274, opened 2021-04-20
WIP: hardening the defaulting mechanism - alternative

tiraboschi:
The defaulting mechanism for certConfig is currently
not working when the user completely omits the whole
spec stanza.
Fixing it and adding tests to avoid regressions.

BugFixes: https://bugzilla.redhat.com/1949795

**Release note**:

```release-note

```


tiraboschi:
not so promising

PR #1275, opened 2021-04-20
[upgrade test] Approve new InstallPlan on upgrade

orenc1:
Ensure that we are waiting for the new InstallPlan to be created when the upgrade is available (upon changing the subscription channel), and then approving it for the upgrade to begin.

**Release note**:

```release-note
NONE
```


nunnatsa (commented):
/approve

PR #1280, opened 2021-04-22
Don't run build-push-images workflow on irrelevant branches and paths

orenc1:
Currently, the branches of release-4.8 and release-4.9 are configured automatically by openshift-ci and are following master.
Thus, we want to prevent merge on them to trigger the workflow of building the pushing the images of hco-operator, hco-webhook,
bundle and index. It will run on our actual release branches (requires to have the workflow file there too).

Also, when files are modified only in the following paths, the workflow won't run, as these paths have no effect on the resulting images:
```
docs/**
images/**
automation/**
cluster/**
tests/**
```

**Reviewer Checklist**

- [x] PR Message
- [x] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
NONE
```

### DIFF
--- a/.github/workflows/build-push-images.yaml
+++ b/.github/workflows/build-push-images.yaml
@@ -4,7 +4,7 @@ on:
   push:
     branches:
       - master
-      - release-*
+      - release-[12]*
 jobs:
   build_push:
     if: (github.repository == 'kubevirt/hyperconverged-cluster-operator')
--- a/.github/workflows/build-push-images.yaml
+++ b/.github/workflows/build-push-images.yaml
@@ -4,7 +4,13 @@ on:
   push:
     branches:
       - master
-      - release-[12]*
+      - release-1*
+    paths-ignore:
+      - 'docs/**'
+      - 'images/**'
+      - 'automation/**'
+      - 'cluster/**'
+      - 'tests/**'
 jobs:
   build_push:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-push-images.yaml
```diff
@@ -4,7 +4,13 @@ on:
   push:
     branches:
       - master
-      - release-*
+      - release-[12]*
```
nunnatsa, line 7: Why 2?

.github/workflows/build-push-images.yaml
```diff
@@ -4,7 +4,13 @@ on:
...
+    paths-ignore:
+      - 'docs/**'
+      - 'images/**'
+      - 'automation/**'
+      - 'cluster/**'
+      - 'tests/**'
```
nunnatsa, line 13: :clap:

.github/workflows/build-push-images.yaml
orenc1: have it, in case we'll move to release-2.Y in the future, according to @tiraboschi .
on the other hand, we can add it when relevant.

PR #1288, opened 2021-04-28
Fix a bug in Obsolete CPU Models

nunnatsa:
HCO exposes the Obsolete CPU Model configuration in it's API, with no default. The KV behavior is that it uses a predefnine hardcoded
list, if the values are not set. This is not visible in the KV CR. The same is for the minCPUModel field.

The problem is that setting the HC by the user, will override the KV defaults.

This PR copied the default values from KV, and always populate the KV's `obsoleteCPUModels` field with these value. The user can only add
more CPU models, but can't remove CPU models from the predefined list. The predefined list is not visible in the HyperConverged CR.

This PR also removes the upgrade handling for the obsolete CPU models, because we're using the defaults anyway, and in the past, the user couldn't
change the values in the configmap, so there is no user modified data.

**Reviewer Checklist**

- [x] PR Message
- [x] Commit Messages
- [ ] How to test
- [x] Unit Tests
- [ ] Functional Tests
- [x] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [x] Backward Compatibility
  
**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/apis/hco/v1beta1/hyperconverged_types.go
+++ b/pkg/apis/hco/v1beta1/hyperconverged_types.go
@@ -233,10 +233,15 @@ type OperandResourceRequirements struct {
 // HyperConvergedObsoleteCPUs allows avoiding scheduling of VMs for obsolete CPU models
 // +k8s:openapi-gen=true
 type HyperConvergedObsoleteCPUs struct {
-	// MinCPUModel is the Minimum CPU model that is used for basic CPU features; e.g. Penryn or Haswell
+	// MinCPUModel is the Minimum CPU model that is used for basic CPU features; e.g. Penryn or Haswell.
+	// The default value for this field is nil, but in KubeVirt, the default value is "Penryn", if nothing else is set.
+	// Use this field to override KubeVirt default value.
+	// +optional
 	MinCPUModel string `json:"minCPUModel,omitempty"`
 	// CPUModels is a list of obsolete CPU models. When the node-labeller obtains the list of obsolete CPU models, it
 	// eliminates those CPU models and creates labels for valid CPU models.
+	// The default values for this field is nil, however, HCO uses opinionated values, and adding values to this list
+	// will add them to the opinionated values.
 	// +optional
 	CPUModels []string `json:"cpuModels,omitempty"`
 }
--- a/pkg/controller/hyperconverged/hyperconverged_controller.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller.go
@@ -861,11 +861,10 @@ const (
 	imsCmName        = "v2v-vmware"
 	liveMigrationKey = "migrations"
 	vddkInitImakeKey = "vddk-init-image"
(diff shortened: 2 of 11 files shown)


pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -880,12 +879,7 @@ func (r *ReconcileHyperConverged) migrateBeforeUpgrade(req *common.HcoRequest) (
 		return false, err
 	}
 
-	cpuPluginConfigModified, err := r.migrateCPUPluginConfigurations(req)
-	if err != nil {
-		return false, err
-	}
-
-	return kvConfigMmodified || cdiConfigModified || imsConfigModified || cpuPluginConfigModified, nil
+	return kvConfigMmodified || cdiConfigModified || imsConfigModified, nil
```
orenc1, line 882: typo in variable name `kvConfigMmodified`

pkg/controller/operands/kubevirt.go
```diff
@@ -265,16 +286,21 @@ func getKVConfig(hc *hcov1beta1.HyperConverged) (*kubevirtv1.KubeVirtConfigurati
 }
 
 func getObsoleteCPUConfig(hcObsoleteCPUConf *hcov1beta1.HyperConvergedObsoleteCPUs) (map[string]bool, string) {
+	obsoleteCPUModels := make(map[string]bool)
+	for _, cpu := range hardcodedObsoleteCPUModels {
+		obsoleteCPUModels[cpu] = true
```
orenc1, line 291: why are you setting all keys on `obsoleteCPUModels` map as `true`? is there a scenario in which a cpu listed in `obsoleteCPUModels` will be `false`?

orenc1 (commented):
general question:
if the user can set `minCPU` for nodes accepting scheduling VMs, is it implying there's a list of CPU models ordered from lowest to highest rank, which contains both Intel and AMD processors?
if so, do you know where it is?

pkg/controller/operands/kubevirt.go
nunnatsa: KubeVirt API - map of trues

pkg/apis/hco/v1beta1/hyperconverged_types.go
```diff
@@ -234,9 +234,13 @@ type OperandResourceRequirements struct {
 // +k8s:openapi-gen=true
 type HyperConvergedObsoleteCPUs struct {
 	// MinCPUModel is the Minimum CPU model that is used for basic CPU features; e.g. Penryn or Haswell
```
tiraboschi, line 236: Penryn or Haswell.
Please add a dot here to make the CRD nicer.

nunnatsa: Done

tiraboschi (approved):

PR #1304, opened 2021-05-02
Switch index image to semver mode in "unstable" workflow

orenc1:
When performing an upgrade in CI from 1.4.0-unstable (upgrade-prev), the CSV version of previous version contains timestamp, thus '1.4.0' does not exist and the upgrade is failing. Moving to semver mode will allow us to upgrade from any previous version, without specifying exact previous version ('replaces') in the CSV.

Modified the build-push-images workflow to always push to `*-unstable` tags, even if we are on release branches.
We'll push to "stable" tags (e.g. `1.4.0` , `1.5.0`) only when publishing the tag in the repo.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
switch to index image in semver mode
```

### DIFF
--- a/.github/workflows/build-push-images.yaml
+++ b/.github/workflows/build-push-images.yaml
@@ -30,15 +30,13 @@ jobs:
           PACKAGE_DIR="./deploy/olm-catalog/community-kubevirt-hyperconverged"
           CSV_VERSION=$(ls -d ${PACKAGE_DIR}/*/ | sort -rV | awk "NR==$((RELEASE_DELTA+1))" | cut -d '/' -f 5)
           echo "CSV_VERSION=${CSV_VERSION}" >> $GITHUB_ENV
-      - name: set tag for master
-        if: ${{ github.ref == 'refs/heads/master' }}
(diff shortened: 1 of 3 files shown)


pkg/components/components.go
```diff
@@ -1146,7 +1146,6 @@ func GetCSVBase(params *CSVBaseParams) *csvv1alpha1.ClusterServiceVersion {
 			Description: params.Description,
 			Keywords:    []string{"KubeVirt", "Virtualization"},
 			Version:     csvVersion.OperatorVersion{Version: params.Version},
-			Replaces:    params.Replaces,
```
nunnatsa, line 1149: If this is the only place that uses `params.Replaces` please remove it completely - from the struct and from the command line arguments

orenc1: I changed it, and the removal of `replaces` field in the CSV occurs only on the workflow

PR #1325, opened 2021-05-06
Increase the initialDelay for the liveness probe

tiraboschi:
The current value for initialDelaySeconds is 5
seconds for both the readiness and liveness probes
so the first checks are going to be executed 5
seconds after the containers (operator and webhook)
have started.
failureThreshold is currently set to 1 so the first
failure will restart the container and this can
potentially cause an endless loop on really overloaded
clusters where our containers are consantly not able
to become live withing 5 seconds.

I'm increasing the initialDelaySeconds for the
livenessProbe to 10 seconds to maintain a certain
responsiveness but raising failureThreshold to 3
so that the pod will not be restarted
during its first 30 seconds.

I'm not going to touch the parameters for the
readinessProbe because it's the mechanism that
we are currently using to report back our status
to the OLM and this can potentially have
side effects.

Fixes: https://bugzilla.redhat.com/1956993

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
Increase the initialDelay for the liveness probe
```

### DIFF
--- a/deploy/index-image/community-kubevirt-hyperconverged/1.5.0/manifests/kubevirt-hyperconverged-operator.v1.5.0.clusterserviceversion.yaml
+++ b/deploy/index-image/community-kubevirt-hyperconverged/1.5.0/manifests/kubevirt-hyperconverged-operator.v1.5.0.clusterserviceversion.yaml
@@ -1853,12 +1853,12 @@ spec:
                 image: +IMAGE_TO_REPLACE+
                 imagePullPolicy: IfNotPresent
                 livenessProbe:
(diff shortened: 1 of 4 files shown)


nunnatsa (approved):
/approve

PR #1356, opened 2021-05-25
Fix upgrade path to unstable versions

orenc1:
When upgrading HCO from a stable version, to an unstable one, by switching the channel, OLM won't start the upgrade becasue the old CSV is not in the new channel.
To bypass this, if we skip all versions prior to the unstable version, the OLM will upgrade when switching the subscription channel.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
NONE
```

### DIFF
--- a/tools/csv-merger/csv-merger.go
+++ b/tools/csv-merger/csv-merger.go
@@ -281,7 +281,7 @@ func getHcoCsv() {
 	csvBase := components.GetCSVBase(csvParams)
 
 	if *enableUniqueSemver {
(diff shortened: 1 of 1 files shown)


nunnatsa (approved):
approve

PR #1388, opened 2021-06-10
Exit the upgrade mode only when nothing is still to be written on the HCO CR

tiraboschi:
Effectively exit the upgrade mode only when we are sure that 
nothing is still to be written on the HCO CR
because we are going to write the status subresource before
really updating the HCO CR and so we don't want to report
a wrong status if the update on the HCO CR is going to fail
for any reason.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
Exit the upgrade mode only when nothing is still to be written on the HCO CR
```


nunnatsa (approved):
/approve

PR #1396, opened 2021-06-14
Remove descriptions from unstored versions of CRDs

tiraboschi:
Remove descriptions from unstored versions of CRDs
to keep the bundle size < 1MiB.

Until https://github.com/operator-framework/enhancements/pull/40
got implemented, a bundle should not exceed 1MiB
because its internally implemented with a configmap.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:
```release-note
Remove descriptions from unstored versions of CRDs
```

### DIFF
--- a/deploy/crds/cluster-network-addons00.crd.yaml
+++ b/deploy/crds/cluster-network-addons00.crd.yaml
@@ -1,4 +1,3 @@
----
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
@@ -10,6 +9,7 @@ spec:
     listKind: NetworkAddonsConfigList
     plural: networkaddonsconfigs
     singular: networkaddonsconfig
+  preserveUnknownFields: false
   scope: Cluster
   versions:
   - name: v1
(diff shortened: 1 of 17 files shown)


deploy/crds/hco01.crd.yaml
```diff
@@ -18,16 +17,16 @@ spec:
         description: V2VVmware is the Schema for the v2vvmwares API
         properties:
           apiVersion:
-            description: |-
-              APIVersion defines the versioned schema of this representation
-                                      of an object. Servers should convert recognized schemas to the latest
-                                      internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#resources
+            description: "APIVersion defines the versioned schema of this representation\n\
```
orenc1, line 20: Why we are modifying that file (and others) if there is only one stored version in the CRD?

tiraboschi: right, good point

nunnatsa (approved):
/approve

PR #1409, opened 2021-06-22
Split writing HyperConverged's status and body

nunnatsa:
split writing HyperConverged's status and body (spec/metadata)

Since writing to the HyperConverged status must be a separate write and we can't write the whole custom resource as an atomic operation, it caused several instabilities and conflicts.

This PR split the right to two calls - if the code modified something in the body (should be with very low frequency, except for during upgrade), HCO will write only the body, and exit with requeue = true to force additional call, then we expect the body not to be modified again, letting HCO write only the status this time.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
None
```

### DIFF
--- a/hack/generate_local_env.py
+++ b/hack/generate_local_env.py
@@ -33,7 +33,7 @@ def get_env_file(outdir, file_format='txt'):
         for row in reader:
             if row['image_var'] in ['VMWARE_IMAGE', 'CONVERSION_IMAGE', 'KUBEVIRT_VIRTIO_IMAGE']:
                 image = f"{row['name']}@sha256:{row['digest']}"
-                env = 'VIRTIOWIN_CONTAINER' if row['image_var'] == 'KUBEVIRT_VIRTIO_IMAGE' else row['image_var']
+                env = 'VIRTIOWIN_CONTAINER' if row['image_var'] == 'KUBEVIRT_VIRTIO_IMAGE' else row['image_var'].replace("_IMAGE", "_CONTAINER")
                 vars_list.append(f"{env}={image}")
 
     vars_list.extend([
--- a/pkg/controller/hyperconverged/hyperconverged_controller.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller.go
@@ -17,7 +17,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/yaml"
-	"k8s.io/client-go/tools/record"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
@@ -93,7 +92,6 @@ func newReconciler(mgr manager.Manager, ci hcoutil.ClusterInfo) reconcile.Reconc
 	return &ReconcileHyperConverged{
 		client:         mgr.GetClient(),
 		scheme:         mgr.GetScheme(),
-		recorder:       mgr.GetEventRecorderFor(hcoutil.HyperConvergedName),
 		operandHandler: operands.NewOperandHandler(mgr.GetClient(), mgr.GetScheme(), ci.IsOpenshift(), hcoutil.GetEventEmitter()),
 		upgradeMode:    false,
 		ownVersion:     ownVersion,
@@ -176,7 +174,6 @@ type ReconcileHyperConverged struct {
 	// that reads objects from the cache and writes to the apiserver
 	client         client.Client
 	scheme         *runtime.Scheme
-	recorder       record.EventRecorder
 	operandHandler *operands.OperandHandler
 	upgradeMode    bool
 	ownVersion     string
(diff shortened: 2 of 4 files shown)


pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -44,6 +43,7 @@ import (
 var (
 	log               = logf.Log.WithName("controller_hyperconverged")
 	randomConstSuffix = ""
+	// Upgrade is ALWAYS to the running software, so it may happened only once.
```
orenc1, line 46: what is the context of this comment?

hack/generate_local_env.py
```diff
@@ -33,7 +33,7 @@ def get_env_file(outdir, file_format='txt'):
         for row in reader:
             if row['image_var'] in ['VMWARE_IMAGE', 'CONVERSION_IMAGE', 'KUBEVIRT_VIRTIO_IMAGE']:
                 image = f"{row['name']}@sha256:{row['digest']}"
-                env = 'VIRTIOWIN_CONTAINER' if row['image_var'] == 'KUBEVIRT_VIRTIO_IMAGE' else row['image_var']
+                env = 'VIRTIOWIN_CONTAINER' if row['image_var'] == 'KUBEVIRT_VIRTIO_IMAGE' else row['image_var'].replace("_IMAGE", "_CONTAINER")
```
orenc1, line 36: why not storing them in `deploy/images.csv` with the proper suffix at first place?

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -366,40 +368,30 @@ func (r *ReconcileHyperConverged) getHyperConverged(req *common.HcoRequest) (*hc
...
 	// Moreover, we need to update the status first, in order to prevent a conflict.
 	// In addition, spec changes are removed by status update, but since status update done first, we need
 	// to store the spec and recover it after status update
 
-	var spec hcov1beta1.HyperConvergedSpec
 	if request.Dirty {
-		request.Instance.Spec.DeepCopyInto(&spec)
+		err := r.updateHyperConvergedSpecMetadata(request)
+		if err != nil {
+			r.logHyperConvergedUpdateError(request, err, "Failed to update HCO CR")
```
orenc1, line 382: maybe be more specific:
"Failed to update HCO CR spec and metadata"

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -456,8 +448,6 @@ func (r *ReconcileHyperConverged) validateNamespace(req *common.HcoRequest) (boo
 
 func (r *ReconcileHyperConverged) setInitialConditions(req *common.HcoRequest) {
 	req.Instance.Status.UpdateVersion(hcoVersionName, r.ownVersion)
-	req.Instance.Spec.Version = r.ownVersion
-	req.Dirty = true
```
orenc1, line 460: why are removing setting the dirty flag to `true` if you update the CR status?

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -366,40 +368,30 @@ func (r *ReconcileHyperConverged) getHyperConverged(req *common.HcoRequest) (*hc
...
 	// In addition, spec changes are removed by status update, but since status update done first, we need
 	// to store the spec and recover it after status update
 
-	var spec hcov1beta1.HyperConvergedSpec
 	if request.Dirty {
-		request.Instance.Spec.DeepCopyInto(&spec)
+		err := r.updateHyperConvergedSpecMetadata(request)
+		if err != nil {
+			r.logHyperConvergedUpdateError(request, err, "Failed to update HCO CR")
+			return false, err
```
orenc1, line 383: why not requeue in case the spec/metadata failed to be updated?

orenc1 (commented):
ok in general, just a few comments / questions

pkg/controller/hyperconverged/hyperconverged_controller.go
nunnatsa: `Dirty` is for the spec/metadata. `StatusDirty` is dedicated to status changes.

pkg/controller/hyperconverged/hyperconverged_controller.go
nunnatsa: need to be removed.

hack/generate_local_env.py
nunnatsa: This is for running locally. This is a development issue. changing the tool rather than the product.

pkg/controller/hyperconverged/hyperconverged_controller.go
nunnatsa: look t line 228. We're queuing only in update conflict, in order to avoid endless loop, if something went terribly wrong.

pkg/controller/hyperconverged/hyperconverged_controller.go
nunnatsa: We're logging the original error. I think it's too much information for the user.

pkg/controller/hyperconverged/hyperconverged_controller.go
orenc1: and if you have something to update in this iteration that was failed not due to conflict error, are we dismissing this change?

nunnatsa: Yes (and this is not a change. This is the current behavior)

orenc1: just wanted to be sure. thanks.
/approve
/lgtm

PR #1410, opened 2021-06-23
Only update the conditions if they are actually modified

nunnatsa:
This PR only updates the conditions if they are changed from the existing conditions. That gives a good understanding on when thing happened, because the timestamp is the time of the last change of the condition's state. If the state was not changed, the timestamp wouldn't be changed either.

API Changes:
The status contains a new `observedGeneration` field. This field reflects the HyperConverged resource generation. If the ObservedGeneration is less than the resource generation in metadata, the status is out of date.

**Reviewer Checklist**

- [x] PR Message
- [x] Commit Messages
- [ ] How to test
- [x] Unit Tests
- [ ] Functional Tests
- [x] User Documentation
- [ ] Developer Documentation
- [x] Upgrade Scenario
- [x] Uninstallation Scenario
- [x] Backward Compatibility
  
**Release note**:

```release-note
None
```

### DIFF
--- a/deploy/crds/hco00.crd.yaml
+++ b/deploy/crds/hco00.crd.yaml
@@ -2377,6 +2377,12 @@ spec:
                   - type
                   type: object
                 type: array
+              observedGeneration:
+                description: ObservedGeneration reflects the HyperConverged resource
+                  generation. If the ObservedGeneration is less than the resource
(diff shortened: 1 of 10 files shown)


orenc1 (approved):
/approve

PR #1412, opened 2021-06-27
Test that VM created before upgrade, is still runninig after the upgrade is done

nunnatsa:
In the upgrade test, create a VM before starting the upgrade and make sure it's running. 

Then, after the upgrade is successfully done, checks that the VM is still running.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
None
```

### DIFF
--- a/hack/upgrade-test-index-image.sh
+++ b/hack/upgrade-test-index-image.sh
@@ -31,7 +31,7 @@
 # and performs various validations against the upgraded version.
 
 
-MAX_STEPS=17
+MAX_STEPS=$(( $(grep -c "Msg " "$0") - 2)) # subtract self line and the function name
 CUR_STEP=1
 RELEASE_DELTA="${RELEASE_DELTA:-1}"
 HCO_DEPLOYMENT_NAME=hco-operator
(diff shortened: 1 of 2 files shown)


hack/upgrade-test-index-image.sh
```diff
@@ -206,6 +224,14 @@ ${CMD} get pod $HCO_CATALOGSOURCE_POD -n ${HCO_CATALOG_NAMESPACE} -o yaml | grep
 
 dump_sccs_after
 
+Msg "make sure that the VM is still running, after the upgrade"
```
tiraboschi, line 227: I think we have also to add a bit of logic to ensure that the VM didn't got restarted in the middle

tiraboschi: we can do this with a second PR

tiraboschi (approved):

PR #1413, opened 2021-06-27
Remove the spec.version field as it's not needed anymore

nunnatsa:
Openshift UI used the spec.version to present the operator version. That feature was removed, so there is no reason for this field anymore.

Placing the version in the spec was problematic for two reasons:
1. the user can get the wrong impression that it is possible to upgrade the HCO version by changing its version, because the spec is for the user to edit.
2. we need to sync the spec.version and the status version, but can't do that as an atomic write, but in two different update operation, and that leads to edge cases.

/hold
@tiraboschi - please review

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
API Change: removed the version field from the HyperConverged's spec
```

### DIFF
--- a/deploy/crds/hco00.crd.yaml
+++ b/deploy/crds/hco00.crd.yaml
@@ -1292,9 +1292,6 @@ spec:
                 description: VDDK Init Image eventually used to import VMs from external
                   providers
                 type: string
-              version:
-                description: operator version
-                type: string
               workloadUpdateStrategy:
                 default:
                   batchEvictionInterval: 1m0s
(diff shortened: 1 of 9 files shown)


nunnatsa:
small coverage reduce because fo removed code
/override coverage/coveralls

tiraboschi (approved):

PR #1418, opened 2021-07-01
add new API field CommonTemplatesNamespace

ksimon1:
add new API field CommonTemplatesNamespace
This new field defines in which namespace will be templates deployed

/cc @tiraboschi @nunnatsa can you please help me, if these are all the places that need to be updated?

Signed-off-by: Karel Šimon <ksimon@redhat.com>

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```
API Change: add commonTemplatesNamespace field to the HyperConverged's spec
```

### DIFF
--- a/pkg/controller/hyperconverged/hyperconverged_controller_test.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller_test.go
@@ -5,10 +5,11 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	apimetav1 "k8s.io/apimachinery/pkg/api/meta"
-	"k8s.io/apimachinery/pkg/util/yaml"
 	"os"
 	"time"
+
+	apimetav1 "k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/util/yaml"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/ginkgo/extensions/table"
@@ -23,6 +24,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/reference"
 	cdiv1beta1 "kubevirt.io/containerized-data-importer/pkg/apis/core/v1beta1"
+	sspv1beta1 "kubevirt.io/ssp-operator/api/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	// TODO: Move to envtest to get an actual api server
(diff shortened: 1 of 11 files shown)


pkg/controller/operands/ssp.go
```diff
@@ -168,13 +168,18 @@ func (h *sspHooks) updateCr(req *common.HcoRequest, client client.Client, exists
 
 func NewSSP(hc *hcov1beta1.HyperConverged, opts ...string) *sspv1beta1.SSP {
```
nunnatsa, line 169: This is a good test, but please also add test/test case in pkg/controller/hyperconverged/hyperconverged_controller_test.go

nunnatsa (changes requested):
Thanks for the PR, it looks good in general.

A few comments:
1. Please extend the PR comment to explain what is this new field and what is the reason to add it.
2. Please run `openapi-gen` and `deepcopy-gen`
3. Please document the new field in docs/cluster-configuration.md

pkg/controller/operands/ssp_test.go
```diff
@@ -79,13 +79,14 @@ var _ = Describe("SSP Operands", func() {
...
 			expectedResource := NewSSP(hco)
 			existingResource := expectedResource.DeepCopy()
 			existingResource.ObjectMeta.SelfLink = fmt.Sprintf("/apis/v1/namespaces/%s/dummies/%s", existingResource.Namespace, existingResource.Name)
 
 			replicas := int32(defaultTemplateValidatorReplicas * 2) // non-default value
 			existingResource.Spec.TemplateValidator.Replicas = &replicas
-			existingResource.Spec.CommonTemplates.Namespace = "foobar"
 			existingResource.Spec.NodeLabeller.Placement = &lifecycleapi.NodePlacement{
 				NodeSelector: map[string]string{"foo": "bar"},
 			}
```
nunnatsa, line 92: This is a good test, but please also add test/test case in pkg/controller/hyperconverged/hyperconverged_controller_test.go

pkg/controller/hyperconverged/hyperconverged_controller_test.go
```diff
@@ -279,6 +283,34 @@ var _ = Describe("HyperconvergedController", func() {
...
+				Eventually(func() bool {
+					err := cl.List(context.TODO(),
+						templatesList,
+						client.InNamespace(namespace),
+						client.MatchingLabels{
+							templateTypeLabel: base,
+						})
+					return len(templatesList.Items) > 0 && err == nil
+				}, 30*time.Second).Should(BeTrue(), "templates should be deployed in different namespace")
+			})
```
nunnatsa, line 313: This test is a unit test, without a running cluster. There is no running SSP to implement the logic under test. Also, as this is a unit test, it should not test other components. Please just make sure the required fields are set in the SSP CR.

pkg/controller/hyperconverged/hyperconverged_controller_test.go
```diff
@@ -494,9 +526,9 @@ var _ = Describe("HyperconvergedController", func() {
 					Expect(requeue).To(BeFalse())
 					checkAvailability(foundResource, metav1.ConditionTrue)
 				})
-				By("Check KV", func() {
+				By("Check SSP", func() {
```
nunnatsa, line 529: What is the reason for this change?

nunnatsa (changes requested):

pkg/controller/hyperconverged/hyperconverged_controller_test.go
ksimon1: because the test is checking ssp not kubevirt

pkg/controller/hyperconverged/hyperconverged_controller_test.go
ksimon1: ah I see, I updated the test

nunnatsa (approved):
/approve

nunnatsa:
Adding a new parameter to the API reduces the coverage of the auto-generated code. 
/override coverage/coveralls

PR #1421, opened 2021-07-02
Add new controller for dashboard configmaps

erkanerol:

This PR adds a new handler for dashboard configmaps. It checks a manifest directory and loads the desired configmaps from that directory. The default location is `./dashboard` and it can be overriden by setting environment variable `DASHBOARD_FILES_LOCATION`.

This PR introduces usage of configuration file for functional tests. It gives the desired spec with the file `/hack/testFiles/test_config.yaml`

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  

**Release note**:
```release-note
Add new controllers for reconciling ConfigMaps of OCP Dashboard
```


nunnatsa (commented):
It seems that the version of the hook interface used here is out of date (e.g. `validate` is no longer part of the interface). Also, implementing the `hcoResourceHooks` interface is enough for a configmap. No need to implement `hcoOperandHooks`.

cmd/hyperconverged-cluster-operator/main.go
```diff
@@ -152,7 +152,6 @@ func getNewManagerCache(operatorNamespace string) cache.NewCacheFunc {
 				},
 				&corev1.ConfigMap{}: {
 					Label: labelSelector,
-					Field: namespaceSelector,
```
nunnatsa, line 155: This is risky. It will cause HCO to hold a cache of all the configmaps in the system.

erkanerol: "all the configmaps in the system" or "all the configmaps in the system which has `app=kubevirt-hyperconverged` labels?

tiraboschi: if we remove the namespace selector the cache will watch only the configmaps with label `app=kubevirt-hyperconverged` but now in whatever namespace.
If we know that we have to watch for configmaps only in a specific additional namespace, we can add it to the namespaceSelector set.

erkanerol: As far as I see  from code of apimachinery, it is not possible to build OR selector for fields. https://github.com/kubernetes/kubernetes/issues/32946

erkanerol: I tried to implement `orTerm` like `andTerm` here https://github.com/kubernetes/apimachinery/blob/master/pkg/fields/selector.go#L179 but it is not easy. Selectors in `fields` package are written for only "equals" or "not equals" operators. I tried to implement a new selector but the interface is bundled with `Requirements` in the same package and the requirements are supposed to have "AND" relation. See https://github.com/kubernetes/apimachinery/blob/master/pkg/fields/requirements.go#L21

On the other hand, `labels` package have some more operators like `in` , `notin` etc. https://github.com/kubernetes/apimachinery/blob/master/pkg/labels/selector.go#L135

erkanerol: At request of @tiraboschi I did an experiment for this issue. I created 100 namespace and 100 configmap per namespace. (total is more than 10K). I deployed a version of hco without this PR, deleted/re-created HCO 30-40 times and checked CPU&memory consumption via Prometheus. Then I deployed a version of hco with this PR and did the same things again. I compared the results and observed that there is no significant increase after this PR.

PR #1441, opened 2021-07-14
Move rbacs for config reader to a separate handler

erkanerol:

With this PR, hco operator w,ill start reconciling `hco.kubevirt.io:config-reader` Role&RoleBinding strictly like other operands. After this PR, changes on these operands will be overwritten immediately without requiring any change on CDI or HCO CRs. 

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1963963

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/operands/cdi.go
+++ b/pkg/controller/operands/cdi.go
@@ -99,8 +99,6 @@ func getDefaultFeatureGates() []string {
 	return []string{HonorWaitForFirstConsumerGate}
 }
 
-func (h *cdiHooks) postFound(req *common.HcoRequest, exists runtime.Object) error { return nil }
-
 func NewCDI(hc *hcov1beta1.HyperConverged, opts ...string) (*cdiv1beta1.CDI, error) {
 	uninstallStrategy := cdiv1beta1.CDIUninstallStrategyBlockUninstallIfWorkloadsExist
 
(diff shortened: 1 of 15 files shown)


nunnatsa:
I think we need to check if we can take it out of the code entirely, let OLM manage it. Can we move it to the CSV?

erkanerol:
> We need to check if we can entirely take it out of the code, letting OLM manage it. The main problem with creating rbacs from the code is that HCO must have permissions for that, and we need to avoid such permmissions. Can we move it to the CSV?

@nunnatsa  My understanding from the discussions in the PR below, it is not possible to create Role&RoleBinding with CSV
https://github.com/kubevirt/hyperconverged-cluster-operator/pull/498

pkg/controller/operands/cdi.go
```diff
@@ -104,19 +99,7 @@ func getDefaultFeatureGates() []string {
...
-	}
-
-	err = h.ensureKubeVirtStorageRoleBinding(req)
-	if err != nil {
-		return err
-	}
-
-	return nil
-}
+func (h *cdiHooks) postFound(req *common.HcoRequest, exists runtime.Object) error { return nil }
```
nunnatsa, line 102: This is the one and an only real implementation of the `postFound` function. Please remove it from the interface and delete all the empty implementations (and the the function call too).

tiraboschi:
> > We need to check if we can entirely take it out of the code, letting OLM manage it. The main problem with creating rbacs from the code is that HCO must have permissions for that, and we need to avoid such permmissions. Can we move it to the CSV?
> 
> @nunnatsa My understanding from the discussions in the PR below, it is not possible to create Role&RoleBinding with CSV
> #498

If I correctly recall the specific problem was that a serviceAccountName was mandatory in the CSV while the console is impersonating the connected user while reading that config map.
Can you please double check if this is still relevant on CSV side?

nunnatsa:
> Can you please double check if this is still relevant on CSV side?

As far as I can see, this is still the case.

nunnatsa (changes requested):
Please see one comment about `postFound`

pkg/controller/operands/cdi.go
erkanerol: Done

nunnatsa (approved):
/approve

PR #1483, opened 2021-08-17
Remove controller-runtime dependency from api package

erkanerol:
Because of webhook codes, we have a dependency on
contoller-runtime in our api package. Api packages should
be as clean as possible. Also, we are hitting an issue*
in test when we try to consume our api package because of
this dependency.

* https://github.com/kubernetes-sigs/controller-runtime/issues/878

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
NONE
```


pkg/webhooks/handler/webhooks.go
```diff
@@ -28,6 +28,12 @@ const (
 	updateDryRunTimeOut = time.Second * 3
 )
 
+type WebhookHandlerIfs interface {
+	v1beta1.ValidatorWebhookHandler
+	Init(logger logr.Logger, cli client.Client, namespace string, isOpenshift bool)
+	HandleMutatingNsDelete(ns *corev1.Namespace, dryRun bool) (bool, error)
```
erkanerol, line 34: @nunnatsa  we may think about moving this nsmutator into another interface in another package.

pkg/webhooks/setup.go
```diff
@@ -0,0 +1,110 @@
...
+
+func SetupWebhookWithManager(ctx context.Context, mgr ctrl.Manager, isOpenshift bool) error {
+	operatorNsEnv, nserr := hcoutil.GetOperatorNamespaceFromEnv()
+	if nserr != nil {
+		logger.Error(nserr, "failed to get operator namespace from the environment")
+		return nserr
+	}
+
+	whHandler := validator.WebhookHandler{}
+	whHandler.Init(logger, mgr.GetClient(), operatorNsEnv, isOpenshift)
```
nunnatsa, line 45: Please replace with a single `NewWHValidator` function. There is no point to construct in two lines.

pkg/webhooks/setup.go
```diff
@@ -0,0 +1,110 @@
...
+	hcov1beta1.SetValidatorWebhookHandler(whHandler)
+
+	nsMutator := mutator.NsMutator{}
+	nsMutator.Init(mgr.GetClient(), operatorNsEnv)
```
nunnatsa, line 49: Here too

nunnatsa (commented):
Looks very good. Added 3 comments, that are actually the same: use a "constructor" (`New...` function) to instantiate the `whHandler` and the `nsMutator`.

PR #1513, opened 2021-08-31
Cancel 2-phase init of some reources

nunnatsa:
The client from the manager can't be used before its cache is populated. That prevents some initialization operation, like setting up the `clientInfo` and the event emitter. The solution so far was to perform the initialization in two phases - one on boot up and the second on the first reconciliation loop.

But it is possible to get a client without cache that will work on boot up. This PR uses such a client to remove the late initialization and perform everything on boot time.

This is mostly preparation for the operator-condition implementation, that must setup the operator condition on boot up.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
None
```

### DIFF
--- a/cmd/cmdcommon/cmdcommon.go
+++ b/cmd/cmdcommon/cmdcommon.go
@@ -109,10 +109,6 @@ func (h HcCmdHelper) ExitOnError(err error, message string, keysAndValues ...int
 	}
 }
 
-func (h HcCmdHelper) IsRunInLocal() bool {
-	return h.runInLocal
-}
-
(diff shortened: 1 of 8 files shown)


nunnatsa:
/override coverage/coveralls

coverall is not working at the moment

tiraboschi (approved):

PR #1530, opened 2021-09-15
Add new recording rule "cluster:vmi_request_cpu_cores:sum"

erkanerol:
We want to add a new metric into cluster-monitoring-operator to track sum of cpu requests of virt-launcher pods.

"cluster" prefix means that the scope of the metric is cluster-wide like the other metrics in the cluster-monitoring-operator. "cnv" is not used here since it is not related to this upstream project 

Details:
- It relies on cpu requests on pod specs.
- This recording rule only accounts virt-launcher pods and ignores pods of control-plane.
- It only accounts "Running" pods. Other states (Pending, Succeeded, Failed) are ignored.
- It ignores initContainers

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
Add  new recording rule ("cnv:vmi_request_cpu_cores:sum" ) to keep track of sum of allocated CPUs to VMIs.
```

### DIFF
--- a/hack/prom-rule-ci/prom-rules-tests.yaml
+++ b/hack/prom-rule-ci/prom-rules-tests.yaml
@@ -269,4 +269,64 @@ tests:
           exp_labels:
             severity: "info"
             annotation_name: "networkaddonsconfigs.kubevirt.io/jsonpatch"
-
+  # Test recording rule
+  - interval: 1m
+    input_series:
+      # take all containers of running virt-launcher pods
+      - series: 'kube_pod_container_resource_requests{pod="virt-launcher-x-y", resource="cpu", container="compute"}'
+        # time:  0   1   2   3   4   5
+        values: "0.1 0.1 0.1 0.1 0.1 0.1"
+      # take all containers of running virt-launcher pods
+      - series: 'kube_pod_container_resource_requests{pod="virt-launcher-x-y", resource="cpu", container="volumecontainerdisk"}'
+        # time:  0    1    2    3    4    5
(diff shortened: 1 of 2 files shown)


erkanerol:
@sradco  Could you please review this PR?

hack/prom-rule-ci/prom-rules-tests.yaml
```diff
@@ -269,4 +269,64 @@ tests:
...
+        # time:  0     1     2     3   4    5
+        values: "stale stale stale 0.01 0.01 0.01"
+      - series: 'kube_pod_labels{label_kubevirt_io="virt-launcher",pod="virt-launcher-new"}'
+        # time: 0      1     2     3 4 5
+        values: "stale stale stale 1 1 1"
+      - series: 'kube_pod_status_phase{phase="Running", pod="virt-launcher-new"}'
+        # time:  0     1     2     3 4 5
+        values: "stale stale stale 1 0 1"
+    promql_expr_test:
+      - expr: 'cnv:vmi_request_cpu_cores:sum'
```
tiraboschi, line 310: I think we should avoid calling it CNV in upstream code

erkanerol: done.

tiraboschi (approved):

PR #1586, opened 2021-11-08
Apply upgrade patches according to a yaml file

tiraboschi:
Apply upgrade patches according to a yaml files
shipped within the container.
The operator will read the yaml file expecting
a fixed struct.

Now it handles only jsonpatches to be applied
on the HCO CR, in the future the same mechanism
could be easily extended to delete leftovers
or patch other objects.

Now it expects something like:
```json
{
  "hcoCRPatchList": [
    {
      "semverRange": ">=1.4.0 <=1.5.0",
      "jsonPatch": [
        {
          "op": "replace",
          "path": "/spec/featureGates/sriovLiveMigration",
          "value": true
        }
      ]
    }
  ]
}
```
where hcoCRPatchList is a list of upgrade patches.
Each hcoCRPatch consists in a semver range
( https://github.com/blang/semver#ranges ) of affected
source versions and a json patch
( https://datatracker.ietf.org/doc/html/rfc6902 )
to be applied during the upgrade if relevant.

Applying it to set
`/spec/featureGates/sriovLiveMigration = true`
when upgrading from 1.4.z and 1.5.0 to the current version.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2018468

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
Apply upgrade patches according to a yaml file
```

### DIFF
--- a/build/Dockerfile.okd
+++ b/build/Dockerfile.okd
@@ -5,6 +5,6 @@ ENV KUBEVIRT_CLIENT_GO_SCHEME_REGISTRATION_VERSION=v1
 COPY hyperconverged-cluster-operator /usr/bin/
 COPY hack/testFiles/test_quickstart.yaml quickStart/
 COPY hack/testFiles/test_dashboard_cm.yaml dashboard/
-COPY assets/dataImportCronTemplates dataImportCronTemplates/
+COPY assets/ .
 
 ENTRYPOINT /usr/bin/hyperconverged-cluster-operator
--- a/pkg/controller/hyperconverged/hyperconverged_controller.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller.go
@@ -2,10 +2,13 @@ package hyperconverged
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"os"
 	"reflect"
 	"strings"
+
+	"github.com/blang/semver/v4"
 
 	jsonpatch "github.com/evanphx/json-patch"
 	"github.com/google/uuid"
@@ -375,6 +378,12 @@ func (r *ReconcileHyperConverged) doReconcile(req *common.HcoRequest) (reconcile
 	req.SetUpgradeMode(r.upgradeMode)
 
 	if r.upgradeMode {
+
+		err = validateUpgradePatches(req)
+		if err != nil {
+			return reconcile.Result{Requeue: true}, err
+		}
+
 		crdStatusUpdated, err := r.updateCrdStoredVersions(req)
 		if err != nil {
 			return reconcile.Result{Requeue: true}, err
@@ -1066,9 +1075,14 @@ func (r *ReconcileHyperConverged) migrateBeforeUpgrade(req *common.HcoRequest) (
 		return false, err
 	}
 
+	upgradePatched, err := r.applyUpgradePatches(req)
+	if err != nil {
+		return false, err
+	}
+
 	removeOldQuickStartGuides(req, r.client, r.operandHandler.GetQuickStartNames())
 
-	return kvConfigModified || cdiConfigModified || defaultsAmended, nil
+	return kvConfigModified || cdiConfigModified || defaultsAmended || upgradePatched, nil
 }
 
 func (r ReconcileHyperConverged) migrateKvConfigurations(req *common.HcoRequest) (bool, error) {
@@ -1118,6 +1132,62 @@ func (r ReconcileHyperConverged) amendBadDefaults(req *common.HcoRequest) (bool,
 		req.Dirty = true
 	}
 	return modified, nil
+}
+
+func (r ReconcileHyperConverged) applyUpgradePatches(req *common.HcoRequest) (bool, error) {
+	modified := false
+
+	knownHcoVersion, _ := req.Instance.Status.GetVersion(hcoVersionName)
+	if knownHcoVersion == "" {
+		knownHcoVersion = "0.0.0"
+	}
+	knownHcoSV, err := semver.Parse(knownHcoVersion)
+	if err != nil {
+		req.Logger.Error(err, "Error!")
+		return false, err
+	}
+
+	hcoJson, err := json.Marshal(req.Instance)
+	if err != nil {
+		return false, err
+	}
+
+	for _, p := range hcoUpgradeChanges.HCOCRPatchList {
+		hcoJson, err = r.applyUpgradePatch(req, hcoJson, knownHcoSV, p)
+		if err != nil {
+			return false, err
+		}
+	}
+
+	tmpInstance := &hcov1beta1.HyperConverged{}
+	err = json.Unmarshal(hcoJson, tmpInstance)
+	if err != nil {
+		return false, err
+	}
+	if !reflect.DeepEqual(tmpInstance.Spec, req.Instance.Spec) {
+		req.Logger.Info("updating HCO spec as a result of upgrade patches")
+		tmpInstance.Spec.DeepCopyInto(&req.Instance.Spec)
+		modified = true
+		req.Dirty = true
+	}
+
+	return modified, nil
+}
+
+func (r ReconcileHyperConverged) applyUpgradePatch(req *common.HcoRequest, hcoJson []byte, knownHcoSV semver.Version, p hcoCRPatch) ([]byte, error) {
+	affectedRange, err := semver.ParseRange(p.SemverRange)
+	if err != nil {
+		return hcoJson, err
+	}
+	if affectedRange(knownHcoSV) {
+		req.Logger.Info("applying upgrade patch", "knownHcoSV", knownHcoSV, "affectedRange", affectedRange, "patches", p.JSONPatch)
+		patchedBytes, err := p.JSONPatch.Apply(hcoJson)
+		if err != nil {
+			return hcoJson, err
+		}
+		return patchedBytes, nil
+	}
+	return hcoJson, nil
 }
 
 func (r *ReconcileHyperConverged) removeConfigMap(req *common.HcoRequest, cm *corev1.ConfigMap) error {
--- a/pkg/controller/hyperconverged/test-files/upgradePatches/upgradePatches.json
+++ b/pkg/controller/hyperconverged/test-files/upgradePatches/upgradePatches.json
@@ -1,0 +1,1 @@
+../../../../../assets/upgradePatches.json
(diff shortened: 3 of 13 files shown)


build/Dockerfile.okd
```diff
@@ -6,5 +6,6 @@ COPY hyperconverged-cluster-operator /usr/bin/
 COPY hack/testFiles/test_quickstart.yaml quickStart/
 COPY hack/testFiles/test_dashboard_cm.yaml dashboard/
 COPY assets/dataImportCronTemplates dataImportCronTemplates/
+COPY assets/upgradePatches.yaml upgradePatches.yaml
```
nunnatsa, line 9: Need to check, but I think 
```
COPY assets/ .
```
will do the same

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -2,7 +2,9 @@ package hyperconverged
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"github.com/blang/semver/v4"
```
nunnatsa, line 7: please `goimports -w -local="kubevirt.io,github.com/kubevirt,github.com/kubevirt/hyperconverged-cluster-operator" pkg/controller/hyperconverged/hyperconverged_controller.go`

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -1120,6 +1127,74 @@ func (r ReconcileHyperConverged) amendBadDefaults(req *common.HcoRequest) (bool,
...
+	}
+
+	err := readUpgradePatchesFromFile(req)
+	if err != nil {
+		return false, err
+	}
+	err = validateUpgradePatches(req)
+	if err != nil {
+		return false, err
+	}
```
nunnatsa, line 1145: Can we do it in a single function, for code simplicity?

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -1120,6 +1127,74 @@ func (r ReconcileHyperConverged) amendBadDefaults(req *common.HcoRequest) (bool,
...
+
+	knownHcoSV, err := semver.Parse(knownHcoVersion)
+	if err != nil {
+		req.Logger.Error(err, "Error!")
+		return false, err
+	}
```
nunnatsa, line 1151: Please move up, just after reading the version. We don't want to do more actions if we know we need to exit anyway.

pkg/controller/hyperconverged/hyperconverged_controller.go
nunnatsa, line 1145: Can we do it only once, when getting into upgrade mode, and cache the result?

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -1120,6 +1127,74 @@ func (r ReconcileHyperConverged) amendBadDefaults(req *common.HcoRequest) (bool,
...
+	if affectedRange(knownHcoSV) {
+		patches, err := jsonpatch.DecodePatch([]byte(p.JSONPatch))
+		if err != nil {
+			return modified, err
+		}
+		req.Logger.Info("applying upgrade patch", "knownHcoSV", knownHcoSV, "affectedRange", affectedRange, "patches", patches)
+		specBytes, err := json.Marshal(req.Instance)
+		if err != nil {
+			return modified, err
+		}
```
nunnatsa, line 1178: Please take it outside the loop

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -1120,6 +1127,74 @@ func (r ReconcileHyperConverged) amendBadDefaults(req *common.HcoRequest) (bool,
...
+		patchedBytes, err := patches.Apply(specBytes)
+		if err != nil {
+			return modified, err
```
nunnatsa, line 1181: maybe `return false, err` ?

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -1120,6 +1127,74 @@ func (r ReconcileHyperConverged) amendBadDefaults(req *common.HcoRequest) (bool,
...
+		}
+		tmpInstance := &hcov1beta1.HyperConverged{}
+		err = json.Unmarshal(patchedBytes, tmpInstance)
+		if err != nil {
+			return modified, err
```
nunnatsa, line 1186: maybe `return false, err` ?

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -1120,6 +1127,74 @@ func (r ReconcileHyperConverged) amendBadDefaults(req *common.HcoRequest) (bool,
...
+		}
+		if !reflect.DeepEqual(tmpInstance.Spec, req.Instance.Spec) {
+			req.Logger.Info("updating HCO spec as a result of upgrade patch", "knownHcoSV", knownHcoSV, "affectedRange", affectedRange, "patches", patches)
+			tmpInstance.Spec.DeepCopyInto(&req.Instance.Spec)
+			modified = true
+			req.Dirty = true
+		}
```
nunnatsa, line 1193: move after the loop

assets/upgradePatches.yaml
```diff
@@ -0,0 +1,11 @@
...
+hcoCRPatchList:
+  - semverRange: '>=1.4.0 <=1.5.0'
+    jsonPatch: |-
+      [
+        {
+          "op": "replace",
+          "path": "/spec/featureGates/sriovLiveMigration",
+          "value": true
+        }
+      ]
```
nunnatsa, line 11: Since everything is json, maybe the file should be a json file as well?

pkg/controller/hyperconverged/upgradePatches.go
```diff
@@ -0,0 +1,92 @@
...
+	// (see: https://datatracker.ietf.org/doc/html/rfc6902 as the format reference).
+	JSONPatch string `json:"jsonPatch"`
+}
+
+type upgradePatches struct {
+	// hcoCRPatchList is a list of upgrade patches.
+	// Each hcoCRPatch consists in a semver range of affected source versions and a json patch to be applied during the upgrade if relevant.
+	HCOCRPatchList []hcoCRPatch `json:"hcoCRPatchList"`
+}
+
```
nunnatsa, line 32: Can also be:

```suggestion
// upgradePatches is a list of upgrade patches.
// Each hcoCRPatch consists in a semver range of affected source versions and a json patch to be applied type 
upgradePatches struct HCOCRPatchList []hcoCRPatch
```

nunnatsa (commented):
I added several comments

pkg/controller/hyperconverged/hyperconverged_controller.go
tiraboschi: done

tiraboschi: done

pkg/controller/hyperconverged/hyperconverged_controller.go
tiraboschi: done

assets/upgradePatches.yaml
tiraboschi: right, done

pkg/controller/hyperconverged/hyperconverged_controller.go
tiraboschi: I think we should keep it in to ensure we are applying each patch on top of the previous ones

pkg/controller/hyperconverged/hyperconverged_controller.go
tiraboschi: done

pkg/controller/hyperconverged/hyperconverged_controller.go
tiraboschi: done

pkg/controller/hyperconverged/hyperconverged_controller.go
tiraboschi: ?

pkg/controller/hyperconverged/upgradePatches.go
tiraboschi: I prefer to keep it in this way to keep places to add other operators like deleting other objects in the future

pkg/controller/hyperconverged/hyperconverged_controller.go
nunnatsa: Marshalling and unmarshaling costs cpu and memory. I think this is notvsomething to do in a loop. Also, here you also do it one on top the other, but on a json representation rather than cr.

pkg/controller/hyperconverged/test-files/upgradePatches/upgradePatches.json
```diff
@@ -0,0 +1 @@
+../../../../../assets/upgradePatches.json
```
nunnatsa, line 1: Is this the desired content of the file?

tiraboschi: no, this is how github renders a symlink to the real file under `../../../../../assets/upgradePatches.json`

pkg/controller/hyperconverged/hyperconverged_controller.go

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -1120,6 +1134,60 @@ func (r ReconcileHyperConverged) amendBadDefaults(req *common.HcoRequest) (bool,
...
+		err = json.Unmarshal(patchedBytes, tmpInstance)
+		if err != nil {
+			return false, err
+		}
+		if !reflect.DeepEqual(tmpInstance.Spec, req.Instance.Spec) {
+			req.Logger.Info("updating HCO spec as a result of upgrade patch", "knownHcoSV", knownHcoSV, "affectedRange", affectedRange, "patches", p.JSONPatch)
+			tmpInstance.Spec.DeepCopyInto(&req.Instance.Spec)
+			modified = true
+			req.Dirty = true
+		}
```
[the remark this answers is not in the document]

PR #1599, opened 2021-11-10
[release-1.5] Apply upgrade patches according to a yaml file (#1586)

tiraboschi:
Apply upgrade patches according to a yaml files
shipped within the container.
The operator will read the yaml file expecting
a fixed struct.

Now it handles only jsonpatches to be applied
on the HCO CR, in the future the same mechanism
could be easily extended to delete leftovers
or patch other objects.

Now it expects something like:
```json
{
  "hcoCRPatchList": [
    {
      "semverRange": ">=1.4.0 <=1.5.0",
      "jsonPatch": [
        {
          "op": "replace",
          "path": "/spec/featureGates/sriovLiveMigration",
          "value": true
        }
      ]
    }
  ]
}
```
where hcoCRPatchList is a list of upgrade patches.
Each hcoCRPatch consists in a semver range
( https://github.com/blang/semver#ranges ) of affected
source versions and a json patch
( https://datatracker.ietf.org/doc/html/rfc6902 )
to be applied during the upgrade if relevant.

Applying it to set
/spec/featureGates/sriovLiveMigration = true
when upgrading from 1.4.z and 1.5.0 to the current version.

This is a manual cherry-pick of #1586

Bump github.com/evanphx/json-patch => v5.6.0+incompatible
to be aligned with the main branch.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2018468

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
Apply upgrade patches according to a yaml file
```

### DIFF
--- a/assets/upgradePatches.json
+++ b/assets/upgradePatches.json
@@ -1,0 +1,14 @@
+{
+  "hcoCRPatchList": [
+    {
+      "semverRange": ">=1.4.0 <=1.5.0",
+      "jsonPatch": [
+        {
+          "op": "replace",
+          "path": "/spec/featureGates/sriovLiveMigration",
+          "value": true
+        }
+      ]
+    }
+  ]
+}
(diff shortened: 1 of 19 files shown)


nunnatsa (approved):
/approve

PR #1629, opened 2021-11-28
Create ImageStreams if they defined in the dataImportCronTemplate yaml files

nunnatsa:
If the dataImportCronTemplate yaml files contain a temaplate with both URL and ImageStream, HCO will create the ImageStream if not exist, and will update SSP with the template without the URL but only with the imageStream.

Specific implementation for updating existing imageStreams (from the unit tests):
1. should create the ImageStream resource if not exists
2. should update the ImageStream resource if the docker image was changed
3. should remove tags if they are not required
4. should not update the imageStream if the tag name and the from.name fields are the same
5. should not update the imageStream if the it not controlled by HCO (even if the details are not the same)

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
None
```


nunnatsa:
/hold

Need to redesign this

tiraboschi (approved):

PR #1639, opened 2021-12-02
Configure Kubevirt for SNO mode

tiraboschi:
Read openshiftconfigv1.Infrastructure and tune
kubevirt accordingly:
- `ControlPlaneTopology=SingleReplica` => `.spec.infra.replicas=1`
- `InfrastructureTopology=SingleReplica` => `LiveMigrationGate=false`

This PR is going to affect only the behaviour on OCP/OKD;
the behaviour on plain k8s is not going to change.
In a future PR we can extend it to properly detect
cluster topology also on k8s.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2026336

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:
```release-note
Configure Kubevirt for SNO mode
```


pkg/controller/operands/kubevirt.go
```diff
@@ -450,29 +450,36 @@ func NewKubeVirtWithNameOnly(hc *hcov1beta1.HyperConverged, opts ...string) *kub
 	}
 }
 
-func hcoConfig2KvConfig(hcoConfig hcov1beta1.HyperConvergedConfig) *kubevirtcorev1.ComponentConfig {
-	if hcoConfig.NodePlacement != nil {
+func hcoConfig2KvConfig(hcoConfig hcov1beta1.HyperConvergedConfig, infrastructureHighlyAvailable bool) *kubevirtcorev1.ComponentConfig {
+	if hcoConfig.NodePlacement != nil || !infrastructureHighlyAvailable {
```
nunnatsa, line 454: not sure it's so important, but I think it's the two cases in this condition handled desperately below with almost no common code (beside the new  `kvConfig` var and the return statement). Maybe we ca save one layer of nesting and make the sonarCloud happier.

pkg/util/cluster.go
```diff
@@ -18,13 +21,17 @@ type ClusterInfo interface {
 	IsRunningLocally() bool
 	GetDomain() string
 	IsManagedByOLM() bool
+	IsControlPlaneHighlyAvailable() bool
```
nunnatsa, line 24: This is not in use in the code. Maybe we want to remove it to simplify the interface?

tiraboschi: we are not using it now with kubevirt, but I'm pretty sure we will start using it in the future with other components like CNAO which is scheduling also on masters.

PR #1647, opened 2021-12-05
Provide virtio-win pullspec in dedicated ConfigMap

orenc1:
Addresses: https://bugzilla.redhat.com/show_bug.cgi?id=2025295

When creating a VirtualMachine with OpenShift UI wizard, the UI searches for a ConfigMap that contains the pullspec for `virtio-win` image. The problem was that the ConfigMap was not accessible by non-admin users.
In previous releases, this data was overloaded onto `v2v-vmware` ConfigMap, which was detached from HCO in current version, along with VM Import Operator (has been separated into independent operator - [MTV / Konveyor](https://www.konveyor.io/)).

This PR adds a new ConfigMap, named `virtio-win` into HCO reconciliation loop, along with corresponding Role and RoleBinding for every authenticated user.

This PR is coupled with another PR against openshift/console, for updating the ConfigMap resolution:
https://github.com/openshift/console/pull/10616

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
Add virtio-win ConfigMap [BugFix for #2025295]
```

### DIFF
--- a/pkg/controller/operands/cdi.go
+++ b/pkg/controller/operands/cdi.go
@@ -4,6 +4,8 @@ import (
 	"errors"
 	"reflect"
 
+	log "github.com/go-logr/logr"
+
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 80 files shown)


nunnatsa (commented):
Looks good.

Two main comments: (details are inline in the code):
Please add handlers to the `NewOperandHandler` if possible. The `FirstUseInitiation` is for case we need to read something from the cluster before creating the handlers. For example, to check if the quickstart CRD is in the cluster. And we do that because we don't have a working client before when `NewOperandHandler` is called. But in general, we want to initiate all the handlers in one place, if possible. So unless I missed something and there is a reason for that, please move the new handlers to `NewOperandHandler`.

The second comment is about the tools/digester/vendor directory. Please check if you didn't run `go mod tidy` and/or `go mod vendor` with go version other than 1.17. When I'm running with 1.17 I get no change in this directory.

pkg/controller/operands/cdi.go
```diff
@@ -251,62 +253,30 @@ func NewKubeVirtStorageConfigForCR(cr *hcov1beta1.HyperConverged, namespace stri
...
-	return &configReaderRoleHandler{
-		Client:                 Client,
-		Scheme:                 Scheme,
-		crType:                 "Role",
-		removeExistingOwner:    false,
-		setControllerReference: true,
-		hooks:                  &configReaderRoleHooks{},
-	}
-}
+func NewConfigReaderRoleHandler(_ log.Logger, Client client.Client, Scheme *runtime.Scheme, hc *hcov1beta1.HyperConverged) ([]Operand, error) {
```
nunnatsa, line 256: please change to return Operator rather than a slice. error is not in use too.

pkg/controller/operands/cdi.go
```diff
@@ -251,62 +253,30 @@ func NewKubeVirtStorageConfigForCR(cr *hcov1beta1.HyperConverged, namespace stri
...
-		util.DeepCopyLabels(&configReaderRole.ObjectMeta, &found.ObjectMeta)
-
-		err := Client.Update(req.Ctx, found)
-		if err != nil {
-			return false, false, err
-		}
-		return true, !req.HCOTriggered, nil
-	}
+// ************** Config Reader Role Binding Handler **************
+func newConfigReaderRoleBindingHandler(_ log.Logger, Client client.Client, Scheme *runtime.Scheme, hc *hcov1beta1.HyperConverged) ([]Operand, error) {
```
nunnatsa, line 266: same

pkg/controller/operands/cdi_test.go
```diff
@@ -1089,14 +1089,14 @@ var _ = Describe("CDI Operand", func() {
...
+				expectedRole := NewCdiConfigReaderRole(hco)
 				expectedLabels := expectedRole.Labels
 				expectedRole.Labels = nil
 
 				cl := commonTestUtils.InitClient([]runtime.Object{hco, expectedRole})
 
-				handler := (*genericOperand)(newConfigReaderRoleHandler(cl, commonTestUtils.GetScheme()))
-				res := handler.ensure(req)
+				handler, _ := NewConfigReaderRoleHandler(logger, cl, commonTestUtils.GetScheme(), hco)
+				res := handler[0].ensure(req)
```
nunnatsa, line 1099: Please consider the following, to make the test a bit safer:

```suggestion
				handlers, err := NewConfigReaderRoleHandler(logger, cl, commonTestUtils.GetScheme(), hco)
				Expect(res.Err).ToNot(HaveOccurred())
				Expect(handlers).ToNot(HaveLen(1))
				res := handler1[0].ensure(req)
```

pkg/controller/operands/cdi_test.go
```diff
@@ -1112,12 +1112,12 @@ var _ = Describe("CDI Operand", func() {
...
 			It("should do nothing if exists", func() {
-				expectedRoleBinding := NewConfigReaderRoleBindingForCR(hco, hco.Namespace)
+				expectedRoleBinding := NewCdiConfigReaderRoleBinding(hco)
 
 				cl := commonTestUtils.InitClient([]runtime.Object{hco, expectedRoleBinding})
 
-				handler := (*genericOperand)(newConfigReaderRoleBindingHandler(cl, commonTestUtils.GetScheme()))
-				res := handler.ensure(req)
+				handler, _ := newConfigReaderRoleBindingHandler(logger, cl, commonTestUtils.GetScheme(), hco)
+				res := handler[0].ensure(req)
```
nunnatsa, line 1120: same

pkg/controller/operands/cdi_test.go
```diff
@@ -1131,14 +1131,14 @@ var _ = Describe("CDI Operand", func() {
...
+				expectedRoleBinding := NewCdiConfigReaderRoleBinding(hco)
 				expectedLabels := expectedRoleBinding.Labels
 				expectedRoleBinding.Labels = nil
 
 				cl := commonTestUtils.InitClient([]runtime.Object{hco, expectedRoleBinding})
 
-				handler := (*genericOperand)(newConfigReaderRoleBindingHandler(cl, commonTestUtils.GetScheme()))
-				res := handler.ensure(req)
+				handler, _ := newConfigReaderRoleBindingHandler(logger, cl, commonTestUtils.GetScheme(), hco)
+				res := handler[0].ensure(req)
```
nunnatsa, line 1141: same

pkg/controller/operands/cdi_test.go
orenc1: Ok, but please not the `Expect(res.Err).ToNot(HaveOccurred())` should be checked only after  `res := handler[0].ensure(req)` is evaluated.

orenc1: fixed

pkg/controller/operands/cdi_test.go
orenc1: fixed

pkg/controller/operands/cdi_test.go
nunnatsa: > Ok, but please note the `Expect(res.Err).ToNot(HaveOccurred())` should be checked only after `res := handler[0].ensure(req)` is evaluated.

typo. should be `Expect(err).ToNot(HaveOccurred())`

pkg/controller/operands/cdi_test.go
orenc1: fixed

orenc1:
SNO lanes failed on timeout while deleting the `kubevirt-hyperconverged` namespace.

/retest

PR #1653, opened 2021-12-10
DataImportCronTemplates references quay.io/containerdisks

dominikholler:
There are maintained containerdisks in quay.io/containerdisks/ , which
are referenced by the DataImportCronTemplates make them
available to the users in a comfortable way.

@nunnatsa Can you please have a look?

cc @rmohr

### DIFF
--- a/ci-test-files/centos8-imagestream.yaml
+++ b/ci-test-files/centos8-imagestream.yaml
@@ -10,7 +10,7 @@ spec:
   - annotations: null
     from:
       kind: DockerImage
-      name: quay.io/kubevirt/centos8-container-disk-images
+      name: quay.io/containerdisks/centos
     importPolicy:
       scheduled: true
-    name: latest
+    name: 8.4
--- a/assets/dataImportCronTemplates/dataImportCronTemplates.yaml
+++ b/assets/dataImportCronTemplates/dataImportCronTemplates.yaml
@@ -6,7 +6,7 @@
       spec:
         source:
           registry:
-            url: docker://quay.io/kubevirt/centos8-container-disk-images
+            url: docker://quay.io/containerdisks/centos:8.4
         storage:
           resources:
             requests:
@@ -21,7 +21,7 @@
       spec:
         source:
           registry:
-            url: docker://quay.io/kubevirt/fedora-container-disk-images
+            url: docker://quay.io/containerdisks/fedora:35
         storage:
           resources:
             requests:
--- a/hack/check_golden_images.sh
+++ b/hack/check_golden_images.sh
@@ -7,7 +7,7 @@ IMAGES_NS=${IMAGES_NS:-kubevirt-os-images}
 if [[ $(${KUBECTL_BINARY} get ssp -n ${INSTALLED_NAMESPACE}) ]]; then
   # Test image streams
   [[ $(${KUBECTL_BINARY} get imageStream centos8  -n ${IMAGES_NS} --no-headers | wc -l) -eq 1 ]]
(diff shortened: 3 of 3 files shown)


nunnatsa (commented):
Thanks for the PR. One question: don't we want to use the `latest` tag?

dominikholler:
> Thanks for the PR. One question: don't we want to use the `latest` tag?

In CentOS 7 and 8 there are not updates anymore expected, and for Fedora it is planed to update the URL, if we support a new Fedora version.

rmohr:
> Thanks for the PR. One question: don't we want to use the `latest` tag?

Let me also point you to my comment a https://bugzilla.redhat.com/show_bug.cgi?id=2031155#c4 why `:latests` is not necessarily a good idea.

nunnatsa:
> In CentOS 7 and 8 there are not updates anymore expected, and for Fedora it is planed to update the URL, if we support a new Fedora version.
> 
> Please find more details in https://bugzilla.redhat.com/show_bug.cgi?id=2031155#c4 .

Thanks @dominikholler. Two things then. First, please sign the commit. Also,do you think there is a way to automate the fedora image tag upgrade?

rmohr:
> Thanks @dominikholler. Two things then. First, please sign the commit. Also,do you think there is a way to automate the fedora image tag upgrade?

Did you consider greating some kind of catalogue? For instance, just because fc36 is out, it does not mean that suddenly people want fc36. I think there are two dimensions here:

 * Updating for e.g. delivering CVE fixes in base images (should happen through the referenced moving tags)
 * Keeping interesting OS versions for their supported lifetime

Do you have some thoughts on the second point?

nunnatsa:
> Did you consider greating some kind of catalogue? For instance, just because fc36 is out, it does not mean that suddenly people want fc36. I think there are two dimensions here:
> 
> * Updating for e.g. delivering CVE fixes in base images (should happen through the referenced moving tags)
> * Keeping interesting OS versions for their supported lifetime
> 
> Do you have some thoughts on the second point?

We can edit this file manually every few months, but if there is any way to automate it, it will be safer.

ci-test-files/centos8-imagestream.yaml
```diff
@@ -10,7 +10,7 @@ spec:
   - annotations: null
     from:
       kind: DockerImage
-      name: quay.io/kubevirt/centos8-container-disk-images
+      name: quay.io/containerdisks/centos:8.4
     importPolicy:
       scheduled: true
     name: latest
```
nunnatsa, line 16: First, this is a test only file.

Also, I'm not sure it will work. I think that the `.name` field should be "8.4" and the `from.name` fiels should be "quay.io/containerdisks/centos", without the tag.

And HCO only supports image stream with `latest` tag.

rmohr:
> We can edit this file manually every few months, but if there is any way to automate it, it will be safer.

I just want to start a conversation if this is good enough. Because doesn't it mean that you suddenly stop updating fc34 as soon as fc35 is out, potentially leaving people behind which would expect to get e.g. potential CVE updates because fc34 is still for a few months?

A first thought would for instance be to read from `quay.io/containerdisks` tags, iterate over them and generate a set of sync jobs and target PVCs.

I am not sure if this is considered an issue though ...

ci-test-files/centos8-imagestream.yaml
dominikholler: Thanks, I adjusted to `name:`.
@nunnatsa What do you mean by 

> And HCO only supports image stream with latest tag.

Does this mean that `quay.io/containerdisks/centos ` requires the `latest` tag?

nunnatsa: We're not using image stream for the community version. This file is only for CI testing, to test the image stream logic.

The actual DataImportCronTemplate files are using URLs.

nunnatsa: @dominikholler - please fix this file (cancel the changes) and add release notes so we can go on with this pr.

nunnatsa:
Closing in favor of #1665

PR #1711, opened 2022-01-17
Remove old HCO metrics services and endpoints when upgrading

machadovilaca:
There is a bug when updating from v1.2.x to versions later than v1.4.x where services `hyperconverged-cluster-operator-metrics` and `hyperconverged-cluster-webhook-metrics`, and endpoint `hyperconverged-cluster-operator-metrics` are not being deleted. This PR ensures those objects are removed.

Signed-off-by: João Vilaça <jvilaca@redhat.com>

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
Remove old HCO metrics services and endpoints when upgrading
```

### DIFF
--- a/pkg/controller/hyperconverged/hyperconverged_controller.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller.go
@@ -1118,33 +1118,25 @@ var (
 	operatorMetrics = "hyperconverged-cluster-operator-metrics"
 	webhookMetrics  = "hyperconverged-cluster-webhook-metrics"
 
-	oldMetricsServices  map[string]corev1.Service
-	oldMetricsEndpoints map[string]corev1.Endpoints
+	oldMetricsObjects map[string]client.Object
 )
 
-func initOldMetricsServices(req *common.HcoRequest) {
-	if oldMetricsServices == nil {
-		oldMetricsServices = map[string]corev1.Service{
-			operatorMetrics: {
+func initOldMetricsObjects(req *common.HcoRequest) {
+	if oldMetricsObjects == nil {
+		oldMetricsObjects = map[string]client.Object{
+			"operatorService": &corev1.Service{
 				ObjectMeta: metav1.ObjectMeta{
 					Name:      operatorMetrics,
 					Namespace: req.Namespace,
 				},
 			},
-			webhookMetrics: {
+			"webhookService": &corev1.Service{
 				ObjectMeta: metav1.ObjectMeta{
 					Name:      webhookMetrics,
 					Namespace: req.Namespace,
 				},
 			},
-		}
-	}
-}
-
-func initOldMetricsEndpoints(req *common.HcoRequest) {
-	if oldMetricsEndpoints == nil {
-		oldMetricsEndpoints = map[string]corev1.Endpoints{
-			operatorMetrics: {
+			"operatorEndpoint": &corev1.Endpoints{
 				ObjectMeta: metav1.ObjectMeta{
 					Name:      operatorMetrics,
 					Namespace: req.Namespace,
@@ -1155,30 +1147,17 @@ func initOldMetricsEndpoints(req *common.HcoRequest) {
 }
 
 func (r ReconcileHyperConverged) removeOldMetricsObjs(req *common.HcoRequest) error {
-	initOldMetricsServices(req)
-	initOldMetricsEndpoints(req)
+	initOldMetricsObjects(req)
 
-	for name, service := range oldMetricsServices {
-		removed, err := r.deleteObj(req, &service)
-
-		if err != nil {
-			return err
-		}
-
-		if removed {
-			delete(oldMetricsServices, name)
-		}
-	}
-
-	for name, endpoint := range oldMetricsEndpoints {
-		removed, err := r.deleteObj(req, &endpoint)
+	for name, object := range oldMetricsObjects {
+		removed, err := r.deleteObj(req, object)
 
 		if err != nil {
 			return err
 		}
 
 		if removed {
-			delete(oldMetricsEndpoints, name)
+			delete(oldMetricsObjects, name)
 		}
 	}
 
--- a/pkg/controller/hyperconverged/hyperconverged_controller.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller.go
@@ -1044,14 +1044,14 @@ func (r *ReconcileHyperConverged) migrateBeforeUpgrade(req *common.HcoRequest) (
 		return false, err
 	}
 
-	oldMetricsSVCRemoved, err := r.removeOldMetricsObjs(req)
+	err = r.removeOldMetricsObjs(req)
 	if err != nil {
 		return false, err
 	}
 
 	removeOldQuickStartGuides(req, r.client, r.operandHandler.GetQuickStartNames())
 
-	return upgradePatched || oldMetricsSVCRemoved, nil
+	return upgradePatched, nil
 }
 
 func (r ReconcileHyperConverged) applyUpgradePatches(req *common.HcoRequest) (bool, error) {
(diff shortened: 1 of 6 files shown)


pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -1100,6 +1105,94 @@ func (r ReconcileHyperConverged) applyUpgradePatch(req *common.HcoRequest, hcoJs
 	return hcoJson, nil
 }
 
+func (r ReconcileHyperConverged) removeOldMetricsObjs(req *common.HcoRequest) (bool, error) {
+	anyServiceRemoved, err := r.removeObjs(req, corev1.Service{}, []string{
+		"hyperconverged-cluster-operator-metrics",
+		"hyperconverged-cluster-webhook-metrics",
```
nunnatsa, line 1111: please use constants

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -1100,6 +1105,94 @@ func (r ReconcileHyperConverged) applyUpgradePatch(req *common.HcoRequest, hcoJs
...
+	})
+	if err != nil {
+		return false, err
+	}
+
+	anyEndpointRemoved, err := r.removeObjs(req, corev1.Endpoints{}, []string{
+		"hyperconverged-cluster-operator-metrics",
```
nunnatsa, line 1118: should use the same constant as above

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -1100,6 +1105,94 @@ func (r ReconcileHyperConverged) applyUpgradePatch(req *common.HcoRequest, hcoJs
...
+}
+
+func (r ReconcileHyperConverged) removeObj(req *common.HcoRequest, objType interface{}, objName string) (bool, error) {
+	obj, err := r.getObj(req, objType, objName)
+	if err != nil {
+		return false, err
+	} else if obj == nil {
+		return false, nil
+	}
+
```
nunnatsa, line 1148: I think `hcoutil.EnsureDeleted` already does this logic. Please check if we can remove these lines and the `getObj` function.

nunnatsa (commented):
I want to suggest something. Maybe, instead of going to the api-server each time to search for the services and the endpoint, converging the resources to object and so on, we can have a static list of these three objects, where we only need to loop over and delete its items? We can maybe also remove the deleted objects from the list so we could skip them next time, then this code will happen again only if the pod was restarted during upgrade.

@machadovilaca, what do you think?

pkg/controller/hyperconverged/hyperconverged_controller_test.go
```diff
@@ -1742,6 +1742,116 @@ var _ = Describe("HyperconvergedController", func() {
 				})
 
 			})
+
+			Context("Remove old metrics services and endpoints", func() {
+				services := []string{
+					"hyperconverged-cluster-operator-metrics",
+					"hyperconverged-cluster-webhook-metrics",
```
nunnatsa, line 1749: use constant from above comment

pkg/controller/hyperconverged/hyperconverged_controller_test.go
```diff
@@ -1742,6 +1742,116 @@ var _ = Describe("HyperconvergedController", func() {
...
+
+				serviceEvents := make([]commonTestUtils.MockEvent, 0)
+				for _, service := range services {
+					serviceEvents = append(serviceEvents, commonTestUtils.MockEvent{
+						EventType: corev1.EventTypeNormal,
+						Reason:    "Killing",
+						Msg:       fmt.Sprintf("Removed %s Service", service),
+					})
+
+				}
```
nunnatsa, line 1760: It will work. I wonder if it won't be simpler just to add these two services as a golang structs.

pkg/controller/hyperconverged/hyperconverged_controller_test.go
```diff
@@ -1742,6 +1742,116 @@ var _ = Describe("HyperconvergedController", func() {
...
+
+				endpointsEvents := make([]commonTestUtils.MockEvent, 0)
+				for _, endpoint := range endpoints {
+					endpointsEvents = append(endpointsEvents, commonTestUtils.MockEvent{
+						EventType: corev1.EventTypeNormal,
+						Reason:    "Killing",
+						Msg:       fmt.Sprintf("Removed %s Endpoints", endpoint),
+					})
+
+				}
```
nunnatsa, line 1774: same

pkg/controller/hyperconverged/hyperconverged_controller_test.go
```diff
@@ -1742,6 +1742,116 @@ var _ = Describe("HyperconvergedController", func() {
...
+						resources = append(resources, &corev1.Service{
+							ObjectMeta: metav1.ObjectMeta{
+								Name:      service,
+								Namespace: namespace,
+								Labels: map[string]string{
+									hcoutil.AppLabel: expected.hco.Name,
+								},
+							},
+						})
+					}
```
nunnatsa, line 1809: again - it's only two services

pkg/controller/hyperconverged/hyperconverged_controller_test.go
```diff
@@ -1742,6 +1742,116 @@ var _ = Describe("HyperconvergedController", func() {
...
+						resources = append(resources, &corev1.Endpoints{
+							ObjectMeta: metav1.ObjectMeta{
+								Name:      endpoint,
+								Namespace: namespace,
+								Labels: map[string]string{
+									hcoutil.AppLabel: expected.hco.Name,
+								},
+							},
+						})
+					}
```
nunnatsa, line 1838: here too

pkg/controller/hyperconverged/hyperconverged_controller_test.go
```diff
@@ -1742,6 +1742,116 @@ var _ = Describe("HyperconvergedController", func() {
...
+					req := commonTestUtils.NewReq(expected.hco)
+
+					modified, err := r.migrateBeforeUpgrade(req)
+					Expect(err).ToNot(HaveOccurred())
+					Expect(modified).To(BeTrue())
+
+					By("Check events")
+					events := r.eventEmitter.(*commonTestUtils.EventEmitterMock)
+					Expect(events.CheckEvents(endpointsEvents)).To(BeTrue())
+				})
```
nunnatsa, line 1853: Please add a check to read back the endpoint and make sure it's not there.

pkg/controller/hyperconverged/hyperconverged_controller_test.go
```diff
@@ -1742,6 +1742,116 @@ var _ = Describe("HyperconvergedController", func() {
...
+
+					req := commonTestUtils.NewReq(expected.hco)
+
+					modified, err := r.migrateBeforeUpgrade(req)
+					Expect(err).ToNot(HaveOccurred())
+					Expect(modified).To(BeTrue())
+
+					By("Check events")
+					events := r.eventEmitter.(*commonTestUtils.EventEmitterMock)
+					Expect(events.CheckEvents(serviceEvents)).To(BeTrue())
```
nunnatsa, line 1823: Please add a check to read back the services and make sure they are not there.

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -1035,9 +1035,14 @@ func (r *ReconcileHyperConverged) migrateBeforeUpgrade(req *common.HcoRequest) (
...
 
+	oldMetricsSVCRemoved, err := r.removeOldMetricsObjs(req)
+	if err != nil {
+		return false, err
+	}
+
 	removeOldQuickStartGuides(req, r.client, r.operandHandler.GetQuickStartNames())
 
-	return upgradePatched, nil
+	return upgradePatched || oldMetricsSVCRemoved, nil
```
nunnatsa, line 1045: I believe this is not needed. This function returns `true` if the **`HyperConverged`** CR is modified, in order to requeue later. This is not neede when removing another resources, as services or endpoints. I think the `removeOldMetricsObjs` function should only return error, but not the boolean value.

nunnatsa (changes requested):
Please review my comment.

machadovilaca:
@nunnatsa updated

nunnatsa:
@machadovilaca - the upgrade tests are failing because HCO dose not have permissions to list/get/delete endpoints. Please add the missing RBAC.

pkg/controller/hyperconverged/hyperconverged_controller.go
```diff
@@ -1100,6 +1105,99 @@ func (r ReconcileHyperConverged) applyUpgradePatch(req *common.HcoRequest, hcoJs
...
+		oldMetricsEndpoints = map[string]corev1.Endpoints{
+			operatorMetrics: {
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      operatorMetrics,
+					Namespace: req.Namespace,
+				},
+			},
+		}
+	}
+}
```
nunnatsa, line 1146: I think it's can be combined into one function, then uses a single loop to delete them. WDYT?

nunnatsa (approved):

PR #1717, opened 2022-01-18
Enhancement proposal for a cluster-wide Crypto Policy Setting

tiraboschi:

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
NONE
```

### DIFF
--- a/docs/enhancements/crypto-api.md
+++ b/docs/enhancements/crypto-api.md
@@ -15,7 +15,7 @@ Starting from OCP/OKD 4.6, a [cluster-wide API](https://github.com/openshift/enh
 ### Non-Goals
 
 1. Describe the technical implementation of TLS security policy for each operand of a sibling operator of HCO (just a quick hint).
-2. The Mozilla Recommended Configurations mandate also `Certificate type` and `Certificate lifespan` as part of each profile while this is not covered by [Openshift implementation](https://github.com/openshift/enhancements/blob/master/enhancements/kube-apiserver/tls-config.md) so it's not covered by this API.  
+2. The Mozilla Recommended Configurations mandate also `Certificate type` and `Certificate lifespan` as part of each profile while this is not covered by [OpenShift implementation](https://github.com/openshift/enhancements/blob/master/enhancements/kube-apiserver/tls-config.md) so it's not covered by this API.
 
 ## Overview
 
@@ -149,8 +149,7 @@ spec:
 ## Proposal
 
 Hyperconverged-cluster-operator (HCO) will read the global configuration for TLS security profile of the APIServer, without storing it in HCO CR, and will propagate the .spec.tlsSecurityProfile stanza to all underlying HCO managed custom resources, and will reconcile them, to achieve alignment and consistency of security profiles between Hyperconverged Kubevirt and OCP/OKD control plane components.
-The idea is that the code to read the Openshift API can be coded once in HCO without the need to replicate it for each single component. Moreover, a few components like kubevirt/kubevirt are pretty independent of Openshift specific APIs so HCO looks like a good candidate for this.  
-
+The idea is that the code to read the OpenShift API can be coded once in HCO without the need to replicate it for each single component. Moreover, a few components like kubevirt/kubevirt are pretty independent of OpenShift-specific APIs so HCO looks like a good candidate for this.
 
 The underlying HCO managed custom resources to which HCO will propagate the security profile stanza to, are:
 * `kubevirt.kubevirt.io/kubevirt-kubevirt-hyperconverged`
@@ -230,9 +229,9 @@ The custom resources that the TLS Security Profile stanza will be propagated to,
 ## Open Questions
 1. In plain kubernetes HCO deployment, will we use the default intermediate profile for all HCO managed components, or give the option for the administrator to set it on the HCO CR, which will propagate to all components?
    1. The TLS Security Profile configuration feature will be available only on OCP/OKD platforms.
-   2. We can decide to expose the API to let the user optionally fine-tune/override the configuration for all the Hyperconverged Kubevirt having HCO reading the Openshift API server one as its default only if the user didn't express custom values. This will work also on plain k8s.
+   2. We can decide to expose the API to let the user optionally fine-tune/override the configuration for all the Hyperconverged Kubevirt having HCO reading the OpenShift API server one as its default only if the user didn't express custom values. This will work also on plain k8s.
 
-2. [Openshift API design](https://github.com/openshift/enhancements/blob/master/enhancements/kube-apiserver/tls-config.md) explicitly mentions API servers, but it never mentions admission webhooks that are also TLS protected.
+2. [OpenShift API design](https://github.com/openshift/enhancements/blob/master/enhancements/kube-apiserver/tls-config.md) explicitly mentions API servers, but it never mentions admission webhooks that are also TLS protected.
 Is there any technical reason for us to ignore the admission webhooks? 
 
 ## Implementation hints
--- a/docs/enhancements/crypto-api.md
+++ b/docs/enhancements/crypto-api.md
@@ -8,6 +8,11 @@ Starting from OCP/OKD 4.6, a [cluster-wide API](https://github.com/openshift/enh
 
 ## Motivation
 
+A really security-conscious cluster-admin, is seeking an option to configure ciphers that would apply to all OpenShift-proper components both for internal and external traffic (on OCP/OKD).
+
+In an OpenShift cluster (OCP/OKD), secure communications is expected between system level pods and application level pods within the cluster (in other words east-west traffic in the control plane, system plane and data plane). In addition, secure communication is expected for traffic that ingress into the cluster from external clients as well as egressing from the cluster to external entities.
+This RFE proposes to have all the HyperConverged Kubevirt components comply with Cluster-wide cryptographic policies on OpenShift (OCP/OKD) that provide a convenient way to ensure that OpenShift system pods and application pods can use cryptographic libraries that do not allow known insecure protocols, ciphers, or algorithms.
+
 ### Goals
 
 1. Describe how a configuration of a TLS security profile for OCP/OKD control plane will be propagated to all the Hyperconverged Kubevirt control plane components, using existing OCP/OKD API.
--- a/docs/enhancements/crypto-api.md
+++ b/docs/enhancements/crypto-api.md
@@ -1,0 +1,247 @@
+# Technical Design Input
+
+# Hyperconverged Kubevirt cluster-wide Crypto Policy Setting
+
+## Abstract
+
+Starting from OCP/OKD 4.6, a [cluster-wide API](https://github.com/openshift/enhancements/blob/master/enhancements/kube-apiserver/tls-config.md) is available for cluster administrators to set TLS profiles for OCP/OKD core components. HCO, as an OCP/OKD layered product, will follow along OCP/OKD crypto policy cluster-wide setting, and use the same profile configured for the cluster’s control plane. Configuration of a TLS security profile ensures that OCP/OKD, as well as HCO and its sibling operators and operands, use cryptographic libraries that do not allow known insecure protocols, ciphers, or algorithms.
+
+## Motivation
+
+### Goals
+
+1. Describe how a configuration of a TLS security profile for OCP/OKD control plane will be propagated to all the Hyperconverged Kubevirt control plane components, using existing OCP/OKD API.
+
+### Non-Goals
+
+1. Describe the technical implementation of TLS security policy for each operand of a sibling operator of HCO (just a quick hint).
+2. The Mozilla Recommended Configurations mandate also `Certificate type` and `Certificate lifespan` as part of each profile while this is not covered by [Openshift implementation](https://github.com/openshift/enhancements/blob/master/enhancements/kube-apiserver/tls-config.md) so it's not covered by this API.  
+
+## Overview
+
+Currently, TLS security policy can be configured for the following OCP/OKD core components:
+
+* Ingress Controller
+* Kubelet
+* API Server
+
+While the TLS security policy for the Ingress Controller and the kubelet are independent, configuring TLS security policy for the API Server is automatically propagating it for the following core components:
+
+* Kubernetes API server
+* Kubernetes controller manager
+* Kubernetes scheduler
+* OpenShift API server
+* OpenShift OAuth API server
+* OpenShift OAuth server
+
+The TLS security profiles are based on [Mozilla Recommended Configurations](https://wiki.mozilla.org/Security/Server_Side_TLS):
+* Old - intended for use with legacy clients of libraries; requires a minimum TLS version of 1.0
+* Intermediate - the default profile for all components; requires a minimum TLS version of 1.2
+* Modern - intended for use with clients that don’t need backward compatibility; requires a minimum TLS version of 1.3. Unsupported in OCP/OKD 4.8 and below.
+
+## API Schema
+
+```bash
+$ oc explain apiserver.spec.tlsSecurityProfile
+KIND:     APIServer
+VERSION:  config.openshift.io/v1
+
+RESOURCE: tlsSecurityProfile <Object>
+
+DESCRIPTION:
+     tlsSecurityProfile specifies settings for TLS connections for externally
+     exposed servers. If unset, a default (which may change between releases) is
+     chosen. Note that only Old, Intermediate and Custom profiles are currently
+     supported, and the maximum available MinTLSVersions is VersionTLS12.
+
+FIELDS:
+   custom	<>
+     custom is a user-defined TLS security profile. Be extremely careful using a
+     custom profile as invalid configurations can be catastrophic. An example
+     custom profile looks like this:
+     ciphers: - ECDHE-ECDSA-CHACHA20-POLY1305 - ECDHE-RSA-CHACHA20-POLY1305 -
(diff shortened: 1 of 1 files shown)


docs/enhancements/crypto-api.md
```diff
@@ -0,0 +1,247 @@
...
+
+
+## Open Questions
+1. In plain kubernetes HCO deployment, will we use the default intermediate profile for all HCO managed components, or give the option for the administrator to set it on the HCO CR, which will propagate to all components?
+   1. The TLS Security Profile configuration feature will be available only on OCP/OKD platforms.
+   2. We can decide to expose the API to let the user optionally fine-tune/override the configuration for all the Hyperconverged Kubevirt having HCO reading the Openshift API server one as its default only if the user didn't express custom values. This will work also on plain k8s.
+
+2. [Openshift API design](https://github.com/openshift/enhancements/blob/master/enhancements/kube-apiserver/tls-config.md) explicitly mentions API servers, but it never mentions admission webhooks that are also TLS protected.
+Is there any technical reason for us to ignore them admission webhooks?
```
jbpratt, line 236:
```suggestion
Is there any technical reason for us to ignore the admission webhooks? 
```
just browsing through :smiley_cat:

tiraboschi: thanks

docs/enhancements/crypto-api.md
```diff
@@ -0,0 +1,247 @@
...
+4. [https://wiki.mozilla.org/Security/Server_Side_TLS#Recommended_configurations](https://wiki.mozilla.org/Security/Server_Side_TLS#Recommended_configurations) also mentions Certificate type and Certificate lifespan but [https://github.com/openshift/enhancements/blob/master/enhancements/kube-apiserver/tls-config.md](https://github.com/openshift/enhancements/blob/master/enhancements/kube-apiserver/tls-config.md) doesn't, so we can safely ignore them
+
+
+## Open Questions
+1. In plain kubernetes HCO deployment, will we use the default intermediate profile for all HCO managed components, or give the option for the administrator to set it on the HCO CR, which will propagate to all components?
+   1. The TLS Security Profile configuration feature will be available only on OCP/OKD platforms.
+   2. We can decide to expose the API to let the user optionally fine-tune/override the configuration for all the Hyperconverged Kubevirt having HCO reading the Openshift API server one as its default only if the user didn't express custom values. This will work also on plain k8s.
```
dankenigsberg, line 233: I like this, as this is how CDI behaves with regards to [importProxy](https://github.com/kubevirt/containerized-data-importer/blob/main/doc/cdi-config.md#options): if it empty it uses OpenShift's defaults. But it lets one set it explicitly for environment which does not have the cluster-wide definition.

docs/enhancements/crypto-api.md
```diff
@@ -0,0 +1,247 @@
...
+      minTLSVersion: VersionTLS11
+```
+
+*When specifying a _Custom_ profile, a ciphers list must be provided, as well as a minimum TLS version.*
+
+
+## Proposal
+
+Hyperconverged-cluster-operator (HCO) will read the global configuration for TLS security profile of the APIServer, without storing it in HCO CR, and will propagate the .spec.tlsSecurityProfile stanza to all underlying HCO managed custom resources, and will reconcile them, to achieve alignment and consistency of security profiles between Hyperconverged Kubevirt and OCP/OKD control plane components.
+The idea is that the code to read the Openshift API can be coded once in HCO without the need to replicate it for each single component. Moreover, a few components like kubevirt/kubevirt are pretty independent of Openshift specific APIs so HCO looks like a good candidate for this.
```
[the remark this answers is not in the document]
tiraboschi: done

docs/enhancements/crypto-api.md
```diff
@@ -0,0 +1,246 @@
+# Technical Design Input
+
+# Hyperconverged Kubevirt cluster-wide Crypto Policy Setting
+
+## Abstract
+
+Starting from OCP/OKD 4.6, a [cluster-wide API](https://github.com/openshift/enhancements/blob/master/enhancements/kube-apiserver/tls-config.md) is available for cluster administrators to set TLS profiles for OCP/OKD core components. HCO, as an OCP/OKD layered product, will follow along OCP/OKD crypto policy cluster-wide setting, and use the same profile configured for the cluster’s control plane. Configuration of a TLS security profile ensures that OCP/OKD, as well as HCO and its sibling operators and operands, use cryptographic libraries that do not allow known insecure protocols, ciphers, or algorithms.
```
mhenriks, line 7: I am not aware of any "Externally Facing Services" in the KubeVirt ecosystem so I'm struggling to understand why we should support this.  None of our components are directly exposed to clients outside the cluster.

APIServers (virt-api, cdi-apiserver) are proxied through via kube-apiserver.

Same with any webhooks

cdi-uploaaproxy is expected to be exposed via a Route/Ingress

So exactly what KubeVirt components are  directly exposed out the cluster?

mhenriks: I don't see anywhere in the document where these settings are expected to be propagated down to subcomponents.

docs/enhancements/crypto-api.md
```diff
@@ -0,0 +1,246 @@
...
+
+
+## Open Questions
+1. In plain kubernetes HCO deployment, will we use the default intermediate profile for all HCO managed components, or give the option for the administrator to set it on the HCO CR, which will propagate to all components?
+   1. The TLS Security Profile configuration feature will be available only on OCP/OKD platforms.
+   2. We can decide to expose the API to let the user optionally fine-tune/override the configuration for all the Hyperconverged Kubevirt having HCO reading the OpenShift API server one as its default only if the user didn't express custom values. This will work also on plain k8s.
+
+2. [OpenShift API design](https://github.com/openshift/enhancements/blob/master/enhancements/kube-apiserver/tls-config.md) explicitly mentions API servers, but it never mentions admission webhooks that are also TLS protected.
+Is there any technical reason for us to ignore the admission webhooks?
```
mhenriks, line 235: I sincerely believe those settings are not expected to be propagated down to services that are not directly accessible from outside the cluster like our webhooks/aggregated apiservers.

davidvossel (commented):
i'd like to see some use cases defined here. It is unclear to me what we're solving and why.

For example, what external hco services are we trying to secure here, and how do these policies around inner component communication help in that regard?

tiraboschi:
> i'd like to see some use cases defined here. It is unclear to me what we're solving and why.
> 
> For example, what external hco services are we trying to secure here, and how do these policies around inner component communication help in that regard?

HCO is not exposing any service by itself, this is simply about having the whole Kubevirt Hyperconverged control plane behaving accordingly to the configuration of the OpenShift one when deployed on OpenShift:
https://docs.openshift.com/container-platform/4.9/security/tls-security-profiles.html#tls-profiles-kubernetes-configuring_tls-security-profiles

docs/enhancements/crypto-api.md
tiraboschi: According to:
https://docs.openshift.com/container-platform/4.9/security/tls-security-profiles.html#tls-profiles-kubernetes-configuring_tls-security-profiles ,
setting the TLS security profile in the APIServer CR propagates the setting to a few OpenShift control plane components:
- Kubernetes API server
- Kubernetes controller manager
- Kubernetes scheduler
- OpenShift API server
- OpenShift OAuth API server
- OpenShift OAuth server
- etcd

This whole proposal is just about having the same setting propagated also to the whole Kubevirt Hyperconverged control plane.

dankenigsberg: If I understand @mhenriks correctly, this proposal is trying to be holier than OKD. There, the purpose is to harden crypto policy for extern services. We have no such services; the purpose of this proposal is to let a cluster admin to harden KubeVirt-internal APIs. Do you know, @tiraboschi, if other complex OLM-controlled operators do that? Do internal OKD operators do that?

tiraboschi: OCP/OKD are aiming to have a configuration option to let the user configure ciphers that would apply to all OpenShift components both for internal and external traffic.
I'm not aware of any OLM-controlled operator that already implemented it, while between CVO-controlled ones CMO already adopted it ( https://github.com/openshift/cluster-monitoring-operator/pull/1380 ) and MCO is currently working in that direction.

dankenigsberg: Good to know. I'd add this background to your proposal. Also, I'd reduce the emphasis on OKD.
We use encryption internally because cluster owners do not trust all workload. Some cluster owners are more suspicious than others and would like to use stronger, more expensive and less compatible crypto. I think that this proposal should be about letting them do that. Just like I can set `/etc/crypto-policies/config` on my laptop, affecting all polite applications, I would like to set my crypto policy once in hco (or okd) and have all components comply.

tiraboschi:
I updated the motivation section trying to explain better the use case.

docs/enhancements/crypto-api.md
```diff
@@ -0,0 +1,251 @@
...
+2. [OpenShift API design](https://github.com/openshift/enhancements/blob/master/enhancements/kube-apiserver/tls-config.md) explicitly mentions API servers, but it never mentions admission webhooks that are also TLS protected.
+Is there any technical reason for us to ignore the admission webhooks? 
+
+## Implementation hints
+
+If the TLS implementation is based on the standard golang crypto library (as expected), the TLS configuration is performed via tls.Config struct: [https://pkg.go.dev/crypto/tls#Config](https://pkg.go.dev/crypto/tls#Config)
+which contains `MinVersion` and `CipherSuites` which should be tuned according to the chosen tlsSecurityProfile.
+
+For instance intermediate will set: [https://wiki.mozilla.org/Security/Server_Side_TLS#Intermediate_compatibility_.28recommended.29](https://wiki.mozilla.org/Security/Server_Side_TLS#Intermediate_compatibility_.28recommended.29)
+If you can/want to import `github.com/openshift/api/config/v1`, the three standard configurations lists are already coded [here](https://github.com/openshift/api/blob/master/config/v1/types_tlssecurityprofile.go#L203)
```
qinqon, line 248: It's ok to import this on k8s projects ?

qinqon (commented):
Just a little question  the rest is ok.

docs/enhancements/crypto-api.md
tiraboschi: I think it's a question for the maintainer of the specific project

qinqon (approved):

acardace (approved):
Looks good to me, thanks @tiraboschi.

borod108:
Looks good to me in general, this maybe a stupid question but - is there any case where we may have a change like this in a middle of a process break it? I mean like some kind of interaction during a vm migration, or during upgrade? 
May it be useful to preserve information about the previous profile used in case we need to resume some kind of connection after the crypto policy changed? (Say have a field in the API with that information)

tiraboschi:
> Looks good to me in general, this maybe a stupid question but - is there any case where we may have a change like this in a middle of a process break it? I mean like some kind of interaction during a vm migration, or during upgrade? May it be useful to preserve information about the previous profile used in case we need to resume some kind of connection after the crypto policy changed? (Say have a field in the API with that information)

Technically a TLS connection can be dropped at any time for various reasons,
the client will eventually establish a new one according to the up-to-date crypto policy settings.
Personally I don't see anything so critical in this regards, but eventually each component operator can decide to postpone applying the new settings if the component is involved in a critical task like an upgrade.

borod108:
@tiraboschi makes sense. So just to clarify, you say it will not be useful to preserve information about the previous profile used?

tiraboschi:
> @tiraboschi makes sense. So just to clarify, you say it will not be useful to preserve information about the previous profile used?

Each profile mandates a set of allowed ciphers and protocol versions,
even if you track the profile, you cannot infer which cipher was actually in use so honestly I don't see that much value there.

borod108:
@tiraboschi thank you for clarifying. no more questions from me then :)

tiraboschi:
@borod108 thanks, so can you please ack this PR?

borod108 (approved):

mhenriks:
I am totally in favor of giving KubeVirt users the explicit ability to set tls config on our operator CRs.  Although I wonder if a single "knob" is appropriate for everything.

I am less sure that what is good for `config.openshift.io. APIServer` is appropriate for all our internal communication by default.  Maybe it's fine for webhooks.  But for CDI network cloning?  Or VM migration?  Typically, external servers have the strictest tls config.  But admins may want to be a little looser internal traffic.  Especially high throughput traffic.

Even if there is some openshift resource (now or later) that is appropriate for our internal traffic, I am even less sure that HCO should push that config down to the other operators.  Yes, the alternative will probably lead to some duplicated code.  But it's not a lot of code.  Especially if using controller runtime.  It would also make sure we honor those values for non HCO based installs.

tiraboschi:
> I am totally in favor of giving KubeVirt users the explicit ability to set tls config on our operator CRs. Although I wonder if a single "knob" is appropriate for everything.
> 
> I am less sure that what is good for `config.openshift.io.APIServer` is appropriate for all our internal communication by default. Maybe it's fine for webhooks. But for CDI network cloning? Or VM migration? Typically, external servers have the strictest tls config. But admins may want to be a little looser for internal traffic. Especially high throughput traffic.

I understand your point, but I tend to think that the vast majority of our user will stick with the default value. I'd expect that only a few really security conscious sysadmin will tweak this knob tolerating to pay it in terms of performances.

So I tend to think that having a single, cluster wide, knob with the ability to override it for the whole Hyper Converged Cluster virtualization is still a good trade off between simplicity and flexibility.
I tend to think that letting the user choose multiple policy level at the same time for different components will be overkill.

Also on RHEL/Centos we have a system wide policy setting in `/etc/crypto-policies/config`.

> Even if there is some openshift resource (now or later) that is appropriate for our internal traffic defaults, I am even less sure that HCO should push that config down to the other operators. Yes, the alternative will probably lead to some duplicated code. But it's not a lot of code. Especially if using controller runtime. It would also make sure we honor those values for non HCO based installs.

My concern about this is that apiserver.spec.tlsSecurityProfile is an OCP/OKD specific feature and it's not available on plain k8s. So I still think we should try to handle it on HCO without introducing any dependency on OCP/OKD specific stuff on other components.

rmohr:
> > Even if there is some openshift resource (now or later) that is appropriate for our internal traffic defaults, I am even less sure that HCO should push that config down to the other operators. Yes, the alternative will probably lead to some duplicated code. But it's not a lot of code. Especially if using controller runtime. It would also make sure we honor those values for non HCO based installs.
> 
> My concern about this is that apiserver.spec.tlsSecurityProfile is an OCP/OKD specific feature and it's not available on plain k8s. So I still think we should try to handle it on HCO without introducing any dependency on OCP/OKD specific stuff on other components.

I think, that the propagation flow via HCO is in principle fine. Not introducing a direct dependency makes sense too. What I would love to explore more in-depth is the comment from @mhenriks that the TLS settings in OpenShift seem to be meant  for external communication and not internal communication. If this is the case, we would basically be the only part in OpenShift which would actually change our behaviour if admins just define allowed ciphers for external apiserver or ingress communication. That would definitely be surprising.

qinqon:
> > > Even if there is some openshift resource (now or later) that is appropriate for our internal traffic defaults, I am even less sure that HCO should push that config down to the other operators. Yes, the alternative will probably lead to some duplicated code. But it's not a lot of code. Especially if using controller runtime. It would also make sure we honor those values for non HCO based installs.
> > 
> > 
> > My concern about this is that apiserver.spec.tlsSecurityProfile is an OCP/OKD specific feature and it's not available on plain k8s. So I still think we should try to handle it on HCO without introducing any dependency on OCP/OKD specific stuff on other components.
> 
> I think, that the propagation flow via HCO is in principle fine. Not introducing a direct dependency makes sense too. What I would love to explore more in-depth is the comment from @mhenriks that the TLS settings in OpenShift seem to be meant for external communication and not internal communication. If this is the case, we would basically be the only part in OpenShift which would actually change our behaviour if admins just define allowed ciphers for external apiserver or ingress communication. That would definitely be surprising.

For example for network team only a pair of webhook will be affected and that internal.

tiraboschi:
We are try to address a feature that states in its goal:
> Goal: Cluster-wide TLS configuration with the ability to configure ciphers that would apply to all OpenShift components both for internal and external traffic.

rmohr:
@tiraboschi again the flow is in general fine, I however still can't understand from your response if the part where internal components are changing their config exists yet and if you are referencing the right objects as source.

tiraboschi:
> @tiraboschi again the flow is in general fine, I however still can't understand from your response if the part where internal components are changing their config exists yet and if you are referencing the right objects as source.

Honestly I'm aware only of MCO and CMO:
https://github.com/openshift/machine-config-operator/pull/2324
https://github.com/openshift/cluster-monitoring-operator/pull/1380
Others are still in the planning phase.

CMO config is also based on the APIServer resource.

rmohr:
> Honestly I'm aware only of MCO and CMO: [openshift/machine-config-operator#2324](https://github.com/openshift/machine-config-operator/pull/2324) [openshift/cluster-monitoring-operator#1380](https://github.com/openshift/cluster-monitoring-operator/pull/1380) Others are still in the planning phase.
> 
> CMO config is also based on the APIServer resource.

could you quickly check with openshift people if this is the desired final behaviour? If yes, this looks definitely fine.

PR #1718, opened 2022-01-19
Expose Kubevirt log verbosity to HCO CR

iholder-redhat:
Expose Kubevirt's [log verbosity](https://kubevirt.io/api-reference/master/definitions.html#_v1_logverbosity) to HCO CR. This will let users that deploy Kubevirt with HCO to tweak log verbosity for different components.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
Expose Kubevirt's log verbosity to HCO CR to tweak Kubevirt's log verbosity for different components
```


pkg/apis/hco/v1beta1/hyperconverged_types.go
```diff
@@ -46,7 +46,7 @@ type HyperConvergedSpec struct {
 
 	// featureGates is a map of feature gate flags. Setting a flag to `true` will enable
 	// the feature. Setting `false` or removing the feature gate, disables the feature.
-	// +kubebuilder:default={"withHostPassthroughCPU": false, "sriovLiveMigration": true, "enableCommonBootImageImport": true}
+	// +kubebuilder:default={"withHostPassthroughCPU": false, "sriovLiveMigration": true, "enableCommonBootImageImport": false}
```
tiraboschi, line 49: how is this related to kubevirt log verbosity?

iholder-redhat: It's auto-generated by `make generated`... not sure why.. =/

iholder-redhat: No idea how that happened.. fixing

iholder-redhat: Not sure what was going on. it's fixed now.

pkg/apis/hco/v1beta1/hyperconverged_types.go
```diff
@@ -119,6 +120,11 @@ type HyperConvergedSpec struct {
 	// +kubebuilder:validation:Enum=RemoveWorkloads;BlockUninstallIfWorkloadsExist
 	// +optional
 	UninstallStrategy *HyperConvergedUninstallStrategy `json:"uninstallStrategy,omitempty"`
+
+	// LogVerbosityConfig configures the verbosity level of Kubevirt's different components. The higher
+	// the value - the higher the log verbosity.
```
nunnatsa, line 125: I'm sure there are limitations. I guess no more than 5?

nunnatsa (changes requested):
1. Please run `make sanity` and commit the changed files.
2. Please add unit tests to the new code in pkg/controller/operands/kubevirt.go
3. Also, please add documentation in docs/cluster-configuration.md

pkg/apis/hco/v1beta1/hyperconverged_types.go
iholder-redhat: No, there is no clear limitation. This can be different for every component and can change in the future.

iholder-redhat:
> 1. Please run `make sanity` and commit the changed files.
> 2. Please add unit tests to the new code in pkg/controller/operands/kubevirt.go
> 3. Also, please add documentation in docs/cluster-configuration.md

Done

pkg/controller/operands/kubevirt.go
```diff
@@ -458,6 +458,9 @@ func getKVDevConfig(hc *hcov1beta1.HyperConverged) (*kubevirtcorev1.DeveloperCon
 	if useKVMEmulation {
 		devConf.UseEmulation = useKVMEmulation
 	}
+	if lv := hc.Spec.LogVerbosityConfig; lv != nil && lv.Kubevirt != nil {
```
tiraboschi, line 461: why also `&& lv.Kubevirt != nil` ?

iholder-redhat: This is because this function is responsible for kubvirt logs only. In this PR the only attribute for `hc.Spec.LogVerbosityConfig` is `Kubevirt`, but In the future log verbosity might be added to more components.

tiraboschi: ok

docs/cluster-configuration.md
```diff
@@ -828,3 +828,22 @@ Labels
...
+kind: HyperConverged
+metadata:
+  name: kubevirt-hyperconverged
+spec:
+  logVerbosityConfig:
+    kubevirt:
+      virtLauncher: 8
+      virtHandler: 4
+      virtController: 1
+```
```
orenc1, line 846: I think an example for `nodeVerbosity` would be helpful, since it's in the API

docs/cluster-configuration.md
```diff
@@ -828,3 +828,22 @@ Labels
     annotation_name="kubevirt.kubevirt.io/jsonpatch"
     severity=info
 ```
+## Log verbosity
+Currently logging verbosity is only supported for Kubevirt.
+
+### Kubevirt
+In order to define logging verbosity for Kubevirt, the following can be done:
```
orenc1, line 835: I think this would sound better:
```suggestion
In order to define logging verbosity for Kubevirt, the following can be configured on HyperConverged CR:
```

iholder-redhat:
Thanks @orenc1 for your review! I've addressed your feedback :+1:

orenc1:
> Thanks @orenc1 for your review! I've addressed your feedback +1

Thanks.
One thing I'm not quite sure I understood correctly:
Is the `nodeVerbosity` overrides the log verbosity for *all* components on the specified node, or the other way around (i.e. the component-specific verbosity level overrides the general verbosity per node, if specified)?

iholder-redhat:
> Thanks. One thing I'm not quite sure I understood correctly: Is the `nodeVerbosity` overrides the log verbosity for _all_ components on the specified node, or the other way around (i.e. the component-specific verbosity level overrides the general verbosity per node, if specified)?

From looking at [the code](https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-config/virt-config.go#L250) it seems that `nodeVerbosity` has the precedence.

In any case, this is a Kubevirt feature that already exists. This PR's scope is to only expose it to HCO CR. If we want to change Kubevirt's behavior we can do that in future work.

Thanks!

orenc1:
> > Thanks. One thing I'm not quite sure I understood correctly: Is the `nodeVerbosity` overrides the log verbosity for _all_ components on the specified node, or the other way around (i.e. the component-specific verbosity level overrides the general verbosity per node, if specified)?
> 
> From looking at [the code](https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-config/virt-config.go#L250) it seems that `nodeVerbosity` has the precedence.
> 
> In any case, this is a Kubevirt feature that already exists. This PR's scope is to only expose it to HCO CR. If we want to change Kubevirt's behavior we can do that in future work.
> 
> Thanks!

Yes, I know this question is out of scope for this PR, I asked for a general knowledge.
Thanks

iholder-redhat:
> > > Thanks. One thing I'm not quite sure I understood correctly: Is the `nodeVerbosity` overrides the log verbosity for _all_ components on the specified node, or the other way around (i.e. the component-specific verbosity level overrides the general verbosity per node, if specified)?
> > 
> > 
> > From looking at [the code](https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-config/virt-config.go#L250) it seems that `nodeVerbosity` has the precedence.
> > In any case, this is a Kubevirt feature that already exists. This PR's scope is to only expose it to HCO CR. If we want to change Kubevirt's behavior we can do that in future work.
> > Thanks!
> 
> Yes, I know this question is out of scope for this PR, I asked for a general knowledge. Thanks

With pleasure :)

docs/cluster-configuration.md
```diff
@@ -828,3 +828,25 @@ Labels
     annotation_name="kubevirt.kubevirt.io/jsonpatch"
     severity=info
 ```
+## Log verbosity
+Currently logging verbosity is only supported for Kubevirt.
+
+### Kubevirt
+In order to define logging verbosity for Kubevirt, the following can be configured on HyperConverged CR:
```
nunnatsa, line 835: As a reader of this documentation, I find it a bit crypt. I would add some more details. For example, what are the options under the `kubevirt` field? should I have to list all of them? What are the numbers? what is the range of them (e.g. 8 seems to be out of range). If you have a link that explains all of it in kubevirt project/api, please add it.

nunnatsa (changes requested):
In addition to the inline comment, please also remove the `##### Higher Log Verbosity` chapter in docs/cluster-configuration.md. This section explains how to change log verbosity using the jsonpatch annotation, and after this PR, it's not needed anymore.

docs/cluster-configuration.md
iholder-redhat: I understand your feeling, but these knobs are defined within Kubevirt as a developer configuration - maybe that's why it isn't really documented or well defined (yet).

> what are the options under the kubevirt field?

See what I at the bottom: "All the values defined [here](https://kubevirt.io/api-reference/master/definitions.html#_v1_logverbosity) can be applied". I've avoided re-writing them here intentionally - if they would change in the future this docs would become obsolete.

> should I have to list all of them?

You're not and it's also in the API link I've added - but I can add it explicitly if you insist - although again I think it's problematic for the same reason. This API could change in the future, we better write it in a single place and reference it from other places.

> What are the numbers? what is the range of them (e.g. 8 seems to be out of range)

As I've said, this is not well defined nor documented, and can change in the future. 8 is a complete valid value.

> In addition to the inline comment, please also remove the ##### Higher Log Verbosity chapter in docs/cluster-
configuration.md

Done :+1:

pkg/controller/operands/kubevirt.go
```diff
@@ -461,6 +461,9 @@ func getKVDevConfig(hc *hcov1beta1.HyperConverged) (*kubevirtcorev1.DeveloperCon
...
```
nunnatsa, line 464: I'm afraid this is not a complete reconciliation. 

We only check that HC field is not nil. But what if it is and the KV is not? In this case we won't force KV field to be nil as well, and then we're entering an endless reconcile loop, where we do find a gap between KV and HCO, but fail to close this gap.

docs/cluster-configuration.md
nunnatsa: I guess I wasn't clear. The questions above are not for the code review, but questions that the readers may ask themselves. I another words, I think we need more verbose documentation.

nunnatsa: @iholder-redhat ^^^

docs/cluster-configuration.md
```diff
@@ -828,3 +801,25 @@ Labels
     annotation_name="kubevirt.kubevirt.io/jsonpatch"
     severity=info
 ```
+## Log verbosity
```
nunnatsa, line 804: Please move up. The HyperConverged formal API should be one after the other. The jsonpatch should remain the last section. Let's put it after the `## Configure custom golden images` section and before the `## Workloads protection on uninstall` section.

pkg/controller/operands/kubevirt.go
nunnatsa: I was wrong. This code will work fine. The comparison is done later.

docs/cluster-configuration.md
iholder-redhat: Done

PR #1799, opened 2022-03-01
Opt-out from cluster-wide default node selector

tiraboschi:
The hyperconverged cluster operator defines
its own node selectors APIs that schedules
deployments and daemon sets on nodes according to
infra and workload classification.
Opting out from cluster-wide default node selector
on Openshift setting a specific annotation
(openshift.io/node-selector=) to prevent conflicts.

Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=2055950

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
Opt-out from cluster-wide default node selector
```

### DIFF
--- a/pkg/controller/hyperconverged/hyperconverged_controller_test.go
+++ b/pkg/controller/hyperconverged/hyperconverged_controller_test.go
@@ -79,10 +79,15 @@ var _ = Describe("HyperconvergedController", func() {
 	Describe("Reconcile HyperConverged", func() {
 		Context("HCO Lifecycle", func() {
 
+			var (
+				hcoNamespace *corev1.Namespace
+			)
+
 			BeforeEach(func() {
 				_ = os.Setenv("VIRTIOWIN_CONTAINER", commonTestUtils.VirtioWinImage)
 				_ = os.Setenv("OPERATOR_NAMESPACE", namespace)
 				_ = os.Setenv(hcoutil.HcoKvIoVersionName, version.Version)
+				hcoNamespace = commonTestUtils.NewHcoNamespace()
 			})
 
 			It("should handle not found", func() {
@@ -106,7 +111,7 @@ var _ = Describe("HyperconvergedController", func() {
 						Conditions: []metav1.Condition{},
 					},
 				}
-				cl := commonTestUtils.InitClient([]runtime.Object{hco})
+				cl := commonTestUtils.InitClient([]runtime.Object{hcoNamespace, hco})
 				r := initReconciler(cl, nil)
 
 				// Do the reconcile
@@ -137,13 +142,12 @@ var _ = Describe("HyperconvergedController", func() {
 			})
 
 			It("should create all managed resources", func() {
-
 				hco := commonTestUtils.NewHco()
 				hco.Spec.FeatureGates = hcov1beta1.HyperConvergedFeatureGates{
 					WithHostPassthroughCPU: true,
 				}
 
-				cl := commonTestUtils.InitClient([]runtime.Object{hco})
+				cl := commonTestUtils.InitClient([]runtime.Object{hcoNamespace, hco})
 
 				r := initReconciler(cl, nil)
 
@@ -511,7 +515,7 @@ var _ = Describe("HyperconvergedController", func() {
 				existingResource.Spec.Infra.NodePlacement.NodeSelector["key1"] = "BADvalue1"
 				existingResource.Spec.Workloads.NodePlacement.NodeSelector["key2"] = "BADvalue2"
 
-				cl := commonTestUtils.InitClient([]runtime.Object{hco, existingResource})
+				cl := commonTestUtils.InitClient([]runtime.Object{hcoNamespace, hco, existingResource})
 				r := initReconciler(cl, nil)
 
 				// mock a reconciliation triggered by a change in secondary CR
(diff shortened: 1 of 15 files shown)


pkg/controller/operands/namespace.go
```diff
@@ -0,0 +1,103 @@
...
+	for k, v := range namespace.Labels {
+		if found_v, hasKey := found.Labels[k]; hasKey {
+			if found_v != v {
+				found.Labels[k] = v
+				needUpdate = true
+			}
+		} else {
+			if found.Labels == nil {
+				found.Labels = make(map[string]string)
+			}
```
machadovilaca, line 57: Couldn't this be made outside the loop? Probably most of the time it would reduce the number of comparisons made.

pkg/controller/operands/namespace.go
```diff
@@ -0,0 +1,103 @@
...
+	for k, v := range namespace.Annotations {
+		if found_v, hasKey := found.Annotations[k]; hasKey {
+			if found_v != v {
+				found.Annotations[k] = v
+				needUpdate = true
+			}
+		} else {
+			if found.Annotations == nil {
+				found.Annotations = make(map[string]string)
+			}
```
machadovilaca, line 71: same as above

pkg/controller/hyperconverged/hyperconverged_controller_test.go
```diff
@@ -96,6 +96,7 @@ var _ = Describe("HyperconvergedController", func() {
 			})
 
 			It("should ignore invalid requests", func() {
+				hcoNamespace := commonTestUtils.NewHcoNamespace()
```
machadovilaca, line 99: This is done in many tests, couldn't this namespace be created only once and be reused in all of them?

tiraboschi: done

pkg/controller/operands/namespace.go
tiraboschi: done

pkg/controller/operands/namespace.go
tiraboschi: done

pkg/controller/hyperconverged/hyperconverged_controller_test.go
```diff
@@ -96,6 +101,7 @@ var _ = Describe("HyperconvergedController", func() {
 			})
 
 			It("should ignore invalid requests", func() {
+
```
machadovilaca, line 104: nit: remove this empty line

pkg/controller/hyperconverged/hyperconverged_controller_test.go
machadovilaca: in `pkg/controller/operands/operandHandler_test.go` there's the same issue

tiraboschi: done

pkg/controller/hyperconverged/hyperconverged_controller_test.go
tiraboschi: done

pkg/controller/operands/namespace.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func (h namespaceHandler) reset( /* No implementation */ ) {}
+
+// NewNamespace creates namespace definition
+func NewNamespace(hc *hcov1beta1.HyperConverged) *corev1.Namespace {
+	return &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:   hc.Namespace,
+			Labels: getLabels(hc, hcoutil.AppComponentDeployment),
```
nunnatsa, line 102: `getLabels` adds the `app.kubernetes.io/managed-by` label - do we want it on the namespace?

pkg/controller/operands/namespace.go
```diff
@@ -0,0 +1,108 @@
...
+
+func newNamespaceHandler(Client client.Client, Scheme *runtime.Scheme) *namespaceHandler {
+	return &namespaceHandler{
+		Client: Client,
+		Scheme: Scheme,
+	}
+}
+
+func (h *namespaceHandler) ensure(req *common.HcoRequest) *EnsureResult {
+	required := NewNamespace(req.Instance)
```
nunnatsa, line 29: Low priority: this is not going to be changed while the software is running. Please consider to compute it in advance, and only once (in `newNamespaceHandler` maybe? but then the ns is not available in the current function interface).

tiraboschi: done

pkg/controller/operands/namespace.go
tiraboschi: Yes, I agree: we are only setting an annotation, `app.kubernetes.io/managed-by` is too much.

tiraboschi:
/override coverage/coveralls

pkg/controller/operands/namespace.go
```diff
@@ -0,0 +1,78 @@
...
+				req.Logger.Error(err, "failed updating the namespace")
+				return &EnsureResult{
+					Err: err,
+				}
+			}
+		}
+		res.SetUpdated()
+		res.SetOverwritten(!req.HCOTriggered)
+		return res
+	}
```
nunnatsa, line 73: this condition and the `needUpdate` variable are not needed. it all can be placed in the previous condition. But we can fix it later.

/lgtm
/approve

nunnatsa (approved):

PR #1801, opened 2022-03-02
Fix reconcile not updating CR when cache misses when fetching

machadovilaca:
If `app` label is removed in `kubevirt-cluster-critical` PriorityClass, since this label is used as a selector for cache, when the hco reconcile loop tries to revert the change, it is not able to fetch the resource, then it tries to recreate it which is throwing an `already exists` error causing the operator to go into an "Reconcile Failed" condition, and the PriorityClass not being reverted.

In this PR, when recreating a resource to revert a change, if we catch an `already exists` error, we then try to update it even if the previous fetch resource action failed.

Signed-off-by: João Vilaça <jvilaca@redhat.com>

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
Fix reconcile not updating CR when cache misses when fetching
```

### DIFF
--- a/pkg/controller/operands/operand.go
+++ b/pkg/controller/operands/operand.go
@@ -93,7 +93,14 @@ func (h *genericOperand) ensure(req *common.HcoRequest) *EnsureResult {
 	found := h.hooks.getEmptyCr()
 	err = h.Client.Get(req.Ctx, key, found)
 	if err != nil {
-		return h.createNewCr(req, err, cr, res)
+		e := h.createNewCr(req, err, cr, res)
+		if e.Err != nil && strings.Contains(e.Err.Error(), "already exists") {
+			res.Err = nil
+			found.SetName(cr.GetName())
+		} else {
+			req.Logger.Error(err, "Failed to create object for "+h.crType)
+			return e
+		}
 	}
 
 	return h.handleExistingCr(req, key, found, cr, res)
@@ -198,7 +205,6 @@ func (h *genericOperand) createNewCr(req *common.HcoRequest, err error, cr clien
 		req.Logger.Info("Creating " + h.crType)
 		err = h.Client.Create(req.Ctx, cr)
 		if err != nil {
-			req.Logger.Error(err, "Failed to create object for "+h.crType)
 			return res.Error(err)
 		}
 		return res.SetCreated()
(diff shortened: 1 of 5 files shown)


pkg/controller/operands/operand.go
```diff
@@ -93,7 +93,14 @@ func (h *genericOperand) ensure(req *common.HcoRequest) *EnsureResult {
 	found := h.hooks.getEmptyCr()
 	err = h.Client.Get(req.Ctx, key, found)
 	if err != nil {
-		return h.createNewCr(req, err, cr, res)
+		e := h.createNewCr(req, err, cr, res)
+		if e.Err != nil && strings.Contains(e.Err.Error(), "already exists") {
```
tiraboschi, line 97: please use `IsAlreadyExists(err)` from `k8s.io/apimachinery/pkg/api/errors`

pkg/controller/operands/operand.go
```diff
@@ -93,7 +93,14 @@ func (h *genericOperand) ensure(req *common.HcoRequest) *EnsureResult {
...
+			res.Err = nil
+			found.SetName(cr.GetName())
```
tiraboschi, line 99: I fear that if the cache is not correctly populated for our object due to missing selectors, also the get on line 94 would fail and so we don't have found at all.
Probably the best and simplest strategy is:
- try creating as now,
- then, if it fails with IsAlreadyExists(err)=true, build a new client using New ( https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/client.go#L75 ) from controller runtime client to get a client that completely bypasses the cache on reads and writes
- and finally use that client to fetch the object again and update it

tiraboschi: cluster.APIReader should also work:
https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/cluster/cluster.go#L69-L72

tiraboschi:
Implemented in https://github.com/kubevirt/hyperconverged-cluster-operator/pull/1813

PR #1813, opened 2022-03-09
Fix cache miss hit on resource recreation

tiraboschi:
The k8s client is configured to use selectors
based on label to watch only a subset of
certain resources to limit the memory consumption.

IF the user explictly removes that label
from one of the watched object, the client
cache will miss it and so the operator will try
recreating it failing then with AlreadyExists.

Let's explictly detect this corner case and
fix it setting the missing label with
a custom client to bypass the cache.

Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=2032837

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
Fix cache miss hit on resource recreation
```

### DIFF
--- a/pkg/controller/operands/operand.go
+++ b/pkg/controller/operands/operand.go
@@ -8,6 +8,8 @@ import (
 	"reflect"
 	"strings"
 
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+
 	conditionsv1 "github.com/openshift/custom-resource-status/conditions/v1"
 
 	jsonpatch "github.com/evanphx/json-patch"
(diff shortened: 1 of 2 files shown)


machadovilaca:
@tiraboschi 
In #1801, in [pkg/controller/commonTestUtils/testClient.go:80](https://github.com/kubevirt/hyperconverged-cluster-operator/pull/1801/files#diff-491dc6316d44c924856cc5dd40f88848452ccad506c72555b60c26f019d2e0aeR80) I added a filter in the `Get` operation to simulate cache behaviour in unit tests.
Do you think it makes sense and that it could be added in this PR?

tiraboschi:
ignoring slightly decreased coverage due to a path that cannot be covered with the fake client
/override coverage/coveralls

pkg/controller/operands/operand.go
```diff
@@ -93,7 +95,43 @@ func (h *genericOperand) ensure(req *common.HcoRequest) *EnsureResult {
...
+				return &EnsureResult{
+					Err: geterr,
+				}
+			}
+			originalClient := h.Client
+			h.Client = apiClient
+			existingcrresult := h.handleExistingCr(req, key, found, cr, res)
+			h.Client = originalClient
+			return existingcrresult
+		}
```
nunnatsa, line 133: 1. Can you put all of this in a new function? the original is too long now.
2. I wonder (not urgent) if we can change it to be "thread safe" if we'll want to call the operand ensure methods in their own goroutine. WDYT?

tiraboschi: done

tiraboschi:
ignoring slightly decreased coverage due to a path that cannot be covered with the fake client
/override coverage/coveralls

nunnatsa (approved):
/approve

PR #1819, opened 2022-03-15
Ignore sriovLiveMigration FG on SNO

tiraboschi:
The default value of sriovLiveMigration is
true but the value is meaningless on SNO
cluster, let's simply ignore it:
the CRD defaulting mechanism on the
APIServer is not so flexible to allow to
specify a default based on the cluster topology.
On the other side, validating it with the webhook is overkilling
for users that simply expect to have the simplest possible
CR always accepted.

Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=2062227

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
Ignore sriovLiveMigration FG on SNO
```

### DIFF
--- a/controllers/operands/kubevirt_test.go
+++ b/controllers/operands/kubevirt_test.go
@@ -30,6 +30,7 @@ import (
 var _ = Describe("KubeVirt Operand", func() {
 	var (
 		basicNumFgOnOpenshift = len(hardCodeKvFgs) + len(sspConditionKvFgs)
+		deltaFGNotSNO         = 1
 	)
 
 	Context("KubeVirt Priority Classes", func() {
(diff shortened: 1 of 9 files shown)


controllers/operands/kubevirt_test.go
```diff
@@ -1271,6 +1285,42 @@ Version: 1.2.3`)
 					})
 				})
 
+				Context("should ignore SRIOVLiveMigration on SNO ", func() {
+					BeforeEach(func() {
+						hcoutil.GetClusterInfo = func() hcoutil.ClusterInfo {
+							return &commonTestUtils.ClusterInfoSNOMock{}
+						}
+					})
```
nunnatsa, line 1293: Do we want to restore `hcoutil.GetClusterInfo` after the test?

nunnatsa (approved):
/approve

controllers/operands/kubevirt_test.go
tiraboschi: it should be covered by the AfterEach on the external context (line 1227)

PR #1823, opened 2022-03-17
Filter out LiveMigrate workloadUpdateMethod on SNO

tiraboschi:
Explicitly filter out LiveMigrate
workloadUpdateMethod on SNO clusters.
On SNO cluster Live Migration FG is globally
disabled because we have a single worker node
but still we allow the user (and it's
the default) to pass
workloadUpdateMethod=["LiveMigrate"]
and this causes an infinite loop in
the upgrade handler in virt-controller
that currently ignores the LiveMigration FG.
Let's explicitly filter out LiveMigrate
workloadUpdateMethod to prevent the issue.

Bug-url: https://bugzilla.redhat.com/show_bug.cgi?id=2065308

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
Filter out LiveMigrate workloadUpdateMethod on SNO
```

### DIFF
--- a/controllers/operands/kubevirt.go
+++ b/controllers/operands/kubevirt.go
@@ -280,15 +280,33 @@ func hcWorkloadUpdateStrategyToKv(hcObject *hcov1beta1.HyperConvergedWorkloadUpd
 			*kvObject.BatchEvictionSize = *hcObject.BatchEvictionSize
 		}
 
-		if size := len(hcObject.WorkloadUpdateMethods); size > 0 {
+		filteredWUMethods := filterWorkloadupdatemethods(hcObject.WorkloadUpdateMethods)
+
+		if size := len(filteredWUMethods); size > 0 {
 			kvObject.WorkloadUpdateMethods = make([]kubevirtcorev1.WorkloadUpdateMethod, size)
-			for i, updateMethod := range hcObject.WorkloadUpdateMethods {
+			for i, updateMethod := range filteredWUMethods {
 				kvObject.WorkloadUpdateMethods[i] = kubevirtcorev1.WorkloadUpdateMethod(updateMethod)
 			}
 		}
 	}
 
 	return kvObject
+}
+
+func filterWorkloadupdatemethods(workloadUpdateMethods []string) []string {
+	var filteredWUMethods []string
+
+	if hcoutil.GetClusterInfo().IsInfrastructureHighlyAvailable() {
+		return workloadUpdateMethods
+	} else {
+		for _, method := range workloadUpdateMethods {
+			if method != string(kubevirtcorev1.WorkloadUpdateMethodLiveMigrate) {
+				filteredWUMethods = append(filteredWUMethods, method)
+			}
+		}
+	}
+
+	return filteredWUMethods
 }
 
 func getKVConfig(hc *hcov1beta1.HyperConverged) (*kubevirtcorev1.KubeVirtConfiguration, error) {
--- a/api/v1beta1/hyperconverged_types.go
+++ b/api/v1beta1/hyperconverged_types.go
@@ -357,6 +357,7 @@ type HyperConvergedWorkloadUpdateStrategy struct {
 	// precedence over more disruptive methods. For example if both LiveMigrate and Evict
 	// methods are listed, only VMs which are not live migratable will be restarted/shutdown.
 	// An empty list defaults to no automated workload updating.
+	// LiveMigrate is ignored on SNO clusters.
 	//
 	// +listType=atomic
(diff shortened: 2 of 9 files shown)


controllers/operands/kubevirt.go
```diff
@@ -280,9 +280,11 @@ func hcWorkloadUpdateStrategyToKv(hcObject *hcov1beta1.HyperConvergedWorkloadUpd
 			*kvObject.BatchEvictionSize = *hcObject.BatchEvictionSize
 		}
 
-		if size := len(hcObject.WorkloadUpdateMethods); size > 0 {
+		filterWorkloadUpdateMethods := filterWorkloadupdatemethods(hcObject.WorkloadUpdateMethods)
```
nunnatsa, line 283: Please rename to prevent name collision with the function, and improve readability.

controllers/operands/kubevirt.go
```diff
@@ -280,9 +280,11 @@ func hcWorkloadUpdateStrategyToKv(hcObject *hcov1beta1.HyperConvergedWorkloadUpd
...
+
+		if size := len(filterWorkloadUpdateMethods); size > 0 {
 			kvObject.WorkloadUpdateMethods = make([]kubevirtcorev1.WorkloadUpdateMethod, size)
-			for i, updateMethod := range hcObject.WorkloadUpdateMethods {
+			for i, updateMethod := range filterWorkloadUpdateMethods {
 				kvObject.WorkloadUpdateMethods[i] = kubevirtcorev1.WorkloadUpdateMethod(updateMethod)
 			}
 		}
```
nunnatsa, line 290: check if the `copy` builtin function can be used here.

controllers/operands/kubevirt.go
```diff
@@ -291,6 +293,20 @@ func hcWorkloadUpdateStrategyToKv(hcObject *hcov1beta1.HyperConvergedWorkloadUpd
 	return kvObject
 }
 
+func filterWorkloadupdatemethods(workloadUpdateMethods []string) []string {
+	var filterWorkloadUpdateMethods []string
```
nunnatsa, line 297: please rename. Same name as the function - I find it confusing

controllers/operands/kubevirt.go
```diff
@@ -291,6 +293,20 @@ func hcWorkloadUpdateStrategyToKv(hcObject *hcov1beta1.HyperConvergedWorkloadUpd
...
+		if workloadUpdateMethods[i] == string(kubevirtcorev1.WorkloadUpdateMethodLiveMigrate) {
+			if hcoutil.GetClusterInfo().IsInfrastructureHighlyAvailable() {
+				filterWorkloadUpdateMethods = append(filterWorkloadUpdateMethods, workloadUpdateMethods[i])
+			}
+		} else {
+			filterWorkloadUpdateMethods = append(filterWorkloadUpdateMethods, workloadUpdateMethods[i])
+		}
+	}
+	return filterWorkloadUpdateMethods
+}
```
nunnatsa, line 308: What do you think about this?

```suggestion
        if hcoutil.GetClusterInfo().IsInfrastructureHighlyAvailable() {
            copy(workloadUpdateMethods, workloadUpdateMethods)
        } else {
            for _, method := range workloadUpdateMethods {
                if method != string(kubevirtcorev1.WorkloadUpdateMethodLiveMigrate) {
                    filterWorkloadUpdateMethods = append(filterWorkloadUpdateMethods, method)
                }
            }
        }
	return filterWorkloadUpdateMethods
}
```

nunnatsa (commented):

controllers/operands/kubevirt.go
tiraboschi: done

controllers/operands/kubevirt.go
tiraboschi: no because we also need a typecast between string and kubevirtcorev1.WorkloadUpdateMethod

controllers/operands/kubevirt.go
tiraboschi: done

controllers/operands/kubevirt.go
tiraboschi: nice, and we can even do better avoiding the copy at all if not on SNO

nunnatsa (approved):
/approve

PR #1824, opened 2022-03-18
Remove leftovers during upgrades

tiraboschi:
Remove leftovers during upgrades
Extend the UpgradePatches mechanism
to be able to remove leftovers during
upgrades according to a json configuration
file.
Start consuming it by removing v2v-vmware and
vm-import-controller-config (not owned by HCO)
ConfigMap when upgrading from versions <=1.6.0

In a future PR, other pieces of current code
(removal of old SSP CRDs, old metrics,
old quickstarts...) can be migrated to
this generic mechanism.

Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=2063991

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
Remove leftovers during upgrades
```

### DIFF
--- a/controllers/hyperconverged/hyperconverged_controller.go
+++ b/controllers/hyperconverged/hyperconverged_controller.go
@@ -7,6 +7,8 @@ import (
 	"os"
 	"reflect"
 
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
 	"github.com/blang/semver/v4"
 	jsonpatch "github.com/evanphx/json-patch"
 	"github.com/google/uuid"
@@ -264,6 +266,11 @@ func (r *ReconcileHyperConverged) Reconcile(ctx context.Context, request reconci
 
 	if r.firstLoop {
 		r.firstLoopInitialization(hcoRequest)
+		if err := validateUpgradePatches(hcoRequest); err != nil {
+			logger.Error(err, "Failed validating upgrade patches file")
+			r.eventEmitter.EmitEvent(hcoRequest.Instance, corev1.EventTypeWarning, "Failed validating upgrade patches file", err.Error())
+			os.Exit(1)
+		}
 	}
 
 	result, err := r.doReconcile(hcoRequest)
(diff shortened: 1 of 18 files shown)


nunnatsa (changes requested):
Not a change in this PR, but need to be fixed:

the current code calls `validateUpgradePatches` for every reconcile loop during upgrade, This function reads from a file over and over again while the file is not expected to be modified.

Please consider to move the function call a few lines above, into the condition with the comment `// get into upgrade mode`. this one happens only once, when initiating upgrade mode.

tiraboschi:
> Not a change in this PR, but need to be fixed:
> 
> the current code calls `validateUpgradePatches` for every reconcile loop during upgrade, This function reads from a file over and over again while the file is not expected to be modified.
> 
> Please consider to move the function call a few lines above, into the condition with the comment `// get into upgrade mode`. this one happens only once, when initiating upgrade mode.

The file read and the decoding of the json are actually cached, but yes we can validate only once in the beginning.

controllers/hyperconverged/hyperconverged_controller.go
```diff
@@ -355,7 +357,10 @@ func (r *ReconcileHyperConverged) doReconcile(req *common.HcoRequest) (reconcile
 		// get into upgrade mode
 
 		r.upgradeMode = true
-
+		err := validateUpgradePatches(req)
+		if err != nil {
+			return reconcile.Result{Requeue: true}, err
```
nunnatsa, line 362: Maybe I'm wrong here, but returning requeue=true  won't solve the problem. The error will happen if we can't open the file, if its format is wrong. Another option is when the HyperConverged version is with a wrong format.

For the first cases WDYT about validating on boot, and fail the whole process? that way we could make sure we're releasing a valid file, because no test will run.

For the last one - maybe we can check the version in the create/update webhooks?

tiraboschi: If the json file is corrupted/wrong, requeue=true will not fix it for us.
But for sure the upgrade lanes will detect it so the risk of releasing with a broken file is really low.
On the other side, the earlier we fail the better it is.

nunnatsa: Still, returning requeue=true here may cause an endless loop - because nothing can fix these problems.

PR #1830, opened 2022-03-23
Add metrics types to HCO metrics documentation

assafad:
When querying Prometheus it is extremely important to know the metric type: gauge/counter/histogram/summary. 
This PR Adds metrics types to HCO metrics documentation.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
None
```


assafad:
Hi @machacekondra, @sradco. Can you please have a look?

assafad:
@nunnatsa can you please have a look?

tools/metricsdocs/metricsdocs.go
```diff
@@ -42,18 +42,20 @@ func writeToFile(metricsList metricList) {
 type metric struct {
 	name        string
 	description string
+	mType       string
```
orenc1, line 45: If the `m` prefix for this field stands for "metric", I would say it is redundant, since the struct named `metric` so it is obvious the type of the metric. Same comment for:
https://github.com/kubevirt/hyperconverged-cluster-operator/pull/1830/files#diff-4bc91a74bc47b8467cdcd25a7e3fe1651dbe44907f478377eca31b982f444f23R20

assafad: I added this prefix to avoid confusion with Golang's saved word - `type`.
Do you have any suggestions for this field?

pkg/metrics/metrics.go
```diff
@@ -175,13 +178,14 @@ func getLabelsForUnsafeAnnotation(unsafeAnnotation string) prometheus.Labels {
...
 	Help   string
+	Type   string
 }
 
 func (hm hcoMetrics) GetMetricDesc() []MetricDescription {
 	res := make([]MetricDescription, len(hm.metricDescList))
 	i := 0
 	for _, md := range hm.metricDescList {
-		res[i] = MetricDescription{FqName: md.fqName, Help: md.help}
+		res[i] = MetricDescription{FqName: md.fqName, Help: md.help, Type: md.mType}
```
orenc1, line 188: Not a part of this PR, but you already using the `range` iterator which provides you also the index, which is being discarded by using `_`.
You can assign it to `i`, and then you won't need to: 1) initialize it with 0; 2) increment it by 1 at every iteration.
```
	for i, md := range hm.metricDescList {
		res[i] = MetricDescription{FqName: md.fqName, Help: md.help, Type: md.mType}
	}
```

Same comment goes for line 89 above.

tools/metricsdocs/metricsdocs.go
orenc1: yes, you're right. we don't want to export this field by capitalizing it, and using `_type` instead won't improve much.

pkg/metrics/metrics.go
assafad: Correct me if I'm wrong, but I think in both cases the loops iterate over a map, so `range` provides keys and values.

orenc1: Yes, `hm.metricList` is a map. sorry :(

What do you think about the following code instead (it will relief the need for using an indexing variable):
```
	var res []MetricDescription
	for _, md := range hm.metricDescList {
		res = append(res, MetricDescription{FqName: md.fqName, Help: md.help, Type: md.mType})
	}
```

orenc1: This is just a style nit. I'm good with the current code

assafad: Agree, it looks better. updated

nunnatsa (approved):
/approve

PR #1832, opened 2022-03-24
Allow editing and disabling common DataImportCronTemplates

nunnatsa:
HCO now supports the removing of a specific DataImportCronTemplate from the list in SSP. Also, HCO allow modying the storage field.

### New field: `status.dataImportCronTemplates`
This field is a list of the actual `dataImportCronTemplates` as HCO updated the SSP CR. The list contains both the common and the user-defined templates (the list in the spec field contains only the user-defined templates). 

The purpose of this field is for transparency, so the user could see the actual templates. It should also ease the modification of the templates, because it will give the information about the common-templates, which are currently not visible to the user.

The templates in the `status.dataImportCronTemplates` list contains a new status object with two fields:
* `commonTemplate`; boolean; indicates whether this is a common template (`true`), or a custom one (`false`).
* `modified`; boolean; indicates if a common template was customized. Always false for custom templates.

For example:
```yaml
kind: HyperConverged
metadata:
  name: kubevirt-hyperconverged
spec:
  …
status:
  dataImportCronTemplates:
  - metadata:
      name: rhel8-image-cron
    spec:
      …
    status:
      commonTemplate: true
      Modified: false
  - metadata:
      name: my-castom-image
    spec:
      …
    status:
      commonTemplate: false
```
### New annotation: `dataimportcrontemplate.kubevirt.io/enable`
`spec.dataImportCronTemplates[*].metadata.annotations["dataimportcrontemplate.kubevirt.io/enable"]`
Boolean string. Default = `"true"`

This new annotation in the `dataImportCronTemplate` object, if set to `false`, signals HCO not to set this template in the SSP CR, or remove the template from the SSP CR if it is currently listed there. For common templates, the specific template is recognized by its name. For example, in order to disable the common "rhel8-image-cron" template, add this to the spec:
```yaml
kind: HyperConverged
metadata:
  name: kubevirt-hyperconverged
spec:
  dataImportCronTemplates:
  - metadata:
      annotations:
        dataimportcrontemplate.kubevirt.io/enable: false
      name: rhel8-image-cron
```

## HCO Changes
Currently, the HCO webhook rejects changes to the HyperConverged CR, if the spec.dataImportCronTemplates list contains a common template. After this change, HCO will not reject such requests. Instead, HCO allows to modify the following fields in the common templates:

* spec.dataImportCronTemplates[*].spec.template.spec.storage

For example
```yaml
kind: HyperConverged
metadata:
  name: kubevirt-hyperconverged
spec:
  dataImportCronTemplates:
  - metadata:
      name: rhel8-image-cron
    spec:
      template:
        spec:
          storageClassName: "some-name"
```

If this field is added to the HyperConverged spec, HCO will replace the storage field if exist, or add it if it is missing.

Any other field will be ignored.

* If the DataImportCronTemplate’s dataimportcrontemplate.kubevirt.io/enable  annotation is set to false, HCO will not set this template in the SSP CR, or delete it from SSP CR if it is there.

HCO will copy the template list, as written to the SSP CR, to the new `status.dataImportCronTemplates` field, with addition of two optional fields: `commonTemplate` and `modified`, as described above. If the template is set to “disabled” in the spec, it won’t appear in the status.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
HCO now supports the removing of a specific DataImportCronTemplate from the list in SSP. Also, HCO allow modying the storage field.

```


api/v1beta1/hyperconverged_types.go
```diff
@@ -426,6 +430,22 @@ type LogVerbosityConfiguration struct {
 	Kubevirt *v1.LogVerbosity `json:"kubevirt,omitempty"`
 }
 
+// DataImportCroneStatus is the status field of the DIC template
+type DataImportCroneStatus struct {
```
tiraboschi, line 434:
```suggestion
type DataImportCronStatus struct {
```

tiraboschi (approved):

orenc1 (approved):

PR #1835, opened 2022-03-24
Fix a race condition on object recreation

tiraboschi:
If the cluster admin deletes one of the
HCO managed resource, HCO will try to
recreate it but (according to the delete
instant) HCO could potentially consume
a cached in-memoery reppresentation
of the object that can include a
resourceVersion. In that case HCO
was going to fail in a loop with:
"resourceVersion should not be set on objects to be created".
Let's ensure we never try to re-create an object
with a value for resourceVersion.

Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=2062321

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
Fix a race condition on object recreation
```

### DIFF
--- a/controllers/operands/operand.go
+++ b/controllers/operands/operand.go
@@ -97,14 +97,17 @@ func (h *genericOperand) ensure(req *common.HcoRequest) *EnsureResult {
 	found := h.hooks.getEmptyCr()
 	err = h.Client.Get(req.Ctx, key, found)
 	if err != nil {
-		res = h.createNewCr(req, err, cr, res)
-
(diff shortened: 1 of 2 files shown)


nunnatsa (approved):
/approve

PR #1953, opened 2022-05-17
Fix alert bug

nunnatsa:
Fix Prometheus alert bug, where the KubevirtHyperconvergedClusterOperatorInstallationNotCompletedAlert alert is not working when the HyperConverged CR is not present.

This alert is all about the existence of the HyperConverged CR, and so when it removed with the CR, the alert is not working. The root cause is that HCO manages the PrometheusRule resource as any other resource, and set its owner resource to the HyperConverged CR. In addition, the code that reconciles the PrometheusRule resource is only run when the HypeConverged CR is found. 

This PR moves the PrometheusRule handling code to a new package name `alerts`. Then the new PrometheusRule reconcile function is called regardless the existence of the HyperConverged CR. Also, the PrometheusRule ownerReference is now set to the hco-operator Deployment, so it will be garbage collected on the operator uninstallation. 

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
Fix Prometheus alert bug, where the KubevirtHyperconvergedClusterOperatorInstallationNotCompletedAlert alert is not working when the HyperConverged CR is not present.
```


controllers/alerts/alerts.go
```diff
@@ -0,0 +1,236 @@
...
+		Labels: map[string]string{
+			severityAlertLabelKey:  "info",
+			partOfAlertLabelKey:    partOfAlertLabelValue,
+			componentAlertLabelKey: componentAlertLabelValue,
+		},
+	}
+}
+
+// Recording rules for openshift/cluster-monitoring-operator
+func createRequestCPUCoresRule() monitoringv1.Rule {
```
tiraboschi, line 231: is this needed?

controllers/hyperconverged/hyperconverged_controller_test.go
```diff
@@ -714,7 +714,7 @@ var _ = Describe("HyperconvergedController", func() {
 				).To(BeNil())
 
 				Expect(foundResource.Status.RelatedObjects).ToNot(BeNil())
-				Expect(len(foundResource.Status.RelatedObjects)).Should(Equal(18))
+				Expect(len(foundResource.Status.RelatedObjects)).Should(Equal(17))
```
tiraboschi, line 717: is this due to the prometheus Rule referencing the deployment?
maybe we should still list it under relatedObjects

controllers/alerts/alerts.go
nunnatsa: yes. it was there before the PR. Just wrapped it in a function

tiraboschi:
/override coverage/coveralls

nunnatsa:
hco-e2e-upgrade-index-aws lane passed
/override ci/prow/hco-e2e-upgrade-index-azure

controllers/hyperconverged/hyperconverged_controller_test.go
nunnatsa: Fixed. I added the PrometheusRule to the relatedObjects list, but the implementation is in the `alerts` package, so it won't affect this test.

tiraboschi (approved):

PR #1992, opened 2022-06-07
Require OKD/OCP console to enable cluster-monitoring on namespace

tiraboschi:
Set operatorframework.io/cluster-monitoring=true annotation
on the CSV to require the console
to set label openshift.io/cluster-monitoring=true
on the install namespace at creation time.
This will ensure that cluster-monitoring can
scrape the metrics on the namespace even before the cluster
admin creates the CR for HCO.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
Require OKD/OCP console to enable cluster-monitoring on namespace
```

### DIFF
--- a/deploy/index-image/community-kubevirt-hyperconverged/1.8.0/manifests/kubevirt-hyperconverged-operator.v1.8.0.clusterserviceversion.yaml
+++ b/deploy/index-image/community-kubevirt-hyperconverged/1.8.0/manifests/kubevirt-hyperconverged-operator.v1.8.0.clusterserviceversion.yaml
@@ -12,6 +12,7 @@ metadata:
     createdAt: "2020-10-23 08:58:25"
     description: A unified operator deploying and controlling KubeVirt and its supporting
       operators with opinionated defaults
+    operatorframework.io/cluster-monitoring: "true"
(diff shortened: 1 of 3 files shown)


orenc1:
Note that this annotation won't automatically add the `openshift.io/cluster-monitoring: 'true'` label to the namespace by default, but will rather add a checkbox on the OLM UI that the user can accept monitoring on the namespace, and in case the checkbox is checked, the label will be applied.

image

nunnatsa:
> Note that this annotation won't automatically add the `openshift.io/cluster-monitoring: 'true'` label to the namespace by default, but will rather add a checkbox on the OLM UI that the user can accept monitoring on the namespace, and in case the checkbox is checked, the label will be applied.
> 
> image

:+1: 

It also does not cover installation from command line. Do we need it in addition to #1993 ?

tiraboschi:
> > Note that this annotation won't automatically add the `openshift.io/cluster-monitoring: 'true'` label to the namespace by default, but will rather add a checkbox on the OLM UI that the user can accept monitoring on the namespace, and in case the checkbox is checked, the label will be applied.
> > image
> 
> +1
> 
> It also does not cover installation from command line. Do we need it in addition to #1993 ?

I think we should drop this PR: the user can be tempted to not click the checkbox on the UI but then, as for #1993 , we are not going to honor that choice. So I think that not exposing it is more consistent with the actual behaviour.

/hold

tiraboschi:
Fixed in #1993, closing this

PR #2011, opened 2022-06-19
Fix panic: "nil pointer dereference"

nunnatsa:
The software craches when updating a metric resource, because then we're
trying to send a nil HyperConverged pointer to a function as an interface.
The function does check for nil, but this is not working, because
`interface(nil) != nil`.

This PR fixes the bug by not sending the nil interface but an actual
nil, and also changes the check to use reflection in order to make sure
that the function parameter is nil.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
Fix sw crash 
```


orenc1 (approved):
/approve

PR #2044, opened 2022-07-29
upgrade_test: use ssh instead of expect

tiraboschi:
Use ssh instead of expect over serial console to check vm uptime:
expect over serial console proved to be no so robust, let's
replace it with an ssh connection.
Generate an ssh keypair, inject it with cloud-init
in the test vm (cirros doesn't really support
yaml based cloud-config but only sh scripts),
connect over ssh (eventually retrying in case of failures).

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
NONE
```


nunnatsa (approved):
/approve

PR #2045, opened 2022-07-31
Apply the ginkgolinter linter

nunnatsa:
Force standards of using ginkgo and gomega.

See additional info in https://github.com/nunnatsa/ginkgolinter

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
None
```

### DIFF
--- a/pkg/webhooks/mutator/mutator_test.go
+++ b/pkg/webhooks/mutator/mutator_test.go
@@ -53,7 +53,7 @@ var _ = Describe("webhooks mutator", func() {
 		sspv1beta1.AddToScheme,
 		corev1.AddToScheme,
 	} {
-		Expect(f(s)).To(BeNil())
+		Expect(f(s)).To(Succeed())
 	}
 
 	codecFactory := serializer.NewCodecFactory(s)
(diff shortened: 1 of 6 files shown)


tiraboschi (approved):

pkg/webhooks/mutator/mutator_test.go
```diff
@@ -53,15 +53,15 @@ var _ = Describe("webhooks mutator", func() {
...
+		Expect(f(s)).To(Succeed())
 	}
 
 	codecFactory := serializer.NewCodecFactory(s)
 	corev1Codec := codecFactory.LegacyCodec(corev1.SchemeGroupVersion)
 
 	Context("Check mutating webhook for namespace deletion", func() {
 		BeforeEach(func() {
-			Expect(os.Setenv("OPERATOR_NAMESPACE", HcoValidNamespace)).To(BeNil())
+			Expect(os.Setenv("OPERATOR_NAMESPACE", HcoValidNamespace)).To(Succeed())
```
orenc1, line 64: Are you sure this is the most appropriate matcher in this case?
From what I understood, `Succeed()` matcher should be used with a function that returns an error-type as its first or only return value.
https://onsi.github.io/gomega/#succeed

nunnatsa: Totally agree. This is what the linter looks for, and here is the os.Setenv code from the standard library:

```golang
func Setenv(key, value string) error {
	err := syscall.Setenv(key, value)
	if err != nil {
		return NewSyscallError("setenv", err)
	}
	return nil
}
```

orenc1: ok, cool 👍

PR #2046, opened 2022-08-01
digester: retry on docker.GetDigest failures

tiraboschi:
Sometimes the image digester tool fails due to
`too many requests to registry` error from the server
and this leads to unrecoverable error in the
Build and Push Images job.
This PR introduces a retry mechanism
to make it more resilient.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
NONE
```

### DIFF
--- a/tools/digester/main.go
+++ b/tools/digester/main.go
@@ -25,12 +25,15 @@ import (
 	"encoding/csv"
 	"flag"
 	"fmt"
+	"math/rand"
 	"os"
 	"strings"
 	"sync"
 	"time"
 
 	"github.com/containers/image/v5/docker"
+	"github.com/containers/image/v5/types"
+	"github.com/opencontainers/go-digest"
 )
 
 const (
@@ -96,7 +99,7 @@ func init() {
 	}
 }
 func main() {
-	ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*60)
 	defer cancel()
 
 	// Single image use-case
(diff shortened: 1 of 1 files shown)


tools/digester/main.go
```diff
@@ -274,13 +276,29 @@ func readOneDigest(ctx context.Context, image *Image, index int, wg *sync.WaitGr
...
+			fmt.Println("retrying after error:", err)
+			time.Sleep(sleep)
+			sleep *= 2
+		}
+		digest, err = docker.GetDigest(ctx, sys, imgRef)
+		if err == nil {
+			return digest, nil
+		}
+	}
+	return "", fmt.Errorf("aborting after %d attempts, last error: %s", attempts, err)
```
nunnatsa, line 299: Please change `%s` to `%w`

```suggestion
	return "", fmt.Errorf("aborting after %d attempts, last error: %w", attempts, err)
```

tools/digester/main.go
```diff
@@ -274,13 +276,29 @@ func readOneDigest(ctx context.Context, image *Image, index int, wg *sync.WaitGr
...
-	digest, err := docker.GetDigest(ctx, nil, imgRef)
+	digest, err := retryGetDigest(5, 1, ctx, nil, imgRef)
+
 	exitOnError(err, "failed to get digest from image")
 
 	ch <- message{index: index, digest: digest.Hex(), fullName: fullName}
 	wg.Done()
 }
 
+func retryGetDigest(attempts int, sleep time.Duration, ctx context.Context, sys *types.SystemContext, imgRef types.ImageReference) (digest digest.Digest, err error) {
```
nunnatsa, line 287: please move ctx to be the first parameter
```suggestion
func retryGetDigest(ctx context.Context, attempts int, sleep time.Duration, sys *types.SystemContext, imgRef types.ImageReference) (digest digest.Digest, err error) {
```

tools/digester/main.go
```diff
@@ -274,13 +276,29 @@ func readOneDigest(ctx context.Context, image *Image, index int, wg *sync.WaitGr
 	imgRef, err := docker.ParseReference(fullName)
 	exitOnError(err, "failed to parse container reference")
 
-	digest, err := docker.GetDigest(ctx, nil, imgRef)
+	digest, err := retryGetDigest(5, 1, ctx, nil, imgRef)
```
nunnatsa, line 279: Duration of 1 is one nanosecond - isn't it too fast? maybe start from 1 second? Just need to make sure that the final duration won't cause timeout.

tools/digester/main.go
```diff
@@ -25,6 +25,8 @@ import (
 	"encoding/csv"
 	"flag"
 	"fmt"
+	"github.com/containers/image/v5/types"
+	"github.com/opencontainers/go-digest"
```
nunnatsa, line 29: please move down - to not be with the standard library packages.

nunnatsa (commented):
Added several comments. Most of them are cosmetic. But please consider to change the initial sleep time.

nunnatsa (approved):
/approve

tools/digester/main.go
tiraboschi: done, thanks

tools/digester/main.go
tiraboschi: done, thanks

tools/digester/main.go
tiraboschi: done, thanks

tools/digester/main.go
tiraboschi: Now it can potentially range from:
0, 1, 3, 9, 27 = 40 seconds up to:
0, 1+0.5, 3+1.5, 9+4.5, 27+13.5 = 60 seconds

PR #2051, opened 2022-08-05
Expose the webhook with configured TLS config

tiraboschi:
Properly expose the the validating webhook
with TLS as configured on HCO CR or
at cluster level.
The configuration is cached in the webhook
and refreshed each time the cluster admin tries to create
or refresh the HCO CR.

Test it e2e on plain k8s and OCP/OKD.
Please also notice that with FIPS enabled golang
builders we can get only a subset of
allowed ciphers for the same TLS profile.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
Expose the webhook with configured TLS config
```


pkg/webhooks/validator/validator.go
```diff
@@ -28,25 +38,88 @@ const (
...
-func (wh WebhookHandler) ValidateCreate(hc *v1beta1.HyperConverged) error {
+// Handle admits a pod if a specific annotation exists.
+func (wh *WebhookHandler) Handle(ctx context.Context, req admission.Request) admission.Response {
+
+	ctx = admission.NewContextWithRequest(ctx, req)
+
+	// Get the object in the request
+	obj := &v1beta1.HyperConverged{}
+
+	dryRun := *req.DryRun
```
nunnatsa, line 68: will panic if `req.DryRun == nil`

tiraboschi: done, thanks

nunnatsa (approved):
/approve
:100:

PR #2054, opened 2022-08-16
Migrate index generation process to File-Based Catalog

orenc1:
SQLite-Based catlog format, which we were using so far, is being deprecated in favour of the new format of file-based catalog.
With this change, the 'opm index add --bundles' command no longer works, and we need to change the workflow to work with json files inside the catalog image.
References:
- https://olm.operatorframework.io/docs/reference/file-based-catalogs/
- https://github.com/k8s-operatorhub/community-operators/discussions/505

Signed-off-by: Oren Cohen <ocohen@redhat.com>

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
migrate to FBC in index image workflows.
```

### DIFF
--- a/hack/build-index-image.sh
+++ b/hack/build-index-image.sh
@@ -14,8 +14,8 @@ set -ex
 
 PROJECT_ROOT="$(readlink -e "$(dirname "${BASH_SOURCE[0]}")"/../)"
 DEPLOY_DIR="${PROJECT_ROOT}/deploy/olm-catalog"
+OUT_DIR="${PROJECT_ROOT}/_out"
 PACKAGE_NAME="community-kubevirt-hyperconverged"
-INDEX_IMAGE_PARAM=
 IMAGE_REGISTRY=${IMAGE_REGISTRY:-quay.io}
 REGISTRY_NAMESPACE=${REGISTRY_NAMESPACE:-kubevirt}
 BUNDLE_REGISTRY_IMAGE_NAME=${BUNDLE_REGISTRY_IMAGE_NAME:-hyperconverged-cluster-bundle}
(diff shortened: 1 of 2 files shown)


hack/build-index-image.sh
```diff
@@ -35,22 +34,64 @@ function create_index_image() {
...
+  then
+    # The index image is already in file-based format. Extracting its catalog file
+    oc image extract ${INDEX_IMAGE_NAME} --file /configs/catalog.json
+  else
+    # The migration took place
+    mv fbc-catalog/community-kubevirt-hyperconverged/catalog.json catalog.json
+  fi
+
+  ${OPM} render ${BUNDLE_IMAGE_NAME} > bundle.json
+  CSV_NAME=$(jq .name bundle.json | tr -d '"')
```
nunnatsa, line 57: you can use the `-r` flag instead of `tr -d`:

```suggestion
  CSV_NAME=$(jq -r .name bundle.json)
```

hack/build-index-image.sh
```diff
@@ -35,22 +34,64 @@ function create_index_image() {
...
 
   # Referencing the unstable bundle digest in the index image, rather than the floating tag, to avoid
   # unalignment between cached index image and fetched bundle image.
   BUNDLE_IMAGE_NAME=$("${PROJECT_ROOT}/tools/digester/digester" --image "${BUNDLE_IMAGE_NAME}")
 
-  # shellcheck disable=SC2086
-  ${OPM} index add --bundles "${BUNDLE_IMAGE_NAME}" ${INDEX_IMAGE_PARAM} --tag "${INDEX_IMAGE_NAME}" -u podman --mode semver
+
+  # File-Based Catalog handling
+  ${OPM} migrate ${INDEX_IMAGE_NAME} fbc-catalog || true
```
nunnatsa, line 46: I think we should avoid writing to the root directory of the project. What do you think about output to `_out/fbc-catalog`?

If you do that, please add an env var for this path, in order to use it later in the script.

nunnatsa (commented):
Looks very good.

Added some comments.

hack/build-index-image.sh
orenc1: I think it would be better to change the working directory in this part to `${PROJECT_ROOT}/_out` to avoid adding the `_out/ ` prefix for every file reference.

hack/build-index-image.sh
```diff
@@ -35,22 +35,60 @@ function create_index_image() {
...
 
   # Referencing the unstable bundle digest in the index image, rather than the floating tag, to avoid
   # unalignment between cached index image and fetched bundle image.
   BUNDLE_IMAGE_NAME=$("${PROJECT_ROOT}/tools/digester/digester" --image "${BUNDLE_IMAGE_NAME}")
 
-  # shellcheck disable=SC2086
-  ${OPM} index add --bundles "${BUNDLE_IMAGE_NAME}" ${INDEX_IMAGE_PARAM} --tag "${INDEX_IMAGE_NAME}" -u podman --mode semver
+
+  # File-Based Catalog handling
+  cd "${OUT_DIR}"
```
nunnatsa, line 47: This will work. But I'm not sure I'm happy with `cd` and `cd` back. What do you think about placing all of this in a new function, then call it like this?

```shell
(cd "${OUT_DIR}" && my_new_function)
```

orenc1: valid point.
updated, thanks.

PR #2068, opened 2022-08-28
Add HCO API to set CDIConfig DataVolumeTTLSeconds

arnongilboa:
Signed-off-by: Arnon Gilboa <agilboa@redhat.com>

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:
```release-note
Add HCO API to set CDIConfig DataVolumeTTLSeconds
```

### DIFF
--- a/controllers/operands/cdi.go
+++ b/controllers/operands/cdi.go
@@ -122,6 +122,7 @@ func NewCDI(hc *hcov1beta1.HyperConverged, opts ...string) (*cdiv1beta1.CDI, err
 			spec.Config.InsecureRegistries = make([]string, length)
 			copy(spec.Config.InsecureRegistries, hc.Spec.StorageImport.InsecureRegistries)
 		}
+		spec.Config.DataVolumeTTLSeconds = hc.Spec.StorageImport.DataVolumeTTLSeconds
 	}
 
 	if hc.Spec.Infra.NodePlacement != nil {
--- a/controllers/operands/cdi_test.go
+++ b/controllers/operands/cdi_test.go
@@ -603,11 +603,14 @@ var _ = Describe("CDI Operand", func() {
 
 		Context("Test StorageImport", func() {
 
-			It("should add InsecureRegistries if exists in HC and missing in CDI", func() {
+			testTTL := int32(1)
+
+			It("should add missing CDIConfig fields as set in HCO CR", func() {
 				existingResource, err := NewCDI(hco)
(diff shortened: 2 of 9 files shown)


controllers/operands/cdi.go
```diff
@@ -122,6 +122,7 @@ func NewCDI(hc *hcov1beta1.HyperConverged, opts ...string) (*cdiv1beta1.CDI, err
 			spec.Config.InsecureRegistries = make([]string, length)
 			copy(spec.Config.InsecureRegistries, hc.Spec.StorageImport.InsecureRegistries)
 		}
+		spec.Config.DataVolumeTTLSeconds = hc.Spec.StorageImport.DataVolumeTTLSeconds
```
akalenyu, line 125: Do we want GC on by default in near term? if so, maybe force 0 if not defined?

controllers/operands/cdi_test.go
```diff
@@ -653,15 +659,18 @@ var _ = Describe("CDI Operand", func() {
...
 				Expect(foundCDI.Spec.Config.InsecureRegistries).To(BeNil())
+				Expect(foundCDI.Spec.Config.DataVolumeTTLSeconds).To(BeNil())
 			})
 
-			It("should modify InsecureRegistries according to HCO CR", func() {
+			It("should modify CDIConfig according to HCO CR", func() {
 				existingCDI, err := NewCDI(hco)
 				Expect(err).ToNot(HaveOccurred())
 				existingCDI.Spec.Config.InsecureRegistries = []string{"first:5000", "second:5000", "third:5000"}
+				existingCDI.Spec.Config.DataVolumeTTLSeconds = &[]int32{2}[0]
```
nunnatsa, line 669: Please use `pointer.Int32(2)` instead:
```suggestion
				existingCDI.Spec.Config.DataVolumeTTLSeconds = pointer.Int32(2)
```

arnongilboa:
/hold

We decided to hold it for now, as we prefer having as few CDI knobs as possible in HCO, and this seems like something that someone would disable only for debugging or if there was a bug. `DataVolumeTTLSeconds` is configured in the CDI CR, so we can use a jsonpatch to change it if there are extenuating circumstances.

PR #2072, opened 2022-09-01
Don't increase the overwrittenModifications metric during upgrade

nunnatsa:
Some of the underline custom resources are getting changed as part of the upgrade process.
This causes HCO to increase the `overwrittenModifications` metric, while this is not a user
action.

This PR changes this behaviour to not increase the metric during upgrade, and so changes made
during upgrade are not count.

Also, added some clean code in separate commits.

```release-note
None
```


tiraboschi:
/override coverage/coveralls 

PR #2089, opened 2022-09-20
Propagate TLSSecurityProfile to Kubevirt

tiraboschi:
Propagate TLSSecurityProfile to Kubevirt
translating the configuration to its internal
API.
Add unit tests.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
Propagate TLSSecurityProfile to Kubevirt
```

### DIFF
--- a/controllers/hyperconverged/hyperconverged_controller_test.go
+++ b/controllers/hyperconverged/hyperconverged_controller_test.go
@@ -1036,7 +1036,24 @@ var _ = Describe("HyperconvergedController", func() {
 				checkAvailability(foundResource, metav1.ConditionTrue)
 				Expect(foundResource.Spec.TLSSecurityProfile).To(BeNil(), "TLSSecurityProfile on HCO CR should still be nil")
 
-				// TODO: check also Kubevirt once managed
+				By("Verify that Kubevirt was properly configured with initialTLSSecurityProfile", func() {
+					kv := operands.NewKubeVirtWithNameOnly(foundResource)
+					Expect(
+						cl.Get(context.TODO(),
+							types.NamespacedName{Name: kv.Name, Namespace: kv.Namespace},
+							kv),
+					).To(Succeed())
+
+					Expect(kv.Spec.Configuration.TLSConfiguration.MinTLSVersion).To(Equal(kubevirtcorev1.VersionTLS12))
+					Expect(kv.Spec.Configuration.TLSConfiguration.Ciphers).To(Equal([]string{
+						"TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256",
+						"TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256",
+						"TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384",
+						"TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384",
+						"TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256",
+						"TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256",
+					}))
+				})
 				By("Verify that CDI was properly configured with initialTLSSecurityProfile", func() {
 					cdi := operands.NewCDIWithNameOnly(foundResource)
 					Expect(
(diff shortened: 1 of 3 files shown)


tiraboschi:
this temporanearly relies on untagged Kubevirt from main while we wait for v0.58.0,
avoid merging it.
/hold

controllers/hyperconverged/hyperconverged_controller_test.go
```diff
@@ -1036,17 +1036,23 @@ var _ = Describe("HyperconvergedController", func() {
 				checkAvailability(foundResource, metav1.ConditionTrue)
 				Expect(foundResource.Spec.TLSSecurityProfile).To(BeNil(), "TLSSecurityProfile on HCO CR should still be nil")
 
-				// TODO: check also Kubevirt once managed
 				By("Verify that CDI was properly configured with initialTLSSecurityProfile", func() {
-					cdi := operands.NewCDIWithNameOnly(foundResource)
+					kv := operands.NewKubeVirtWithNameOnly(foundResource)
```
fossedihelm, line 1040: Is it correct that CDI test does not exist anymore? Thanks

controllers/operands/kubevirt_test.go
```diff
@@ -2998,4 +3000,157 @@ Version: 1.2.3`)
...
+				oldTLSSecurityProfile,
+				kubevirtcorev1.VersionTLS10,
+				kvOldCiphers,
+			),
+			Entry("Setting Modern TLSSecurityProfile on HCO",
+				modernTLSSecurityProfile,
+				kubevirtcorev1.VersionTLS13,
+				kvModernCiphers,
+			),
+			Entry("Setting Modern TLSSecurityProfile on HCO",
```
fossedihelm, line 3099: leftover: Here you are testing `Custom` profile.

fossedihelm (commented):
Thanks. I know this is temporary, but in the meanwhile, I left two small comments. Thanks

controllers/operands/kubevirt_test.go
tiraboschi: thanks

controllers/operands/kubevirt.go
```diff
@@ -454,6 +459,47 @@ func hcLiveMigrationToKv(lm hcov1beta1.LiveMigrationConfigurations) (*kubevirtco
...
+func hcTLSSecurityProfileToKv(profile *openshiftconfigv1.TLSSecurityProfile) *kubevirtcorev1.TLSConfiguration {
+	var profileCiphers []string
+	var profileMinVersion openshiftconfigv1.TLSProtocolVersion
+
+	if profile == nil {
+		return nil
+	}
+
+	if profile.Custom != nil {
+		profileCiphers = profile.Custom.TLSProfileSpec.Ciphers
```
fossedihelm, line 490: Is there a check that prevents to specify ciphers when using 1.3 as `minTLSVersion`?
1.3 ciphers will be skipped by `crypto.OpenSSLToIANACipherSuites(profileCiphers)` and this is ok;
but what happens if a user uses a wrong configuration? i.e. if I specify 
```
minTLSVersion: VersionTLS13,
ciphers: <one compatible tls1.2 cipher>
```
this will result in a reject from kv webhook.
More in general, the question is: how hco handle wrong custom configurations?
Thanks

fossedihelm (commented):
Thanks @tiraboschi! Looks good overall. I have a doubt about how HCO handles wrong tls configurations. Thanks

controllers/operands/kubevirt.go
tiraboschi: HCO webhook is always going to try validating each change with all the components in dry-run mode, so each configuration change will be accepted only if all the components are going to accept it.
On the other side, if kv webhook will refuse the change, HCO webhook will refuse as a consequence.

fossedihelm: Thanks for the clear explanation. In the light of that, it seems to me that everything is great! How about adding a test for this? WDYT? Or is it already tested in another place? Thanks

tiraboschi: Not on this specific failure from kw webhook, but in general the mechanism is already tested in validator_test.go,
see: `should return error if dry-run update of KV CR returns error`

fossedihelm: Sounds great! No other objection. Thanks

fossedihelm:
@tiraboschi Feel free to ping me for an lgtm when the PR is ready. Thanks!

tiraboschi:
/unhold
@fossedihelm ping: we should be ready consuming Kubevirt v0.58.0-rc.0

PR #2119, opened 2022-10-18
[Clean code] Tiny JSON file reading improvement

nunnatsa:
Use the json.Decoder to read from a json file, instead of reading to a buffer and then unmarshal this buffer.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

/sig clean-code

```release-note
None
```

### DIFF
--- a/controllers/hyperconverged/upgradePatches.go
+++ b/controllers/hyperconverged/upgradePatches.go
@@ -3,7 +3,6 @@ package hyperconverged
 import (
 	"encoding/json"
 	"errors"
-	"io"
 	"os"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


orenc1:
@nunnatsa , would it be possible also to do it for yaml parsing at:
https://github.com/tiraboschi/hyperconverged-cluster-operator/blob/588650bb363995ac5a89546da56674c82525adb4/pkg/util/file.go#L12
by using:
https://pkg.go.dev/gopkg.in/yaml.v2#NewDecoder
?

nunnatsa:
> @nunnatsa , would it be possible also to do it for yaml parsing at: https://github.com/tiraboschi/hyperconverged-cluster-operator/blob/588650bb363995ac5a89546da56674c82525adb4/pkg/util/file.go#L12 by using: https://pkg.go.dev/gopkg.in/yaml.v2#NewDecoder ?

I played with it a bit. It seems like the yaml.Decoder uses the yaml tags of type's fields, and does support the json tag as the other yaml library does. This is not useful for us, because we're trying to populate external types that we can't modify to add the yaml tags.

orenc1 (approved):
/approve

PR #2151, opened 2022-11-22
Correctly set defaults for golang clients

tiraboschi:
Correctly set defaults for generated golang clients
using also '// +default' comment marker.
Please notice that +default and +kubebuilder:default
syntaxes are slightly different.
See:
https://book.kubebuilder.io/reference/markers.html#marker-syntax
and
https://github.com/kubernetes/kube-openapi/blob/master/test/integration/testdata/defaults/default.go

Fixing the usage of pointers/direct values in the golang
API remebering that:
- pointer allows distinguishing unset from the zero value for that type
- structs are not omitted from encoder output even where omitempty is specified

Exception to this general rules are required on pointers
to typedef aliases beacuse generated defaults lose type
information for aliases.
See:
https://github.com/kubernetes/gengo/issues/235

Avoid duplicating and hardcording defaults in
pkg/components/components.go since now we can properly
consume them at golang level.

Amend a few tests to properly consume the expected defaults.

This is not going to break backward compatibility at CRD level,
while it will break golang one but, on the other side,
we know that golang clients were not able to properly
consume defaults in a clean way so this is definitively
an enhancement.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
Correctly set defaults for golang clients
```


pkg/components/components.go
```diff
@@ -29,6 +29,8 @@ import (
 	hcov1beta1 "github.com/kubevirt/hyperconverged-cluster-operator/api/v1beta1"
 	"github.com/kubevirt/hyperconverged-cluster-operator/pkg/util"
 	hcoutil "github.com/kubevirt/hyperconverged-cluster-operator/pkg/util"
+
+	"k8s.io/apimachinery/pkg/runtime"
```
0xFelix, line 33: Import could be placed above?

pkg/components/components.go
```diff
@@ -664,51 +666,19 @@ func GetOperatorCRD(relPath string) *extv1.CustomResourceDefinition {
...
-			WorkloadUpdateStrategy: &hcov1beta1.HyperConvergedWorkloadUpdateStrategy{
-				WorkloadUpdateMethods: stringListToSlice("LiveMigrate"),
-				BatchEvictionSize:     &batchEvictionSize,
-				BatchEvictionInterval: &batchEvictionInterval,
-			},
-			LocalStorageClassName: "",
-		},
-	}
+		}}
+	defaultScheme.Default(defaultHco)
```
0xFelix, line 680: A really nice reduction of the code!

0xFelix (commented):
Looks good, just one little nit.

api/v1beta1/zz_generated.deepcopy.go
```diff
@@ -25,9 +25,9 @@
 package v1beta1
 
 import (
-	v1 "github.com/openshift/api/config/v1"
+	configv1 "github.com/openshift/api/config/v1"
 	corev1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
nunnatsa, line 30: Why?

Issue #2153, opened 2022-11-22
Unable to build the hyperconverged-cluster-operator


jcanocan:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line:
>

/kind bug

**What happened**:

I'm trying to build the project following [this](https://github.com/kubevirt/hyperconverged-cluster-operator#create-a-cluster--launch-the-hco). I'm able to `make cluster-up`. However, in the `make cluster-sync` I'm getting the following error:

```
$ make cluster-sync
./cluster/sync.sh
+ source ./hack/common.sh
++ set -e
++++ dirname './hack/common.sh[0]'
+++ readlink -e ./hack/../
++ PROJECT_ROOT=/home/jcano/Repos/hyperconverged-cluster-operator
++ source /home/jcano/Repos/hyperconverged-cluster-operator/hack/config
+++ KUBEVIRT_VERSION=v0.59.0-alpha.0
+++ CDI_VERSION=v1.55.0
+++ NETWORK_ADDONS_VERSION=v0.80.0
+++ SSP_VERSION=v0.16.1
+++ TTO_VERSION=v0.4.1
+++ HPPO_VERSION=v0.14.0
+++ HPP_VERSION=v0.14.0
+++ HPP_CSI_VERSION=v0.12.0
+++ NODE_DRIVER_REG_VERSION=v2.2.0
+++ LIVENESS_PROBE_VERSION=v2.3.0
+++ CSI_SNAPSHOT_VERSION=v4.2.1
+++ CSI_SIG_STORAGE_PROVISIONER_VERSION=v2.2.1
+++ KUBE_RBAC_PROXY_VERSION=4.10
+++ KUBEVIRT_CONSOLE_PLUGIN_VERSION=v4.12.0-3
+++ CONTAINER_REGISTRY=quay.io/kubevirt
+++ RELEASE_DELTA=0
+++ PACKAGE_DIR=./deploy/olm-catalog/community-kubevirt-hyperconverged
++++ sort -rV
[...]

PR #2155, opened 2022-11-23
Fix bug and change cri selection procedure of cluster-sync

jcanocan:
The command for getting the registry port gets more than one output, which breaks following commands in the deploying scripts. The output is now parsed to pickup just the first port.

Also the container runtime selection was hardcoded, now it uses the cri selection script of kubervirtci.

Further details can be found at [issue 2153](https://github.com/kubevirt/hyperconverged-cluster-operator/issues/2153)
**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
Fix make cluster-sync issue and changes the cri selection procedure
```

### DIFF
--- a/cluster/sync.sh
+++ b/cluster/sync.sh
@@ -1,15 +1,13 @@
 #!/bin/bash -ex
 
 source ./hack/common.sh
+source ./_kubevirtci/cluster-up/cluster/ephemeral-provider-common.sh
 
-container_command=podman
-
-registry_port=$(${container_command} ps | grep -Po '(?<=0.0.0.0:)\d+(?=->5000\/tcp)' || echo "")
-
+registry_port=$(${_cri_bin} ps | grep -Po '(?<=0.0.0.0:)\d+(?=->5000\/tcp)' | head -n 1)
 if [ -z "$registry_port" ]
 then
-      container_command=docker
-      registry_port=$(${container_command} ps | grep -Po '(?<=0.0.0.0:)\d+(?=->5000\/tcp)')
+      >&2 echo "unable to get the registry port"
+      exit 1
 fi
 
 registry=localhost:$registry_port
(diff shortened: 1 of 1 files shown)


cluster/sync.sh
```diff
@@ -1,15 +1,15 @@
 #!/bin/bash -ex
 
 source ./hack/common.sh
+source ${PWD}/_kubevirtci/cluster-up/cluster/ephemeral-provider-common.sh
```
0xFelix, line 4: Use same path style as above? `.` instead of `${PWD}`?

cluster/sync.sh
```diff
@@ -1,15 +1,15 @@
...
 
-container_command=podman
-
-registry_port=$(${container_command} ps | grep -Po '(?<=0.0.0.0:)\d+(?=->5000\/tcp)' || echo "")
+container_command=$_cri_bin
```
0xFelix, line 6: Why not replace `container_command` with `_cri_bin` everywhere?

cluster/sync.sh
```diff
@@ -1,15 +1,15 @@
...
 
+registry_port=$(${container_command} ps | grep -Po '(?<=0.0.0.0:)\d+(?=->5000\/tcp)' | head -n 1)
```
0xFelix, line 8: Did you check if the regex works for both podman and docker?

cluster/sync.sh
jcanocan: I agree.

cluster/sync.sh
jcanocan: Makes sense for me too.

nunnatsa (approved):

nunnatsa:
Coverage reduction seems not related to this PR, because it does not change code nor unit tests.

/override coverage/coveralls
/approve

PR #2163, opened 2022-12-02
WIP: Add KubeVirt rate limiting profiles through configmap

### LINKED ISSUE #2152: Expose new APIs to adjust KubeVirt rate limiting
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line:
>
> /kind bug

/kind enhancement

**What happened**:
User expects to run as [many](https://docs.openshift.com/container-platform/4.11/scalability_and_performance/planning-your-environment-according-to-object-maximums.html) VMs in a OpenShift cluster as pods. However, Kubernetes API clients comes a token bucket rate limiter which avoids to congest the kube-apiserver bandwidth. The token bucket rate limiter is configurable through `bust` and `Query Per Second (QPS)` parameters. These parameters [limit the total amount](https://cloud.redhat.com/blog/kubevirt-scale-test-creating-400-vmis-per-node) of VMs that may be scheduled in the cluster. Also, it is required to increase the number of `virt-api` replicas to support the extra requests.

As it is now, users need to apply the following patch to increase the limits:

`$ oc patch hco -n openshift-cnv kubevirt-hyperconverged --type=merge -p '{"metadata":{"annotations":{"kubevirt.kubevirt.io/jsonpatch":"[{\"op\":\"add\",\"path\":\"/spec/configuration\",\"value\":{\"apiConfiguration\":{\"restClient\":{\"rateLimiter\":{\"tokenBucketRateLimiter\":{\"burst\":200,\"qps\":100}}}},\"controllerConfiguration\":{\"restClient\":{\"rateLimiter\":{\"tokenBucketRateLimiter\":{\"burst\":400,\"qps\":200}}}},\"handlerConfiguration\":{\"restClient\":{\"rateLimiter\":{\"tokenBucketRateLimiter\":{\"burst\":400,\"qps\":200}}}},\"webhookConfiguration\":{\"restClient\":{\"rateLimiter\":{\"tokenBucketRateLimiter\":{\"burst\":400,\"qps\":200}}}}}}]"}}}'`

**What you expected to happen**:
I would be more clear and easy for users if the would be an API to set the values, e.g.
``` yaml
Kind: HCO
Spec:
  tuning:
    Static:  # Easy
       burst: 200
       qps: 100
```
or
``` yaml
Kind: HCO
Spec:
  tuning:
    Static:  # Easy
      apiConfiguration:
         burst: 200
         qps: 100
       controllerConfiguration:
         burst: 400
         qps: 200
       handlerConfiguration:
         burst: 400
         qps: 200
       webhookConfiguration:
         burst: 400
         qps: 200
```

I'm not sure which one would be more appropriated. Also, regarding the replica set increasing I'm not sure if it would be good to include it in the same configuration file or just leave it as a separate patch. Another possibility is that this is no longer required because of [this](https://github.com/kubevirt/kubevirt/pull/7401).

Issue tracked in: [CNV-21596](https://issues.redhat.com/browse/CNV-21596)
**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- HCO version (use `oc get csv -n kubevirt-hyperconverged`): 4.11.0
- Kubernetes version (use `kubectl version`): 1.24.0
- Cloud provider or hardware configuration: Any
- Install tools:
- Others:

jcanocan:
The following approach has a downside: when the configmap it's updated, you need to restart/reload/update manually the hco to be aware of the configmap changes. Is this acceptable?

I would also appreciate some hits about mocking the cluster in the testing if we agree to go forward with this approach.

Implements: [Issue 2152](https://github.com/kubevirt/hyperconverged-cluster-operator/issues/2152)
Thanks in advance.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note

```
Fixes #2152


jcanocan:
@nunnatsa  Can you please take a look at this? Thanks a lot.

tiraboschi:
> The following approach has a downside: when the configmap it's updated, you need to restart/reload/update manually the hco to be aware of the configmap changes. Is this acceptable?

HCO is already watching also configmaps, see:
https://github.com/kubevirt/hyperconverged-cluster-operator/blob/main/controllers/hyperconverged/hyperconverged_controller.go#L180

but its cache's ListWatch is already configured to filter only ConfigMaps with specific labels to control the memory consumption, see:
https://github.com/kubevirt/hyperconverged-cluster-operator/blob/main/cmd/hyperconverged-cluster-operator/main.go#L199-L201

We can probably tolerate to watch all the configmaps in openshift-cnv namespaces regardless of any specific label (tune the cache from `Label: labelSelector` to `Field: namespaceSelector`).

controllers/operands/kubevirt.go
```diff
@@ -294,7 +295,71 @@ func hcWorkloadUpdateStrategyToKv(hcObject *hcov1beta1.HyperConvergedWorkloadUpd
...
+		},
+	}, nil
+
+}
+
+func hcoTuning2Kv(hcTuning *hcov1beta1.Tuning, hcNamespace string) (*kubevirtcorev1.ReloadableComponentConfiguration, error) {
+	var kvReloadable *kubevirtcorev1.ReloadableComponentConfiguration
+	if hcTuning != nil {
+		if hcTuning.Policy != nil {
+			if hcTuning.Policy.Name != "" {
```
tiraboschi, line 340: if hcTuning.Policy.Name == "configmap" (please use a const in api/v1beta1/hyperconverged_types.go)

api/v1beta1/hyperconverged_types.go
```diff
@@ -25,6 +25,18 @@ const (
 	HyperConvergedUninstallStrategyBlockUninstallIfWorkloadsExist HyperConvergedUninstallStrategy = "BlockUninstallIfWorkloadsExist"
 )
```
tiraboschi, line 26: Do we need only the name of the policy or also the capability to fine tune predefined policies?
If just the name (which in the beginning could only be "ConfigMap" pointing to a configmap with a fixed name in the operator namespace), why not:

```
type HyperConvergedTuningPolicy string
...
HyperConvergedTuningPolicyConfigMap                HyperConvergedTuningPolicy = "ConfigMap"
...
// +kubebuilder:validation:Enum=ConfigMap
// +optional
TuningPolicy HyperConvergedTuningPolicy `json:"tuningPolicy,omitempty"`
```

api/v1beta1/hyperconverged_types.go
```diff
@@ -25,6 +25,18 @@ const (
...
 
+// TODO: Not sure if it needs the openapi-gen this should be a corev1.ConfigMap
+
+type Policy struct {
+	Name string `json:"name,omitempty"`
```
nunnatsa, line 31: What do you think about using a pointer instead of a string?
```suggestion
	Name *string `json:"name,omitempty"`
```

controllers/operands/kubevirt.go
```diff
@@ -1,8 +1,10 @@
 package operands
 
 import (
+	"context"
 	"errors"
 	"fmt"
+	"k8s.io/client-go/kubernetes"
```
nunnatsa, line 7: move this line down

controllers/operands/kubevirt.go
```diff
@@ -294,7 +295,71 @@ func hcWorkloadUpdateStrategyToKv(hcObject *hcov1beta1.HyperConvergedWorkloadUpd
 	return kvObject
 }
 
-func getKVConfig(hc *hcov1beta1.HyperConverged) (*kubevirtcorev1.KubeVirtConfiguration, error) {
+func getRatesFromConfigMap(configMap *corev1.ConfigMap) (*kubevirtcorev1.ReloadableComponentConfiguration, error) {
+
+	qps, ok := configMap.Data["qps"]
```
nunnatsa, line 300: What's `qps`? if we're using it for a user facing API, maybe we want a human readable name?

controllers/operands/kubevirt.go
```diff
@@ -294,7 +295,71 @@ func hcWorkloadUpdateStrategyToKv(hcObject *hcov1beta1.HyperConvergedWorkloadUpd
...
+
+	if !ok {
+		return nil, errors.New("qps parameter not found in configmap " + configMap.Name)
```
nunnatsa, line 303: Is it an error case?

controllers/operands/kubevirt.go
```diff
@@ -294,7 +295,71 @@ func hcWorkloadUpdateStrategyToKv(hcObject *hcov1beta1.HyperConvergedWorkloadUpd
...
+	}
+
+	qpsConverted, err := strconv.ParseFloat(qps, 32)
+
+	if err != nil {
+		return nil, err
```
nunnatsa, line 309: Please use verbose error message. Maybe wrap the error?
```suggestion
		return nil, fmt.Errorf("more verbose error message; %w", err)
```

controllers/operands/kubevirt.go
```diff
@@ -294,7 +295,71 @@ func hcWorkloadUpdateStrategyToKv(hcObject *hcov1beta1.HyperConvergedWorkloadUpd
...
+	}
+
+	burst, ok := configMap.Data["burst"]
+	if !ok {
+		return nil, errors.New("burst parameter not found in configmap" + configMap.Name)
```
nunnatsa, line 314: missing space
```suggestion
		return nil, errors.New("burst parameter not found in configmap " + configMap.Name)
```

controllers/operands/kubevirt.go
```diff
@@ -294,7 +295,71 @@ func hcWorkloadUpdateStrategyToKv(hcObject *hcov1beta1.HyperConvergedWorkloadUpd
...
+	}
+
+	burstConverted, err := strconv.Atoi(burst)
+
+	if err != nil {
+		return nil, err
```
nunnatsa, line 320: please add more verbose error message.

controllers/operands/kubevirt_test.go
```diff
@@ -3,7 +3,10 @@ package operands
 import (
 	"context"
 	"fmt"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
```
nunnatsa, line 7: move down please

controllers/operands/kubevirt_test.go
```diff
@@ -2524,6 +2527,90 @@ Version: 1.2.3`)
...
+					},
+					Data: configMapData,
+				}
+
+				//WIP: not able to create a mock cluster for testing
+				resources := []runtime.Object{}
+				cl := commonTestUtils.InitClient(resources)
+
+				ci := commonTestUtils.ClusterInfoMock{}
+				Expect(ci.IsOpenshift()).To(BeTrue())
```
nunnatsa, line 2587: Is this check needed here?

controllers/operands/kubevirt_test.go
```diff
@@ -2524,6 +2527,90 @@ Version: 1.2.3`)
...
+				mockmgr, _ := mgr.(*commonTestUtils.ManagerMock)
+
+				//	config, err := cfg.GetConfig()
+				//	Expect(err).ToNot(HaveOccurred())
+				clientSet, err := kubernetes.NewForConfig(mockmgr.GetConfig())
+				//	Expect(err).ToNot(HaveOccurred())
+
+				mockmgr.GetClient()
+				_, err = clientSet.CoreV1().ConfigMaps(hco.Namespace).Create(context.TODO(), &configMap, metav1.CreateOptions{})
+				Expect(err).ToNot(HaveOccurred())
```
nunnatsa, line 2600: The easiest way to create the config map is to create a CM object and add it to the resources; e.g.
```suggestion
				//WIP: not able to create a mock cluster for testing
				resources := []runtime.Object{configMap}
				cl := commonTestUtils.InitClient(resources)
```

nunnatsa (commented):
Thanks @jcanocan for the PR. Added a few comments.

api/v1beta1/hyperconverged_types.go
jcanocan: I agree, I have used this approach. I think it's much more cleaner now.

controllers/operands/kubevirt.go
jcanocan: Done.

api/v1beta1/hyperconverged_types.go
jcanocan: I have changed a bit the approach on this.

controllers/operands/kubevirt.go
jcanocan: I agree. I have also used a const for referring these parameters.

controllers/operands/kubevirt.go
jcanocan: Maybe an error is not the best approach. The idea here is if one of these parameters is not present in the configmap, we do not set the rate limits. If you have a suggestion about this, it would be much appreciated. Thanks!

controllers/operands/kubevirt.go
jcanocan: Done. I have also changed `errors.New` by `fmt.Errorf` for the sake of consistency. Please, let me know if this is appropriated.

controllers/operands/kubevirt.go
jcanocan: Done.

controllers/operands/kubevirt_test.go
jcanocan: Done.

jcanocan:
I have noticed that changing the WatchList selector in https://github.com/kubevirt/hyperconverged-cluster-operator/blob/main/cmd/hyperconverged-cluster-operator/main.go#L199-L201 to tolerate all configmaps in the namespace, it causes a reconciliation hot loop. I was able to find the root of this: Some operators use a leader election mechanism throw configmaps. Basically, the elected leader pod of each operator updates the a configmap with a timestamp periodically after a given amount of time to inform the rest that he still owns the leadership. The downside is that because the fact that I have configured the WatchList to trigger the reconcile loop when any configmap in the namespace is modified, the reconcile loop is spammed constantly. 
We have a couple of options to address this: 

1) To leave it was and only trigger the reconcile loop within configmaps with the label `app: kubevirt-hyperconverged`.  The downside is if users forgot to add the label they will never notice. Maybe with a proper documentation this can be addressed.
2) To filter somehow the leader election configmaps. The downside is we will need to maintain this filter list/mechanism. Take in account that the configmap does not have to have any specific label, name or structure. In sort, I do not see a reliable way to distinguish them from the rest.
3) To use a `Lease` object in the affected operators instead a configmap. This will force us to update a couple of operators and to be careful that new operators do not use the configmaps for the leader election.
4) Use a different namespace for the rate liming policy or operators deployment.

In my humble opinion, option 1) seems to be the best one. However, I would like to know your opinion on this. Thanks in advance!

controllers/hyperconverged/hyperconverged_controller.go
```diff
@@ -262,6 +262,27 @@ type ReconcileHyperConverged struct {
...
+
+		if err != nil {
+			return err
+		}
+		// If the configMap has changed kubevirt object have to be updated, otherwise the cache object is used
+		r.operandHandler.Reset()
+	}
+	return nil
+}
+
```
nunnatsa, line 285: please consider moving this logic to the `operands` package. maybe in the kubevirt.go file

api/v1beta1/hyperconverged_types.go
```diff
@@ -19,10 +19,12 @@ import (
 const HyperConvergedName = "kubevirt-hyperconverged"
 
 type HyperConvergedUninstallStrategy string
+type HyperConvergedTuningPolicy string
 
 const (
 	HyperConvergedUninstallStrategyRemoveWorkloads                HyperConvergedUninstallStrategy = "RemoveWorkloads"
 	HyperConvergedUninstallStrategyBlockUninstallIfWorkloadsExist HyperConvergedUninstallStrategy = "BlockUninstallIfWorkloadsExist"
+	HyperConvergedTuningPolicyConfigMap                           HyperConvergedTuningPolicy      = "static-policy"
```
nunnatsa, line 27: please split the `HyperConvergedUninstallStrategy` consts and the `HyperConvergedTuningPolicy` const. Please move the `HyperConvergedTuningPolicy` type definition to be after the const, and add the new const below it.

controllers/hyperconverged/hyperconverged_controller.go
```diff
@@ -80,13 +80,24 @@ const (
...
+// TriggeredBy informs which resource has triggered the Reconcile loop
+type TriggeredBy int8
+
+const (
+	Hco TriggeredBy = iota
+	Api
+	Cm
+	Other
+)
+
```
nunnatsa, line 100: please use more declarative names, and please don't expose these const (they are not needed outside of this package):

For example:
```suggestion
const (
	triggeredByHCO TriggeredBy = iota
	triggeredByAPI
	triggeredByCM
	triggeredByOther
)
```

controllers/hyperconverged/hyperconverged_controller.go
```diff
@@ -80,13 +80,24 @@ const (
...
+	configMapCRPrefix = "cm-policy-cr-"
 
 	// These group are no longer supported. Use these constants to remove unused resources
 	v2vGroup = "v2v.kubevirt.io"
 
 	requestedStatusKey = "requested status"
 )
 
+// TriggeredBy informs which resource has triggered the Reconcile loop
+type TriggeredBy int8
```
nunnatsa, line 92: We don't need to expose this type. It's only used within the package.

controllers/hyperconverged/hyperconverged_controller.go
```diff
@@ -334,55 +374,84 @@ func (r *ReconcileHyperConverged) Reconcile(ctx context.Context, request reconci
...
 	}
 	if apiServerCRTriggered {
 		logger.Info("Triggered by ApiServer CR, refreshing it")
 		err = hcoutil.GetClusterInfo().RefreshAPIServerCR(ctx, r.client)
 		if err != nil {
-			return reconcile.Request{}, hcoTriggered, err
+			if hcoTriggered {
+				return reconcile.Request{}, Hco, err
+			}
+			return reconcile.Request{}, Other, err
```
nunnatsa, line 407: This case is not reachable, because we already check for it above, and exciting. Please drop this `if` sentence.

controllers/hyperconverged/hyperconverged_controller.go
```diff
@@ -1433,6 +1502,20 @@ func getApiServerCRPlaceholder() (types.NamespacedName, error) {
...
+	namespace, err := hcoutil.GetOperatorNamespaceFromEnv()
+
+	if err != nil {
+		return fakeHco, err
+	}
+	fakeHco.Namespace = namespace
+
+	return fakeHco, nil
+}
+
```
nunnatsa, line 1518: I think we should have a special handling for the rate limit CM, What do you think about returning the new namespaced value only for this CM, and for the rest of the CMs, return the one that used for "other"? you can pass the already computed value as a parameter to this function.

controllers/operands/kubevirt.go
```diff
@@ -232,8 +240,8 @@ func (*kubevirtHooks) updateCr(req *common.HcoRequest, Client client.Client, exi
 
 func (*kubevirtHooks) justBeforeComplete(_ *common.HcoRequest) { /* no implementation */ }
 
-func NewKubeVirt(hc *hcov1beta1.HyperConverged, opts ...string) (*kubevirtcorev1.KubeVirt, error) {
-	config, err := getKVConfig(hc)
+func NewKubeVirt(ctx context.Context, client client.Client, hc *hcov1beta1.HyperConverged, opts ...string) (*kubevirtcorev1.KubeVirt, error) {
+	config, err := getKVConfig(ctx, client, hc, getNamespace(hc.Namespace, opts))
```
nunnatsa, line 244: We don't need the `getNamespace` here. Please drop.

controllers/operands/kubevirt.go
```diff
@@ -294,7 +301,73 @@ func hcWorkloadUpdateStrategyToKv(hcObject *hcov1beta1.HyperConvergedWorkloadUpd
 	return kvObject
 }
 
-func getKVConfig(hc *hcov1beta1.HyperConverged) (*kubevirtcorev1.KubeVirtConfiguration, error) {
+func getRatesFromConfigMap(configMap *corev1.ConfigMap) (*kubevirtcorev1.ReloadableComponentConfiguration, error) {
+
+	qps, ok := configMap.Data[qps]
```
nunnatsa, line 306: `qps` is already used in this file as a constant. please change the name of one of them (e.g. make the constant more readable).

controllers/operands/kubevirt.go
```diff
@@ -294,7 +301,73 @@ func hcWorkloadUpdateStrategyToKv(hcObject *hcov1beta1.HyperConvergedWorkloadUpd
...
+		return nil, fmt.Errorf("queryPerSeconds parameter not found in configmap " + configMap.Name)
+	}
+
+	qpsConverted, err := strconv.ParseFloat(qps, 32)
+
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert queryPerSeconds parameter to float; %w", err)
+	}
+
+	burst, ok := configMap.Data[burst]
```
nunnatsa, line 318: `brust` is already defined in this package as a constant. Please rename.

controllers/operands/kubevirt_test.go
```diff
@@ -2524,6 +2523,104 @@ Version: 1.2.3`)
...
+
+			})
+
+			It("Should create the fields and populate them when requested", func() {
+				hco.Spec.TuningPolicy = hcov1beta1.HyperConvergedTuningPolicyConfigMap
+				//burst is missing
+				configMapData := make(map[string]string, 0)
+				configMapData[qps] = "100"
+				configMapData[burst] = "200"
+
```
nunnatsa, line 2591: maybe this would be more clear?

```suggestion
				configMapData := map[string]string {
				    qps: "100",
				    burst: "200",
				}
```

controllers/operands/kubevirt.go
```diff
@@ -307,6 +380,13 @@ func getKVConfig(hc *hcov1beta1.HyperConverged) (*kubevirtcorev1.KubeVirtConfigu
 
 	obsoleteCPUs, minCPUModel := getObsoleteCPUConfig(hc.Spec.ObsoleteCPUs)
 
+	if client != nil {
+		rateLimiter, err = hcoTuning2Kv(ctx, client, &hc.Spec.TuningPolicy, hcNamespace)
```
nunnatsa, line 384: Not sure about returning error here. Unlike other cases here, this error is not coming from a wrong configuration of the  HyperConverged CR, but from the CM. I don't want to kill the reconcile if the CM is not there (yet). @tiraboschi, what do you think?

nunnatsa (commented):
Thanks, great work. I added some more comments. 

About returning an error if the new CM is missing or wrong, I'm not sure it's the best approach. I'm not sure I want to link HCO with the CM, in a way that will actually make HCO stop working.

Also, as result of this error, setting of the new `TuningPolicy` field in HCO will fail if the CM is not exist or with a wrong data, because the call to the `NewKubevirt` function in the webhook will fail. We can maybe say that this is the desired behaviour, and that it even solve the issue above (returning error), but I think it's a risk; e.g when removing or editing the existing CM.

Anyway, if we're going there, it must be very well documented, both in the cluster_config doc AND in the field documentation in the CRD. @tiraboschi - WDYT?

nunnatsa:
Also, please run `make goimport`

api/v1beta1/hyperconverged_types.go
jcanocan: I think I changed it as you requested. Please, let me know.

controllers/hyperconverged/hyperconverged_controller.go
jcanocan: Done.

controllers/hyperconverged/hyperconverged_controller.go
jcanocan: I have addressed this using a different approach. I didn't find convenient modifying the function since `isTriggerdByConfigMap()` and `isTriggeredByHyperconverged` are using this function too. Please, let me know if the approach used is acceptable and clean. Thanks.

controllers/operands/kubevirt.go
jcanocan: Done.

controllers/operands/kubevirt.go
jcanocan: Done.

controllers/operands/kubevirt.go
jcanocan: Done.

controllers/operands/kubevirt_test.go
jcanocan: I agree. Thanks for the suggestion.

controllers/operands/kubevirt.go
jcanocan: I think you are right and we can just throw a warning to inform the user that something happened and remove the `tuningPolicy` from the HCO spec. Just to not confuse users if they check the hco definition and they see that the `tuningPolicy` is present but the rates are not being configured (even if we notify them that the CM is missing or missconfigured and therefore this is not going to take effect).

jcanocan:
> Also, please run `make goimport`

I ran it, but nothing has changed :/

controllers/hyperconverged/hyperconverged_controller.go
```diff
@@ -80,13 +80,24 @@ const (
...
 	requestedStatusKey = "requested status"
 )
 
+// triggeredBy informs which resource has triggered the Reconcile loop
+type triggeredBy int8
+
+const (
+	triggeredByHco triggeredBy = iota
+	triggeredByApi
+	triggeredByCm
```
nunnatsa, line 97: Nit:

HCO, API and CM are acronyms. Please rename:
 ```suggestion
	triggeredByHCO triggeredBy = iota
	triggeredByAPI
	triggeredByCM
```

File: controllers/hyperconverged/hyperconverged_controller.go (hunk 18 above)
Comment (nunnatsa) [reply controllers/hyperconverged/hyperconverged_controller.go:1522]: You actually did the right logic in the `add` function :+1: 

I think that what confused me was the name of this function, and the related triggerdBy constant. Just "configMap" or "CM" is maybe too general. Could you rename them?

e.g. `getRateLimitConfigMapPlaceholder` or something like that. 
and also `triggeredByRateLimitCM`, for example. I'll add a comment there too.

File: controllers/hyperconverged/hyperconverged_controller.go (hunk 24 above)
Comment (nunnatsa) [inline controllers/hyperconverged/hyperconverged_controller.go:97]: After reading the code again, this constant is only used for the rate limit CM. I think this name is confusing. Could you please rename? e.g. `triggeredByRateLimitCM` or something like this?

File: controllers/operands/kubevirt.go (hunk 23 above)
Comment (nunnatsa) [reply controllers/operands/kubevirt.go:384]: I don't want to edit the HyperConverged CR, if possible.

Maybe we need to catch it in the webhook. But this will only solve the creating or editing the HC CR use cases. But it won't help for the case of deleting or wrongly editing the CM, so I'm not sure.

File: controllers/hyperconverged/hyperconverged_controller.go (hunk 18 above)
Comment (jcanocan) [reply controllers/hyperconverged/hyperconverged_controller.go:1522]: What do you think about `getStaticRateLimitConfigMapPlaceholder()` and `triggeredByStaticRateLimitCM`? Naming is sometimes hard :smiling_face_with_tear:

**Comment by nunnatsa** [comment]
> I ran it, but nothing has changed :/

Right. There was a problem in one of the file earlier, but I guess you fixed it. Still, the sanity test is failing. You'll need to run `make generate-doc` as well, and commit the docs/api.md file.

File: controllers/operands/kubevirt.go (hunk 23 above)
Comment (nunnatsa) [reply controllers/operands/kubevirt.go:384]: OK, let's do the following:
If the policy is "configmap" and the CM is missing or if its content is wrong, we'll write a warning to the log + trigger an event to notify that. @tiraboschi - we can also add a new metric (more general one: wrong config or something) + an alert.

In this case, HCO will ignore the policy field. This should be well documented.

Comment (jcanocan) [reply controllers/operands/kubevirt.go]: Addressed.

File: controllers/hyperconverged/hyperconverged_controller.go (hunk 24 above)
Comment (jcanocan) [reply controllers/hyperconverged/hyperconverged_controller.go]: Addressed.

Comment (jcanocan) [reply controllers/hyperconverged/hyperconverged_controller.go]: I have used `triggeredByStaticRateLimitCM`. I find it more self-explanatory.

**Comment by jcanocan** [comment]
> > I ran it, but nothing has changed :/
> 
> Right. There was a problem in one of the file earlier, but I guess you fixed it. Still, the sanity test is failing. You'll need to run `make generate-doc` as well, and commit the docs/api.md file.

Done. I have also extended a bit the description of the new parameter.

File: pkg/webhooks/validator/validator.go [hunk 25]
```diff
@@ -286,8 +292,11 @@ func (wh WebhookHandler) updateOperatorCr(ctx context.Context, hc *v1beta1.Hyper
 
 	switch existing := exists.(type) {
 	case *kubevirtcorev1.KubeVirt:
-		required, err := operands.NewKubeVirt(hc)
-		if err != nil {
+		required, err := operands.NewKubeVirt(ctx, wh.cli, hc)
+
+		if errT, ok := err.(*operands.RateTuningError); ok {
+			wh.logger.Error(errT, "failed to set the rate limit")
```
nunnatsa, line 298: Nit: can drop `errT`. logger.Error will work with the existing `err` variable.

pkg/webhooks/validator/validator.go
```diff
@@ -168,13 +172,15 @@ func (wh *WebhookHandler) ValidateCreate(ctx context.Context, dryrun bool, hc *v
...
+func (wh *WebhookHandler) getOperands(ctx context.Context, requested *v1beta1.HyperConverged) (*kubevirtcorev1.KubeVirt, *cdiv1beta1.CDI, *networkaddonsv1.NetworkAddonsConfig, error) {
 	if err := wh.validateCertConfig(requested); err != nil {
 		return nil, nil, nil, err
 	}
 
-	kv, err := operands.NewKubeVirt(requested)
-	if err != nil {
+	kv, err := operands.NewKubeVirt(ctx, wh.cli, requested)
+	if errT, ok := err.(*operands.RateTuningError); ok {
+		wh.logger.Error(errT, "failed to set the rate limit")
```
nunnatsa, line 182: Nit: can drop `errT` var. logger.Error will work with the existing `err` variable.

pkg/webhooks/validator/validator.go
```diff
@@ -145,8 +145,12 @@ func (wh *WebhookHandler) ValidateCreate(ctx context.Context, dryrun bool, hc *v
 		return err
 	}
 
-	if _, err := operands.NewKubeVirt(hc); err != nil {
-		return err
+	if _, err := operands.NewKubeVirt(ctx, wh.cli, hc); err != nil {
+		if errT, ok := err.(*operands.RateTuningError); ok {
+			wh.logger.Error(errT, "failed to set the rate limit")
```
nunnatsa, line 150: here too

controllers/operands/kubevirt_test.go
```diff
@@ -2524,6 +2524,121 @@ Version: 1.2.3`)
...
+				hco.Spec.TuningPolicy = hcov1beta1.HyperConvergedTuningPolicyConfigMap
+				cl := commonTestUtils.InitClient([]runtime.Object{})
+
+				kv, err := NewKubeVirt(context.TODO(), cl, hco)
+
+				errT, ok := err.(*RateTuningError)
+
+				Expect(err).To(HaveOccurred())
+				Expect(ok).To(BeTrue())
+				Expect(errT).To(HaveOccurred())
```
nunnatsa, line 2556: Not sure we need to check that if `ok` is `true`. If you want, you can check that `errT` is of the expected type. See here: https://onsi.github.io/gomega/#matcherrorexpected-interface

controllers/operands/kubevirt_test.go
```diff
@@ -2524,6 +2524,121 @@ Version: 1.2.3`)
...
+				resources := []runtime.Object{&configMap}
+
+				cl := commonTestUtils.InitClient(resources)
+
+				kv, err := NewKubeVirt(context.TODO(), cl, hco)
+
+				errT, ok := err.(*RateTuningError)
+				Expect(err).To(HaveOccurred())
+				Expect(ok).To(BeTrue())
+				Expect(errT).To(HaveOccurred())
```
nunnatsa, line 2594: same

controllers/operands/kubevirt.go
```diff
@@ -17,16 +18,27 @@ import (
...
-
 	hcov1beta1 "github.com/kubevirt/hyperconverged-cluster-operator/api/v1beta1"
 	"github.com/kubevirt/hyperconverged-cluster-operator/controllers/common"
 	hcoutil "github.com/kubevirt/hyperconverged-cluster-operator/pkg/util"
+	kubevirtcorev1 "kubevirt.io/api/core/v1"
 )
 
+type RateTuningError struct {
+	msg string
+}
```
nunnatsa, line 33: I think the struct here is overkill. You can use directly string:
```suggestion
type RateTuningError string
```

controllers/operands/kubevirt.go
```diff
@@ -294,7 +324,75 @@ func hcWorkloadUpdateStrategyToKv(hcObject *hcov1beta1.HyperConvergedWorkloadUpd
...
+
+	if !ok {
+
+		return nil, NewTuningPolicyError("queryPerSeconds parameter not found in configmap " + configMap.Name)
+	}
+
+	qpsConverted, err := strconv.ParseFloat(qpsRead, 32)
+
+	if err != nil {
+		return nil, NewTuningPolicyError(fmt.Sprintf("failed to convert queryPerSeconds parameter to float; " + err.Error()))
```
nunnatsa, line 339: Not sure there is reason to use `fmt.Sprintf` for a single string w/o parameters. Please fix to one of:
```suggestion
		return nil, NewTuningPolicyError("failed to convert queryPerSeconds parameter to float; " + err.Error())
```

Or (preferred):
```suggestion
		return nil, NewTuningPolicyError(fmt.Sprintf("failed to convert queryPerSeconds parameter to float; %s", err.Error()))
```

controllers/operands/kubevirt.go
```diff
@@ -294,7 +324,75 @@ func hcWorkloadUpdateStrategyToKv(hcObject *hcov1beta1.HyperConvergedWorkloadUpd
...
+	}
+
+	burstRead, ok := configMap.Data[burst]
+	if !ok {
+
+		return nil, NewTuningPolicyError(fmt.Sprintf("burstRead parameter not found in configmap " + configMap.Name))
```
nunnatsa, line 345: same

controllers/operands/kubevirt.go
```diff
@@ -294,7 +324,75 @@ func hcWorkloadUpdateStrategyToKv(hcObject *hcov1beta1.HyperConvergedWorkloadUpd
...
+	}
+
+	burstConverted, err := strconv.Atoi(burstRead)
+
+	if err != nil {
+		return nil, NewTuningPolicyError(fmt.Sprintf("failed to convert burstRead parameter to integer; " + err.Error()))
```
nunnatsa, line 351: same

pkg/webhooks/validator/validator.go
jcanocan: Done.

pkg/webhooks/validator/validator.go
jcanocan: Done.

pkg/webhooks/validator/validator.go
jcanocan: Done.

controllers/operands/kubevirt_test.go
jcanocan: Please, check if this is addressed properly.

controllers/operands/kubevirt.go
jcanocan: I agree. Changed.

controllers/operands/kubevirt.go
jcanocan: I have used the second option. Check if it is fine now.

nunnatsa:
Thanks @jcanocan - great work - it wasn't easy.

Please don't forget to update the [docs/cluster-configuration.md](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/main/docs/cluster-configuration.md) file with the user guide for the new field, and the new configMap.

Also, please consider to add a functional test to make sure the whole feature is working. Either similar to [this bash test](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/main/hack/check_defaults.sh), or in the golang functional tests.

jcanocan:
> Thanks @jcanocan - great work - it wasn't easy.
> 
> Please don't forget to update the [docs/cluster-configuration.md](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/main/docs/cluster-configuration.md) file with the user guide for the new field, and the new configMap.
> 
> Also, please consider to add a functional test to make sure the whole feature is working. Either similar to [this bash test](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/main/hack/check_defaults.sh), or in the golang functional tests.

Many thanks for your kind words and you support through the process. Much appreciated. Sure, I will try to have both things ASAP.

jcanocan:
> > Thanks @jcanocan - great work - it wasn't easy.
> > Please don't forget to update the [docs/cluster-configuration.md](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/main/docs/cluster-configuration.md) file with the user guide for the new field, and the new configMap.
> > Also, please consider to add a functional test to make sure the whole feature is working. Either similar to [this bash test](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/main/hack/check_defaults.sh), or in the golang functional tests.
> 
> Many thanks for your kind words and you support through the process. Much appreciated. Sure, I will try to have both things ASAP.

Hi @nunnatsa! I have published the documentation and the functional test. Please, check if they are fine. Many thanks!

PR #2182, opened 2022-12-13
[release-1.7] Avoid omitempty on WorkloadUpdateMethods (#2180)

tiraboschi:
In our API for WorkloadUpdateMethods we read:
...
An empty list defaults to no automated workload updating.

and a default of: ["LiveMigrate"]

but in the past the json representation of
WorkloadUpdateMethods was defined with omitempty
but, as for:
https://go.dev/src/encoding/json/encode.go
// The "omitempty" option specifies that the field should be omitted // from the encoding if the field has an empty value, defined as // false, 0, a nil pointer, a nil interface value, and any empty array, // slice, map, or string.

the golang json encoder is omitting also empty array and slice and so an empty list on WorkloadUpdateMethods gets omitted at json level and so the APIServer will then replace it with the default value (["LiveMigrate"]).

Avoid declaring omitempty on WorkloadUpdateMethods to make an empty list an acceptable value.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2151693

This is a manual cherry-pick of
https://github.com/kubevirt/hyperconverged-cluster-operator/pull/2180

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
Avoid omitempty on WorkloadUpdateMethods
```

### DIFF
--- a/api/v1beta1/hyperconverged_types.go
+++ b/api/v1beta1/hyperconverged_types.go
@@ -395,8 +395,7 @@ type HyperConvergedWorkloadUpdateStrategy struct {
 	//
 	// +listType=atomic
 	// +kubebuilder:default={"LiveMigrate"}
-	// +optional
-	WorkloadUpdateMethods []string `json:"workloadUpdateMethods,omitempty"`
+	WorkloadUpdateMethods []string `json:"workloadUpdateMethods"`
 
 	// BatchEvictionSize Represents the number of VMIs that can be forced updated per
 	// the BatchShutdownInterval interval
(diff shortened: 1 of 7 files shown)


nunnatsa (approved):
/approve

PR #2186, opened 2022-12-15
Support KubeSecondaryDNS

oshoval:
Undeploy will be supported with CNAO v0.83.1

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:
```release-note
Support deployKSD
```


oshoval:
/hold

until https://github.com/kubevirt/hyperconverged-cluster-operator/pull/2184 is merged

oshoval:
/hold cancel

Beside undeploy it works (bug is on CNAO, and there is a new release soon for that)

orenc1:
@oshoval , this PR introduces a vast code duplication - there are several code blocks that are pretty much the same, one for `deployOVS` and another for `deployKSD`.
This is DRY anti-pattern. I would prefer to generalize the code to suite both annotations as parameters.

oshoval:
> @oshoval , this PR introduces a vast code duplication - there are several code blocks that are pretty much the same, one for `deployOVS` and another for `deployKSD`. This is DRY anti-pattern. I would prefer to generalize the code to suit both annotations as parameters.

Done

oshoval:
Maybe this is related to the fact CNAO has a bug which doesn't remove KSD ?
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/kubevirt_hyperconverged-cluster-operator/2186/pull-ci-kubevirt-hyperconverged-cluster-operator-main-hco-e2e-image-index-sno-aws/1605490603888480256

We fixed it, but need a new CNAO release and to bump it on HCO

controllers/operands/networkAddons.go
```diff
@@ -113,28 +119,40 @@ func (*cnaHooks) updateSpec(req *common.HcoRequest, found *networkaddonsv1.Netwo
...
+func checkComponent(spec networkaddonsshared.NetworkAddonsConfigSpec, annotation string) bool {
+	if annotation == deployOVS {
+		return spec.Ovs != nil
+	}
+	if annotation == deployKSD {
+		return spec.KubeSecondaryDNS != nil
+	}
+
+	logger.Error(errors.New("checkComponent error"), "unknown annotation")
+	return false
```
oshoval, line 153: maybe panic here instead?

controllers/operands/networkAddons.go
```diff
@@ -113,28 +119,40 @@ func (*cnaHooks) updateSpec(req *common.HcoRequest, found *networkaddonsv1.Netwo
 	return false
 }
 
-// If deployOVS annotation doesn't exists prior the upgrade - set this annotation to true;
+// If annotation doesn't exists prior the upgrade - set this annotation to true;
 // Otherwise - remain the value as it is.
-func setDeployOvsAnnotation(req *common.HcoRequest, found *networkaddonsv1.NetworkAddonsConfig) {
+func setAnnotation(req *common.HcoRequest, found *networkaddonsv1.NetworkAddonsConfig, annotation string) {
```
nunnatsa, line 124: This name is too generic. Please consider renaming it.

oshoval: `setCnaoAnnotation` ?

oshoval: changed to `setCnaoAnnotation`

controllers/operands/networkAddons.go
```diff
@@ -113,28 +119,40 @@ func (*cnaHooks) updateSpec(req *common.HcoRequest, found *networkaddonsv1.Netwo
...
+				req.Instance.Annotations[annotation] = "false"
+				req.Logger.Info(annotation + " annotation is set to false.")
 			}
 
 			req.Dirty = true
 		}
 	}
 }
 
+func checkComponent(spec networkaddonsshared.NetworkAddonsConfigSpec, annotation string) bool {
```
nunnatsa, line 144: I think that this name  is too generic as well. @oshoval, could you rename this one too?

controllers/operands/networkAddons.go
nunnatsa: I think that panic from a server should be done for a very good reason. I'm not sure this specific case. Anyway, if you decide to change it, please add a detailed comment.

controllers/operands/networkAddons.go
```diff
@@ -113,28 +119,40 @@ func (*cnaHooks) updateSpec(req *common.HcoRequest, found *networkaddonsv1.Netwo
...
+	if annotation == deployOVS {
+		return spec.Ovs != nil
+	}
+	if annotation == deployKSD {
+		return spec.KubeSecondaryDNS != nil
+	}
+
+	logger.Error(errors.New("checkComponent error"), "unknown annotation")
+	return false
```
nunnatsa, line 153: Maybe switch-case-default will be more readable? WDYT?

controllers/operands/networkAddons.go
oshoval: done

controllers/operands/networkAddons.go
oshoval: possible, but on the other hand i didnt prefer it because it looks like default is allowed
but default actually isn't allowed unless we also add panic there (but you didn't like it according the previous comment)

nunnatsa: I think it's very common patter to return error (OK, or panic) from default case.

oshoval: done

controllers/operands/networkAddons_test.go
```diff
@@ -507,6 +507,58 @@ var _ = Describe("CNA Operand", func() {
...
+			Entry("should not have ovs if deployKSD annotation is not set to true", ksdAnnotationParams{
+				annotationExists:  true,
+				annotationValue:   "false",
+				ksdDeployExpected: false,
+			}),
+			Entry("should not have ovs if deployKSD annotation does not exist", ksdAnnotationParams{
+				annotationExists:  false,
+				annotationValue:   "",
+				ksdDeployExpected: false,
+			}),
```
nunnatsa, line 559: Missing test case? When `desplayKSD` is *not* set, but now it change to `"true"`. Is it a valid case? I know it can't be tested by the current test function.

oshoval: if the annotation doesnt exists it can't be set true
please see the test itself
it is not about upgrade, it is about if to deploy or not deploy the component in the current version

nunnatsa: I thought about a test like this. WDYT?

1. create NetworkAddons object from `hco`: 
    ```go
    existingResource, err := NewNetworkAddons(hco)
    ```
2. check the created object
    ```go
    Expect(existingResource.Spec.KubeSecondaryDNS).To(BeNil(), "KSD spec should not be added")
    ```
3. modify `hco` to add the annotation
4. initialize the client with the "old" NetworkAddons and the modified hco.
5. run `handler.ensure(req)`
6. check the result - 
    ```go
    Expect(existingResource.Spec.KubeSecondaryDNS).To(BeNil(), "KSD spec should not be added")
    ```

oshoval:
Replaced typo ovs with ksd in a string

controllers/operands/networkAddons_test.go
oshoval: We can just add to the first case that it will check it is not on the CR before it adds it right ?

nunnatsa: I mean the use case where the NetworkAddons is w/o the `Spec.KubeSecondaryDNS` field, and the HC is w/o the annotation, and the user now sets the annotation. I don't think its covered.

oshoval: If we break down this secanario, case 2 and 3 are checking the first part of the sentence
"NetworkAddons is w/o the Spec.KubeSecondaryDNS field, and the HC is w/o the annotation"
and then case 1 checks that if you add it, it is deployed
do you agree?

I don't think we need another one that check both, as the building blocks are checked

oshoval: Thanks, as we spoke once https://github.com/kubevirt/hyperconverged-cluster-operator/pull/2191 is merged
Will adapt this PR

oshoval:
Closing this PR as will use a feature gate instead of annotation after speaking with Oren and Nahshon

oshoval:
Created PR with feature gate instead
https://github.com/kubevirt/hyperconverged-cluster-operator/pull/2193

PR #2189, opened 2022-12-21
cluster-up: Fix tag updating

oshoval:
In case the _kubevirtci folder exists,
changing the tag won't reclone the folder.
The cluster-up folder won't be updated and it can lead to bugs in case the folder is changed.

Fix it by deleting the folder in case of tag mismatch.
It will enforce a reclone.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
None
```

### DIFF
--- a/cluster/kubevirtci.sh
+++ b/cluster/kubevirtci.sh
@@ -15,8 +15,24 @@
 export KUBEVIRT_PROVIDER=${KUBEVIRT_PROVIDER:-'k8s-1.23'}
 export KUBEVIRTCI_TAG=$(curl -L -Ss https://storage.googleapis.com/kubevirt-prow/release/kubevirt/kubevirtci/latest)
 KUBEVIRTCI_PATH="${PWD}/_kubevirtci"
+KUBEVIRTCI_REPO='https://github.com/kubevirt/kubevirtci.git'
+
+function cluster::_get_repo() {
+    git --git-dir ${KUBEVIRTCI_PATH}/.git remote get-url origin
+}
+
+function cluster::_get_tag() {
+    git -C ${KUBEVIRTCI_PATH} describe --tags
+}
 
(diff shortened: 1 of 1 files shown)


cluster/kubevirtci.sh
```diff
@@ -15,8 +15,24 @@
...
+}
+
+function cluster::_get_tag() {
+    git -C ${KUBEVIRTCI_PATH} describe --tags
+}
 
 function kubevirtci::install() {
+    # Remove cloned kubevirtci repository if it does not match the requested one
+    if [ -d ${KUBEVIRTCI_PATH} ]; then
+        if [ $(cluster::_get_repo) != ${KUBEVIRTCI_REPO} -o $(cluster::_get_tag) != ${KUBEVIRTCI_TAG} ]; then
```
nunnatsa, line 31: Nit: As far as I understand, the `||` operator is preferred, and I also think it's more readable than `-o`. Please consider to change it.

nunnatsa:
Thanks @oshoval - good catch! Added an optional nit comment.

cluster/kubevirtci.sh
oshoval: why is preferable ?
i prefer to not change it, as it is the same thing we are doing on all the repos tbh
if you are fine with that
I know that "this what we do everywhere" is not an excuse, but as it is a nit i raised it

oshoval:
Addressed comments

nunnatsa (approved):
/approve

PR #2191, opened 2022-12-25
Fix the "when reconciling ovs-cni" unit test

nunnatsa:
The "when reconciling ovs-cni" unit test in
controllers/operators/networkAddons_test.go is wrong. It runs a scenario, but check the result against the wrong object - the original object before the change, instead of the object the it reads back from the client.

This bug hides a wrong test scenario, where the OVS field is expected to be found but it's not.

This PR fixes the test to check the NetworkAddons CR as read from the client, and the wrong test scenario.

This PR also changes `Expect(cl.Get(...)).ToNot(HaveOccurred())` to `Expect(cl.Get(...)).To(Succeed())`, for better readability.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
None
```

### DIFF
--- a/controllers/operands/networkAddons_test.go
+++ b/controllers/operands/networkAddons_test.go
@@ -50,7 +50,7 @@ var _ = Describe("CNA Operand", func() {
 				cl.Get(context.TODO(),
 					types.NamespacedName{Name: expectedResource.Name, Namespace: expectedResource.Namespace},
 					foundResource),
-			).ToNot(HaveOccurred())
+			).To(Succeed())
 			Expect(foundResource.Name).To(Equal(expectedResource.Name))
 			Expect(foundResource.Labels).Should(HaveKeyWithValue(hcoutil.AppLabel, commonTestUtils.Name))
 			Expect(foundResource.Namespace).To(Equal(expectedResource.Namespace))
@@ -112,7 +112,7 @@ var _ = Describe("CNA Operand", func() {
 				cl.Get(context.TODO(),
 					types.NamespacedName{Name: existingResource.Name, Namespace: existingResource.Namespace},
 					foundResource),
-			).ToNot(HaveOccurred())
+			).To(Succeed())
 			Expect(foundResource.Spec.ImagePullPolicy).To(BeEmpty())
 
 			Expect(req.Conditions).To(BeEmpty())
@@ -147,7 +147,7 @@ var _ = Describe("CNA Operand", func() {
 				cl.Get(context.TODO(),
 					types.NamespacedName{Name: existingResource.Name, Namespace: existingResource.Namespace},
 					foundResource),
-			).ToNot(HaveOccurred())
+			).To(Succeed())
 
 			Expect(existingResource.Spec.PlacementConfiguration).To(BeNil())
 			Expect(foundResource.Spec.PlacementConfiguration).ToNot(BeNil())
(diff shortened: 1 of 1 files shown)


controllers/operands/networkAddons_test.go
```diff
@@ -456,52 +456,58 @@ var _ = Describe("CNA Operand", func() {
...
 		},
-			Entry("should have ovs if deployOVS annotation is set to true", ovsAnnotationParams{
-				annotationExists:  true,
+			Entry("should have OVS if deployOVS annotation is set to true", ovsAnnotationParams{
+				ovsExists:         false,
+				setAnnotation:     true,
 				annotationValue:   "true",
 				ovsDeployExpected: true,
 			}),
 			Entry("should not have ovs if deployOVS annotation is not set to true", ovsAnnotationParams{
```
oshoval, line 502: Please consider "is not set to true" -> "is set to false"

oshoval (approved):
Thanks

controllers/operands/networkAddons_test.go
orenc1: not necessarily. OVS is deployed if the annotation value is exactly `"true"`, otherwise it is considered as false and OVS won't be deployed.

oshoval: right but this case does set, and it set to false, so better be explicit

nunnatsa: Fixed. Added two new cases:
"should not have ovs if deployOVS annotation is not set to true" with some other value, and
"should not have ovs if deployOVS annotation is empty" with an empty string.

controllers/operands/networkAddons_test.go
```diff
@@ -456,52 +456,70 @@ var _ = Describe("CNA Operand", func() {
...
+				setAnnotation:     true,
 				annotationValue:   "false",
 				ovsDeployExpected: false,
 			}),
+			Entry("should not have ovs if deployOVS annotation is not set to true", ovsAnnotationParams{
+				ovsExists:         true,
+				setAnnotation:     true,
+				annotationValue:   "someValue",
+				ovsDeployExpected: false,
+			}),
```
oshoval, line 513: I think this one is overkill 
but whatever you prefer

controllers/operands/networkAddons_test.go
```diff
@@ -456,52 +456,70 @@ var _ = Describe("CNA Operand", func() {
...
+			Entry("should not have ovs if deployOVS annotation is empty", ovsAnnotationParams{
```
oshoval, line 514: "if deployOVS annotation doesn't exist"
instead ?

Worth to emphasize that ovsExists is true, so basically this test case checks
undeploy

nunnatsa: But the annotation does exist. its value is empty.

orenc1 (approved):
/approve
/lgtm
/thanks

controllers/operands/networkAddons_test.go
oshoval: ah right mixed with the next one
what about the 2nd part 
"Worth to emphasize that ovsExists is true, so basically this test case checks
undeploy" ?

nunnatsa:
hco-e2e-upgrade-prev-index-aws lane passed

/override ci/prow/hco-e2e-upgrade-prev-index-azure

PR #2192, opened 2022-12-27
Add `operator_health_impact` label to HCO alerts

assafad:
This PR adds to each of HCO alerts a label named `operator_health_impact`, which indicates how the alerts impact the operator health. This will enable us filtering alerts based on their impact on the operator health, and to later have an operator health metric, which will tell the user what is the aggregated operator health.

This label gets one of the following values:
- `critical` indicates that the alert fires due to an issue that impacts the operator's functionality badly, and an action must be taken immediately in order to solve it.
-  `warning` indicates that the alert fires due to an issue that soon might impact the operator's functionality badly, and the user should be warned in order to solve the issue.
- `none` indicates that the alert fires due to an issue that doesn't affect the operator health, and in most cases is related to the workload rather than the operator itself.

**Which issue(s) this PR fixes**:
https://issues.redhat.com/browse/CNV-18923

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
None
```

### DIFF
--- a/controllers/alerts/alerts.go
+++ b/controllers/alerts/alerts.go
@@ -25,6 +25,7 @@ const (
 	unsafeModificationAlert       = "KubevirtHyperconvergedClusterOperatorUSModification"
 	installationNotCompletedAlert = "KubevirtHyperconvergedClusterOperatorInstallationNotCompletedAlert"
 	severityAlertLabelKey         = "severity"
+	healthImpactAlertLabelKey     = "operator_health_impact"
 	partOfAlertLabelKey           = "kubernetes_operator_part_of"
 	partOfAlertLabelValue         = "kubevirt"
 	componentAlertLabelKey        = "kubernetes_operator_component"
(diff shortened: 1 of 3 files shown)


controllers/alerts/alerts.go
```diff
@@ -148,9 +150,10 @@ func createUnsafeModificationAlertRule() monitoringv1.Rule {
 			"runbook_url": unsafeModificationRunbookUrl,
 		},
 		Labels: map[string]string{
-			severityAlertLabelKey:  "info",
-			partOfAlertLabelKey:    partOfAlertLabelValue,
-			componentAlertLabelKey: componentAlertLabelValue,
+			severityAlertLabelKey:     "info",
```
sradco, line 153: @nunnatsa Hi, In the runbook it says that this alert indicates that it may lead to unexpected results or an unstable environment and upgrade is dangerous. Why is severity only "info"?

nunnatsa: This alert is fired when the user is using the JSON annotation to modify one of the sub-resources CRs. We want to warn the user in this case, but on the other hand, this is a valid way to go if there is no other option. In this cases we don't want to alert with too urgent severity.

assafad:
Hi @nunnatsa, Can you please review this PR?

nunnatsa (approved):

nunnatsa:
hco-e2e-image-index-sno-aws lane passed

/override ci/prow/hco-e2e-image-index-sno-azure

PR #2195, opened 2022-12-29
WIP, DNM: Test KSD placement

oshoval:
**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note

```


oshoval:
It seems it doesnt run the placement test
as it had to fail, because CNAO v0.83.1 doesnt support it

oshoval:
/cc @nunnatsa 
FYI (see please comment above)

oshoval:
I might be wrong, and latest CNAO does have it, validating

oshoval:
The current CNAO version already include it
and if we deploy KSD and update the placement test, the test does work 

[https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_hyperconverged[…]converged-cluster-operator-e2e-k8s-1.23/1608402270632808448](https://prow.ci.kubevirt.io/view/gs/kubevirt-prow/pr-logs/pull/kubevirt_hyperconverged-cluster-operator/2195/pull-hyperconverged-cluster-operator-e2e-k8s-1.23/1608402270632808448)
Lines 2657, 2673 and 2697

Closing the PR, as we don't want to always deploy KSD, it was just in order to test placement
and since we are not deploying KSD on HCO we cant update the test

oshoval:
reopened, will check the following:
do not deploy KSD
do change the test

expect to see test fails - just to know the test is fine

oshoval:
@nunnatsa 
FYI something is weird, i reverted v0.83.1 which was the only one that support placement on KSD
deployed KSD, updated test to expect it to be on workload
and the test still pass
what do i do wrong, or there is a bug ?

tests/func-tests/node_placement_test.go
```diff
@@ -63,6 +63,7 @@ var _ = Describe("[rfe_id:4356][crit:medium][vendor:cnv-qe@redhat.com][level:sys
 				//"kube-multus":     false,
 				"ovs-cni-marker": false,
 				"virt-handler":   false,
+				"secondary-dns":  false,
```
oshoval, line 66: need to do also like https://github.com/kubevirt/hyperconverged-cluster-operator/pull/2195/commits/0a925dd25b95fbd6ad9c3c397f1c392e4e62d6de#diff-ed5f0c7b958324db1687cbf20e18758696850ec0fc30c638cf913f523a3452edR71

so in case we dont deploy it, the test will pass

oshoval:
Test pass because the default worker node has the pods, it isn't due to the placement
It was assumed it seems that in case the test fails it means that it was scheduled to a wrong node
but it doesnt mean that the affinity works, as it is the default anyhow and affinity might be broken

We concluded that a deeper look would be done to the test, atm taking some of the changes to the KSD PR

PR #2197, opened 2023-01-03
Add KubeVirt rate liming profiles through annotation

### LINKED ISSUE #2152: Expose new APIs to adjust KubeVirt rate limiting

jcanocan:
The following PR implements KubeVirt rate Limiter setters through annotation.

Implements: [Issue 2152](https://github.com/kubevirt/hyperconverged-cluster-operator/issues/2152)
**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
NONE
```
Fixes #2152


controllers/operands/kubevirt.go
```diff
@@ -270,6 +276,35 @@ func NewKubeVirt(hc *hcov1beta1.HyperConverged, opts ...string) (*kubevirtcorev1
...
+func hcoTuning2Kv(hc *hcov1beta1.HyperConverged) *kubevirtcorev1.ReloadableComponentConfiguration {
+	if hc.Spec.TuningPolicy == hcov1beta1.HyperConvergedAnnotationTuningPolicy {
+		if annotation, ok := hc.Annotations[common.TuningPolicyAnnotationName]; ok {
+
+			var rates rateLimits
+			//get values
+			err := json.Unmarshal([]byte(annotation), &rates)
+			if err != nil {
+				logger.Error(err, "failed to get qps and burst parameters")
+				return nil
```
nunnatsa, line 288: Since in this implementation we're dealing with the HyperConverged CR itself (not another resource like a ConfigMap), I think it would be better to exit with error here. It will also prevent setting of wrong values, as this function also called from the webhooks.

controllers/operands/kubevirt.go
```diff
@@ -183,6 +184,11 @@ type kubevirtHooks struct {
 	cache *kubevirtcorev1.KubeVirt
 }
 
+type rateLimits struct {
+	Qps   int
```
nunnatsa, line 188: will it work to change it to `float32`, to get rid of the casting below?

nunnatsa (commented):
Looks good in general. Way simpler than the alternative. Added a few comments.

/ok-to-test

controllers/operands/kubevirt.go
jcanocan: Done. Please, take a look a it now, I'm returning an error if the `tunningPolicy` is set but the annotation is not present. I'm not sure if this is right :/

controllers/operands/kubevirt.go
jcanocan: Done.

jcanocan:
> TODO:
> 
>     1. add unit tests
> 
>     2. add documentation in docs/cluster-configurations.md

Thanks for the remainder. I'm working on it as we speak :)

controllers/operands/kubevirt.go
```diff
@@ -270,6 +276,34 @@ func NewKubeVirt(hc *hcov1beta1.HyperConverged, opts ...string) (*kubevirtcorev1
...
 }
 
+func hcoTuning2Kv(hc *hcov1beta1.HyperConverged) (*kubevirtcorev1.ReloadableComponentConfiguration, error) {
+	var err error
+	if hc.Spec.TuningPolicy == hcov1beta1.HyperConvergedAnnotationTuningPolicy {
+		if annotation, ok := hc.Annotations[common.TuningPolicyAnnotationName]; ok {
+
+			var rates rateLimits
+			err := json.Unmarshal([]byte(annotation), &rates)
+			if err != nil {
```
nunnatsa, line 286: @jcanocan, maybe you also want to check that both fields are populated.

jcanocan: You are right! I've missed that, I'm not sure if as it is, it's the best way to check it. Please, let me know.

controllers/operands/kubevirt_test.go
```diff
@@ -2524,6 +2524,70 @@ Version: 1.2.3`)
...
+
+			})
+
+			It("Should be empty if annotation is not present", func() {
+				hco.Spec.TuningPolicy = hcov1beta1.HyperConvergedAnnotationTuningPolicy
+
+				kv, err := NewKubeVirt(hco)
+
+				Expect(err).To(HaveOccurred())
+				Expect(err).Should(MatchError("tuning policy set but annotation not present or wrong"))
```
nunnatsa, line 2551: Not sure about this error. It does make sense though. Please remember to well document this behaviour.

controllers/operands/kubevirt_test.go
```diff
@@ -2524,6 +2524,70 @@ Version: 1.2.3`)
...
+				Expect(kv.Spec.Configuration.APIConfiguration.RestClient.RateLimiter.TokenBucketRateLimiter.Burst).To(Equal(200))
+				Expect(kv.Spec.Configuration.ControllerConfiguration.RestClient.RateLimiter.TokenBucketRateLimiter.QPS).To(Equal(float32(100)))
+				Expect(kv.Spec.Configuration.ControllerConfiguration.RestClient.RateLimiter.TokenBucketRateLimiter.Burst).To(Equal(200))
+				Expect(kv.Spec.Configuration.WebhookConfiguration.RestClient.RateLimiter.TokenBucketRateLimiter.QPS).To(Equal(float32(100)))
+				Expect(kv.Spec.Configuration.WebhookConfiguration.RestClient.RateLimiter.TokenBucketRateLimiter.Burst).To(Equal(200))
+				Expect(kv.Spec.Configuration.HandlerConfiguration.RestClient.RateLimiter.TokenBucketRateLimiter.QPS).To(Equal(float32(100)))
+				Expect(kv.Spec.Configuration.HandlerConfiguration.RestClient.RateLimiter.TokenBucketRateLimiter.Burst).To(Equal(200))
+			})
+		})
+
```
nunnatsa, line 2590: Maybe add another test case for corrupted json?

controllers/operands/kubevirt_test.go
jcanocan: Sure!

controllers/operands/kubevirt_test.go
jcanocan: Done. I've also added a new functional test

controllers/operands/kubevirt_test.go
```diff
@@ -2524,6 +2524,82 @@ Version: 1.2.3`)
...
+				kv, err := NewKubeVirt(hco)
+
+				Expect(err).To(HaveOccurred())
+				Expect(err).Should(MatchError("tuning policy set but annotation not present or wrong"))
+
+				Expect(kv).To(BeNil())
+
+			})
+			It("Should be empty if the annotation is present but the parameters are wrong", func() {
```
nunnatsa, line 2556: Maybe: "Should return error if..."

controllers/operands/kubevirt_test.go
```diff
@@ -2524,6 +2524,82 @@ Version: 1.2.3`)
...
+
+				kv, err := NewKubeVirt(hco)
+
+				Expect(err).To(HaveOccurred())
+				Expect(err).Should(MatchError("burst parameter not found in annotation"))
+				Expect(kv).To(BeNil())
+
+			})
+
+			It("Should be empty if the json annotation is corrupted", func() {
```
nunnatsa, line 2571: same

controllers/operands/kubevirt_test.go
```diff
@@ -2524,6 +2524,82 @@ Version: 1.2.3`)
...
+				Expect(err).ToNot(HaveOccurred())
+				Expect(hco.Spec.TuningPolicy).To(BeEmpty())
+				Expect(kv.Spec.Configuration.APIConfiguration).To(BeNil())
+				Expect(kv.Spec.Configuration.ControllerConfiguration).To(BeNil())
+				Expect(kv.Spec.Configuration.WebhookConfiguration).To(BeNil())
+				Expect(kv.Spec.Configuration.HandlerConfiguration).To(BeNil())
+
+			})
+
+			It("Should be empty if annotation is not present", func() {
```
nunnatsa, line 2545: same

controllers/operands/kubevirt_test.go
jcanocan: Done.

controllers/operands/kubevirt_test.go
jcanocan: Done.

controllers/operands/kubevirt_test.go
jcanocan: Done.

docs/cluster-configuration.md
```diff
@@ -1018,3 +1018,45 @@ Labels
...
+For instance, in the above example the `QPS` parameter is set to 100 and the `burst` parameter is set to 200.
+The annotation can be created with the following command:
+
+```bash
+kubectl annotate -n kubevirt-hyperconverged hco kubevirt-hyperconverged hco.kubevirt.io/tuningPolicy='{"qps":100,"burst":200}'
+```
+
+> **_Note_:**  HCO will not configure the rate limiters if both the annotation and the `spec.tuningPolicy` are not populated correctly.
+> Moreover, if `spec.tuningPolicy` is set but the annotation is not present, HCO will throw an error. 
+> In case that the annotation is defined but the field `spec.tuningPolicy`, HCO will not throw an error.
```
nunnatsa, line 1056: > but the field `spec.tuningPolicy`,

Missing word?

Also, I would change `HCO will throw an error` by something like "HCO will reject the changes in the HyperConverged CR", and change `HCO will not throw an error` with something like "HCO will ignore the rate limit configurations".

nunnatsa (commented):
I think we're almost there...

My only comment is that the new `hack/check_tuningPolicy.sh` script is never called.

jcanocan:
> I think we're almost there...
> 
> My only comment is that the new `hack/check_tuningPolicy.sh` script is never called.

I've added [this](https://github.com/kubevirt/hyperconverged-cluster-operator/pull/2197/files#diff-f0ae78aa7707b773d72b208e676f2e41b814231dfb6d4bee843313d9fe99460fR56-R58), I'm not 100% sure if this is the right way.

nunnatsa:
/lgtm
/approve 

@jcanocan please remember to write the release notes in the PR description.

The PR is currently blocked because the release notes are missing.

PR #2198, opened 2023-01-03
ksd: Support passing BaseDomain on Openshift

oshoval:
When running OpenShift the cluster name is available via cluster DNS config resource.
Fetch it and pass it to KSD if it is deployed.
KSD will use it in order to build the FQDN.
This way different clusters will have different FQDN suffix.

Tested on CRC with HCO that the DOMAIN env var of KSD is updated according:
`oc get dnses.config.openshift.io cluster -o json | jq .spec.baseDomain`

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
NONE
```

### DIFF
--- a/controllers/operands/networkAddons_test.go
+++ b/controllers/operands/networkAddons_test.go
@@ -526,12 +526,14 @@ var _ = Describe("CNA Operand", func() {
 		)
 
 		type ksdAnnotationParams struct {
-			ksdExists         bool
-			setFeatureGate    bool
-			featureGateValue  bool
-			ksdDeployExpected bool
+			ksdExists          bool
+			setFeatureGate     bool
+			featureGateValue   bool
+			ksdDeployExpected  bool
+			expectedBaseDomain string
 		}
-		DescribeTable("when reconciling kube-secondary-dns", func(o ksdAnnotationParams) {
+
+		ksdTester := func(o ksdAnnotationParams) {
 			existingCNAO, err := NewNetworkAddons(hco)
 			Expect(err).ToNot(HaveOccurred())
 			if o.ksdExists {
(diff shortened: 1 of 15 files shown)


oshoval:
Note that we have again the action item that
`tests/vendor/github.com/kubevirt/hyperconverged-cluster-operator/pkg/util/cluster.go`
would need to be updated.

pkg/components/components.go
```diff
@@ -524,15 +524,20 @@ func GetClusterPermissions() []rbacv1.PolicyRule {
...
 		{
-			APIGroups: stringListToSlice("config.openshift.io"),
+			APIGroups: stringListToSlice(configOpenshiftIO),
 			Resources: stringListToSlice("apiservers"),
 			Verbs:     stringListToSlice("get", "list", "watch"),
 		},
+		{
+			APIGroups: stringListToSlice(configOpenshiftIO),
+			Resources: stringListToSlice("dnses"),
+			Verbs:     stringListToSlice("get"),
```
nunnatsa, line 539: I hope it's enough. If it won't work, please consider to add also `list` (or even better, add it to the list in line 528).

nunnatsa (approved):

pkg/components/components.go
oshoval: it worked with just get if thats what you mean
thanks !

oshoval:
Added another commit that is the result of `make sanity`, it updates the year
Check it please if you wish to keep it, or that we need to fix the template to not advance the year
(not due to this PR i believe, saw it on kubevirt as well)

Want as well to see if now sanity passes

oshoval:
Coveralls shows lots of stuff which isn't due to this PR
Please advise what AI to do

nunnatsa:
Ignoring non-relevant coverage reduction. 
/override coverage/coveralls

pkg/util/cluster_test.go
```diff
@@ -95,7 +104,7 @@ var _ = Describe("test clusterInfo", func() {
 		os.Setenv(OperatorConditionNameEnvVar, "aValue")
 		cl := fake.NewClientBuilder().
 			WithScheme(testScheme).
-			WithRuntimeObjects(clusterVersion, infrastructure, ingress, apiServer).
+			WithRuntimeObjects(clusterVersion, infrastructure, ingress, apiServer, dns).
```
nunnatsa, line 107: Maybe we can add a test for `GetBaseDomain` and `getClusterBaseDomain` here, or below?

oshoval: I saw it is missing indeed, but saw that there are no UT for the `GetDomain`, `getClusterDomain`
so first wanted to see first comments
Thanks

oshoval:
Added UT

oshoval:
/retest

something happened the the bootstrap ?

oshoval:
Maybe try this image for the presubmits

quay.io/kubevirtci/bootstrap:v20221222-8f66e7e
note it uses podman
this what we use allover

oshoval:
will push in a moment the updated timestamp of the CSV

nunnatsa (approved):
/approve

controllers/operands/networkAddons_test.go
```diff
@@ -558,29 +560,76 @@ var _ = Describe("CNA Operand", func() {
...
+				getClusterInfo := hcoutil.GetClusterInfo
+
+				hcoutil.GetClusterInfo = func() hcoutil.ClusterInfo {
+					return &commonTestUtils.ClusterInfoMock{}
+				}
+
+				DeferCleanup(func() {
+					hcoutil.GetClusterInfo = getClusterInfo
+				})
+			})
```
oshoval, line 607: Possible idea:
Have the original GetClusterInfo always saved in a parameter of the core code,
and a function that will restore it
This way we can just call the function without creating a function, nor saving it

If there can be even a BeforeEach that we can call that does it all with the Defer it is even better
Can be adapted all over the code

oshoval:
Thanks for the review and the help

PR #2204, opened 2023-01-11
Add HCO health metric

assafad:
Add a metric which indicates the health of HCO and its secondary resources, based on the aggregated conditions. The proposed metric is exposed both as a Prometheus metric (named `kubevirt_hco_system_health_status`), and as a field in HCO status (named `systemHealthStatus`).
Its value is set according to the following logic:
- If at least one resource, out of the resources maintained by HCO, is degraded or not available:
`kubevirt_hco_system_health_status` = 2 and `systemHealthStatus` = error

- If all resources maintained by HCO are available and not degraded, but either one of the resources is progressing, or the reconciliation wasn't completed:
`kubevirt_hco_system_health_status `= 1 and `systemHealthStatus` = warning

- If all resources maintained by HCO are available, not degraded, not progressing and the reconciliation was completed:
`kubevirt_hco_system_health_status` = 0 and `systemHealthStatus` = healthy

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
Add HCO health metric
```

Jira-Ticket: https://issues.redhat.com/browse/CNV-24648


assafad:
@sradco, @nunnatsa Can you please have a look?
It doesn't include automated testing yet, working on that

controllers/common/hcoConditions.go
```diff
@@ -59,3 +59,11 @@ func (hc HcoConditions) GetCondition(conditionType string) (metav1.Condition, bo
 	cond, found := hc[conditionType]
 	return cond, found
 }
+
+func (hc HcoConditions) IsStatusConditionTrue(conditionType string) bool {
+	if hc.HasCondition(conditionType) {
+		return hc[conditionType].Status == metav1.ConditionTrue
+	}
+
+	return false
```
nunnatsa, line 68: With this implementation, the code access the map twice to get the same information. You can access the map only once. e.g.
```suggestion
    cond, found := hc[conditionType]
    return found && cond.Status == metav1.ConditionTrue
```

controllers/hyperconverged/hyperconverged_controller.go
```diff
@@ -1045,6 +1055,34 @@ func (r *ReconcileHyperConverged) detectTaintedConfiguration(req *common.HcoRequ
...
+		}
+		return systemHealthStatusWarning
+	}
+
+	if metricErr := metrics.HcoMetrics.SetHCOMetricSystemHealthStatusHealthy(); metricErr != nil {
+		req.Logger.Error(metricErr, "failed to update the systemHealthStatus metric")
+	}
+	return systemHealthStatusHealthy
+}
+
```
nunnatsa, line 1085: Not sure about this function. First, its name: it promises to get the status, but it also sets the metric. In general, function should do one thing. What do you think about splitting it to two functions, one to get the status and one to switch-case the status to set the right metric?

controllers/hyperconverged/hyperconverged_controller.go
```diff
@@ -1045,6 +1055,34 @@ func (r *ReconcileHyperConverged) detectTaintedConfiguration(req *common.HcoRequ
 	}
 }
 
+func (r *ReconcileHyperConverged) getSystemHealthStatus(req *common.HcoRequest) string {
+	isConditionAvailableTrue := req.Conditions.IsStatusConditionTrue(hcov1beta1.ConditionAvailable)
+	isConditionDegradedTrue := req.Conditions.IsStatusConditionTrue(hcov1beta1.ConditionDegraded)
+	isConditionProgressingTrue := req.Conditions.IsStatusConditionTrue(hcov1beta1.ConditionProgressing)
+	isConditionReconcileCompleteTrue := req.Conditions.IsStatusConditionTrue(hcov1beta1.ConditionReconcileComplete)
```
nunnatsa, line 1062: each one of these variables is used only once. Not sure we need them, or to call `IsStatusConditionTrue` instead.

controllers/hyperconverged/hyperconverged_controller.go
```diff
@@ -1045,6 +1055,34 @@ func (r *ReconcileHyperConverged) detectTaintedConfiguration(req *common.HcoRequ
...
+
+	isSystemHealthStatusError := !isConditionAvailableTrue || isConditionDegradedTrue
+	if isSystemHealthStatusError {
```
nunnatsa, line 1065: The `isSystemHealthStatusError` variable is only used here. I think we can drop it, and to use the condition directly for the if; e.g.
```suggestion
	if !isConditionAvailableTrue || isConditionDegradedTrue {
```

Another alternative is to write "micro-functions" so we'll preserve the readability; e.g.
```golang
func isSystemHealthStatusError(conditions HcoConditions) {
    return !req.Conditions.IsStatusConditionTrue(hcov1beta1.ConditionAvailable) || req.Conditions.IsStatusConditionTrue(hcov1beta1.ConditionDegraded)
}

...
    if isSystemHealthStatusError(req.Conditions) { 
        ...
```

What do you think?

api/v1beta1/hyperconverged_types.go
```diff
@@ -480,6 +480,10 @@ type HyperConvergedStatus struct {
 	// DataImportCronTemplates is a list of the actual DataImportCronTemplates as HCO update in the SSP CR. The list
 	// contains both the common and the custom templates, including any modification done by HCO.
 	DataImportCronTemplates []DataImportCronTemplateStatus `json:"dataImportCronTemplates,omitempty"`
+
+	// SystemHealthStatus reflects the health of HCO and its secondary resources, based on the aggregated conditions.
+	// +optional
+	SystemHealthStatus string `json:"systemHealthStatus,omitempty"`
```
nunnatsa, line 486: I wonder why do we need this field. K8s is getting away from a single status fields. This is why we have conditions.

controllers/hyperconverged/hyperconverged_controller.go
nunnatsa, line 1065: Not sure about this logic. The result will be that we'll maybe get false negative, mostly during setup, until the system is fully function.

nunnatsa (commented):
Added a few comments and questions.

api/v1beta1/hyperconverged_types.go
assafad: The motivation was to have a way for users that don’t use Prometheus to check the operator's health. 

@sradco is this field necessary for these users? maybe checking the metrics endpoint (which will include the new metric), without accessing Prometheus UI could be enough?

controllers/hyperconverged/hyperconverged_controller.go
```diff
@@ -988,6 +991,13 @@ func (r *ReconcileHyperConverged) updateConditions(req *common.HcoRequest) {
 		req.Instance.Status.Conditions = conditions
 		req.StatusDirty = true
 	}
+
+	systemHealthStatus := r.getSystemHealthStatus(req)
+
+	if !reflect.DeepEqual(systemHealthStatus, req.Instance.Status.SystemHealthStatus) {
```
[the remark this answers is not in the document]
assafad: Done

controllers/common/hcoConditions.go
assafad: Done

api/v1beta1/hyperconverged_types.go
nunnatsa: This is why we have the conditions.

nunnatsa:
@assafad - please notice that you have 1 code smell.

controllers/hyperconverged/hyperconverged_controller.go
assafad: Took the second approach. thanks

pkg/metrics/metrics.go
```diff
@@ -229,6 +248,26 @@ func (hm *hcoMetrics) IsHCOMetricHyperConvergedExists() (bool, error) {
...
+
+// SetHCOMetricSystemHealthStatusWarning sets the gauge to 1 (warning)
+func (hm *hcoMetrics) SetHCOMetricSystemHealthStatusWarning() error {
+	return hm.SetMetric(HCOMetricSystemHealthStatus, nil, SystemHealthStatusWarning)
+}
+
+// SetHCOMetricSystemHealthStatusError sets the gauge to 2 (error)
+func (hm *hcoMetrics) SetHCOMetricSystemHealthStatusError() error {
+	return hm.SetMetric(HCOMetricSystemHealthStatus, nil, SystemHealthStatusError)
+}
```
nunnatsa, line 264: Can we have one function that will get the status as a parameter? Not sure what is the advantage to write three functions. The reason I'm rising it is the `updateSystemHealthStatusMetric` function, that can be much simpler - just pass the current status to a single function.

controllers/hyperconverged/hyperconverged_controller.go
assafad: Can you elaborate on what would be the reason for this false negative?
Is it related to the conditions' logic, or to the location in the reconciliation in which we update the system health?

pkg/metrics/metrics.go
assafad: Sure, makes sense. 
Should we then keep `updateSystemHealthStatusMetric()` in the controller, or to have all of this logic in the metrics package?

controllers/hyperconverged/hyperconverged_controller.go
nunnatsa: HyperConverged won't be available during setup. According to this code, we'll get an error metric in this case. Is that what we want? @sradco

api/v1beta1/hyperconverged_types.go
nunnatsa: @fabiand - what do you think?

controllers/hyperconverged/hyperconverged_controller.go
assafad: During setup, and only then, if I understand correctly, `ReconcileComplete=Unknown` ([hyperconverged_controller.go#L641](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/main/controllers/hyperconverged/hyperconverged_controller.go#L641)).
Would it be accurate to first check if `ReconcileComplete=Unknown`, and if so, to give the metric `warning`/`healthy`?

@sradco

api/v1beta1/hyperconverged_types.go
fabiand: Is there a condition which is providing this information (healthy or not). If there is a condition, then this is enough imo

pkg/metrics/metrics.go
assafad: Done

pkg/metrics/metrics.go
```diff
@@ -16,9 +16,14 @@ const (
...
 	HCOMetricHyperConvergedExists     = "HyperConvergedCRExists"
+	HCOMetricSystemHealthStatus       = "systemHealthStatus"
 
 	HyperConvergedExists    = float64(1)
 	HyperConvergedNotExists = float64(0)
+
+	SystemHealthStatusHealthy = float64(0)
+	SystemHealthStatusWarning = float64(1)
+	SystemHealthStatusError   = float64(2)
 )
```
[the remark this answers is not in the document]
assafad: Done without the new type. Let me know what you think

pkg/metrics/metrics.go
```diff
@@ -16,11 +16,18 @@ const (
...
 	HCOMetricUnsafeModifications      = "unsafeModifications"
 	HCOMetricHyperConvergedExists     = "HyperConvergedCRExists"
+	HCOMetricSystemHealthStatus       = "systemHealthStatus"
 
 	HyperConvergedExists    = float64(1)
 	HyperConvergedNotExists = float64(0)
 )
 
+const (
+	SystemHealthStatusHealthy float64 = iota
```
dharmit, line 26: Why are we setting this to `float64` type and not `int`?

pkg/metrics/metrics.go
```diff
@@ -80,6 +87,20 @@ var HcoMetrics = func() hcoMetrics {
 				)
 			},
 		},
+		HCOMetricSystemHealthStatus: {
+			fqName:          "kubevirt_hco_system_health_status",
+			help:            "Indicates whether the system health status is healthy (0), warning (1), or error (2)",
```
dharmit, line 92: Is "HCO system health status" same as "system health status"? If not, should we add that bit in the help?

pkg/metrics/metrics.go
```diff
@@ -229,6 +250,16 @@ func (hm *hcoMetrics) IsHCOMetricHyperConvergedExists() (bool, error) {
 	return val == HyperConvergedExists, nil
 }
 
+// SetHCOMetricSystemHealthStatus sets the gauge to status
+func (hm *hcoMetrics) SetHCOMetricSystemHealthStatus(status float64) error {
+	return hm.SetMetric(HCOMetricSystemHealthStatus, nil, status)
+}
+
+// GetHCOMetricSystemHealthStatus returns current value of gauge. If error is not nil than value is undefined
```
dharmit, line 258:
```suggestion
// GetHCOMetricSystemHealthStatus returns current value of gauge. If error is not nil then value is undefined
```

pkg/metrics/metrics.go
nunnatsa: Because Prometheus values are `float64`

pkg/metrics/metrics.go
assafad: The metric aggregates the health of the system - HCO and its secondary resources. We use `hco_` prefix for all HCO metrics, in order to identify and group metrics that are generated by it. 
But sure, it is possible to add more information to the `help`. @sradco WDYT about this description?

pkg/metrics/metrics.go
assafad: Are you sure? I think it's not a matter of time (then), but more "if x happens, than y happens"

dharmit: IMO, "than" is used when comparing things:
- sooner rather than later
- x is greater than y

In this case, I feel (correct me if I'm wrong) that we aren't comparing anything. It's more about what happens ("value is undefined") when a certain condition ("error is not nil") is met.

Apologies for making this about English language rather than Go. 😄

assafad: I guess you're right, thanks. Updated

api/v1beta1/hyperconverged_types.go
sradco: @fabiand  There is no single condition for this

pkg/metrics/metrics.go
sradco: @assafad Please add this sentence "The metric aggregates the health of the system - HCO and its secondary resources. " to the help message.

assafad: updated

PR #2206, opened 2023-01-15
Kubevirt: Enforce limits & requests by a configurable ratio

iholder101:
This PR brings a mechanism to align Kubevirt `virt-launcher` Pods with ResourceQuotas that are applied to the namespace.

As an example, let's look at the following VMI:
```yaml
apiVersion: kubevirt.io/v1
kind: VirtualMachineInstance
metadata:
  labels:
    special: vmi-fedora
  name: vmi-fedora
spec:
  domain:
    devices:
      disks:
      - disk:
          bus: virtio
        name: containerdisk
      - disk:
          bus: virtio
        name: cloudinitdisk
      rng: {}
    resources:
      requests:
        memory: 1024M
  terminationGracePeriodSeconds: 0
  volumes:
  - containerDisk:
      image: quay.io/kubevirt/fedora-with-test-tooling-container-disk:20230115_2baf5303d 
    name: containerdisk
  - cloudInitNoCloud:
      userData: |-
        #cloud-config
        password: fedora
        chpasswd: { expire: False }
    name: cloudinitdisk
```

When this VMI is created, the following `virt-launcher` pod is created (some details are omitted for simplicity):
```yaml
kind: Pod
metadata:
  name: virt-launcher-vmi-fedora-lzzn6
spec:
  containers:
    name: compute
    resources:
      limits:
        devices.kubevirt.io/kvm: "1"
        devices.kubevirt.io/tun: "1"
        devices.kubevirt.io/vhost-net: "1"
      requests:
        cpu: 100m
        devices.kubevirt.io/kvm: "1"
        devices.kubevirt.io/tun: "1"
        devices.kubevirt.io/vhost-net: "1"
        ephemeral-storage: 50M
        memory: "1279755392"
```

As can be seen, this `virt-launcher` has only CPU and memory requests - but not limits. This means that if this VMI is being created in a namespace that has a ResourceQuota defined in it - the virt-launcher Pod won't be able to start. This now can be solved by the mechanism that's presented in this PR. 

To enable this mechanism, first a ratio between memory/CPU limits to request needs to be defined as an annotation in HCO object:
```yaml
kind: HyperConverged
metadata:
  annotations:
    kubevirt.io/cpu-limit-to-request-ratio: "2"
    kubevirt.io/memory-limit-to-request-ratio: "1.5"
```

In addition, a ResourceQuota needs to exist on the relevant namespace. As an example, it's possible to create the following object:
```yaml
apiVersion: v1
kind: ResourceQuota
metadata:
  name: test-rq 
spec:
  hard:
    limits.cpu: "200"
    limits.memory: "2000G"
```
Please take into account that if a ResourceQuota only sets a limit on `limits.cpu` or `limits.cpu`, then CPU/memory limits will be set accordingly. If multiple ResourceQuota exist within the relevant namespace, it takes only one of then to limit CPU/memory limits in order to enforce these limits.

When these annotations are enabled along with a ResourceQuota object, a mutating webhook that's targeted to virt-launcher pods will enforce limits on the pod. It would now look like the following:
```yaml
    resources:
      limits:
        cpu: 200m
        devices.kubevirt.io/kvm: "1"
        devices.kubevirt.io/tun: "1"
        devices.kubevirt.io/vhost-net: "1"
        memory: "1919633088"
      requests:
        cpu: 100m
        devices.kubevirt.io/kvm: "1"
        devices.kubevirt.io/tun: "1"
        devices.kubevirt.io/vhost-net: "1"
        ephemeral-storage: 50M
        memory: "1279755392"
```

Bear in mind that the new webhook would be active only if a request is set but a limit is not.

**Reviewer Checklist**

- [X] PR Message
- [X] Commit Messages
- [X] How to test
- [X] Unit Tests
- [ ] Functional Tests
- [X] User Documentation
- [X] Developer Documentation
- [X] Upgrade Scenario
- [X] Uninstallation Scenario
- [X] Backward Compatibility
- [X] Troubleshooting Friendly
  
**Release note**:

```release-note
Kubevirt: Enforce limits & requests by a configurable ratio
```


pkg/webhooks/mutator/virt-launcher-mutator.go
```diff
@@ -0,0 +1,117 @@
...
+
+func NewVirtLauncherMutator(cli client.Client, hcoNamespace string) *VirtLauncherMutator {
+	return &VirtLauncherMutator{
+		cli:          cli,
+		hcoNamespace: hcoNamespace,
+		logger:       log.Log.WithName("virt-launcher mutator"),
+	}
+}
+
+func (m *VirtLauncherMutator) Handle(_ context.Context, req admission.Request) admission.Response {
```
nunnatsa, line 38: Please use a real parameter for the context. You'll need it below.

pkg/webhooks/mutator/virt-launcher-mutator.go
```diff
@@ -0,0 +1,117 @@
...
+
+	launcherPod := &k8sv1.Pod{}
+	err := m.decoder.Decode(req, launcherPod)
+	if err != nil {
+		m.logErr(err, "cannot decode virt-launcher pod")
+		return admission.Errored(http.StatusBadRequest, err)
+	}
+	originalPod := launcherPod.DeepCopy()
+
+	hco, err := getHcoObject(m.cli, m.hcoNamespace)
```
nunnatsa, line 54: please us the context here.

pkg/webhooks/mutator/utils.go
```diff
@@ -0,0 +1,35 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/kubevirt/hyperconverged-cluster-operator/api/v1beta1"
+	hcoutil "github.com/kubevirt/hyperconverged-cluster-operator/pkg/util"
+)
+
+func getHcoObject(cli client.Client, namespace string) (*v1beta1.HyperConverged, error) {
+	ctx := context.TODO()
```
nunnatsa, line 15: Please define the context as the first function parameter and use it.

nunnatsa (commented):
Added three comments about context, but they are all the same one, actually.

pkg/webhooks/mutator/utils.go
iholder101: Will do.
FYI, the other webhooks are implemented in the same way (context isn't being used).

iholder101: @nunnatsa The ns mutator also calls this function, but it dismisses the context parameter.
You can see [here](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/main/pkg/webhooks/mutator/mutator.go#L97) how I'm simply using the same logic that already exists and following the current pattern.

Does this mean that the already code is wrong?
Should I also add the context for other webhooks as well?

pkg/webhooks/mutator/utils.go
```diff
@@ -0,0 +1,34 @@
...
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/kubevirt/hyperconverged-cluster-operator/api/v1beta1"
+	hcoutil "github.com/kubevirt/hyperconverged-cluster-operator/pkg/util"
+)
+
+func getHcoObject(cli client.Client, ctx context.Context, namespace string) (*v1beta1.HyperConverged, error) {
```
nunnatsa, line 14: The golang convention is that the context is always the first parameter in the function. Could you please reorder?

pkg/webhooks/mutator/utils.go
```diff
@@ -0,0 +1,34 @@
...
+	err := cli.Get(ctx, client.ObjectKeyFromObject(hco), hco)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			logger.Info("HCO CR doesn't not exist, allow hcoNamespace deletion")
+			return nil, err
+		}
+
+		logger.Error(err, "failed getting HyperConverged CR")
+		return nil, err
+	}
```
nunnatsa, line 31: Is that the right error handling and error message, now that it's a generic function?

pkg/webhooks/mutator/virt-launcher-mutator.go
```diff
@@ -0,0 +1,178 @@
...
+	err := m.decoder.Decode(req, launcherPod)
+	if err != nil {
+		m.logErr(err, "cannot decode virt-launcher pod")
+		return admission.Errored(http.StatusBadRequest, err)
+	}
+	originalPod := launcherPod.DeepCopy()
+
+	hco, err := getHcoObject(m.cli, ctx, m.hcoNamespace)
+	if err != nil {
+		m.logErr(err, "cannot get hco object")
```
nunnatsa, line 64: maybe 
```suggestion
		m.logErr(err, "cannot get the HyperConverged object")
```

pkg/webhooks/mutator/virt-launcher-mutator_test.go
```diff
@@ -0,0 +1,159 @@
...
+
+		launcherPod.Spec.Containers[0].Resources = podResources
+		err := mutator.handleVirtLauncherCreation(launcherPod, hco)
+		Expect(err).ToNot(HaveOccurred())
+
+		resources := launcherPod.Spec.Containers[0].Resources
+		Expect(resources.Limits[v1.ResourceCPU].Equal(expectedResources.Limits[v1.ResourceCPU])).To(BeTrue())
+		Expect(resources.Requests[v1.ResourceCPU].Equal(expectedResources.Requests[v1.ResourceCPU])).To(BeTrue())
+		Expect(resources.Limits[v1.ResourceMemory].Equal(expectedResources.Limits[v1.ResourceMemory])).To(BeTrue())
+		Expect(resources.Requests[v1.ResourceMemory].Equal(expectedResources.Requests[v1.ResourceMemory])).To(BeTrue())
```
nunnatsa, line 67: Isn't the gomega's `Equals` matcher work here?

pkg/webhooks/mutator/virt-launcher-mutator_test.go
```diff
@@ -0,0 +1,159 @@
...
+	"github.com/kubevirt/hyperconverged-cluster-operator/controllers/commonTestUtils"
+	"github.com/kubevirt/hyperconverged-cluster-operator/pkg/util"
+)
+
+const (
+	hcoNamespace = HcoValidNamespace
+	podNamespace = "fake-namespace"
+)
+
+var _ = Describe("virt-launcher webhook mutator", func() {
```
nunnatsa, line 24: Please add error cases; e.g. no HyperConverged, negative ratio, zero ratio etc.

pkg/webhooks/mutator/utils.go
nunnatsa: If you don't mind to fix it, please do. You're refactoring this code anyway.

nunnatsa (commented):
Added a few comments

pkg/webhooks/mutator/virt-launcher-mutator.go
```diff
@@ -0,0 +1,178 @@
...
+	k8sv1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	"github.com/kubevirt/hyperconverged-cluster-operator/api/v1beta1"
+)
+
+// TODO: Is that needed?
```
nunnatsa, line 23: This is a way to promise in compile time, that the `VirtLauncherMutator` type actually implements the `admission.Handler` interface.

iholder101: Makes sense! Thanks a lot!

pkg/webhooks/mutator/virt-launcher-mutator_test.go
iholder101: Unfortunately not :(
With Gomega's Equals I get errors that are similar to:
```
  [FAILED] Expected
      <resource.Quantity>: {
          i: {value: 150000000000, scale: -3},
          d: {Dec: nil},
          s: "150M",
          Format: "DecimalSI",
      }
  to equal
      <resource.Quantity>: {
          i: {value: 150, scale: 6},
          d: {Dec: nil},
          s: "150M",
          Format: "DecimalSI",
      }
```

Although the two quantities are effectively equal (to `150M`).

pkg/webhooks/mutator/utils.go
iholder101: :+1:

pkg/webhooks/mutator/virt-launcher-mutator_test.go
iholder101: Done!
Added tests for no HCO and invalid ratio. IMHO we don't need to test the library that parses the annotation as it's beyond this "unit's" scope.

iholder101:
> @iholder101 Hey :) Thank you for the PR! A general question, I can't find the place where we filter the pods based on those that belong to the actual virt-launcher, can you please refer me to the place?

Gladly :)
[Here](https://github.com/kubevirt/hyperconverged-cluster-operator/pull/2206/files#diff-f1531df30500812ec42a33a52023c9ceca8a93d4972132fd241ad143b107c063R824) you can find the `WebhookDescription` object being defined. Please look at the following segment:
```go
ObjectSelector: &metav1.LabelSelector{
  MatchLabels: map[string]string{"kubevirt.io": "virt-launcher"},
},
```

This ensures that only virt-launcher pods are being handled.
This specific label is being added to every virt-launcher pod by Kubevirt, you can see it [here](https://github.com/kubevirt/kubevirt/blob/v0.59.0-alpha.2/pkg/virt-controller/services/template.go#L1360).

pkg/webhooks/mutator/virt-launcher-mutator.go
```diff
@@ -0,0 +1,177 @@
...
+		}
+	}
+
+	return nil
+}
+
+func (m *VirtLauncherMutator) setResourceRatio(launcherPod *k8sv1.Pod, ratioStr string, resourceName k8sv1.ResourceName) error {
+	ratio, err := strconv.ParseFloat(ratioStr, 64)
+	if err != nil {
+		return fmt.Errorf("cann't parse %s to float: %v", ratioStr, err)
```
vladikr, line 104: s/cann't/can't

pkg/webhooks/mutator/virt-launcher-mutator.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	if ratio <= 0 {
+		return fmt.Errorf("negative ratio (%v) is not supported", ratio)
```
vladikr, line 108: I think these error messages should be more self-explanatory. The final error won't have any addition context.

pkg/webhooks/mutator/virt-launcher-mutator.go
```diff
@@ -0,0 +1,177 @@
...
+
+	for i, container := range launcherPod.Spec.Containers {
+		request, requestExists := container.Resources.Requests[resourceName]
+		limit, limitExists := container.Resources.Limits[resourceName]
+
+		if requestExists && !limitExists {
+			launcherPod.Spec.Containers[i].Resources.Limits[resourceName] = m.multiplyResource(request, ratio)
+		}
+		if !requestExists && limitExists {
+			launcherPod.Spec.Containers[i].Resources.Requests[resourceName] = m.multiplyResource(limit, 1/ratio)
```
vladikr, line 119: there is no need to set resource requests if only limits are provided. When users provide only the limits it means that requests will be equal to limits.

vladikr:
@iholder101 Didn't we intend to apply limits only to pods in namespaces that actually require this? (i.e. have resourceQuota)

iholder101:
@vladikr I believe I've addressed all of your comments. Now the mechanism is enabled only if a ResourceQuota exists on the namespace that limits cpu/memory limits.

In addition, I've:
- Added documentation
- Added a unit test to ensure that the mechanism doesn't work if only limits are set
- Updated PR description

Can you PTAL?
FYI @nunnatsa

vladikr:
Look good to me. Thank @iholder101 
/lgtm

docs/cluster-configuration.md
```diff
@@ -1075,4 +1075,77 @@ The `tuningPolicy` feature can be enabled using the following patch:
...
+      requests:
+        cpu: 100m
+        devices.kubevirt.io/kvm: "1"
+        devices.kubevirt.io/tun: "1"
+        devices.kubevirt.io/vhost-net: "1"
+        ephemeral-storage: 50M
+        memory: "1279755392"
+```
+
+Bear in mind that the new webhook would be active only if a request is set but a limit is not.
```
nunnatsa, line 1151: Not sure the user cares about the webhook. We can maybe write something like this:
"Bear in mind that if the limit is already set, HCO will not fix it according to the annotations." or something like this.

WDYT?

pkg/components/components.go
```diff
@@ -490,6 +490,11 @@ func GetClusterPermissions() []rbacv1.PolicyRule {
 			Resources: stringListToSlice("namespaces"),
 			Verbs:     stringListToSlice("get", "list", "watch", "patch", "update"),
 		},
+		{
+			APIGroups: emptyAPIGroup,
+			Resources: stringListToSlice("resourcequotas"),
+			Verbs:     stringListToSlice("get", "list", "watch"),
+		},
```
nunnatsa, line 497: nit: The rback section for pods shares the same setting. it is possible to merge them, to get shorter CSV.

pkg/webhooks/mutator/virt-launcher-mutator.go
```diff
@@ -0,0 +1,234 @@
...
+func (m *VirtLauncherMutator) setResourceRatio(launcherPod *k8sv1.Pod, ratioStr, annotationKey string, resourceName k8sv1.ResourceName) error {
+	ratio, err := strconv.ParseFloat(ratioStr, 64)
+	if err != nil {
+		return fmt.Errorf("%s can't parse ratio %s to float: %v. The ratio is the value of annotation key %s", launcherMutatorStr, ratioStr, err, annotationKey)
+	}
+
+	if ratio <= 0 {
+		return fmt.Errorf("%s doesn't support negative ratio: %v. The ratio is the value of annotation key %s", launcherMutatorStr, ratio, annotationKey)
+	}
+
```
nunnatsa, line 122: What do you think about adding this validation to the validation webhook? Not sure who called first, but if the validation webhook is the first one, it's better to fail fast.

iholder101: Mutating webhooks are actually called first :) (and with a good reason - the validating webhooks get the final manifest version that would actually get into etcd). Therefore I think it's better to fail here.

docs/cluster-configuration.md
iholder101: :+1:

pkg/components/components.go
iholder101: :+1:

pkg/webhooks/mutator/virt-launcher-mutator.go
```diff
@@ -0,0 +1,234 @@
...
+		}
+	}
+
+	return nil
+}
+
+func (m *VirtLauncherMutator) setResourceRatio(launcherPod *k8sv1.Pod, ratioStr, annotationKey string, resourceName k8sv1.ResourceName) error {
+	ratio, err := strconv.ParseFloat(ratioStr, 64)
+	if err != nil {
+		return fmt.Errorf("%s can't parse ratio %s to float: %v. The ratio is the value of annotation key %s", launcherMutatorStr, ratioStr, err, annotationKey)
```
nunnatsa, line 116: please use `%w` to embed the error in `fmt.Errorf`.

pkg/webhooks/mutator/virt-launcher-mutator.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	if ratio <= 0 {
+		return fmt.Errorf("%s doesn't support negative ratio: %v. The ratio is the value of annotation key %s", launcherMutatorStr, ratio, annotationKey)
```
nunnatsa, line 120: same

docs/cluster-configuration.md
```diff
@@ -1075,4 +1075,77 @@ The `tuningPolicy` feature can be enabled using the following patch:
 
 ```bash
 kubectl patch -n kubevirt-hyperconverged hco kubevirt-hyperconverged --type=json -p='[{"op": "add", "path": "/spec/tuningPolicy", "value": "annotation"}]'
-```
\ No newline at end of file
+```
+
+## Enforce CPU/memory limits on namespaces with ResourceQuotas
```
nunnatsa, line 1080: Great documentation. Very clear and helpful. :1st_place_medal:

orenc1:
upgrade lanes failed due to an x509 certificate validation of the validating webhook upon HCO CR deletion:
`x509: certificate signed by unknown authority`
which resembles the bug we're opened against OLM a while ago:
https://bugzilla.redhat.com/show_bug.cgi?id=1907290

/retest

docs/cluster-configuration.md
```diff
@@ -1075,4 +1075,77 @@ The `tuningPolicy` feature can be enabled using the following patch:
...
+apiVersion: v1
+kind: ResourceQuota
+metadata:
+  name: test-rq 
+spec:
+  hard:
+    limits.cpu: "200"
+    limits.memory: "2000G"
+```
+Please take into account that if a ResourceQuota only sets a limit on `limits.cpu` or `limits.cpu`, then CPU/memory limits will be set accordingly. If multiple ResourceQuota exist within the relevant namespace, it takes only one of then to limit CPU/memory limits in order to enforce these limits.
```
orenc1, line 1131: did you mean this?
```suggestion
Please take into account that if a ResourceQuota only sets a limit on `limits.cpu` or `limits.memory`, then CPU/memory limits will be set accordingly. If multiple ResourceQuota exist within the relevant namespace, it takes only one of then to limit CPU/memory limits in order to enforce these limits.
```

pkg/webhooks/mutator/utils.go
```diff
@@ -0,0 +1,34 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      hcoutil.HyperConvergedName,
+			Namespace: namespace,
+		},
+	}
+
+	err := cli.Get(ctx, client.ObjectKeyFromObject(hco), hco)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			logger.Info("HCO CR doesn't not exist")
```
orenc1, line 25:
```suggestion
			logger.Info("HyperConverged CR does not exist")
```

pkg/webhooks/mutator/virt-launcher-mutator.go
```diff
@@ -0,0 +1,234 @@
...
+		hcoNamespace: hcoNamespace,
+		logger:       log.Log.WithName("virt-launcher mutator"),
+	}
+}
+
+func (m *VirtLauncherMutator) Handle(ctx context.Context, req admission.Request) admission.Response {
+	m.logInfo("Starting virt-launcher mutator handling")
+
+	if req.Operation != admissionv1.Create {
+		m.logInfo("not a pod creation - ignoring")
```
orenc1, line 51: if we got here, i think it is an error in the webhook definition, since it is defined to forward the request to the webhook handler only on pods creation:
https://github.com/kubevirt/hyperconverged-cluster-operator/pull/2206/files#diff-a9655738a5e2e547cd22e5e3a3b56329022658b0ae27e94c49f54c18d12ce362R3853-R3861

pkg/webhooks/mutator/virt-launcher-mutator.go
```diff
@@ -0,0 +1,234 @@
...
+	if enforceMemoryLimits {
+		memRatioStr = hco.Annotations[memoryLimitToRequestRatioAnnotation]
+	}
+
+	if enforceCpuLimits {
+		err := m.setResourceRatio(launcherPod, cpuRatioStr, cpuLimitToRequestRatioAnnotation, k8sv1.ResourceCPU)
+		if err != nil {
+			return err
+		}
+	}
```
orenc1, line 102: why this block is not at the same block at lines 90-91 above?
same comment for the other block of:
`if enforceMemoryLimits {`

pkg/webhooks/mutator/virt-launcher-mutator.go
```diff
@@ -0,0 +1,234 @@
...
+	return nil
+}
+
+func (m *VirtLauncherMutator) setResourceRatio(launcherPod *k8sv1.Pod, ratioStr, annotationKey string, resourceName k8sv1.ResourceName) error {
+	ratio, err := strconv.ParseFloat(ratioStr, 64)
+	if err != nil {
+		return fmt.Errorf("%s can't parse ratio %s to float: %w. The ratio is the value of annotation key %s", launcherMutatorStr, ratioStr, err, annotationKey)
+	}
+
+	if ratio <= 0 {
```
orenc1, line 119: i would even say that a ratio < 1 is not supported, because a limit that is smaller than the request makes no sense.

pkg/webhooks/mutator/virt-launcher-mutator.go
```diff
@@ -0,0 +1,234 @@
...
+		return fmt.Errorf("%s doesn't support negative ratio: %v. The ratio is the value of annotation key %s", launcherMutatorStr, ratio, annotationKey)
+	}
+
+	for i, container := range launcherPod.Spec.Containers {
+		request, requestExists := container.Resources.Requests[resourceName]
+		_, limitExists := container.Resources.Limits[resourceName]
+
+		if requestExists && !limitExists {
+			newQuantity := m.multiplyResource(request, ratio)
+			m.logInfo("Replacing old quantity (%s) with new quantity (%s) for pod %s/%s, UID: %s, accodring to ratio: %v",
```
orenc1, line 129: i think it would be nice to specify in the log if the quantity that was replaced is memory or cpu.

pkg/webhooks/mutator/namespace_mutator.go
```diff
@@ -7,23 +7,19 @@ import (
...
-	"github.com/kubevirt/hyperconverged-cluster-operator/api/v1beta1"
-	hcoutil "github.com/kubevirt/hyperconverged-cluster-operator/pkg/util"
 )
 
 const (
 	ignoreOperationMessage   = "ignoring other operations"
-	admittingDeletionMessage = "the namespace doesn't contain HyperConverged CR, admitting its deletion"
-	deniedDeletionMessage    = "HyperConverged CR is still present, please remove it before deleting the containing namespace"
+	admittingDeletionMessage = "the hcoNamespace doesn't contain HyperConverged CR, admitting its deletion"
+	deniedDeletionMessage    = "HyperConverged CR is still present, please remove it before deleting the containing hcoNamespace"
```
orenc1, line 18: **IMPORTANT**
you changed here the error message, replacing `namespace` to `hcoNamespace`.
the delete namespace test checks the exact error message, thus you need to update it also there:
https://github.com/kubevirt/hyperconverged-cluster-operator/blob/43945370a6322b92404a817d6cc8e99292552ab7/hack/test_delete_ns.sh#L41

pkg/webhooks/mutator/virt-launcher-mutator.go
iholder101: Yeah that's right. I just prefer to also have it here as a second line of defense. It also serves as "documentation", since the configuration is located in a different file that would be hard to find without context. In addition, the current namespace mutator does the same thing with deletions.

pkg/webhooks/mutator/namespace_mutator.go
iholder101: Thanks a lot!!!
For the record, I think it's a bad idea to expect to the full error message. As we saw, it leads to surprising errors and enforces the developer to duplicate the error code in two places. I think it should be fixed in future PRs :+1:

iholder101:
Thanks a lot @orenc1!
I've addressed your comments above

pkg/webhooks/mutator/virt-launcher-mutator.go
```diff
@@ -0,0 +1,230 @@
...
+	return nil
+}
+
+func (m *VirtLauncherMutator) setResourceRatio(launcherPod *k8sv1.Pod, ratioStr, annotationKey string, resourceName k8sv1.ResourceName) error {
+	ratio, err := strconv.ParseFloat(ratioStr, 64)
+	if err != nil {
+		return fmt.Errorf("%s can't parse ratio %s to float: %w. The ratio is the value of annotation key %s", launcherMutatorStr, ratioStr, err, annotationKey)
+	}
+
+	if ratio <= 1 {
```
orenc1, line 115: i think `ratio == 1` should be allowed, although it is not so common to have the requests the same as limits, but it is valid configuration.

pkg/webhooks/mutator/namespace_mutator.go
orenc1: it's just because you've decided to change `namespace` to `hcoNamespace` :)
this error message was the same for years

pkg/webhooks/mutator/virt-launcher-mutator.go
iholder101: :+1:

orenc1:
closing and reopening to release the stuck lanes
/close

orenc1:
since it is a temporary solution for the resource limits problem in VirtualMachines, and it is going to get a permanent fix in kubevirt/kubevirt side, i'm ok with disregarding the coverage decrease.
/override coverage/coveralls
/lgtm
/approve

PR #2212, opened 2023-01-22
ksd: Support NameServerIP

oshoval:
Add support of NameServerIP field to HCO CR spec (`kubeSecondaryDNSNameServerIP`).
In case the value is set, and KSD is deployed
it will be populated to CNAO CR and from there to KSD deployment.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
None
```

### DIFF
--- a/controllers/operands/networkAddons.go
+++ b/controllers/operands/networkAddons.go
@@ -9,6 +9,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/utils/net"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	networkaddonsshared "github.com/kubevirt/cluster-network-addons-operator/pkg/apis/networkaddonsoperator/shared"
@@ -143,11 +144,16 @@ func NewNetworkAddons(hc *hcov1beta1.HyperConverged, opts ...string) (*networkad
 		KubeMacPool: &networkaddonsshared.KubeMacPool{},
 	}
 
+	nameServerIP, err := getKSDNameServerIP(hc.Spec.KubeSecondaryDNSNameServerIP)
+	if err != nil {
+		return nil, err
+	}
+
 	if hc.Spec.FeatureGates.DeployKubeSecondaryDNS != nil && *hc.Spec.FeatureGates.DeployKubeSecondaryDNS {
-		cnaoSpec.KubeSecondaryDNS = &networkaddonsshared.KubeSecondaryDNS{}
 		baseDomain := hcoutil.GetClusterInfo().GetBaseDomain()
-		if baseDomain != "" {
-			cnaoSpec.KubeSecondaryDNS.Domain = baseDomain
+		cnaoSpec.KubeSecondaryDNS = &networkaddonsshared.KubeSecondaryDNS{
+			Domain:       baseDomain,
+			NameServerIP: nameServerIP,
 		}
 	}
 
@@ -172,6 +178,18 @@ func NewNetworkAddons(hc *hcov1beta1.HyperConverged, opts ...string) (*networkad
 	}
 
 	return cna, nil
+}
+
+func getKSDNameServerIP(nameServerIPPtr *string) (string, error) {
+	var nameServerIP string
+	if nameServerIPPtr != nil {
+		nameServerIP = *nameServerIPPtr
+		if nameServerIP != "" && !net.IsIPv4String(nameServerIP) {
+			return "", errors.New("kubeSecondaryDNSNameServerIP isn't a valid IPv4")
+		}
+	}
+
+	return nameServerIP, nil
 }
 
 func NewNetworkAddonsWithNameOnly(hc *hcov1beta1.HyperConverged, opts ...string) *networkaddonsv1.NetworkAddonsConfig {
--- a/controllers/operands/networkAddons_test.go
+++ b/controllers/operands/networkAddons_test.go
@@ -540,9 +540,11 @@ var _ = Describe("CNA Operand", func() {
 				existingCNAO.Spec.KubeSecondaryDNS = &networkaddonsshared.KubeSecondaryDNS{}
 			}
 
+			kubeSecondaryDNSNameServerIP := "127.0.0.1"
 			if o.setFeatureGate {
 				deployKubeSecondaryDNS := o.featureGateValue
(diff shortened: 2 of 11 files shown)


docs/cluster-configuration.md
```diff
@@ -166,6 +166,9 @@ For additional information, see here: [KubeSecondaryDNS](https://github.com/kube
 
 **Default**: `false`
 
+In order to set KSD's NameServerIP, set it on HyperConverged CR spec.KubeSecondaryDNSNameServerIP field.
+Default: empty string. Value is a string representation of IPv4.
```
orenc1, line 170: i think the value for that field should be validated at it is a valid IPv4 address, and reject update/creation of HCO CR otherwise.
the user should get immediate feedback if the value is invalid.

oshoval: Done

controllers/operands/networkAddons.go
```diff
@@ -149,6 +150,12 @@ func NewNetworkAddons(hc *hcov1beta1.HyperConverged, opts ...string) (*networkad
 		if baseDomain != "" {
 			cnaoSpec.KubeSecondaryDNS.Domain = baseDomain
 		}
+		if nameServerIP := *hc.Spec.KubeSecondaryDNSNameServerIP; nameServerIP != "" {
+			if !net.IsIPv4String(nameServerIP) {
+				return nil, errors.New("kubeSecondaryDNSNameServerIP isn't a valid IPv4")
```
orenc1, line 155: are you sure the error will be shown to the user in realtime, or just CNAO CR won't be created and this will be shown only in the HCO logs?

oshoval: The ValidateCreate would fail

controllers/operands/networkAddons.go
```diff
@@ -149,6 +150,12 @@ func NewNetworkAddons(hc *hcov1beta1.HyperConverged, opts ...string) (*networkad
 		if baseDomain != "" {
 			cnaoSpec.KubeSecondaryDNS.Domain = baseDomain
 		}
+		if nameServerIP := *hc.Spec.KubeSecondaryDNSNameServerIP; nameServerIP != "" {
```
nunnatsa, line 153: Is it working when the field is nil? I think it will cause segmentation violation error. See here for example: https://go.dev/play/p/UjC-oXlIn-e

I think we must check for nil before accessing this field.

controllers/operands/networkAddons_test.go
```diff
@@ -540,9 +540,11 @@ var _ = Describe("CNA Operand", func() {
 				existingCNAO.Spec.KubeSecondaryDNS = &networkaddonsshared.KubeSecondaryDNS{}
 			}
 
+			kubeSecondaryDNSNameServerIP := "127.0.0.1"
 			if o.setFeatureGate {
 				deployKubeSecondaryDNS := o.featureGateValue
 				hco.Spec.FeatureGates.DeployKubeSecondaryDNS = &deployKubeSecondaryDNS
+				hco.Spec.KubeSecondaryDNSNameServerIP = &kubeSecondaryDNSNameServerIP
```
nunnatsa, line 547: nit: you can use `pointer.String()` for that, e.g.
```suggestion
			if o.setFeatureGate {
				deployKubeSecondaryDNS := o.featureGateValue
				hco.Spec.FeatureGates.DeployKubeSecondaryDNS = &deployKubeSecondaryDNS
				hco.Spec.KubeSecondaryDNSNameServerIP = pointer.String("127.0.0.1")
```

docs/cluster-configuration.md
nunnatsa, line 170: Please add an example, as most of the sections in this document have.

nunnatsa (changes requested):
Added a few comments. One of them seems like a real bug. please check.

controllers/operands/networkAddons_test.go
oshoval: but i also use the value on the EXPECT

oshoval:
Addressed comments
and extracted logic to a function, which validates the value even if KSD is not enabled
so if the changes of the CR are incremental it will still validate it on each step

docs/cluster-configuration.md
```diff
@@ -166,6 +166,9 @@ For additional information, see here: [KubeSecondaryDNS](https://github.com/kube
 
 **Default**: `false`
 
+In order to set KSD's NameServerIP, set it on HyperConverged CR spec.KubeSecondaryDNSNameServerIP field.
+Default: empty string. Value is a string representation of IPv4 (i.e "127.0.0.1").
+
```
nunnatsa, line 171: We have section for each field. I understand that the `deployKubeSecondaryDNS` FG and the new `NameServerIP` are connected, and that there is no point to write the same documentation twice, but we could add a new section for the `NameServerIP` with a reference to this section.

Also, still missing an example (and there is also no example to the `deployKubeSecondaryDNS` FG).

docs/cluster-configuration.md
```diff
@@ -166,6 +166,9 @@ For additional information, see here: [KubeSecondaryDNS](https://github.com/kube
 
 **Default**: `false`
 
+In order to set KSD's NameServerIP, set it on HyperConverged CR spec.KubeSecondaryDNSNameServerIP field.
```
nunnatsa, line 169: `spec.KubeSecondaryDNSNameServerIP` ==> `spec.kubeSecondaryDNSNameServerIP` (lowercase "k")

oshoval: yes saw thanks

oshoval:
addressed comments
moved to its own section in readme

docs/cluster-configuration.md
```diff
@@ -636,6 +636,22 @@ spec:
       - "private-registry-example-2:5000"
       ...
 ```
+
+## KubeSecondaryDNS Name Server IP
+In order to set KSD's NameServerIP, set it on HyperConverged CR under spec.kubeSecondaryDNSNameServerIP field.
+Default: empty string. Value is a string representation of IPv4 (i.e "127.0.0.1").
+For more info see "deployKubeSecondaryDNS Feature Gate".
```
nunnatsa, line 643: Could you please make it a link?:
```
For more info see [deployKubeSecondaryDNS Feature Gate](#deploykubesecondarydns-feature-gate).
```

oshoval:
addressed comments

nunnatsa (approved):

PR #2226, opened 2023-02-01
Provide better guidance for HCO on CRC

tiraboschi:
Provide better guidance for new users
that want to easilyand quickly try HCO on CRC.

Signed-off-by: Simone Tiraboschi <stirabos@redhat.com>

**Reviewer Checklist**

> Reviewers are supposed to review the PR for every aspect below one by one. To check an item means the PR is either "OK" or "Not Applicable" in terms of that item. All items are supposed to be checked before merging a PR. 

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
- [ ] Troubleshooting Friendly
  
**Release note**:

```release-note
NONE
```

PR #2234, opened 2023-02-08
Doc: Add a document explaining how to consume HCO metrics locally

jcanocan:
This PR adds a new entry in the documentation section explaining how to set Prometheus for consuming HCO metrics in a local cluster.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
None
```

### DIFF
--- a/docs/deploy_metrics_locally.md
+++ b/docs/deploy_metrics_locally.md
@@ -1,0 +1,92 @@
+# Consume metrics locally
+This document aims to explain how to deploy and consume all metrics exposed by the Hyperconverged Cluster
+Operator (HCO) locally, i.e. using kubevirtci.
+The exposed metrics are documented [here](metrics.md).
+
+Metrics are supported in CRC out-of-the-box, but at least 20GB of free memory is recommended.
+To deploy metrics, you can target the following command:
+```bash
+$ crc config set enable-cluster-monitoring true
+```
+## Deploying kube-prometheus
+
+In order to deploy `kube-prometheus`, it is required to pick a compatible version of `kube-prometheus`
+with your local kubernetes cluster version. Refer to the [compatibility section](https://github.com/prometheus-operator/kube-prometheus#compatibility) 
+of `kube-prometheus`.
+
+> *Note:* You can set a specific kubernetes cluster version by setting the environmental variable `KUBEVIRT_PROVIDER`.
+
+The next step is to deploy `kube-prometheus` in your local cluster. You can follow the [Quickstart section](https://github.com/prometheus-operator/kube-prometheus#quickstart) of 
+`kube-prometheus` to achieve it.
+
+
+
+## Deploying HCO Service Monitor
+
+Service monitors tells Prometheus where to scrape the metrics. In this case, we want to get HCO metrics.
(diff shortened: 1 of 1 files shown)


tiraboschi:
In https://github.com/kubevirt/hyperconverged-cluster-operator/pull/2226 we are providing better guidance for CRC.
Can you please mention that with CRC executing
```$ crc config set enable-cluster-monitoring true```
(and giving it much more RAM) will be enough?

jcanocan:
> In #2226 we are providing better guidance for CRC. Can you please mention that with CRC executing `$ crc config set enable-cluster-monitoring true` (and giving it much more RAM) will be enough?

Done. We can also wait until #2226 is merged and reference the section in this document.

machadovilaca:
@jcanocan @tiraboschi 

This makes sense, I already found myself having to do this to get metrics to work on `kubevirtci`, but personally, I think we should solve this another way.
Right now, HCO only provides support for monitoring in OpenShift clusters. In my opinion, I think we should, at least, create and reconcile the Service to expose metrics and the Prometheus ServiceMonitor on plain Kubernetes clusters too.

Then, the only thing needed to get metrics working on both `crc` and `kubevirtci` out of the box would be to make sure the Prometheus is running in each, with `$ crc config set enable-cluster-monitoring true` or `$ export KUBEVIRT_DEPLOY_PROMETHEUS=true`, respectively.

jcanocan:
> @jcanocan @tiraboschi
> 
> This makes sense, I already found myself having to do this to get metrics to work on `kubevirtci`, but personally, I think we should solve this another way. Right now, HCO only provides support for monitoring in OpenShift clusters. In my opinion, I think we should, at least, create and reconcile the Service to expose metrics and the Prometheus ServiceMonitor on plain Kubernetes clusters too.
> 
> Then, the only thing needed to get metrics working on both `crc` and `kubevirtci` out of the box would be to make sure the Prometheus is running in each, with `$ crc config set enable-cluster-monitoring true` or `$ export KUBEVIRT_DEPLOY_PROMETHEUS=true`, respectively.

Hi! Many thanks for the feedback.

I agree. I would like to open this discussion to alerts. As far as I know, as it is now, alerts are not working on `kubevirtci` out-of-the-box. We can either document how to deploy them or to make HCO deploy them. This can be done either as part of Prometheus deploy, using the var `KUBEVIRT_DEPLOY_PROMETHEUS` or separately using another var like `KUBEVIRT_DEPLOY_PROMETHEUS_ALERTS=true`. What do you think?

jcanocan:
>Right now, HCO only provides support for monitoring in OpenShift clusters. In my opinion, I think we should, at least, create and reconcile the Service to expose metrics and the Prometheus ServiceMonitor on plain Kubernetes clusters too.

Hi @machadovilaca @tiraboschi Do we have any plans to support this in the sort-term? If no, we can push this forward and drop it or updated it later. WDYT?

PR #2254, opened 2023-02-20
Add `kubevirt_hyperconverged_operator_health_status` recording rule

assafad:
This PR adds a new recording rule - `kubevirt_hyperconverged_operator_health_status` with the label `name=kubevirt-hyperconvered`. 
This recording rule indicates the total health of HCO and its secondary resources, based on the following:
- The firing alerts impact on the operator's health, by using their `operator_health_impact` label.
- The conditions which are reported by the different components, by using `kubevirt_hco_system_health_status` metric which aggregates this.

Its value is set according to the following logic:
-  If there are firing alerts with `operator_health_impact = critical` or `kubevirt_hco_system_health_status = 2`:
`kubevirt_hyperconverged_operator_health_status = 2`
-  Else, if there are firing alerts with `operator_health_impact = warning` or `kubevirt_hco_system_health_status = 1`:
`kubevirt_hyperconverged_operator_health_status = 1`
- Else, `kubevirt_hyperconverged_operator_health_status = 0`

jira-ticket: [CNV-14700](https://issues.redhat.com/browse/CNV-14700)

**Reviewer Checklist**

> Reviewers are supposed to review the PR for every aspect below one by one. To check an item means the PR is either "OK" or "Not Applicable" in terms of that item. All items are supposed to be checked before merging a PR. 

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
- [ ] Troubleshooting Friendly
  
**Release note**:

```release-note
Add `kubevirt_hyperconverged_operator_health_status` recording rule 
```

PR #2259, opened 2023-02-22
Add e2e-test to operator_health_status recording rule

assafad:
Add e2e-test to [kubevirt_hyperconverged_operator_health_status ](https://github.com/kubevirt/hyperconverged-cluster-operator/pull/2254) recording rule.

jira-ticket: https://issues.redhat.com/browse/CNV-25164

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
Add e2e-test to operator_health_status recording rule
```

### DIFF
--- a/tests/func-tests/monitoring_test.go
+++ b/tests/func-tests/monitoring_test.go
@@ -7,7 +7,9 @@ import (
 	"flag"
 	"fmt"
 	"io/ioutil"
+	"math"
 	"net/http"
+	"strconv"
 	"time"
 
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -30,6 +32,7 @@ import (
 	promApi "github.com/prometheus/client_golang/api"
 	promApiv1 "github.com/prometheus/client_golang/api/prometheus/v1"
 	promConfig "github.com/prometheus/common/config"
+	promModel "github.com/prometheus/common/model"
 
 	"github.com/kubevirt/hyperconverged-cluster-operator/api/v1beta1"
 	tests "github.com/kubevirt/hyperconverged-cluster-operator/tests/func-tests"
@@ -41,6 +44,12 @@ const (
 	containerImage = "kubevirt/cirros-container-disk-demo"
 )
 
+const (
+	noneImpact float64 = iota
+	warningImpact
+	criticalImpact
+)
+
 var _ = Describe("[crit:high][vendor:cnv-qe@redhat.com][level:system]Monitoring", func() {
 	flag.Parse()
 
@@ -48,6 +57,7 @@ var _ = Describe("[crit:high][vendor:cnv-qe@redhat.com][level:system]Monitoring"
 	var virtCli kubecli.KubevirtClient
 	var promClient promApiv1.API
 	var prometheusRule monitoringv1.PrometheusRule
+	var initialOperatorHealthMetricValue float64
 
 	runbookClient.Timeout = time.Second * 3
 
@@ -58,6 +68,8 @@ var _ = Describe("[crit:high][vendor:cnv-qe@redhat.com][level:system]Monitoring"
 		tests.SkipIfNotOpenShift(virtCli, "Prometheus")
 		promClient = initializePromClient(getPrometheusUrl(virtCli), getAuthorizationTokenForPrometheus(virtCli))
 		prometheusRule = getPrometheusRule(virtCli)
+
+		initialOperatorHealthMetricValue = getMetricValue(promClient, "kubevirt_hyperconverged_operator_health_status")
 	})
 
 	It("Alert rules should have all the required annotations", func() {
@@ -103,6 +115,7 @@ var _ = Describe("[crit:high][vendor:cnv-qe@redhat.com][level:system]Monitoring"
 		Expect(err).ShouldNot(HaveOccurred())
 
 		verifyAlertExists(promClient, "KubevirtHyperconvergedClusterOperatorCRModification", 60*time.Second)
+		verifyOperatorHealthMetricValue(promClient, initialOperatorHealthMetricValue, warningImpact)
 	})
 
 	It("KubevirtHyperconvergedClusterOperatorUSModification alert should fired when there is an jsonpatch annotation to modify an operand CRs", func() {
@@ -115,6 +128,7 @@ var _ = Describe("[crit:high][vendor:cnv-qe@redhat.com][level:system]Monitoring"
 		updateHCO(virtCli, hco)
 
 		verifyAlertExists(promClient, "KubevirtHyperconvergedClusterOperatorUSModification", 60*time.Second)
+		verifyOperatorHealthMetricValue(promClient, initialOperatorHealthMetricValue, warningImpact)
 	})
 
 	It("NodeStatusMaxImagesExceeded alert should fired when node max images is reached", func() {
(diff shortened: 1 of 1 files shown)


assafad:
Hi @machadovilaca, @sradco, Can you please review?

tests/func-tests/monitoring_test.go
```diff
@@ -118,6 +119,26 @@ var _ = Describe("[crit:high][vendor:cnv-qe@redhat.com][level:system]Monitoring"
...
+		kubevirtHyperconvergedOperatorHealthStatusValue := getMetricValue(promClient, "kubevirt_hyperconverged_operator_health_status")
+
+		if criticalAlertsAreFiring || kubevirtHCOSystemHealthStatusValue == "2" {
+			Expect(kubevirtHyperconvergedOperatorHealthStatusValue).To(Equal("2"))
+		} else if warningAlertsAreFiring || kubevirtHCOSystemHealthStatusValue == "1" {
+			Expect(kubevirtHyperconvergedOperatorHealthStatusValue).To(Equal("1"))
+		} else {
+			Expect(kubevirtHyperconvergedOperatorHealthStatusValue).To(Equal("0"))
+		}
+	})
```
machadovilaca, line 141: I think it would be better if you either

1) integrate this test with other alert tests, and verify that, for example, when `KubevirtHyperconvergedClusterOperatorUSModification` is fired, `kubevirt_hyperconverged_operator_health_status` value is 1.

2) maintain the test separately, but you intentionally cause an alert to fire and verify that the metric changes accordingly

machadovilaca (changes requested):

tests/func-tests/monitoring_test.go
assafad: We considered it, but there are downsides for each of these options. 

Option 1:
At the moment, there are only e2e tests for `warning` impact alerts, as we don't have a `none` alert, and the only `critical` impact alert (`KubevirtHyperconvergedClusterOperatorInstallationNotCompletedAlert`) doesn't have an e2e test. Therefore, the integration will miss these cases.

Option 2:
1. This recording rule depends on many alerts. We can trigger a `warning`alert, and expect the recording rule to equal 1, but other things that we didn't initiate can trigger `critical` alerts and fail the test, although the recording rule functions as expected (i.e. its value is 2).
2. We probably can't trigger all the operator alerts (HCO and its secondary components), so even if we trigger a HCO alert and check the recording rule, this doesn't promise that other alerts will make the recording rule behave the same.

Therefore, @sradco and I agreed that the test will only check an existing state of the cluster, and verify the recording rule value according to it.

machadovilaca: anyway, I think it is better to use option 1. You can after all the `verifyAlertExists` check the value of `kubevirt_hyperconverged_operator_health_status`.

If it was 0 before and we triggered a warning alert it should be 1.
If it was 1 or 2 before and we triggered a warning alert it should stay the same.
If it was 0 or 1 before and we triggered a critical alert it should be 2.
If it was 2 before and we triggered a critical alert it should stay the same.

Related to the critical, I missed the test for `KubevirtHyperconvergedClusterOperatorInstallationNotCompletedAlert`, but I can add it later, with this new check

machadovilaca: this could maybe even be done in the `beforeeach`/`aftereach`, you check the alerts before and that the metric is correct, and you check the alerts after and that the metric is updated/correct

assafad: +1. 
Since we don't have a `none` alert, and your suggestion doesn't include changes in components conditions, I suggest to also keep the new test case as well. It will be the only place we verify that the recording rule behaves as expected considering the conditions.

assafad: The new test verifies that the recording rule follows `kubevirt_hco_system_health_status` metric, which is based on the conditions. Your suggestion is to  check how the alerts impact the recording rule, but don't check `kubevirt_hco_system_health_status` metric.

assafad: I will try to integrate it with your suggestion. Thanks!

machadovilaca: but if you don't provoke any "real" change that triggers the metrics to change, that interdependence between `kubevirt_hco_system_health_status` and `kubevirt_hyperconverged_operator_health_status` is already tested in the Prometheus unit tests

assafad: updated

tests/func-tests/monitoring_test.go
```diff
@@ -46,8 +49,8 @@ var _ = Describe("[crit:high][vendor:cnv-qe@redhat.com][level:system]Monitoring"
 
 	var err error
 	var virtCli kubecli.KubevirtClient
-	var promClient promApiv1.API
 	var prometheusRule monitoringv1.PrometheusRule
+	var initialHealthMetricValue int
```
machadovilaca, line 53: you should set up the value of `initialHealthMetricValue` in the `BeforeEach`

machadovilaca (changes requested):

tests/func-tests/monitoring_test.go
assafad: Done

tests/func-tests/monitoring_test.go
```diff
@@ -301,7 +334,7 @@ func getNodeStatusMaxImages(virtCli kubecli.KubevirtClient, nodeName string) (in
 	return kc.Kubeletconfig.NodeStatusMaxImages, nil
 }
 
-func verifyAlertExists(promClient promApiv1.API, alertName string, forDuration time.Duration) {
+func verifyAlertExists(alertName string, forDuration time.Duration) {
```
nunnatsa, line 337: why did you remove this parameter?

tests/func-tests/monitoring_test.go
```diff
@@ -30,14 +32,19 @@ import (
...
 	promApiv1 "github.com/prometheus/client_golang/api/prometheus/v1"
 	promConfig "github.com/prometheus/common/config"
+	promModel "github.com/prometheus/common/model"
 
 	"github.com/kubevirt/hyperconverged-cluster-operator/api/v1beta1"
 	tests "github.com/kubevirt/hyperconverged-cluster-operator/tests/func-tests"
 )
 
 var runbookClient = http.DefaultClient
+var promClient promApiv1.API
```
nunnatsa, line 42: why did you move it here?

tests/func-tests/monitoring_test.go
```diff
@@ -38,6 +41,9 @@ import (
 var runbookClient = http.DefaultClient
 
 const (
+	noneImpact float64 = iota
+	warningImpact
+	criticalImpact
```
nunnatsa, line 46: Please don't mix `iota` consts with other consts. Can you split it into two const sentenses?

assafad: Done

nunnatsa (approved):

nunnatsa:
hco-e2e-upgrade-index-sno-azure lane passed.
/override hco-e2e-upgrade-index-sno-aws

PR #2267, opened 2023-02-27
Prepare for the new version of the ginkgolinter

nunnatsa:
Fix new linter warnings for wrong boolean comparisons; e.g.

`Expect(x == 10).Equal(true)` should be `Expect(x).Equal(10)`

Also, set the golangci-lint version to `v1.51.2` (current latest) so we won't be surprised by new linter rules, but we'll can control when we adopt them.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
None
```

### DIFF
--- a/.github/workflows/pr-sanity.yaml
+++ b/.github/workflows/pr-sanity.yaml
@@ -34,6 +34,7 @@ jobs:
       - name: Run golangci-lint
         uses: golangci/golangci-lint-action@v3
         with:
+          version: v1.51.2
           skip-pkg-cache: true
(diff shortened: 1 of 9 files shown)


tiraboschi (approved):

nunnatsa:
/override coverage/coveralls

False negative

PR #2276, opened 2023-03-05
Fix junit output from tests

nunnatsa:
The custom reporter are deprecated in ginkgo and are ignore. As result, the report files are not created.

This PR restores the junit report, as it supported by ginkgo.

Polarian report can't work for now. Future solution will be to add json report and translate it to polarion.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
  
**Release note**:

```release-note
None
```

### DIFF
--- a/hack/run-tests.sh
+++ b/hack/run-tests.sh
@@ -28,7 +28,8 @@ fi
 source ./hack/check_operator_condition.sh
 printOperatorCondition
 
-${TEST_OUT_PATH}/func-tests.test -ginkgo.v -junit-output="${TEST_OUT_PATH}/output/junit.xml" -installed-namespace="${INSTALLED_NAMESPACE}" -cdi-namespace="${INSTALLED_NAMESPACE}" "$@" "${KUBECONFIG_FLAG}"
+mkdir "${TEST_OUT_PATH}/output"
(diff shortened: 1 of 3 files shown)


hack/run-tests.sh
```diff
@@ -28,7 +28,8 @@ fi
 source ./hack/check_operator_condition.sh
 printOperatorCondition
 
-${TEST_OUT_PATH}/func-tests.test -ginkgo.v -junit-output="${TEST_OUT_PATH}/output/junit.xml" -installed-namespace="${INSTALLED_NAMESPACE}" -cdi-namespace="${INSTALLED_NAMESPACE}" "$@" "${KUBECONFIG_FLAG}"
+mkdir "${TEST_OUT_PATH}/output"
```
tiraboschi, line 31: Please use 
```suggestion
mkdir -p "${TEST_OUT_PATH}/output"
```
to make it more robust (no error if existing, make parent directories as needed)

PR #2291, opened 2023-03-16
Add metrics linter

avlitman:
This linter will lint the metrics created in "github.com/kubevirt/hyperconverged-cluster-operator/pkg/metrics". In order to run the linter:
make lint-metrics

jira-ticket: https://issues.redhat.com/browse/CNV-28665

**Reviewer notes**:
Linter output:
```
kubevirt_hco_out_of_band_modifications_count: counter metrics should have "_total" or "_timestamp_seconds" suffix
kubevirt_hco_out_of_band_modifications_count: non-histogram and non-summary metrics should not have "_count" suffix
kubevirt_hco_unsafe_modification_count: non-histogram and non-summary metrics should not have "_count" suffix
kubevirt_hyperconverged_operator_health_status: name need to start with "kubevirt_hco_"
make: *** [Makefile:227: lint-metrics] Error 1

```

note that we ignore all current errors until the metrics names will be fixed in:https://github.com/kubevirt/hyperconverged-cluster-operator/pull/2358

**Release note**:

```
Add a metrics name linter. In order to monitor metrics naming and make sure they aligned with the Prometheus metrics naming best practices. 

```

Falied sanity for review purposes: https://github.com/kubevirt/hyperconverged-cluster-operator/actions/runs/5209824760/jobs/9400149372?pr=2291


hack/metrics_linter/metrics_linter.go
```diff
@@ -0,0 +1,258 @@
...
+		"inches": "meters",
+		"yards":  "meters",
+		"miles":  "meters",
+		// Bytes
+		"bits": "bytes",
+		// Energy
+		"calories": "joules",
+		// Mass
+		"pounds": "grams",
+		"ounces": "grams",
```
nunnatsa, line 79: do we need these?

avlitman: @nunnatsa I tried to import it from here [promlint] (https://github.com/prometheus/client_golang/tree/v1.14.0/prometheus/testutil/promlint) but it's not working with just import.

Makefile
```diff
@@ -223,6 +223,9 @@ deploy_hpp:
 validate-no-offensive-lang:
 	./hack/validate-no-offensive-lang.sh
 
+generate-metric-linter:
```
machadovilaca, line 226: why not call this `validate-metric-names`?

machadovilaca (dismissed):
Two questions:

- why are we not using the [Prometheus lint package](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus/testutil/promlint) for the more basic validations such as suffixes and units, and on top of that adding the prefixes validations?

- why are we on HCO validating the metric names for all the components and failing our CI if they have the wrong names? It is probably better to add this check on their repositories so that their CI fails when they add a metric with incorrect names.

Makefile
avlitman: Right. it sound better, fixing.

avlitman:
> Two questions:
> 
> * why are we not using the [Prometheus lint package](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus/testutil/promlint) for the more basic validations such as suffixes and units, and on top of that adding the prefixes validations?
 tried to import the units from here [promlint] (https://github.com/prometheus/client_golang/tree/v1.14.0/prometheus/testutil/promlint) but it's not working with just import. I need help with that.

> * why are we on HCO validating the metric names for all the components and failing our CI if they have the wrong names? It is probably better to add this check on their repositories so that their CI fails when they add a metric with incorrect names.
why for all components? we check only hco metrics see: https://github.com/kubevirt/hyperconverged-cluster-operator/pull/2291/files#diff-0340ff53219188bfce205be02fde17c6ab4dad09d02ab7927c426e5b0827564dR241.

And I opened a bug to update the metrics names: https://bugzilla.redhat.com/show_bug.cgi?id=2179660.
@sradco should I fix the names in this pr?

Makefile
avlitman: @machadovilaca Note that we also check that the metric have help section and that the type is valid.

machadovilaca: do you think `lint-metrics` could be better maybe?

machadovilaca:
@avlitman based on the requirements in the PR description:

> - The metric name should start with the operator kubevirt, followed by a sub-operator hco (only if operator!=sub-operator).

we need to implement

> - The metric type should be one of counter, gauge, histogram, or summary.

given by the necessary conversion to the [metric family structs](https://github.com/prometheus/client_model/blob/master/go/metrics.pb.go)

> - The metric should have a description.

[promlint](https://github.com/prometheus/client_golang/blob/254e5468413f19fb75cdad45f5ddc0b8c975188c/prometheus/testutil/promlint/promlint.go#L135)

> - The metric name should be in snake_case.

[promlint](https://github.com/prometheus/client_golang/blob/254e5468413f19fb75cdad45f5ddc0b8c975188c/prometheus/testutil/promlint/promlint.go#L257)

> - The metric name should not contain colons or metric types.

[promlint 1](https://github.com/prometheus/client_golang/blob/254e5468413f19fb75cdad45f5ddc0b8c975188c/prometheus/testutil/promlint/promlint.go#L228)

[promlint 2](https://github.com/prometheus/client_golang/blob/254e5468413f19fb75cdad45f5ddc0b8c975188c/prometheus/testutil/promlint/promlint.go#L246)

> - The metric name should **contain** `total` for non-counter metrics, and should not **end** in `count`, or `sum` for non-histogram and non-summary metrics, and not **end** in `bucket` for non-histogram metrics.

[promlint 1](https://github.com/prometheus/client_golang/blob/254e5468413f19fb75cdad45f5ddc0b8c975188c/prometheus/testutil/promlint/promlint.go#L168)

[promlint 2](https://github.com/prometheus/client_golang/blob/254e5468413f19fb75cdad45f5ddc0b8c975188c/prometheus/testutil/promlint/promlint.go#L187)

> - The metric name should have a unit, unless it ends in total, count, or sum.

We need to implement, but I would like more to see us creating a PR to add this in promlint

> - If the metric name has a unit, it should use the base unit, and not any abbreviation.

[promlint](https://github.com/prometheus/client_golang/blob/254e5468413f19fb75cdad45f5ddc0b8c975188c/prometheus/testutil/promlint/promlint.go#L274)

I think we can take advantage of their work and them maintaining those rules, am I missing something?

hack/metric-linter/metric_name_linter.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+PROJECT_ROOT="$(readlink -e "$(dirname "${BASH_SOURCE[0]}")"/../)"
+cd "${PROJECT_ROOT}/metric-linter"
+
+# Run the git command to get the remote origin URL and extract the operator and sub-operator values
+url=$(git config --get remote.origin.url | awk -F.com/ '{print $2}' | awk -F. '{print $1}')
+operator=$(echo "$url" | awk -F/ '{print $1}')
+subOperator=$(echo "$url" | awk -F/ '{print $2}')
```
sradco, line 9: @avlitman @machadovilaca I'm not sure if this is good practice to set the operator and sub operator name. It might be ok, but only for our specific CNV use case. What will happen if we add a new operator? Are we sure we would still use the same logic for the sub operator name?
I think we should pass it as a parameter. What do you think?

hack/metric-linter/metric_name_linter.sh
```diff
@@ -0,0 +1,24 @@
...
+
+# If the sub-operator contains more than one word use an abbreviation
+if echo "$subOperator" | grep -q "-"; then
+  # Extract first letter of each word separated by hyphen and combine them to form new subOperator
+  subOperator=$(echo "$subOperator" | awk -F- '{for(i=1;i<=NF;i++) printf substr($i,1,1)}')
+fi
```
sradco, line 15: Same as above

hack/metric-linter/metric_name_linter.sh
sradco: I think we should run the command with something like:
metric_name_linter.sh --operator='kubevirt' --suboperator='hco'
Is that possible?

avlitman: Done

hack/prom_metric_linter.sh
```diff
@@ -0,0 +1,65 @@
...
+        ;;
+    esac
+done
+
+# Get the metrics list
+go build -o _out/prom-metrics-collector "$METRICS_COLLECTOR_PATH/..."
+json_output=$(_out/prom-metrics-collector 2>/dev/null)
+
+# Run the linter by using the prom-metrics-linter Docker container
+errors=$(docker run -i "quay.io/kubevirt/prom-metrics-linter:$linter_image_tag" \
```
avlitman, line 56: @orenc1 @nunnatsa do we have a parameter for docker/podman I used docker for no special reason.

nunnatsa: > @orenc1 @nunnatsa do we have a parameter for docker/podman I used docker for no special reason.

Yes: e.g. here: https://github.com/kubevirt/hyperconverged-cluster-operator/blob/1064220644c06831cb625ce09e8f598c68671260/hack/upgrade-test-clusterserviceversion.sh#L15-L17

avlitman: @nunnatsa I was able to implement but output added since this change:

before output is:
```
[alitman@alitman hyperconverged-cluster-operator]$ make lint-metrics
./hack/prom_metric_linter.sh --operator-name="kubevirt" --sub-operator-name="hco"
```

Now output is:
```
[alitman@alitman hyperconverged-cluster-operator]$ make lint-metrics
./hack/prom_metric_linter.sh --operator-name="kubevirt" --sub-operator-name="hco"
++ podman ps
hack/cri-bin.sh: line 5: podman: command not found
++ docker ps
++ _cri_bin=docker
++ echo 'selecting docker as container runtime'
selecting docker as container runtime
++ CRI_BIN=docker
++ CRI_INSECURE=
++ docker run -i quay.io/kubevirt/prom-metrics-linter:v0.0.1 '--metric-families=[{"name":"kubevirt_hco_out_of_band_modifications_total","help":"Count of out-of-band modifications overwritten by HCO","type":0},{"name":"kubevirt_hco_number_of_unsafe_modifications","help":"Count of unsafe modifications in the HyperConverged annotations","type":1},{"name":"kubevirt_hco_hyperconverged_cr_exists","help":"Indicates whether the HyperConverged custom resource exists (1) or not (0)","type":1},{"name":"kubevirt_hco_system_health_status","help":"Indicates whether the system health status is healthy (0), warning (1), or error (2), by aggregating the conditions of HCO and its secondary resources","type":1}]' --operator-name=kubevirt --sub-operator-name=hco
+ errors=
+ [[ '' != '' ]]
```

Do you know why this output added?

tools/prom-metrics-collector/metrics_collector.go
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * Copyright 2023 Red Hat, Inc.
+ *
+ */
+
+package main
+
+import (
+	"github.com/kubevirt/hyperconverged-cluster-operator/pkg/metrics"
+	"github.com/kubevirt/monitoring/test/metrics/prom-metrics-linter/metrics-parser"
```
nunnatsa, line 24: Please rename to avoid underscore names bellow.

e.g.
```suggestion
	parser "github.com/kubevirt/monitoring/test/metrics/prom-metrics-linter/metrics-parser"
```

or even
```suggestion
	metricsparser "github.com/kubevirt/monitoring/test/metrics/prom-metrics-linter/metrics-parser"
```

avlitman: Done

nunnatsa (approved):
/approve

PR #2299, opened 2023-03-26
Remove HPP usage from CI

orenc1:
We can probably use the storage class provided by the cloud provider for the VMs backend storage.

**What this PR does / why we need it**:

**Reviewer Checklist**

> Reviewers are supposed to review the PR for every aspect below one by one. To check an item means the PR is either "OK" or "Not Applicable" in terms of that item. All items are supposed to be checked before merging a PR. 

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
- [ ] Troubleshooting Friendly

**Jira Ticket**:

```jira-ticket

```

**Release note**:

```release-note
remove HPP usage from CI
```

PR #2303, opened 2023-03-28
Fix CI

nunnatsa:
## fix HHP deployment
Up until now, we were using the legacy way of deploying hostpath-provisioner.
We can start consuming HPP-CSI instead, which does not require setting up MachineConfigs and can take advantage of the underlying storage backend provided by the cloud provider.

## Fix flacky functional test
the verifyOperatorHealthMetricValue sometimes fials. trying to use
`Eventually` in order to retry.

```release-note
None
```


nunnatsa:
closed as #2299 was merged.

Issue #2295, opened 2023-03-23
Create KubeVirt rate limiting static profiles


jcanocan:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind enhancement

**What happened**:
In https://github.com/kubevirt/hyperconverged-cluster-operator/pull/2197 we have introduced a user-friendly mechanism to adjust the rate limiting `Query Per Second (QPS)` and `burst` values. Now, we want to create static profiles such as `highBurst` hiding the raw configuration values.

**What you expected to happen**:
It would be more easy for users to high scale their clusters allowing them to create a high number of VMs in a bulk if we have a `highBurst` profile which hides the raw numbers:

```yaml
apiVersion: hco.kubevirt.io/v1beta1
kind: hco
spec:
  tuningPolicy: highBurst
```
Values behind the `highBurst` profile are under discussion/testing in order to figure out possible downsides of increasing these values.

Issue tracked in: [CNV-14170](https://issues.redhat.com/browse/CNV-14170)
**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- HCO version (use `oc get csv -n kubevirt-hyperconverged`): 4.14.0
- Kubernetes version (use `kubectl version`): any
- Cloud provider or hardware configuration: any
- Install tools:
- Others:

tiraboschi:
@jcanocan is #2318 enough to consider this closed or are we targeting more profiles?

jcanocan:
> @jcanocan is #2318 enough to consider this closed or are we targeting more profiles?
As far as I know, there's not plans for
We are exploring if a dynamic profile that follows the kubelet rate limits configuration would be useful. You can find further details in: https://issues.redhat.com/browse/CNV-27873. Nevertheless, we can close this issue for now since this would be a more dynamic approach

jcanocan:
> @jcanocan is #2318 enough to consider this closed or are we targeting more profiles?

As far as I know, there's not plans for implementing more static profiles.

We are exploring if a dynamic profile that follows the kubelet rate limits configuration would be useful. You can find further details in: https://issues.redhat.com/browse/CNV-27873. Nevertheless, we can close this issue for now since this would be a more dynamic approach and I will open a new issue in the future.

Many thanks for the remainder.

PR #2318, opened 2023-04-14
Add static rate limit profile `highBurst`

jcanocan:
This PR adds the rate limit profile `highBurst`. This profile hides the configuration values `Query Per Second (QPS)` and `burst` of KubeVirt components: `API`, `Webhook`, `Controller` and `Handler`.

Currently, in order to set up these values, it is required to use the following path:

```
$ oc patch hco -n openshift-cnv kubevirt-hyperconverged --type=merge -p '{"metadata":{"annotations":{"kubevirt.kubevirt.io/jsonpatch":"[{\"op\":\"add\",\"path\":\"/spec/configuration\",\"value\":{\"apiConfiguration\":{\"restClient\":{\"rateLimiter\":{\"tokenBucketRateLimiter\":{\"burst\":200,\"qps\":100}}}},\"controllerConfiguration\":{\"restClient\":{\"rateLimiter\":{\"tokenBucketRateLimiter\":{\"burst\":400,\"qps\":200}}}},\"handlerConfiguration\":{\"restClient\":{\"rateLimiter\":{\"tokenBucketRateLimiter\":{\"burst\":400,\"qps\":200}}}},\"webhookConfiguration\":{\"restClient\":{\"rateLimiter\":{\"tokenBucketRateLimiter\":{\"burst\":400,\"qps\":200}}}}}}]"}}}'
```
or use the `tuningPolicy: annotation` profile. However, this profile has two main downsides: requires users to know the best suitable values for their use-case and it does not allow to configure the values of each component.
This new feature solves both problems while avoiding the use of the patch.

Implements: https://github.com/kubevirt/hyperconverged-cluster-operator/issues/2295
**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-14170
```

**Release note**:

```release-note
KubeVirt rate limit profile `highBurst`. This profile helps to create and maintain a high number of VMs in the same cluster.
```

### DIFF
--- a/api/v1beta1/hyperconverged_types.go
+++ b/api/v1beta1/hyperconverged_types.go
@@ -29,7 +29,10 @@ type HyperConvergedTuningPolicy string
 
 // HyperConvergedAnnotationTuningPolicy defines a static configuration of the kubevirt query per seconds (qps) and burst values
 // through annotation values.
-const HyperConvergedAnnotationTuningPolicy HyperConvergedTuningPolicy = "annotation"
+const (
+	HyperConvergedAnnotationTuningPolicy HyperConvergedTuningPolicy = "annotation"
+	HyperConvergedHighBurstProfile       HyperConvergedTuningPolicy = "highBurst"
+)
 
 // HyperConvergedSpec defines the desired state of HyperConverged
 // +k8s:openapi-gen=true
@@ -45,7 +48,7 @@ type HyperConvergedSpec struct {
 	// If TuningPolicy is not present the default kubevirt values are used.
 	// It can be set to `annotation` for fine-tuning the kubevirt queryPerSeconds (qps) and burst values.
 	// Qps and burst values are taken from the annotation hco.kubevirt.io/tuningPolicy
-	// +kubebuilder:validation:Enum=annotation
+	// +kubebuilder:validation:Enum=annotation;highBurst
 	// +optional
 	TuningPolicy HyperConvergedTuningPolicy `json:"tuningPolicy,omitempty"`
 
--- a/controllers/operands/kubevirt.go
+++ b/controllers/operands/kubevirt.go
@@ -100,6 +100,11 @@ const (
 	kvKubevirtSeccompProfile = "KubevirtSeccompProfile"
 )
 
+const (
+	highBurstProfileBurst = 400
+	highBurstProfileQPS   = 200
+)
+
 var (
 	hardCodeKvFgs = []string{
 		kvDataVolumesGate,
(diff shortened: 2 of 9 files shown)


jcanocan:
Note: Marked as WIP since we are still the exploring the most suitable configuration values.

jcanocan:
v2:

- Fix failing unit test.

jcanocan:
Hi @nunnatsa and @tiraboschi! Perf-Scale team just confirmed that these are the values that we want for this profile. The values have been tested and they have explored the potential drawbacks. As far as they have been able to test, they are safe. You can find more information [here](https://docs.google.com/document/d/14GfgBt_tCZZvxXaCIR8my88Oi40ku7ce7iABrpgehRs/edit#heading=h.66y4kqbj468a).

controllers/operands/kubevirt_test.go
```diff
@@ -2617,6 +2617,47 @@ Version: 1.2.3`)
...
+				Expect(kv.Spec.Configuration.WebhookConfiguration).To(BeNil())
+				Expect(kv.Spec.Configuration.HandlerConfiguration).To(BeNil())
+
+			})
+
+			It("Should create the fields and populate them using the HighBurst profile values", func() {
+				hco.Spec.TuningPolicy = hcov1beta1.HyperConvergedHighBurstProfile
+				kv, err := NewKubeVirt(hco)
+				expectedQps := float32(200)
+				expectedBurst := 400
```
nunnatsa, line 2646: nit: const?

api/v1beta1/hyperconverged_types.go
```diff
@@ -29,7 +29,10 @@ type HyperConvergedTuningPolicy string
 
 // HyperConvergedAnnotationTuningPolicy defines a static configuration of the kubevirt query per seconds (qps) and burst values
 // through annotation values.
-const HyperConvergedAnnotationTuningPolicy HyperConvergedTuningPolicy = "annotation"
+const (
+	HyperConvergedAnnotationTuningPolicy HyperConvergedTuningPolicy = "annotation"
+	HyperConvergedHighBurstProfile       HyperConvergedTuningPolicy = "HighBurst"
```
nunnatsa, line 34: let's be consistent here. If `annotation` starts with lower cap, can we do the same for `HighBurst`?

controllers/operands/kubevirt_test.go
jcanocan: Addressed.

controllers/operands/kubevirt.go
```diff
@@ -280,41 +280,62 @@ func NewKubeVirt(hc *hcov1beta1.HyperConverged, opts ...string) (*kubevirtcorev1
...
-			err = fmt.Errorf("tuning policy set but annotation not present or wrong")
-		}
+			},
+		}, nil
+
+	} else {
+		err = fmt.Errorf("tuning policy set but annotation not present or wrong")
 	}
+
 	return nil, err
```
nunnatsa, line 314: I other branches in this function, we just return the error.

I think it will be the same logic, but a bit less confusing, to remove the `else` and just return the error. In this case, we don't need the `err` variable.

controllers/operands/kubevirt.go
```diff
@@ -280,41 +280,62 @@ func NewKubeVirt(hc *hcov1beta1.HyperConverged, opts ...string) (*kubevirtcorev1
...
 }
 
+func getHcoHighBurstProfileTuningValues() (*kubevirtcorev1.ReloadableComponentConfiguration, error) {
+	return &kubevirtcorev1.ReloadableComponentConfiguration{
+		RestClient: &kubevirtcorev1.RESTClientConfiguration{
+			RateLimiter: &kubevirtcorev1.RateLimiter{
+				TokenBucketRateLimiter: &kubevirtcorev1.TokenBucketRateLimiter{
+					QPS:   200,
+					Burst: 400,
```
nunnatsa, line 323: Please consider using constants with meaningful names, instead of a magic numbers.

controllers/operands/kubevirt_test.go
```diff
@@ -2617,6 +2617,47 @@ Version: 1.2.3`)
...
+			It("Should be empty by default", func() {
+				kv, err := NewKubeVirt(hco)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(hco.Spec.TuningPolicy).To(BeEmpty())
+				Expect(kv.Spec.Configuration.APIConfiguration).To(BeNil())
+				Expect(kv.Spec.Configuration.ControllerConfiguration).To(BeNil())
+				Expect(kv.Spec.Configuration.WebhookConfiguration).To(BeNil())
+				Expect(kv.Spec.Configuration.HandlerConfiguration).To(BeNil())
+
+			})
```
nunnatsa, line 2640: We have the exact same test above. Please remove.

If you want, you can maybe wrap the annotation context and the highBurst context with additional context, and move this test to the common part. But I'm not sure it's not an overkill.

deploy/olm-catalog/community-kubevirt-hyperconverged/1.9.0/manifests/hco00.crd.yaml
```diff
@@ -2432,6 +2432,7 @@ spec:
                   Qps and burst values are taken from the annotation hco.kubevirt.io/tuningPolicy
                 enum:
                 - annotation
+                - HighBurst
```
nunnatsa, line 2435: this is 1.9.0 CRD. It should not be supported here. Also, please avoid modify this file manually.

nunnatsa (commented):
@jcanocan, This looks good in general. Added a few comments.

Thanks!

controllers/operands/kubevirt.go
jcanocan: Addressed.

controllers/operands/kubevirt_test.go
jcanocan: Done. Check if it's better now. I've also added a new test covering the situation where the `hco.kubevirt.io/tuningPolicy` annotation is set but the profile is set to `highBurst`.

hack/check_tuningPolicy.sh
```diff
@@ -45,4 +45,22 @@ for jpath in "${RATE_LIMITS[@]}"; do
      exit 1
   fi
   sleep 2
-done
\ No newline at end of file
+done
+
+echo "Check that the TuningPolicy highBurst is configuring the KV object as expected"
+
+./hack/retry.sh 10 3 "(${KUBECTL_BINARY} patch -n \"${INSTALLED_NAMESPACE}\" hco kubevirt-hyperconverged --type=json -p='[{"op": "replace", "path": "/spec/tuningPolicy", "value": "highBurst"}]')"
```
nunnatsa, line 52: I think it's failing because of the new check we added. The annotation is still there.

hack/check_tuningPolicy.sh
```diff
@@ -45,4 +45,25 @@ for jpath in "${RATE_LIMITS[@]}"; do
      exit 1
   fi
   sleep 2
-done
\ No newline at end of file
+done
+
+./hack/retry.sh 10 3 "(${KUBECTL_BINARY} patch -n \"${INSTALLED_NAMESPACE}\" hco kubevirt-hyperconverged --type=json -p='[{"op": "remove", "path": "/spec/tuningPolicy", "value": "annotation"}]')"
+./hack/retry.sh 10 3 "(${KUBECTL_BINARY} annotate -n \"${INSTALLED_NAMESPACE}\" hco kubevirt-hyperconverged hco.kubevirt.io/tuningPolicy-)"
```
nunnatsa, line 51: My only concern is that I think that it should be done as an atomic operation, but if it's working, then it's OK.

/lgtm

nunnatsa:
hco-e2e-upgrade-prev-index-sno-aws lane passed

/override ci/prow/hco-e2e-upgrade-prev-index-sno-azure

PR #2321, opened 2023-04-18
fix sanity test

nunnatsa:
Before this PR, running `make generate` in the CI, generates the files in the wrong place, and so the `git diff` missed the changes if the author forgot to regenerate before opening the PR. The result is that we could push API changes without regenerate, and the sanity test missed that.

This PR fixes this issue. Now the files are generated at the right place, and so the sanity test does check the auto generated files.

**Release note**:

```release-note
None
```


orenc1 (approved):
/approve
/thanks!

PR #2350, opened 2023-05-23
Check if deployment is nil to avoid panic

dharmit:
Fixes local runs when working with OCP cluster. We need it for HCO development.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
NONE
```

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/util/own_resources.go
+++ b/pkg/util/own_resources.go
@@ -59,14 +59,14 @@ func findOwnResources(ctx context.Context, cl client.Reader, logger logr.Logger)
 	}
 
 	operatorNs, err := GetOperatorNamespace(logger)
-	if err != nil {
+	if err != nil && !errors.Is(err, ErrRunLocal) {
 		logger.Error(err, "Can't get operator namespace")
 		return
 	}
 
 	or.deployment, err = getDeploymentFromPod(or.pod, cl, operatorNs, logger)
 	if err != nil {
-		logger.Error(err, "Can't the deployment")
+		logger.Error(err, "Can't get the deployment")
 		return
 	}
 	if GetClusterInfo().IsOpenshift() {
@@ -117,13 +117,22 @@ func getPod(ctx context.Context, c client.Reader, logger logr.Logger) (*corev1.P
 }
 
 func getDeploymentFromPod(pod *corev1.Pod, c client.Reader, operatorNs string, logger logr.Logger) (*appsv1.Deployment, error) {
+	deployment := &appsv1.Deployment{}
 	if pod == nil {
-		return nil, nil
+		// we might be running locally
+		if GetClusterInfo().IsRunningLocally() {
+			err := c.Get(context.TODO(), client.ObjectKey{
+				Name:      "hyperconverged-cluster-operator",
+				Namespace: "kubevirt-hyperconverged",
+			}, deployment)
+			if err != nil {
+				return nil, err
+			}
+			return deployment, nil
+		} else {
+			return nil, nil
+		}
 	}
-	//operatorNs, err := GetOperatorNamespace(logger)
-	//if err != nil {
-	//	return nil, err
-	//}
 	rsReference := metav1.GetControllerOf(pod)
 	if rsReference == nil || rsReference.Kind != "ReplicaSet" {
 		err := errors.New("failed getting HCO replicaSet reference")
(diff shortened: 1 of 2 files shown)


pkg/util/util.go
```diff
@@ -64,6 +64,11 @@ func IsRunModeLocal() bool {
 
 // GetOperatorNamespace returns the namespace the operator should be running in.
 var GetOperatorNamespace = func(logger logr.Logger) (string, error) {
+	ns, err := GetWatchNamespace()
```
tiraboschi, line 67: OperatorNamespace and WatchNamespace are two distinct values, please avoid this.

pkg/util/own_resources.go
```diff
@@ -160,6 +160,9 @@ func getDeploymentFromPod(pod *corev1.Pod, c client.Reader, operatorNs string, l
 }
 
 func getCSVFromDeployment(d *appsv1.Deployment, c client.Reader, operatorNs string, logger logr.Logger) (*csvv1alpha1.ClusterServiceVersion, error) {
+	if d == nil {
+		return nil, nil
+	}
```
dharmit, line 165: Added this check because of the panicking unit tests for `pkg/webhooks/validator/validator_test.go`.

pkg/util/util.go
```diff
@@ -64,6 +64,10 @@ func IsRunModeLocal() bool {
 
 // GetOperatorNamespace returns the namespace the operator should be running in.
 var GetOperatorNamespace = func(logger logr.Logger) (string, error) {
+	ns, found := os.LookupEnv(OperatorNamespaceEnv)
+	if found {
+		return ns, nil
+	}
```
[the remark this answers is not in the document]
dharmit: Yes, and that's what I would like to understand better, because if it's running locally (referring to the `main` branch), it returns empty value for namespace and an error. Now, I'm not sure what exactly the behaviour should be. I was inclined to do `os.LookupEnv(OperatorNamespaceEnv)` inside the `if IsRunModeLocal()` instead of returning an empty string and an error. What do you recommend, or what would be the right thing to 
do?

```go
if IsRunModeLocal() {
	ns, found := os.LookupEnv(OperatorNamespaceEnv)
	if found {
		return ns, nil
	}
	return "", ErrRunLocal
}
```

pkg/util/own_resources.go
dharmit: The way I came to make this change was by running the test through GoLand and stepping through the code. It was panicking at the `d.GetOwnerReferences()` in below: https://github.com/kubevirt/hyperconverged-cluster-operator/blob/a3e3030b3f9d0dc7000ff603131adb68f1e1c4f0/pkg/util/own_resources.go#L162-L164 

So I added a `nil` check. The call originates from https://github.com/kubevirt/hyperconverged-cluster-operator/blob/a3e3030b3f9d0dc7000ff603131adb68f1e1c4f0/pkg/webhooks/validator/validator_test.go#L1500

I don't claim to know this 100%, but it doesn't look like fake client is being used here.

Separately, a similar `nil` check exists for another function in the same file: https://github.com/kubevirt/hyperconverged-cluster-operator/blob/a3e3030b3f9d0dc7000ff603131adb68f1e1c4f0/pkg/util/own_resources.go#L119-L122

tiraboschi: > I don't claim to know this 100%, but it doesn't look like fake client is being used here.

please see:
https://github.com/kubevirt/hyperconverged-cluster-operator/blob/a3e3030b3f9d0dc7000ff603131adb68f1e1c4f0/pkg/webhooks/validator/validator_test.go#L1439

dharmit: I meant that the line in the validator_test.go file that's panicking doesn't seem to me like a direct call to the client. It's calling some util function which is working on a (nil) pointer to a Deployment

dharmit: Even though a fake client is being passed, the Deployment resource, which is the cause of the panic, doesn't look related to the fake client. But I could be wrong. :)

PR #2355, opened 2023-05-30
Fix local execution on OCP

tiraboschi:
When we deploy locally, HCO binary
is executed without a pod (by design) so we cannot
get a reference to its deployment and to
the owning CSV starting from the pod we are
running in.
On the other side, when deploying locally,
we know in advance the name of the deployment
object (with replica=0) and the expected
namespace from an environmental variable.
Let's get the deployment object from there.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
NONE
```

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ hack/cloud-init.sh
 hack/test_ssh*
 tools/release-notes/vendor/
 tools/release-notes/git/vendor/
+upgradePatches.json
--- a/hack/make_local.sh
+++ b/hack/make_local.sh
@@ -21,6 +21,8 @@ sed -i "s|-f https://raw.githubusercontent.com/kubevirt/hyperconverged-cluster-o
 
 chmod +x _local/deploy.sh
 
+ln -s assets/upgradePatches.json upgradePatches.json
+
 kubectl config set-context --current --namespace=${hco_namespace}
 _local/deploy.sh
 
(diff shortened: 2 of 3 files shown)


controllers/alerts/reconciler.go
```diff
@@ -37,9 +38,23 @@ type MonitoringReconciler struct {
...
+	if ci.IsRunningLocally() {
+		owner = metav1.OwnerReference{
+			APIVersion:         appsv1.SchemeGroupVersion.String(),
+			Kind:               "Deployment",
+			Name:               "local",
+			UID:                uuid.NewUUID(),
+			BlockOwnerDeletion: pointer.Bool(false),
+			Controller:         pointer.Bool(false),
+		}
+		namespace = "kubevirt-hyperconverged"
```
nunnatsa, line 52: Can we get the actual NS somehow? Not sure it will be "kubevirt-hyperconverged" in all cases.

tiraboschi: done

dharmit:
/approve

Approving since it works for me locally. Don't know if I can approve/lgtm a PR on this repo. :)

nunnatsa (approved):
/approve

PR #2369, opened 2023-06-05
Expose SCSI persistent reservation FG

tiraboschi:
Expose a Feature Gate to configure
SCSI persistent reservation on Kubevirt.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-11239
```

**Release note**:

```release-note
Expose SCSI persistent reservation FG
```

### DIFF
--- a/api/v1beta1/hyperconverged_types.go
+++ b/api/v1beta1/hyperconverged_types.go
@@ -356,6 +356,15 @@ type HyperConvergedFeatureGates struct {
 	// +kubebuilder:default=false
 	// +default=false
 	DisableMDevConfiguration *bool `json:"disableMDevConfiguration,omitempty"`
+
+	// Enable persistent reservation of a LUN through the SCSI Persistent Reserve commands on Kubevirt.
+	// In order to issue privileged SCSI ioctls, the VM requires activation of the persistent reservation flag.
+	// Once this feature gate is enabled, then the additional container with the qemu-pr-helper is deployed inside the virt-handler pod.
+	// Enabling (or removing) the feature gate causes the redeployment of the virt-handler pod.
+	// +optional
+	// +kubebuilder:default=false
+	// +default=false
+	PersistentReservation *bool `json:"persistentReservation,omitempty"`
 }
 
 // PermittedHostDevices holds information about devices allowed for passthrough
--- a/api/v1beta1/zz_generated.deepcopy.go
+++ b/api/v1beta1/zz_generated.deepcopy.go
@@ -245,6 +245,11 @@ func (in *HyperConvergedFeatureGates) DeepCopyInto(out *HyperConvergedFeatureGat
 		*out = new(bool)
 		**out = **in
 	}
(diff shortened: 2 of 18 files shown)


tiraboschi:
@alicefr FYI

alicefr:
it looks great! Many thanks @tiraboschi 
/lgtm

api/v1beta1/hyperconverged_types.go
```diff
@@ -356,6 +356,15 @@ type HyperConvergedFeatureGates struct {
...
 	DisableMDevConfiguration *bool `json:"disableMDevConfiguration,omitempty"`
+
+	// Enable persistent reservation of a LUN through the SCSI Persistent Reserve commands on Kubevirt.
+	// In order to issue privileged SCSI ioctls, the VM requires activation of the persistent reservation flag.
+	// Once this feature gate is enabled, then the additional container with the qemu-pr-helper is deployed inside the virt-handler pod.
+	// Enabling (or removing) the feature gate causes the redeployment of the virt-handler pod.
+	// +optional
+	// +kubebuilder:default=false
+	// +default=false
+	PersistentReservation *bool `json:"persistentReservation,omitempty"`
```
aglitke, line 367: @alicefr What are the risks for having this always enabled?  I'm always in favor of fewer knobs.  Is there a good reason that some people would like to have this off?  If so, should the default be on or off?

tiraboschi:
/hold
let's understand if we really have to expose it or if an opinionated value will be better

api/v1beta1/hyperconverged_types.go
tiraboschi: If we don't see real reasons why users should avoid it, I'd also prefer to have it as an opinionated value.
I started filing https://github.com/kubevirt/hyperconverged-cluster-operator/pull/2374 as an alternative to have this as an opinionated value.

alicefr:
> /hold let's understand if we really have to expose it or if an opinionated value will be better

IMO, we should. I would activate this feature only if customers really want to use it (e.g if they plan to use Windows Shared Cluster Filesystem or another setup that requires SCSI reservation)

@aglitke WDYT?

tiraboschi:
> IMO, we should. I would activate this feature only if users really want to use it (e.g if they plan to use Windows Shared Cluster Filesystem or another setup that requires SCSI reservation)

thanks
/unhold

tiraboschi:
okd-hco-e2e-image-index-gcp lane succeeded.
/override ci/prow/okd-hco-e2e-image-index-aws

tiraboschi:
hco-e2e-kv-smoke-gcp lane succeeded.

PR #2374, opened 2023-06-06
Set PersistentReservation as an opinionated FG

tiraboschi:
Set PersistentReservation as an opinionated FG
on Kubevirt.

It enables persistent reservation of a LUN through the SCSI Persistent Reserve commands on Kubevirt.
In order to issue privileged SCSI ioctls,
the VM requires activation of the persistent
reservation flag so that an additional container
with the qemu-pr-helper is deployed inside
the virt-handler pod.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-11239
```

**Release note**:

```release-note
Set PersistentReservation as an opinionated FG
```

### DIFF
--- a/controllers/hyperconverged/hyperconverged_controller_test.go
+++ b/controllers/hyperconverged/hyperconverged_controller_test.go
@@ -218,6 +218,7 @@ var _ = Describe("HyperconvergedController", func() {
 					"DisableCustomSELinuxPolicy",
 					"KubevirtSeccompProfile",
 					"HotplugNICs",
+					"PersistentReservation",
 				}
 				// Get the KV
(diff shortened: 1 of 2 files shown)


tiraboschi:
/hold
let's understand if we really have to expose it or if an opinionated value will be better

PR #2071, opened 2022-08-31
Introduce cluster level EvictionStrategy

acardace:
use LiveMigrate as a default value and a couple of unit tests.

Signed-off-by: Antonio Cardace <acardace@redhat.com>

**Reviewer Checklist**

> Reviewers are supposed to review the PR for every aspect below one by one. To check an item means the PR is either "OK" or "Not Applicable" in terms of that item. All items are supposed to be checked before merging a PR. 

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
- [ ] Troubleshooting Friendly
  
**Release note**:

```release-note
Introduce cluster level EvictionStrategy field.
```

PR #2376, opened 2023-06-08
Handle cluster level eviction strategy on SNO as a special case

tiraboschi:
With #2071 we introduced a knob to let the cluster
admin configure a default EvictionStrategy at
cluster level.
The default value there was `LiveMigrate` for all
the kinds of clusters but this could potentially
introduce side effects (alerts and potentially
also inability to upgrade the cluster without
manually recovery actions) on SNO clusters.

Let's set a different default value (`None`) for
cluster level eviction strategy when we detect
that we are on a cluster with a single worker node
where `LiveMigrate` is not an option.

The code will still honour explicit user choice if there.

Add unit and e2e tests.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2212289

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-29479
```

**Release note**:

```release-note
Handle cluster level eviction strategy on SNO as a special case
```

### DIFF
--- a/api/v1beta1/hyperconverged_types.go
+++ b/api/v1beta1/hyperconverged_types.go
@@ -175,8 +175,7 @@ type HyperConvergedSpec struct {
 
 	// EvictionStrategy defines at the cluster level if the VirtualMachineInstance should be
 	// migrated instead of shut-off in case of a node drain. If the VirtualMachineInstance specific
-	// field is set it overrides the cluster level one.
-	// +kubebuilder:default=LiveMigrate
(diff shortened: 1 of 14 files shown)


acardace (approved):

nunnatsa (approved):
/approve

PR #2378, opened 2023-06-11
Debug: understand e2e failure in verifyOperatorHealthMetricValue

nunnatsa:
**What this PR does / why we need it**:

**Reviewer Checklist**

> Reviewers are supposed to review the PR for every aspect below one by one. To check an item means the PR is either "OK" or "Not Applicable" in terms of that item. All items are supposed to be checked before merging a PR. 

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
- [ ] Troubleshooting Friendly

**Jira Ticket**:

```jira-ticket

```

**Release note**:

```release-note

```

PR #2379, opened 2023-06-11
Fix `verifyOperatorHealthMetricValue()`

assafad:

The current health metric test is flaky, and it might be due to the fact that the expected health metric value is based on the metric initial value, and the latter might change in the time we wait for alerts to start firing.
This PR fixes this flakiness, by enabling health metric assertion only if `initialOperatorHealthMetricValue <= alertImpact`, and then the expected value is not based on the initial value. This way, the test verifies the alert’s impact on the health metric value, only if the alert harms the current health.
It should be less flaky since the test is not  based on `initialOperatorHealthMetricValue` anymore.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Release note**:

```release-note
None
```

### DIFF
--- a/tests/func-tests/monitoring_test.go
+++ b/tests/func-tests/monitoring_test.go
@@ -143,13 +143,19 @@ func getAlertByName(alerts promApiv1.AlertsResult, alertName string) *promApiv1.
 }
 
 func verifyOperatorHealthMetricValue(promClient promApiv1.API, initialOperatorHealthMetricValue, alertImpact float64) {
-	Eventually(func() bool {
-		systemHealthMetricValue := getMetricValue(promClient, "kubevirt_hco_system_health_status")
-		operatorHealthMetricValue := getMetricValue(promClient, "kubevirt_hyperconverged_operator_health_status")
+	Eventually(func(g Gomega) {
+		if alertImpact >= initialOperatorHealthMetricValue {
+			systemHealthMetricValue := getMetricValue(promClient, "kubevirt_hco_system_health_status")
+			operatorHealthMetricValue := getMetricValue(promClient, "kubevirt_hyperconverged_operator_health_status")
+			expectedOperatorHealthMetricValue := math.Max(alertImpact, systemHealthMetricValue)
 
-		expectedOperatorHealthMetricValue := math.Max(alertImpact, math.Max(systemHealthMetricValue, initialOperatorHealthMetricValue))
-		return operatorHealthMetricValue == expectedOperatorHealthMetricValue
-	}, 60*time.Second, 5*time.Second).Should(BeTrue())
+			g.Expect(operatorHealthMetricValue).To(Equal(expectedOperatorHealthMetricValue),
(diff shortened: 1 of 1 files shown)


tests/func-tests/monitoring_test.go
```diff
@@ -144,11 +144,15 @@ func getAlertByName(alerts promApiv1.AlertsResult, alertName string) *promApiv1.
...
-		systemHealthMetricValue := getMetricValue(promClient, "kubevirt_hco_system_health_status")
-		operatorHealthMetricValue := getMetricValue(promClient, "kubevirt_hyperconverged_operator_health_status")
+		if initialOperatorHealthMetricValue <= alertImpact {
+			systemHealthMetricValue := getMetricValue(promClient, "kubevirt_hco_system_health_status")
+			operatorHealthMetricValue := getMetricValue(promClient, "kubevirt_hyperconverged_operator_health_status")
+			expectedOperatorHealthMetricValue := math.Max(alertImpact, systemHealthMetricValue)
 
-		expectedOperatorHealthMetricValue := math.Max(alertImpact, math.Max(systemHealthMetricValue, initialOperatorHealthMetricValue))
-		return operatorHealthMetricValue == expectedOperatorHealthMetricValue
+			return operatorHealthMetricValue == expectedOperatorHealthMetricValue
```
nunnatsa, line 152: Please use another aay to compare these values, in a way that will produce meaningful error message  in case of failure.

You can follow the implementation in #2378

tests/func-tests/monitoring_test.go
```diff
@@ -143,13 +143,19 @@ func getAlertByName(alerts promApiv1.AlertsResult, alertName string) *promApiv1.
...
-		operatorHealthMetricValue := getMetricValue(promClient, "kubevirt_hyperconverged_operator_health_status")
+	Eventually(func(g Gomega) {
+		if alertImpact >= initialOperatorHealthMetricValue {
+			systemHealthMetricValue := getMetricValue(promClient, "kubevirt_hco_system_health_status")
+			operatorHealthMetricValue := getMetricValue(promClient, "kubevirt_hyperconverged_operator_health_status")
+			expectedOperatorHealthMetricValue := math.Max(alertImpact, systemHealthMetricValue)
+
+			g.Expect(operatorHealthMetricValue).To(Equal(expectedOperatorHealthMetricValue),
+				"kubevirt_hyperconverged_operator_health_status value is %d, but its expected value is %d, "+
+					"while kubevirt_hco_system_health_status value is %d.",
```
nunnatsa, line 154: please replace `%d` with `%f` - the vars are float64, and in case of error, it will trigger a format error.

assafad: done, thanks

PR #2399, opened 2023-06-25
Preparation for running functional test after upgrade

nunnatsa:
Currently, many e2e tests are implemented by bash files. We want to move them to golang. But some of these tests are also running on the upgrade test lanes, so this PR removes these bash tests from the upgrade test, and add the full e2e test instead. That will allow us to move most of the bash tests to golang.

This PR remove several tests from the upgrade lanes. These test will be called from the lane configurations in openshift/release, as implemented in https://github.com/openshift/release/pull/40643.

**Jira Ticket**:

```jira-ticket
None
```

**Release note**:

```release-note
None
```

### DIFF
--- a/hack/run-tests-in-container.sh
+++ b/hack/run-tests-in-container.sh
@@ -119,3 +119,6 @@ echo "Exiting... Exit code: $exitCode"
 # exit non-zero if exit code of functest is not zero
 [[ "${exitCode}" == "0" ]]
 
+# Brutally delete HCO removing the namespace where it's running"
+source hack/test_delete_ns.sh
+test_delete_ns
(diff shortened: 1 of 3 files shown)


nunnatsa:
/test pull-hyperconverged-cluster-operator-e2e-k8s-1.27

No code nor unit tests were changed
/override coverage/coveralls

tiraboschi:
>  and add the full e2e test instead. 

Is this going to be addressed in a different PR?
Can we do both the steps atomically?

nunnatsa:
The plan is to add the e2e step in the lane configurations, after the upgrade step. So it's in another repo.

orenc1 (approved):
/approve

PR #2400, opened 2023-06-25
cnao: Deploy linux bridge only for non Openshift

oshoval:
We used to have two bridge CNIs,
bridge which is the official, and cnv-bridge which we maintained for additional features.
Since there is only one now, in case it is already exists, we don't need to deploy it (as long as there is a softlink
between bridge and cnv-bridge, for backward compability). 
For Openshift it is already deployed, so skip deployment in such case.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket

```

**Release note**:

```release-note
None
```

### DIFF
--- a/controllers/operands/networkAddons.go
+++ b/controllers/operands/networkAddons.go
@@ -140,8 +140,11 @@ func NewNetworkAddons(hc *hcov1beta1.HyperConverged, opts ...string) (*networkad
 
 	cnaoSpec := networkaddonsshared.NetworkAddonsConfigSpec{
 		Multus:      &networkaddonsshared.Multus{},
-		LinuxBridge: &networkaddonsshared.LinuxBridge{},
 		KubeMacPool: &networkaddonsshared.KubeMacPool{},
+	}
+
+	if !hcoutil.GetClusterInfo().IsOpenshift() {
+		cnaoSpec.LinuxBridge = &networkaddonsshared.LinuxBridge{}
 	}
 
 	nameServerIP, err := getKSDNameServerIP(hc.Spec.KubeSecondaryDNSNameServerIP)
--- a/controllers/operands/networkAddons_test.go
+++ b/controllers/operands/networkAddons_test.go
@@ -595,6 +595,12 @@ var _ = Describe("CNA Operand", func() {
 					expectedBaseDomain: "",
 				}),
 			)
+
+			It("Should deploy linux-bridge", func() {
+				existingCNAO, err := NewNetworkAddons(hco)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(existingCNAO.Spec.LinuxBridge).ToNot(BeNil(), "Linux Bridge spec should be added")
+			})
 		})
 
 		Context("With Openshift Mock", func() {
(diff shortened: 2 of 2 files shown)


nunnatsa:
Undid lgtm 

@oshoval - could you please add a check in (even an existing) unit test?

oshoval:
Thanks
Added tests

controllers/operands/networkAddons_test.go
```diff
@@ -595,6 +595,26 @@ var _ = Describe("CNA Operand", func() {
...
+
+				foundCNAO := &networkaddonsv1.NetworkAddonsConfig{}
+				Expect(
+					cl.Get(context.TODO(),
+						types.NamespacedName{Name: existingCNAO.Name, Namespace: existingCNAO.Namespace},
+						foundCNAO),
+				).To(Succeed())
+
+				Expect(foundCNAO.Spec.LinuxBridge).ToNot(BeNil(), "Linux Bridge spec should be added")
+			})
```
nunnatsa, line 617: Not sure this is not a bit overkill.

What about just call `NewNetworkAddons()`, and check if the `LinuxBridge` field is there?

oshoval: > Not sure this is not a bit overkill. I don't we need to initiate a reconcile loop here.
> 
> What do you think about just call `NewNetworkAddons()`, and check if the `LinuxBridge` field is there?

Indeed, done

> If we also want to check the new field in reconcile, we can remove the field before placing it in the fake client, then check that it restored by the handler. But I think this is a new test.

Imo the current is enough, wdyt?

oshoval:
Addressed comments

controllers/operands/networkAddons_test.go
nunnatsa: > Imo the current is enough, wdyt?

:+1:

oshoval:
Thanks

controllers/operands/networkAddons.go
```diff
@@ -140,10 +140,13 @@ func NewNetworkAddons(hc *hcov1beta1.HyperConverged, opts ...string) (*networkad
 
 	cnaoSpec := networkaddonsshared.NetworkAddonsConfigSpec{
 		Multus:      &networkaddonsshared.Multus{},
-		LinuxBridge: &networkaddonsshared.LinuxBridge{},
 		KubeMacPool: &networkaddonsshared.KubeMacPool{},
 	}
 
+	if !hcoutil.GetClusterInfo().IsOpenshift() {
+		cnaoSpec.LinuxBridge = &networkaddonsshared.LinuxBridge{}
```
phoracek, line 147: We still need to deploy this, even on OCP. Otherwise we never create the needed symlinks downstream, no?

phoracek (changes requested):

controllers/operands/networkAddons.go
oshoval: Right, will update as we talked offline

PR #2462, opened 2023-08-07
Force runbooks names to include alerts names

assafad:
This PR adds a check to the alerts rules test, which verifies that the alerts `runbook_url` annotations contain the alerts names.

**Which issue(s) this PR fixes**: https://issues.redhat.com/browse/CNV-30798

**Release note**:

```release-note
None
```

### DIFF
--- a/tests/func-tests/monitoring_test.go
+++ b/tests/func-tests/monitoring_test.go
@@ -68,8 +68,8 @@ var _ = Describe("[crit:high][vendor:cnv-qe@redhat.com][level:system]Monitoring"
 				if rule.Alert != "" {
 					Expect(rule.Annotations).To(HaveKeyWithValue("summary", Not(BeEmpty())),
 						"%s summary is missing or empty", rule.Alert)
-					Expect(rule.Annotations).To(HaveKeyWithValue("runbook_url", Not(BeEmpty())),
-						"%s runbook_url is missing or empty", rule.Alert)
+					Expect(rule.Annotations).To(HaveKeyWithValue("runbook_url", ContainSubstring(rule.Alert)),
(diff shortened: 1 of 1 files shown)


assafad:
@nunnatsa can you please take a look?

tests/func-tests/monitoring_test.go
```diff
@@ -68,8 +68,8 @@ var _ = Describe("[crit:high][vendor:cnv-qe@redhat.com][level:system]Monitoring"
 				if rule.Alert != "" {
 					Expect(rule.Annotations).To(HaveKeyWithValue("summary", Not(BeEmpty())),
 						"%s summary is missing or empty", rule.Alert)
-					Expect(rule.Annotations).To(HaveKeyWithValue("runbook_url", Not(BeEmpty())),
-						"%s runbook_url is missing or empty", rule.Alert)
+					Expect(rule.Annotations).To(HaveKeyWithValue("runbook_url", ContainSubstring(rule.Alert)),
+						"%s runbook_url is missing or doesn't contain alert name", rule.Alert)
```
sradco, line 72: @assafad  What if a runbook_url will include the alert name but with a prefix or a suffix?
I think you need to remove the prefix and then check that the rest is equal to the alert name.

assafad: I will submit a fix PR, but beware that we also check that the runbook URL is accessible, and it must be a suffix for that. And we are already add the `runbook_url` automatically:
https://github.com/kubevirt/hyperconverged-cluster-operator/blob/928efeddcff4b2c6481e7a3b0f843a7038c440e4/controllers/alerts/alerts.go#L149

assafad: (but not across all repos)

PR #2463, opened 2023-08-07
Add support for automatic CPU limits

jean-edouard:
See https://github.com/kubevirt/kubevirt/pull/9768

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
NONE
```

**Release note**:

```release-note
New CR option to enable auto CPU limits for virt-launcher on some namespaces
```

### DIFF
--- a/api/v1beta1/hyperconverged_types.go
+++ b/api/v1beta1/hyperconverged_types.go
@@ -91,6 +91,7 @@ type HyperConvergedSpec struct {
 
 	// ResourceRequirements describes the resource requirements for the operand workloads.
 	// +kubebuilder:default={"vmiCPUAllocationRatio": 10}
+	// +kubebuilder:validation:XValidation:rule="!has(self.vmiCPUAllocationRatio) || self.vmiCPUAllocationRatio != 1",message="Automatic CPU limits are incompatible with a VMI CPU allocation ratio of 1"
 	// +optional
 	ResourceRequirements *OperandResourceRequirements `json:"resourceRequirements,omitempty"`
 
@@ -503,6 +504,13 @@ type OperandResourceRequirements struct {
 	// +default=10
 	// +optional
 	VmiCPUAllocationRatio *int `json:"vmiCPUAllocationRatio,omitempty"`
+
+	// When set, AutoCPULimitNamespaceLabelSelector will set a CPU limit on virt-launcher for VMIs running inside
+	// namespaces that match the label selector.
+	// The CPU limit will equal the number of requested vCPUs.
+	// This setting does not apply to VMIs with dedicated CPUs.
+	// +optional
+	AutoCPULimitNamespaceLabelSelector *metav1.LabelSelector `json:"autoCPULimitNamespaceLabelSelector,omitempty"`
 }
 
 // HyperConvergedObsoleteCPUs allows avoiding scheduling of VMs for obsolete CPU models
--- a/controllers/operands/kubevirt.go
+++ b/controllers/operands/kubevirt.go
@@ -458,6 +458,10 @@ func getKVConfig(hc *hcov1beta1.HyperConverged) (*kubevirtcorev1.KubeVirtConfigu
 		config.VirtualMachineOptions = &kubevirtcorev1.VirtualMachineOptions{DisableFreePageReporting: &kubevirtcorev1.DisableFreePageReporting{}}
 	}
 
+	if hc.Spec.ResourceRequirements != nil {
+		config.AutoCPULimitNamespaceLabelSelector = hc.Spec.ResourceRequirements.AutoCPULimitNamespaceLabelSelector.DeepCopy()
+	}
+
 	return config, nil
 }
 
--- a/tests/func-tests/defaults_test.go
+++ b/tests/func-tests/defaults_test.go
@@ -29,21 +29,15 @@ var _ = Describe("Check Default values", Label("defaults"), Serial, func() {
 
 	BeforeEach(func() {
 		var err error
+
 		cli, err = kubecli.GetKubevirtClient()
 		Expect(cli).ToNot(BeNil())
 		Expect(err).ToNot(HaveOccurred())
 
 		ctx = context.Background()
-	})
 
-	restoreDefaults := func(cli kubecli.KubevirtClient) {
-		Eventually(tests.PatchHCO).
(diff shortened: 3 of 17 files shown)


tests/func-tests/defaults_test.go
```diff
@@ -249,4 +251,32 @@ var _ = Describe("Check Default values", Label("defaults"), Serial, func() {
...
+				VmiCPUAllocationRatio: allocationRatio,
+				AutoCPULimitNamespaceLabelSelector: &metav1.LabelSelector{
+					MatchLabels: map[string]string{"someLabel": "true"},
+				},
+			}
+			Eventually(func() error {
+				var err error
+				hc, err = tests.UpdateHCO(ctx, cli, hc)
+				if err != nil {
+					Expect(err.Error()).NotTo(ContainSubstring("the object has been modified; please apply your changes to the latest version and try again"))
```
nunnatsa, line 272: What is this check? Why do we need it?

Also, failed Expect will break the eventually loop.

tests/vendor/github.com/kubevirt/hyperconverged-cluster-operator/api/v1beta1/hyperconverged_types.go
```diff
@@ -91,6 +91,7 @@ type HyperConvergedSpec struct {
 
 	// ResourceRequirements describes the resource requirements for the operand workloads.
 	// +kubebuilder:default={"vmiCPUAllocationRatio": 10}
+	// +kubebuilder:validation:XValidation:rule="!has(self.vmiCPUAllocationRatio) || self.vmiCPUAllocationRatio != 1",message="Automatic CPU limits are incompatible with a VMI CPU allocation ratio of 1"
```
nunnatsa, line 94: Please add an e2e test for this rule

api/v1beta1/hyperconverged_types.go
```diff
@@ -91,6 +91,7 @@ type HyperConvergedSpec struct {
 
 	// ResourceRequirements describes the resource requirements for the operand workloads.
 	// +kubebuilder:default={"vmiCPUAllocationRatio": 10}
+	// +kubebuilder:validation:XValidation:rule="!has(self.vmiCPUAllocationRatio) || self.vmiCPUAllocationRatio != 1",message="Automatic CPU limits are incompatible with a VMI CPU allocation ratio of 1"
```
nunnatsa, line 94: Please add an e2e test for this rule

controllers/operands/kubevirt.go
```diff
@@ -458,6 +458,10 @@ func getKVConfig(hc *hcov1beta1.HyperConverged) (*kubevirtcorev1.KubeVirtConfigu
 		config.VirtualMachineOptions = &kubevirtcorev1.VirtualMachineOptions{DisableFreePageReporting: &kubevirtcorev1.DisableFreePageReporting{}}
 	}
 
+	if hc.Spec.ResourceRequirements != nil {
+		config.AutoCPULimitNamespaceLabelSelector = hc.Spec.ResourceRequirements.AutoCPULimitNamespaceLabelSelector
```
nunnatsa, line 462: Maybe use deep copy here, just for case?

nunnatsa (commented):
Please add human readable documentation in docs/cluster-configuration.md

tests/func-tests/defaults_test.go
jean-edouard: This is to ensure that we don't count update errors as the error we expect in the failure case.
Are you sure it breaks the loop? I thought it "continue"d it. Either way, I can change that to a `Consistently()` call for the failure case, that probably makes more sense.

api/v1beta1/hyperconverged_types.go
jean-edouard: That's what I wrote the e2e test for... Did I get it wrong?

tests/func-tests/defaults_test.go
nunnatsa: I think we don't care about the error. We need to wait  for no error. See other tests case in the same file

jean-edouard: Well I do want the update to fail when the ratio is 1. Anyway, will improve, thanks for the review!

nunnatsa: OK, after reading again, I understand your point. 

However:
1. I would move this test to another file. This test file is all about default values of the HyperConverged CR. I find it a bit confusing to check the CRD rule here. I'm not sure there is an existing better place to put it, so please consider to move this test to a new file.
2. As said above, using `Expect` within `Eventually`, break the loop in case of failure. There are several ways to address it. for example, using this pattern:
    ```go
    Eventually(func(g Gomega){
        ...
        g.Expect(...).To(...)
        ...
    }).To(...)
    ```
3. the result of `tests.UpdateHCO(ctx, cli, hc)` is assigned back to the `hc` variable, overriding its value (probably to nil, in case of error). So if we do get a conflict error and we'll want to retry, the retry will fail. I would use a new variable for that, defined outside of the Eventually loop, then we could check if the value was set correctly. We can also use `_` - that fine too. 
4. optional: maybe, instead of `Not(Succeed())`, we can use the `MatchError()` matcher? not sure if it will work. But if it will, we won't need the conflict error logic at all.
5. Please remove the `one` and `two` variables, and use `pointer.Int(1)` and `pointer.Int(2)` instead.

controllers/operands/kubevirt.go
```diff
@@ -458,6 +458,10 @@ func getKVConfig(hc *hcov1beta1.HyperConverged) (*kubevirtcorev1.KubeVirtConfigu
...
+	}
+
```
nunnatsa, line 464: Please add a unit test for this,  in controllers/operands/kubevirt_test.go

tests/func-tests/defaults_test.go
jean-edouard: Thanks a lot, all your points make sense, will update later today!

jean-edouard:
@nunnatsa I believe all your comments have been addressed, thanks again for the reviews, PTAL!

nunnatsa (approved):
/approve
/cherry-pick release-1.10

PR #2466, opened 2023-08-09
Fix runbook URL test

assafad:
Fix `runbook_url` test to expect it to have the alert name as a suffix. 

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

**Release note**:

```release-note
None
```

### DIFF
--- a/tests/func-tests/monitoring_test.go
+++ b/tests/func-tests/monitoring_test.go
@@ -68,8 +68,10 @@ var _ = Describe("[crit:high][vendor:cnv-qe@redhat.com][level:system]Monitoring"
 				if rule.Alert != "" {
 					Expect(rule.Annotations).To(HaveKeyWithValue("summary", Not(BeEmpty())),
 						"%s summary is missing or empty", rule.Alert)
-					Expect(rule.Annotations).To(HaveKeyWithValue("runbook_url", ContainSubstring(rule.Alert)),
-						"%s runbook_url is missing or doesn't contain alert name", rule.Alert)
+					Expect(rule.Annotations).To(HaveKey("runbook_url"),
+						"%s runbook_url is missing", rule.Alert)
+					Expect(rule.Annotations).To(HaveKeyWithValue("runbook_url", HaveSuffix(rule.Alert)),
+						"%s runbook_url is not equal to alert name", rule.Alert)
 					checkRunbookURLAvailability(rule)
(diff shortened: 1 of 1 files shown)


assafad:
/hold
waiting for @sradco to review as well

tests/func-tests/monitoring_test.go
```diff
@@ -68,8 +68,8 @@ var _ = Describe("[crit:high][vendor:cnv-qe@redhat.com][level:system]Monitoring"
 				if rule.Alert != "" {
 					Expect(rule.Annotations).To(HaveKeyWithValue("summary", Not(BeEmpty())),
 						"%s summary is missing or empty", rule.Alert)
-					Expect(rule.Annotations).To(HaveKeyWithValue("runbook_url", ContainSubstring(rule.Alert)),
-						"%s runbook_url is missing or doesn't contain alert name", rule.Alert)
+					Expect(rule.Annotations).To(HaveKeyWithValue("runbook_url", HaveSuffix(rule.Alert)),
+						"%s runbook_url is missing or runbook is not equal to alert name", rule.Alert)
```
sradco, line 72: @assafad I think it would need 2 tests here. One for if the runbook_url exists and one if its name is equal to the alert name. Having them together makes fixing the issue harder.

assafad: This test checks them both. Can you explain what would be harder? I think it's pretty straightforward to check if an alert is missing the `runbook_url` field, or it's named incorrectly. We also do it for all the rest of the fields, for example:
```
Expect(rule.Annotations).To(HaveKeyWithValue("summary", Not(BeEmpty())),
"%s summary is missing or empty", rule.Alert)
``` 
It can fail either if `summary` field doesn't exist, or if it's exists and empty.

assafad: Updated. Now the workflow is:
- check `runbook_url` exists (and not empty)
- check its suffix
- check the URL is accessible
@sradco Please confirm this sounds good and I'll update the other PRs as well.

PR #2494, opened 2023-08-29
For tests, use imagePullPolicy=Always

nunnatsa:
In order to redeploy HCO with `make cluster-sync`, change the pull policy for HCO images to Always.

**Jira Ticket**:

```jira-ticket
None
```

**Release note**:

```release-note
None
```


machadovilaca (commented):
Maybe this could be a `--pull-policy=` flag, the same as when generating CSVs

nunnatsa:
> Maybe this could be a `--pull-policy=` flag, the same as when generating CSVs

I think it's just for deployment in test environment, so I think it's safe to set all HCO images policy to `Always`.

I don't think anyone calls the manifest-templator tool outside of the make-manifests anyway, so will add the new flag there - and it's the only usage - I think there is no point

nunnatsa:
no code changes
/override coverage/coveralls

PR #2515, opened 2023-09-19
Fix a possible bug setting finalizers

tiraboschi:
The code that was adding the finalizer
was not really writing it
if the finalizer was the single change
in the reconciliation loop.
As soon as one of the following reconciliation loops
is touching also something else, the finalizer is really set.

So this is basically not affecting real
world executions but it could potentially
hit unit tests if we verify after a single reconciliation loop.

Only a safer fake client like the
controller-runtime v0.16.2 one is showing this.

Let's also remove obsolete code to remove an old finalizer name
since we can safely assume that upgrades to the latest N
released had enough time to amend it.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
NONE
```

**Release note**:

```release-note
NONE
```

### DIFF
--- a/controllers/hyperconverged/hyperconverged_controller.go
+++ b/controllers/hyperconverged/hyperconverged_controller.go
@@ -61,8 +61,7 @@ var (
 const (
 	// We cannot set owner reference of cluster-wide resources to namespaced HyperConverged object. Therefore,
 	// use finalizers to manage the cleanup.
-	FinalizerName    = "kubevirt.io/hyperconverged"
-	badFinalizerName = "hyperconvergeds.hco.kubevirt.io"
+	FinalizerName = "kubevirt.io/hyperconverged"
 
 	// OpenshiftNamespace is for resources that belong in the openshift namespace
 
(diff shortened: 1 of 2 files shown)


controllers/hyperconverged/hyperconverged_controller.go
```diff
@@ -1529,9 +1529,8 @@ func init() {
 }
 
 func checkFinalizers(req *common.HcoRequest) bool {
-	finDropped := false
-
 	if hcoutil.ContainsString(req.Instance.ObjectMeta.Finalizers, badFinalizerName) {
+		finDropped := false
```
nunnatsa, line 1533: This is a very old code, dealing with an old finalizer name, Do we still need to check it?

tiraboschi: yes, it's the time for that!

nunnatsa (approved):

tiraboschi:
hco-e2e-operator-sdk-gcp lane succeeded.
/override ci/prow/hco-e2e-operator-sdk-azure

PR #2402, opened 2023-06-27
Deprecate mediatedDevicesTypes in favor of mediatedDeviceTypes

tiraboschi:
**What this PR does / why we need it**:
With github.com/kubevirt/kubevirt/pull/8525
mediatedDevicesTypes got deprecated in favor of mediatedDeviceTypes on kubevirt/kubevirt.
Let's do the same on the HCO side for consistency.

The mutating webhook will take care of the migration to the new API and the validating webhook will
prevent inconsistencies between the two APIs.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2054863

**Reviewer Checklist**

> Reviewers are supposed to review the PR for every aspect below one by one. To check an item means the PR is either "OK" or "Not Applicable" in terms of that item. All items are supposed to be checked before merging a PR. 

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
- [ ] Troubleshooting Friendly

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-16422
```

**Release note**:

```release-note
CR option mediatedDevicesTypes is deprecated in favor of mediatedDeviceTypes
```

PR #2534, opened 2023-10-02
Do not require nodemediatedDeviceTypes to prevent broken upgrades

tiraboschi:
With #2402 we deprecated MediatedDevicesTypes and
NodeMediatedDeviceTypes.MediatedDevicesTypes for
MediatedDeviceTypes and NodeMediatedDeviceTypes.MediatedDeviceTypes. We also introduced a mutating webhook to handle the conversion. Unfortunately make the two new names mandatory is breaking the validation of the new version of the CRD on OLM upgrades since only the previous names were there with the old CRD.

We already have func test covering this area,
but they are fooled by the mutating webhook that
instead is not involved into the CRD validation process on the OLM side.
On the other side, add an additional upgrade lane just for this is probably too expensive.

Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=2241327

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-33570
```

**Release note**:

```release-note
Do not require nodemediatedDeviceTypes to prevent broken upgrades
```

### DIFF
--- a/api/v1beta1/hyperconverged_types.go
+++ b/api/v1beta1/hyperconverged_types.go
@@ -458,7 +458,9 @@ type MediatedHostDevice struct {
 
 // MediatedDevicesConfiguration holds information about MDEV types to be defined, if available
 // +k8s:openapi-gen=true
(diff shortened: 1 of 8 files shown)


nunnatsa (approved):
/approve

PR #2596, opened 2023-11-06
Add DeployCommonInstanceTypesByVirtOperator feature gate

akrejcir:
Added `DeployCommonInstanceTypesByVirtOperator` feature gate to the HCO CR.
This feature gate is used to specify that common instance types should be deployed by `virt-operator` instead of SSP operator.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-31875
```

**Release note**:

```release-note
Added new feature gate: DeployCommonInstanceTypesByVirtOperator
```


akrejcir:
> Is that something we want to let the user to decide? Look like something that shoild be controlled by hco; e.g. if it's an openshift set the fg to false; else, set it to true.

Right, this is not relevant to users. I will post a different change.

/hold

akrejcir:
@nunnatsa , do you think it is a bad design, to add this feature gate even if it is not useful for most users?
It could be used to enable the feature sooner, before it's enabled by default, or it can be sued to disable it in later versions if there is an issue.

nunnatsa:
> @nunnatsa , do you think it is a bad design, to add this feature gate even if it is not useful for most users? It could be used to enable the feature sooner, before it's enabled by default, or it can be sued to disable it in later versions if there is an issue.

We're trying to keep the HCO API as minimal as possible, and do what ever we can with the information we have. If we don't have a choice, we can expand the API. But that should be the last choice.

akrejcir:
Ok, I will close this PR and post a smaller one to only set the FG on SSP, without API change.

/close

akrejcir:
I've opened PR: https://github.com/kubevirt/hyperconverged-cluster-operator/pull/2600

PR #2684, opened 2023-12-18
expose VM live update configuration bits in the HCO API

enp0s3:

Manage Kubevirt VM live update features with HCO operator

**Release note**:
```release-note
enable the VMLiveUpdateFeatures FG, expose the LiveUpdateConfiguration in HCO API
```

### DIFF
--- a/api/v1beta1/hyperconverged_types.go
+++ b/api/v1beta1/hyperconverged_types.go
@@ -227,6 +227,10 @@ type HyperConvergedSpec struct {
 	// Those bindings can be used when defining virtual machine interfaces.
 	// +optional
 	NetworkBinding map[string]v1.InterfaceBindingPlugin `json:"networkBinding,omitempty"`
+
+	// Configuration of the VM live update features
+	// +optional
+	VMLiveUpdateConfiguration *v1.LiveUpdateConfiguration `json:"vmLiveUpdateConfiguration,omitempty"`
 }
 
 // CertRotateConfigCA contains the tunables for TLS certificates.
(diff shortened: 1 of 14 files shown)


api/v1beta1/hyperconverged_types.go
```diff
@@ -227,6 +227,10 @@ type HyperConvergedSpec struct {
 	// Those bindings can be used when defining virtual machine interfaces.
 	// +optional
 	NetworkBinding map[string]v1.InterfaceBindingPlugin `json:"networkBinding,omitempty"`
+
+	// Configuration of the VM live update features
+	// +optional
+	VMLiveUpdateConfiguration *v1.LiveUpdateConfiguration `json:"vmLiveUpdateConfiguration,omitempty"`
```
tiraboschi, line 233: I know that we did something like this also in the past (`KSMConfiguration` for instance), but I'm now thinking if blindly exposing a struct from another project as part of our APIs is  a good idea.
One one side this will implicitly handle any API introduced on the kubevirt/kubevirt repo for `LiveUpdateConfiguration`, on the other side it will prevent any kind of control on that specific stanza of the APIs from HCO side.

nunnatsa: I agree. Mostly because it create strong coupling to KV.

enp0s3: @nunnatsa @tiraboschi Thank you for taking a look at it! how would you suggest to perform the API conversion?

tiraboschi: Have a local struct like https://github.com/kubevirt/hyperconverged-cluster-operator/blob/main/api/v1beta1/hyperconverged_types.go#L296

and handle the conversion with something like:
https://github.com/kubevirt/hyperconverged-cluster-operator/blob/744c5d5889abb99d7e369ee3a7617124937d61bb/controllers/operands/kubevirt.go#L584

enp0s3:
/close
Closing this as this feature will be deprecated in favor to VM rollout strategy

PR #2685, opened 2023-12-18
Add AlignCPUs behavior by default on kubevirt CR

RamLavi:
The `AlignCPUs` feature gate and `alpha.kubevirt.io/EmulatorThreadCompleteToEvenParity` annotation is proposed [0] and introduced [1][2][3][4] on Kubevirt in order to allow support in SMT enabled nodes while using IsolateEmulatorThread on the VMI [5].
The configuration is set cluster-wide, using the kubevirt CR (annotation + kubevirt feature gate).
Introducing the feature-gate and annotation to HCO in order to allow the behavior to be set by deafult.

Important note: Setting this behavior to default introduces little to no concern, as it only adds support in VMIs in a certain scenario that was blocked before this fix.

[0] https://github.com/kubevirt/community/pull/247
[1] https://github.com/kubevirt/kubevirt/pull/10593
[2] https://github.com/kubevirt/kubevirt/pull/10783
[3] https://github.com/kubevirt/kubevirt/pull/10839
[4] https://github.com/kubevirt/kubevirt/pull/10872
[5] https://bugzilla.redhat.com/show_bug.cgi?id=2228103

**Reviewer Checklist**

- [x] PR Message
- [x] Commit Messages
- [x] How to test
- [x] Unit Tests
- [x] Functional Tests
- [x] User Documentation
- [x] Developer Documentation
- [x] Upgrade Scenario
- [x] Uninstallation Scenario
- [x] Backward Compatibility

**Jira Ticket**:

```jira-ticket

```

**Release note**:
```release-note
Setting AlignCPUs feature-gate and EmulatorThreadCompleteToEvenParity annotation by default on kubevirt CR
```


RamLavi:
/hold

Until https://github.com/kubevirt/kubevirt/pull/10872 is merged and the entire solution is backported to the kubevirt release-1.1 stable branch.

api/v1beta1/hyperconverged_types.go
```diff
@@ -450,6 +450,17 @@ type HyperConvergedFeatureGates struct {
...
+	// Enable KubeVirt emulator thread to assign two extra CPUs if needed to complete even parity.
+	// If IsolateEmulatorThread is set on the VMI, kubevirt assigns extra CPUs to the emulator thread.
+	// By default, KubeVirt only assigns one emulator thread CPU to the virt-launcher pod.
+	// When this feature gate is enabled, and the VMI enables IsolateEmulatorThread, KubeVirt will
+	// complete to assign even parity of CPUs.
+	// Note: this feature is in Developer Preview.
+	// +optional
+	// +kubebuilder:default=false
+	// +default=false
+	AlignCPUs *bool `json:"AlignCPUs,omitempty"`
```
nunnatsa, line 463: Please start json names with lower-cases; i.e.
```suggestion
	AlignCPUs *bool `json:"alignCPUs,omitempty"`
```

controllers/operands/kubevirt.go
```diff
@@ -680,12 +681,24 @@ func getKVSeccompConfig() *kubevirtcorev1.SeccompConfiguration {
...
+
 func NewKubeVirtWithNameOnly(hc *hcov1beta1.HyperConverged, opts ...string) *kubevirtcorev1.KubeVirt {
 	return &kubevirtcorev1.KubeVirt{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      "kubevirt-" + hc.Name,
-			Labels:    getLabels(hc, hcoutil.AppComponentCompute),
-			Namespace: getNamespace(hc.Namespace, opts),
+			Name:        "kubevirt-" + hc.Name,
+			Labels:      getLabels(hc, hcoutil.AppComponentCompute),
+			Annotations: getKubevirtCRAnnotations(hc),
```
nunnatsa, line 700: Please move to NewKubeVirt function, keeping this one producing the minimal version of KV CR.

controllers/operands/kubevirt.go
```diff
@@ -680,12 +681,24 @@ func getKVSeccompConfig() *kubevirtcorev1.SeccompConfiguration {
...
+func getKubevirtCRAnnotations(hc *hcov1beta1.HyperConverged) map[string]string {
+	var annotations map[string]string
+
+	if hc.Spec.FeatureGates.AlignCPUs != nil && *hc.Spec.FeatureGates.AlignCPUs {
+		annotations = make(map[string]string)
+		annotations["alpha.kubevirt.io/EmulatorThreadCompleteToEvenParity"] = ""
+	}
+
+	return annotations
+}
```
nunnatsa, line 693: Not sure about this function. It's ok for create KV, but for the upgrade usecase we should be more careful, not to delete user annotations. We'll need to review the `func (*kubevirtHooks) updateCr()` method to make sure we merge it to the existing CR.

nunnatsa (commented):
thanks for the PR @RamLavi !

Added a few inline comments. In addition, please add human readable documentation of the new field, to docs/cluster-configuration.md + yaml example.

api/v1beta1/hyperconverged_types.go
RamLavi: DONE

controllers/operands/kubevirt.go
RamLavi: I agree. moved this logic outside of this function.

controllers/operands/kubevirt.go
RamLavi: That's a great point! Adding the necessary logic to the UpdateCR flow.
DONE

RamLavi:
[Change](https://github.com/kubevirt/hyperconverged-cluster-operator/compare/e55dfb6c1fd6ae5a85ab565175710eb904e964b3..010ee1ff4887f118ab07292594b48a3313c1497d): Removed the HCO feature gate and made the behavior default by always setting the annotation and feature gate.

controllers/operands/kubevirt.go
```diff
@@ -136,6 +139,7 @@ var (
 		kvHotplugNicsGate,
 		kvVMPersistentState,
 		kvHNetworkBindingPluginsGate,
+		kvAlignCPUsGate,
```
EdDev, line 142: A FG that is added here implies that the U/S FG has been graduated or is scheduled to graduation very soon (and there is commitment that it will actually happen).

If that is not the case, then we have a problem of too early adoption.
I would expect HCO to track the graduation & lifecycle of the U/S project, as setting the FG by default is in practice declaring it GA.

If the intention is for the user not to be bothered, then the U/S should graduate the feature first.
If the intention is to release it at tech-preview, then the FG needs to be defaulted to "false" and exposed through the HCO config.

Please avoid hacking the FG system to reduce UX friction on D/S. The problem is on U/S and there the problem should be solved.

controllers/operands/kubevirt.go
```diff
@@ -246,13 +250,15 @@ func (*kubevirtHooks) updateCr(req *common.HcoRequest, Client client.Client, exi
 		return false, false, errors.New("can't convert to KubeVirt")
 	}
 	if !reflect.DeepEqual(found.Spec, virt.Spec) ||
-		!reflect.DeepEqual(found.Labels, virt.Labels) {
+		!reflect.DeepEqual(found.Labels, virt.Labels) ||
+		!reflect.DeepEqual(found.Annotations, virt.Annotations) {
```
nunnatsa, line 254: We need to be careful here. Updating the annotations as it done now, will cause `outOfBandUpdate` Alert, each time the user will annotate the KV CR. We need to somehow perform the annotation change, but to not return true in the 2nd return value in line 267, if this is the only change. It's tricky, I know...

controllers/operands/kubevirt.go
RamLavi: Thank you for this important notice.
@dankenigsberg - keep me honest her.

The FG we are introducing as default enabled is so that users can run DPDK VMIs. The behavior this FG is setting is of very low risk, so I believe we are OK with setting it ON on D/S. 
In the future we may need to change kubevirt API and eventually remove this FG. IIUC this can done by both removing the FG on Kubevirt and here, thus making the behavior constant for both U/S and D/S.

controllers/operands/kubevirt.go
```diff
@@ -246,13 +250,15 @@ func (*kubevirtHooks) updateCr(req *common.HcoRequest, Client client.Client, exi
...
 		if req.HCOTriggered {
 			req.Logger.Info("Updating existing KubeVirt's Spec to new opinionated values")
 		} else {
 			req.Logger.Info("Reconciling an externally updated KubeVirt's Spec to its opinionated values")
 		}
 		hcoutil.DeepCopyLabels(&virt.ObjectMeta, &found.ObjectMeta)
+		hcoutil.DeepCopyAnnotations(&virt.ObjectMeta, &found.ObjectMeta)
 		virt.Spec.DeepCopyInto(&found.Spec)
 		err := Client.Update(req.Ctx, found)
 		if err != nil {
```
nunnatsa, line 264: Maybe something like this?

(Please notice that the suggested code is longer than the original code. this is because of github limitations)
```suggestion
	if outOfBound := !reflect.DeepEqual(found.Spec, virt.Spec) ||
		!reflect.DeepEqual(found.Labels, virt.Labels); 
		outOfBound || !reflect.DeepEqual(found.Annotations, virt.Annotations) {
		if req.HCOTriggered {
			req.Logger.Info("Updating existing KubeVirt's Spec to new opinionated values")
		} else {
			req.Logger.Info("Reconciling an externally updated KubeVirt's Spec to its opinionated values")
		}
		hcoutil.DeepCopyLabels(&virt.ObjectMeta, &found.ObjectMeta)
		hcoutil.DeepCopyAnnotations(&virt.ObjectMeta, &found.ObjectMeta)
		virt.Spec.DeepCopyInto(&found.Spec)
		err := Client.Update(req.Ctx, found)
		if err != nil {
            			return false, false, err
		}

        	overwritten = outOfBound && !req.HCOTriggered
		return true, overwritten, nil
	}
	return false, false, nil

```

controllers/operands/kubevirt.go
dankenigsberg: @RamLavi if you have to expose the feature gate here in HCO, that would be sad, but acceptable. Make sure to document it very clearly.

RamLavi: @dankenigsberg I'm afraid I don't follow you. Do you mean we should add a "HCO FG/enabler" that would set the annotation+FG on kubevirt CR (like we originally did)?

dankenigsberg: I don't know what you did originally (I saw this PR only today). But yes, if you think this feature is a risky and think that it may still be reverted, you can add a "HCO FG/enabler" that would set the annotation+FG on kubevirt CR.

RamLavi: As we discussed offline yesterday - there is no real need for such an "enabler" from HCO: most users won't notice this change, and those that do and really oppose can choose the jsonPatch solution in order to remove the annotation.

dankenigsberg: I am happy about not tiring HCO users about this kubevirt detail.

controllers/operands/kubevirt.go
RamLavi: After discussing offline - I adopted this change in a separate commit, so that it's well documented (Changed naming a bit - please tell me what you think).

controllers/operands/kubevirt.go
xpivarc: Are json patches generally viable? I think their usage triggers some kind of alert.

RamLavi: they are not. I was only mentioning that if all goes wrong - we should make sure that they are able to save us (currently they are agnostic to kubevirt CR annotations for obvious reasons)

PR #2689, opened 2023-12-20
Allow modifiying operand annotations using json-patch annotation

nunnatsa:
PR #2685 sets a new annotation to the KV CR. Currently, the user can't remove this annotation using the HCO json-patch annotation, in order to remove this new annotation in case it causes issues in their clusters, because HCO rejects json-patch annotations if they are trying to modify fieds that are not under `/spec`.

This PR removes this limitation, allowing the users to modify also annotations.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
None
```

**Release note**:

```release-note
None
```

### DIFF
--- a/controllers/operands/operand.go
+++ b/controllers/operands/operand.go
@@ -2,10 +2,8 @@ package operands
 
 import (
 	"encoding/json"
-	"errors"
 	"fmt"
 	"os"
-	"strings"
 
 	jsonpatch "github.com/evanphx/json-patch/v5"
 	conditionsv1 "github.com/openshift/custom-resource-status/conditions/v1"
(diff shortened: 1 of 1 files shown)


RamLavi (approved):
/lgtm

Is there an e2e to the json patch?

RamLavi:
I suggest we also change the function name: `applyPatchToSpec` to not include only the spec.

RamLavi:
Also, if the patch is allowed (as you implement in this PR) but the annotation is not copied from required to found on `UpdateCr()` - then is the patch set at all eventually?
I would expect that all `UpdateCr()` methods to also need to apply the changes. Am I wrong?

nunnatsa:
@RamLavi - I thought about your comment.

I second though, we don't need this PR. Annotations are not controlled by existing resource handlers, so the user may use the regular k8s mechanisms to annotate or unannotate resources.

PR #2694, opened 2023-12-27
Remove redundant variables

nunnatsa:
use `ptr.To()` instead of defining a variable just to use it's address. That way we can get rid of many variables.

**Jira Ticket**:

```jira-ticket
None
```

**Release note**:

```release-note
None
```

### DIFF
--- a/controllers/hyperconverged/hyperconverged_controller_test.go
+++ b/controllers/hyperconverged/hyperconverged_controller_test.go
@@ -8,8 +8,6 @@ import (
 	"os"
 	"time"
 
-	"k8s.io/utils/ptr"
-
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	openshiftconfigv1 "github.com/openshift/api/config/v1"
@@ -28,6 +26,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/reference"
+	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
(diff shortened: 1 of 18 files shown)


controllers/hyperconverged/hyperconverged_controller_test.go
```diff
@@ -718,12 +717,11 @@ var _ = Describe("HyperconvergedController", func() {
 				existingResource.Kind = kubevirtcorev1.KubeVirtGroupVersionKind.Kind // necessary for metrics
 
 				// now, modify KV's node placement
-				seconds3 := int64(3)
 				existingResource.Spec.Infra.NodePlacement.Tolerations = append(hco.Spec.Infra.NodePlacement.Tolerations, corev1.Toleration{
-					Key: "key3", Operator: "operator3", Value: "value3", Effect: "effect3", TolerationSeconds: &seconds3,
+					Key: "key3", Operator: "operator3", Value: "value3", Effect: "effect3", TolerationSeconds: ptr.To[int64](3),
```
avlitman, line 721: here since we use it 4 times maybe worth creating the var `seconds3 := ptr.To[int64](3)`?

controllers/hyperconverged/hyperconverged_controller_test.go
```diff
@@ -2126,28 +2125,27 @@ var _ = Describe("HyperconvergedController", func() {
...
+					expected.hco.Spec.LiveMigrationConfig.BandwidthPerMigration = ptr.To(badBandwidthPerMigration)
 
 					cl := expected.initClient()
 					_, reconciler, requeue := doReconcile(cl, expected.hco, nil)
 					Expect(requeue).To(BeTrue())
 					foundResource, _, requeue := doReconcile(cl, expected.hco, reconciler)
 					Expect(requeue).To(BeFalse())
 					Expect(foundResource.Spec.LiveMigrationConfig.BandwidthPerMigration).Should(Not(BeNil()))
-					Expect(*foundResource.Spec.LiveMigrationConfig.BandwidthPerMigration).Should(Equal(badBandwidthPerMigration))
+					Expect(foundResource.Spec.LiveMigrationConfig.BandwidthPerMigration).Should(HaveValue(Equal(badBandwidthPerMigration)))
```
avlitman, line 2148: so if we add HaveValue it will fail also if there is no value? asking since before seems we didn't check for value just that it's equal

nunnatsa: yes. The `HaveValue` matcher also checks if the pointer is not nil (so actually, we don't need line 2147).

controllers/hyperconverged/hyperconverged_controller_test.go
nunnatsa: I don't think it saves something. I prefer to not have redundant variables.

orenc1 (approved):
/lgtm

PR #2745, opened 2024-01-18
Improve the creation of index-image

nunnatsa:
1. fix bug that when creating index image for the first time for a version, the script fails.
2. build the index image from scratch, instead of building it from an existing index image, using opm.
3. simplify the script.
4. upgrade opm to v1.35.0

```jira-ticket
None
```

**Release note**:

```release-note
None
```


.github/workflows/build-push-images.yaml
```diff
@@ -20,7 +20,7 @@ jobs:
       QUAY_USERNAME: ${{ secrets.QUAY_USERNAME }}
       QUAY_PASSWORD: ${{ secrets.QUAY_PASSWORD }}
       REGISTRY_NAMESPACE: kubevirt
-      OPM_VERSION: v1.26.2
+      OPM_VERSION: v1.35.0
```
tiraboschi, line 23: We should carefully evaluate this because the output of the opm tool is not guaranteed to be backward compatible with older clusters.

tiraboschi (approved):

PR #2774, opened 2024-01-26
doc: explain two migration options better

dankenigsberg:
Improve the documentation of `completionTimeoutPerGiB` and `allowPostCopy`. Try to explain when they should be configured and how.

**Reviewer Checklist**

- [x] PR Message
- [x] Commit Messages
- [x] How to test
- [x] Unit Tests
- [x] Functional Tests
- [x] User Documentation
- [x] Developer Documentation
- [x] Upgrade Scenario
- [x] Uninstallation Scenario
- [x] Backward Compatibility

**Jira Ticket**:

```jira-ticket
NONE
```

**Release note**:

```release-note
NONE
```

### DIFF
--- a/docs/cluster-configuration.md
+++ b/docs/cluster-configuration.md
@@ -260,10 +260,16 @@ Bandwidth limit of each migrationi, the value is in quantity of bytes per second
 
 ### completionTimeoutPerGiB
 
-The migration will be canceled if it has not completed in this time, in seconds per GiB of memory. For example, a
-virtual machine instance with 6GiB memory will timeout if it has not completed migration in 4800 seconds. If the
-Migration Method is BlockMigration, the size of the migrating disks is included in the calculation. The format is a
-number.
+If a migrating VM is big and busy, while the connection to the destination node 
+is slow, migration may never converge. The completion timeout is calculated
+based on completionTimeoutPerGiB times the size of the guest (both RAM and 
+migrated disks, if any). For example, with completionTimeoutPerGiB set to 800,
+a virtual machine instance with 6GiB memory will timeout if it has not
+completed migration in 1h20m. Use a lower completionTimeoutPerGiB to induce
+quicker failure, so that another destination or post-copy is attempted. Use a
+higher completionTimeoutPerGiB to let workload with spikes in its memory dirty
+rate to converge.
+The format is a number.
 
 **default**: 800
 
@@ -299,9 +305,13 @@ It allows the platform to compromise performance/availability of VMIs to guarant
 
 ### allowPostCopy
 
-It enables post-copy live migrations. Such migrations allow even the busiest VMIs to successfully live-migrate.
-However, events like a network failure can cause a VMI crash.
-If set to true, migrations will still start in pre-copy, but switch to post-copy when CompletionTimeoutPerGiB triggers.
+When enabled, KubeVirt attempts to use post-copy live-migration in case it 
+reaches its completion timeout while attempting pre-copy live-migration.
+Post-copy migrations allow even the busiest VMs to successfully live-migrate.
+However, events like a network failure or a failure in any of the source or
+destination nodes can cause the migrated VM to crash or reach inconsistency.
+Enable this option when evicting nodes is more important than keeping VMs
+alive.
 
 **default**: false
 


docs/cluster-configuration.md
```diff
@@ -299,9 +305,13 @@ It allows the platform to compromise performance/availability of VMIs to guarant
...
 ### allowPostCopy
 
-It enables post-copy live migrations. Such migrations allow even the busiest VMIs to successfully live-migrate.
-However, events like a network failure can cause a VMI crash.
-If set to true, migrations will still start in pre-copy, but switch to post-copy when CompletionTimeoutPerGiB triggers.
+When enabled, KubeVirt attempts to use post-copy live-migration in case it 
+reaches its completion timeout while attempting pre-copy live-migration.
+Post-copy migrations allow even the busiest VMs to successfully live-migrate.
+However, events like a network failure or a failure in any of the source or
+destination nodes can cause the migrated VM to crash or reach inconsistency.
```
dankenigsberg, line 312: /hold

Until I figure out whether the VM can actually reach an inconsistent state in case the destination node crashes.

docs/cluster-configuration.md
```diff
@@ -260,10 +260,16 @@ Bandwidth limit of each migrationi, the value is in quantity of bytes per second
...
 
-The migration will be canceled if it has not completed in this time, in seconds per GiB of memory. For example, a
-virtual machine instance with 6GiB memory will timeout if it has not completed migration in 4800 seconds. If the
-Migration Method is BlockMigration, the size of the migrating disks is included in the calculation. The format is a
-number.
+If a migrating VM is big and busy, while the connection to the destination node 
+is slow, migration may never converge. The completion timeout is calculated
+based on completionTimeoutPerGiB times the size of the guest (both RAM and 
+migrated disks, if any). For example, with completionTimeoutPerGiB set to 800,
+a virtual machine instance with 6GiB memory will timeout if it has not
```
tiraboschi, line 267: Question, is this also going to affect workload upgrades when `spec.workloadUpdateStrategy.workloadUpdateMethods` contains `LiveMigrate`?

See: https://issues.redhat.com/browse/CNV-35196

dankenigsberg: I am trying (with the help of @vladikr ) to document the present situation because the current docs do not make much sense to me.  I'm not an expert of the inner workings. FWIW, I bet this applies to all migrations, regardless of what triggers them. https://kubevirt.io/api-reference/master/definitions.html#_v1_virtualmachineinstancemigrationspec has no way to make a specific migration behave different than the cluster-wide configuration.


https://github.com/kubevirt/hyperconverged-cluster-operator/blob/main/docs/cluster-configuration.md#enable-eventual-launcher-updates-by-default

docs/cluster-configuration.md
xzpeter: A network failure should not crash either side of the VM, but put both sides into a so-called "postcopy-paused" state.  QEMU allows to recover from such state back to "postcopy-active" with the help of an administrator after v3.0.  The problem is that if the current network is down or broken, the admin may need to provide an optional channel to move on the migration until complete.  For a rough idea on how it works (including QEMU APIs) please refer to this page:

https://wiki.qemu.org/Features/PostcopyRecovery

I'm not familiar enough with Libvirt, but I think the corresponding API for libvirt is "--postcopy-resume", as described in the Libvirt documentation:

https://libvirt.org/manpages/virsh.html

xzpeter: I'm not sure what's the best policy here for kubevirt, but a dumb one can be that when postcopy hit network issue (then both QEMUs will go into "postcopy-paused" state), we set a timeout, trying to reconnect using the same old URI to destination QEMU after timeout reached, and keep doing so if it keeps failing until succeed again.  I assume then there's no admin involvement needed if the network can be recovered already automatically via the same route (I'm mostly considering something like an accidentally unplugged wire).

Besides, the admin can overwrite the URI anytime for the case where the old route will be completely broken.  It can be some separate interface besides above auto-reconnect timeout policy, maybe.

dankenigsberg: @xzpeter thanks. This complicates things quite a bit. @vladikr , how does KubeVirt behave in this case?

@xzpeter can you elaborate on what happens to the source or destination if its counterpart dies? To me, the only consistent action is to crash.

dankenigsberg: @xzpeter yes, I agree. A crash is a reasonable and acceptable result. My problem is with other modes of failure, e.g

> If src QEMU died, since some pages still reside on the src QEMU then dest QEMU will have no way to complete such migration. Any access to missing pages will be permanently blocked hence same to a VM crash.

Do you mean that dst would stay paused forever? Who is responsible to kill it so that a fresh VM can restart (maybe KubeVirt already does it)?

xzpeter: > Do you mean that dst would stay paused forever? Who is responsible to kill it so that a fresh VM can restart (maybe KubeVirt already does it)?

Yes it will.  That's the best QEMU can do, IMHO.  It will also be the same case no matter src or dest QEMU died. 

When any side of QEMU crashed, the other side of QEMU should logically go into "postcopy-paused" state just like when there is a network issue - consider when one QEMU died and process recycled, the socket resources will be released by the kernel of that host, so it can have a similar effect of having the socket disconnected to the peer.

Here no matter which side got down, the management layer may need to make the decision on killing the other side of QEMU.  Mgmt apps can have such knowledge of "other side QEMU crashed" by e.g. capturing child exit signals, but QEMU doesn't have that knowledge.  That's why QEMU stays in PAUSED state is what it does right now, as the best effort to maintain whatever data still resides within the QEMU, and with the hope that it could be reconnected in the near future so that data won't get lost.

dankenigsberg: /hold cancel

We need to check this better in KubeVirt. I filed https://issues.redhat.com/browse/CNV-38195 to track it, and added a warning here. I think this PR can be merged now.

docs/cluster-configuration.md
dankenigsberg: @tiraboschi from what I gather, currently there is no differentiation between eviction- and upgrade triggered migration.

tiraboschi (approved):

orenc1:
@dankenigsberg , wouldn't you think the updated description for `completionTimeoutPerGiB` and `allowPostCopy` be also on the HCO CRD itself?
So users will be able to get this explanation in-cluster with `kubectl explain hco.spec.liveMigrationConfig`.
Currently it will show the old text.
in order to update the CRD, these should be updated as well:
https://github.com/kubevirt/hyperconverged-cluster-operator/blob/6c00377fd54efcafd0277e8443be853fcbb58f8b/api/v1beta1/hyperconverged_types.go#L320-L323

https://github.com/kubevirt/hyperconverged-cluster-operator/blob/6c00377fd54efcafd0277e8443be853fcbb58f8b/api/v1beta1/hyperconverged_types.go#L346-L349

tiraboschi:
> @dankenigsberg , wouldn't you think the updated description for `completionTimeoutPerGiB` and `allowPostCopy` be also on the HCO CRD itself? So users will be able to get this explanation in-cluster with `kubectl explain hco.spec.liveMigrationConfig`. 

Sure, #2810 will align it.

PR #2800, opened 2024-02-05
[api] add higher VM workload density API

enp0s3:

Cluster administrators can opt-in for higher VM workload density by configuring 
the memroy overcommit percentage as follows:
```yaml    
spec:
  higherWorkloadDensity:
    memoryOvercommitPrecentage: <intValue> # i.e. 150, 200. default 100
```
As an example when having 150% overcommit, if the VM compute container requests 4Gi memory, then the corresponding VM guest OS will see `4Gi * 150 / 100 = 6Gi`.

If the overcommit percentage is set to lower than 100% the compute container will request more memory than the 
VM guest, therefore the value should be set with precaution.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:
```jira-ticket
https://issues.redhat.com/browse/CNV-23409
```

**Release note**:
```release-note
add higher VM workload density API
```


nunnatsa:
From the PR description:

> The cluster administrator can opt-in for the default by just specifying

I'm not sure about it. I prefer to be explicit. If we want to allow opt-in, let's add a boolean field with a proper name for that. Also, we noticed that there is a UI bug that populates optional fields, so if the HyperConverged CR is created form the UI, the actual result is that the feature is opted in by default.

controllers/operands/kubevirt.go
```diff
@@ -669,13 +673,34 @@ func hcTLSSecurityProfileToKv(profile *openshiftconfigv1.TLSSecurityProfile) *ku
 	}
 }
 
-func getKVDevConfig(hc *hcov1beta1.HyperConverged) *kubevirtcorev1.DeveloperConfiguration {
+func hcMemoryOvercommitToKv(spec *hcov1beta1.HyperConvergedSpec) (percent int, err error) {
+	if spec.HigherWorkloadDensity == nil {
+		return
+	}
+
+	percent = DefaultMemoryOvercommitPercent
```
nunnatsa, line 681: Please use CRD annotation instead. See comment in the API.

api/v1beta1/hyperconverged_types.go
```diff
@@ -757,6 +760,12 @@ type ApplicationAwareConfigurations struct {
 	AllowApplicationAwareClusterResourceQuota bool `json:"allowApplicationAwareClusterResourceQuota,omitempty"`
 }
 
+type HigherWorkloadDensityConfiguration struct {
+	// MemoryOvercommitRatio defines the ratio of memory overcommit for the vitrual machine workloads
+	// The ratio can be used in order to increase virtual machine density
+	MemoryOvercommitRatio *string `json:"memoryOvercommitRatio,omitempty"`
```
nunnatsa, line 766: Please use CRD annotation for the default value; e.g.
```suggestion
	// +kubebuilder:default=1.5
	MemoryOvercommitRatio *string `json:"memoryOvercommitRatio,omitempty"`
```

nunnatsa, line 766: Why using string here? we only support float value anyway, so we won't need to convert later, and the API server will make sure it's a floating number value.
```suggestion
	MemoryOvercommitRatio *float64 `json:"memoryOvercommitRatio,omitempty"`
```

controllers/operands/kubevirt.go
```diff
@@ -669,13 +673,34 @@ func hcTLSSecurityProfileToKv(profile *openshiftconfigv1.TLSSecurityProfile) *ku
 	}
 }
 
-func getKVDevConfig(hc *hcov1beta1.HyperConverged) *kubevirtcorev1.DeveloperConfiguration {
+func hcMemoryOvercommitToKv(spec *hcov1beta1.HyperConvergedSpec) (percent int, err error) {
```
nunnatsa, line 676: Please please please - don't use named returned value. This is so confusing.

controllers/operands/kubevirt.go
```diff
@@ -669,13 +673,34 @@ func hcTLSSecurityProfileToKv(profile *openshiftconfigv1.TLSSecurityProfile) *ku
...
+	if spec.HigherWorkloadDensity == nil {
```
nunnatsa, line 677: This field has actually to purposes: one is to hold the feature configurations and one is to enable/disable the feature. I think is is wrong, mostly because personally, I'm really prefer explicit APIs, but also because it's confusing and error prompt. 

In addition, there is a bug in the UI that auto populate optional fields, making this feature to be always active, if creating the HyperConverged CR from the UI.

What do you think about adding a boolean field for that?

nunnatsa (commented):
Thanks for the PR @enp0s3 

I added a few inline comments.

Also, the sanity test is failing.Please `go mod vendor` from the `tests/` directory to solve it.

api/v1beta1/hyperconverged_types.go
enp0s3: @nunnatsa Hi! I changed to `float64` and got the following error when running `make sanity`:
```
panic: /home/ibezukh/go/src/kubevirt.io/hyperconverged-cluster-operator/api/v1beta1/hyperconverged_types.go:768:24: found float, the usage of which is highly discouraged, as support for them varies across languages. Please consider serializing your float as string instead. If you are really sure you want to use them, re-run with crd:allowDangerousTypes=true
```

enp0s3:
> Thanks for the PR @enp0s3
> 
> I added a few inline comments.
> 
> Also, the sanity test is failing.Please `go mod vendor` from the `tests/` directory to solve it.

@nunnatsa You made me upgrade to FC39 :) since I didn't use `gimme` but rather the supported Go RPMs.

controllers/operands/kubevirt.go
enp0s3: @nunnatsa Thanks! you are correct. I am going to change the API a bit - the field won't optional anymore, it will have a defualt of 1, exactly as it is in kubevirt. And the UI will set it to 1.5 automatically as the user will opt in for memory overcommit.

api/v1beta1/hyperconverged_types.go
nunnatsa: OK. thanks.

controllers/operands/kubevirt.go
```diff
@@ -669,13 +669,21 @@ func hcTLSSecurityProfileToKv(profile *openshiftconfigv1.TLSSecurityProfile) *ku
 	}
 }
 
+func hcMemoryOvercommitToKv(hwd *hcov1beta1.HigherWorkloadDensityConfiguration) int {
+	return hwd.MemoryOvercommitPercentage
+}
+
```
nunnatsa, line 675: Why is this needed? 

usually we need translation functions when KV type is not the same as HCO type, but here both of them are `int`.

nunnatsa (commented):
Added one inline comment

nunnatsa:
/lgtm

@enp0s3 - please fix the [typo](https://github.com/kubevirt/hyperconverged-cluster-operator/pull/2800#discussion_r1512742984) in a following PR

nunnatsa:
hco-e2e-upgrade-prev-operator-sdk-aws lane succeeded
/override ci/prow/hco-e2e-upgrade-prev-operator-sdk-azure

PR #2814, opened 2024-02-15
Adding DaemonSet operand to deploy the swap agent

enp0s3:

**The what**

[wasp-agent ](https://github.com/fabiand/wasp-operator-prototype)allows burstable QoS pods to
use swap, thus allowing for higher VM workload density on the
worker nodes. The Agent is a script that delivered as DaemonSet.
The delivery is gated and its done directly by HCO operator.
The DS need privilege access to the node thus using the kubevirt-handler
service account.

**The why**

Swap support in kuberentes is currently beta1. In order to be enabled
natively in Openshift the swap needs to graduate to beta2 phase. The current
approach with wasp-agent aims to mimic the beta2 functionaliy.

**Upgrade path**

As soon as swap becomes beta2, the wasp-agent will be deprecated.
Users who would like to opt-in for higher density with swap will have
to enable the HigherDensityWithSwap FG in HCO. As soon as swap will
be natively supported in OCP, the cluster admin would be instructed
to validate proper kubelet configuration to allow transpart upgrade.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket

```

**Release note**:

```release-note
Adding DaemonSet operand to deploy the swap agent
```


deploy/olm-catalog/community-kubevirt-hyperconverged/1.12.0/manifests/kubevirt-hyperconverged-operator.v1.12.0.clusterserviceversion.yaml
```diff
@@ -3267,6 +3267,7 @@ spec:
                   value: quay.io/kubevirt-ui/kubevirt-plugin@sha256:4ed1f37116cbaf883edd8783201d091fbbe121fcb84e7184ee531d3b1abd77ab
                 - name: KV_CONSOLE_PROXY_IMAGE
                   value: quay.io/kubevirt-ui/kubevirt-apiserver-proxy@sha256:38293a1050bce9c7cb786cbe8719be1edb91a922a6c583eedb61e56729ed4c72
+                - name: WASP_AGENT_IMAGE
```
orenc1, line 3270: the value of the env var should have been set in the CSV

deploy/images.csv
```diff
@@ -34,3 +34,4 @@ KUBEVIRT_CONSOLE_PROXY_IMAGE,quay.io/kubevirt-ui/kubevirt-apiserver-proxy,KUBEVI
 AAQ_OPERATOR_IMAGE,quay.io/kubevirt/aaq-operator,AAQ_VERSION,2d1c49702fd187030dfa669791f25cefbc29e3103c2f5e0318af4c0ccead0fd2
 AAQ_SERVER_IMAGE,quay.io/kubevirt/aaq-server,AAQ_VERSION,827356ed0ad3857c1319b30cd026775dc8076eb9a1b2b62338f617429af36381
 AAQ_CONTROLLER_IMAGE,quay.io/kubevirt/aaq-controller,AAQ_VERSION,088da1c52a1b14edff7e9d314a46bd9c09c780b39ff82038fb90dc1022ae630d
+WASP_IMAGE,quay.io/ibezukh/wasp-operator-prototype,WASP_VERSION,6092433f8a55216f64b69e4a722e21fef4cedb48c97f2048969bcaf13a2ceb45
```
orenc1, line 37: is it possible to push that image to `quay.io/kubevirt/wasp-operator-prototype` ?
we should avoid referencing personal repos.

hack/config
```diff
@@ -18,6 +18,7 @@ KUBE_RBAC_PROXY_VERSION="4.15"
 
 KUBEVIRT_CONSOLE_PLUGIN_VERSION="v4.15.0"
 KUBEVIRT_CONSOLE_PROXY_VERSION="v0.0.9"
+WASP_VERSION="latest"
```
orenc1, line 21: please use a semver tag for our bump automation.

controllers/operands/wasp.go
```diff
@@ -0,0 +1,225 @@
...
+}
+
+func (*waspHooks) justBeforeComplete(_ *common.HcoRequest) { /* no implementation */ }
+
+func (*waspHooks) getConditions(cr runtime.Object) []metav1.Condition {
+	return dsConditionsToK8s(cr.(*appsv1.DaemonSet).Status.Conditions)
+}
+
+func NewWasp(hc *hcov1beta1.HyperConverged) *appsv1.DaemonSet {
+	waspImage, _ := os.LookupEnv(hcoutil.WaspImageEnvV)
```
orenc1, line 100: please validate this env var actually exists and has a value before even creating the handler.

controllers/operands/wasp.go
```diff
@@ -0,0 +1,225 @@
...
+			Type: appsv1.RollingUpdateDaemonSetStrategyType,
+			RollingUpdate: &appsv1.RollingUpdateDaemonSet{
+				MaxUnavailable: ptr.To(intstr.FromString("10%")),
+				MaxSurge:       ptr.To(intstr.FromString("0")),
+			},
+		},
+		Template: corev1.PodTemplateSpec{
+			ObjectMeta: metav1.ObjectMeta{
+				Labels: map[string]string{
+					"name": "wasp",
```
orenc1, line 118: please use `getLabels(hc, componentName)` for the labels

controllers/operands/wasp.go
```diff
@@ -0,0 +1,225 @@
...
+				RestartPolicy:                 corev1.RestartPolicyAlways,
+				ServiceAccountName:            waspServiceAccount,
+				HostPID:                       true,
+				HostUsers:                     ptr.To(true),
+				TerminationGracePeriodSeconds: ptr.To(int64(30)),
+				Containers: []corev1.Container{
+					{
+						Name:            "wasp-agent",
+						Image:           waspImage,
+						ImagePullPolicy: corev1.PullAlways,
```
orenc1, line 131: why not `IfNotPresent` as all other pods in kubevirt?
this is important for disconnected envs.

controllers/operands/wasp.go
```diff
@@ -0,0 +1,225 @@
...
+							},
+						},
+					},
+				},
+				Volumes: []corev1.Volume{
+					{
+						Name: "host",
+						VolumeSource: corev1.VolumeSource{
+							HostPath: &corev1.HostPathVolumeSource{
+								Path: "/",
```
orenc1, line 176: is it really required to mount all of the host's filesystem? can it be done with the relevant path(s)?

controllers/operands/wasp.go
```diff
@@ -0,0 +1,225 @@
...
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+	return wasp
```
orenc1, line 184: i think you need also to set the node placement for the DaemonSet, based on HCO's `.spec.workloads.nodePlacement`.

controllers/operands/wasp_test.go
```diff
@@ -0,0 +1,80 @@
...
+		cl  client.Client
+	)
+
+	BeforeEach(func() {
+		hco = commontestutils.NewHco()
+		req = commontestutils.NewReq(hco)
+	})
+
+	Context("check FG", func() {
+		It("shoud not deploy wasp when FG is off", func() {
```
orenc1, line 30:
```suggestion
		It("should not deploy wasp when FG is off", func() {
```

tools/csv-merger/csv-merger.go
```diff
@@ -95,6 +95,7 @@ var (
 	cliDownloadsImage   = flag.String("cli-downloads-image-name", "", "Downloads Server image")
 	kvUIPluginImage     = flag.String("kubevirt-consoleplugin-image-name", "", "KubeVirt Console Plugin image")
 	kvUIProxyImage      = flag.String("kubevirt-consoleproxy-image-name", "", "KubeVirt Console Proxy image")
+	waspImage           = flag.String("wasp-image-name", "", "Swap configurtaion agent image")
```
orenc1, line 98:
```suggestion
	waspImage           = flag.String("wasp-image-name", "", "Swap configuration agent image")
```

controllers/operands/wasp.go
```diff
@@ -0,0 +1,225 @@
...
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	hcov1beta1 "github.com/kubevirt/hyperconverged-cluster-operator/api/v1beta1"
+	"github.com/kubevirt/hyperconverged-cluster-operator/controllers/common"
+	hcoutil "github.com/kubevirt/hyperconverged-cluster-operator/pkg/util"
+)
+
+const (
+	pullURL = "quay.io/fdeutsch/wasp-operator-prototype:latest"
```
orenc1, line 23: remove, we're using a pullspec from an env var.

deploy/olm-catalog/community-kubevirt-hyperconverged/1.12.0/manifests/kubevirt-hyperconverged-operator.v1.12.0.clusterserviceversion.yaml
enp0s3: @orenc1 Can you please help me with adding it? I am not sure whether I need to:
* Add it manually
* Run some make target
* Add some missing piece of code/script and then run some make target

deploy/images.csv
orenc1, line 37: should be:
```suggestion
WASP_AGENT_IMAGE,quay.io/ibezukh/wasp-operator-prototype,WASP_VERSION,6092433f8a55216f64b69e4a722e21fef4cedb48c97f2048969bcaf13a2ceb45
```
to get the value added in the HCO env var in the CSV

controllers/operands/wasp.go
nunnatsa: Good point. Can be done when *creating* the handler, into a new field in the handler, then you could use this new field without checking the env var each time.

enp0s3: @orenc1 @nunnatsa I have added:
```
	if ci.IsOpenshift() && ci.IsWaspImageProvided() {
		operands = append(operands, newWaspHandler(client, scheme))
	}
```
and implementation:
```
	waspImageValue, waspImageExists := os.LookupEnv(WaspImageEnvV)
	c.waspImageProvided = waspImageExists && len(waspImageValue) > 0
...
...
...
func (c *ClusterInfoImp) IsWaspImageProvided() bool {
	return c.waspImageProvided
}

```

hack/config
nunnatsa: I don't sure we want it under bump automation. On the other hand, we don't want floating tag either, to avoid pulling a missing digest.

bundle.Dockerfile
nunnatsa, line 1: What is this file? was it added by mistake?

controllers/operands/wasp.go
```diff
@@ -0,0 +1,449 @@
...
+
+	sdkapi "kubevirt.io/controller-lifecycle-operator-sdk/api"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
nunnatsa, line 21: nit: please re-group imports

controllers/operands/wasp.go
```diff
@@ -0,0 +1,449 @@
...
+		}
+		return true, !req.HCOTriggered, nil
+	}
+
+	return false, false, nil
+}
+
+func (*waspHooks) justBeforeComplete(_ *common.HcoRequest) { /* no implementation */ }
+
+func (*waspHooks) getConditions(cr runtime.Object) []metav1.Condition {
```
nunnatsa, line 111: This is a partial implementation of the `hcoOperandHooks` interface. This interface includes two methods, and this PR implements only one of them. That why the linter says this method is unused, because it never called for types that are not fully implementing the interface. 

But actually, we don't need this method at all. this method should collect the conditions from KubeVirt CRs, that are publishing common list of conditions. The DaemonSet does not publish the KV common conditions, obviously, and the is no point to implement the interface.

However, maybe it will be a good idea to add a new condition to HCO to reflect the status of the swap daemonset? It won't be simple though, so we can add it in a following PR. @enp0s3 , @orenc1 , @tiraboschi  - WDYT?

controllers/operands/wasp.go
```diff
@@ -0,0 +1,449 @@
...
+	return &appsv1.DaemonSet{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "wasp-agent",
+			Labels:    getLabels(hc, hcoutil.AppComponentWasp),
+			Namespace: hc.Namespace,
+		},
+	}
+}
+
+func dsConditionsToK8s(conditions []appsv1.DaemonSetCondition) []metav1.Condition {
```
nunnatsa, line 220: See comment above, on the `getConditions` method https://github.com/kubevirt/hyperconverged-cluster-operator/pull/2814/files#r1536748265.

controllers/operands/wasp.go
```diff
@@ -0,0 +1,449 @@
...
+			crType: "DaemonSet",
+			hooks:  &waspHooks{},
+		},
+		getCRWithName: func(hc *hcov1beta1.HyperConverged) client.Object {
+			return NewWaspWithNameOnly(hc)
+		},
+		shouldDeploy: func(hc *hcov1beta1.HyperConverged) bool {
+			// convert from anon to private
+			return hc.Spec.FeatureGates.EnableHigherDensityWithSwap != nil &&
+				*hc.Spec.FeatureGates.EnableHigherDensityWithSwap
```
nunnatsa, line 50: Let's make a function from it, and use it also for the ClusterRoleHandler and the and the ClusterRoleBindingHandler, so we'll have this common logic in one place.

controllers/operands/wasp.go
```diff
@@ -0,0 +1,449 @@
...
+			Scheme: Scheme,
+			crType: "ClusterRole",
+			hooks:  &waspClusterRoleHooks{},
+		},
+		getCRWithName: func(hc *hcov1beta1.HyperConverged) client.Object {
+			return NewWaspClusterRoleWithNameOnly(hc)
+		},
+		shouldDeploy: func(hc *hcov1beta1.HyperConverged) bool {
+			return hc.Spec.FeatureGates.EnableHigherDensityWithSwap != nil &&
+				*hc.Spec.FeatureGates.EnableHigherDensityWithSwap
```
nunnatsa, line 287: same

controllers/operands/wasp.go
```diff
@@ -0,0 +1,449 @@
...
+			Scheme: Scheme,
+			crType: "ClusterRoleBinding",
+			hooks:  &waspClusterRoleBindingHooks{},
+		},
+		getCRWithName: func(hc *hcov1beta1.HyperConverged) client.Object {
+			return NewWaspClusterRoleBindingWithNameOnly(hc)
+		},
+		shouldDeploy: func(hc *hcov1beta1.HyperConverged) bool {
+			return hc.Spec.FeatureGates.EnableHigherDensityWithSwap != nil &&
+				*hc.Spec.FeatureGates.EnableHigherDensityWithSwap
```
nunnatsa, line 374: same

nunnatsa (commented):
Thanks @enp0s3. 

In addition, we applied the new ginkgolinter rue to force Expect-To, so please change all `Expect(...).Should/Not(...)`, to `Expect(...).To/Not()`.

you can do that automatically using:
```shell
go install github.com/nunnatsa/ginkgolinter/cmd/ginkgolinter@latest
ginkgolinter -fix --force-expect-to=true ./controllers/...
cd tests
ginkgolinter -fix --force-expect-to=true ./...
```

controllers/operands/wasp.go
enp0s3: @nunnatsa @tiraboschi Hi. To panic makes sense to me as well, but I also tried to be consistent with the way the console plugin is being deployed. If we agree on panic I can defer it to another PR so we won't drift with the current one which is big enough already.

PR #2823, opened 2024-02-27
Add `downwardMetrics` feature gate

jcanocan:
It adds the `downwardMetrics` feature gate to HCO. It allows users to disable this feature in KubeVirt.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket

```

**Release note**:

```release-note
Add `downwardMetrics` feature gate
```


api/v1beta1/hyperconverged_types.go
```diff
@@ -387,6 +387,10 @@ type VirtualMachineOptions struct {
 // by default yet.
 // +k8s:openapi-gen=true
 type HyperConvergedFeatureGates struct {
+	// Allow to expose a limited set of host metrics to guests.
+	// +optional
+	DownwardMetrics *bool `json:"downwardMetrics,omitempty"`
```
tiraboschi, line 392: We need to understand why we want to do this:
with something like this the feature can only be fully on/fully off at cluster level.
If we think that having the DownwardMetrics always enabled can introduce some security concern, we should probably introduce something more fine grained to let the cluster admin selectively enable it just for trusted VMs/namespaces.

In this way, if just a single VM on the cluster need this for a valid reason, this could only be enabled for all the VMs on the cluster and so probably we are not reaching what we want to offer to the cluster admin.

nunnatsa, line 392: Please add default value

controllers/operands/kubevirt.go
```diff
@@ -880,8 +878,19 @@ func getMandatoryKvFeatureGates(isKVMEmulation bool) []string {
 // get list of feature gates or KV FG list
 func getKvFeatureGateList(fgs *hcov1beta1.HyperConvergedFeatureGates) []string {
 	checks := getFeatureGateChecks(fgs)
+
 	res := make([]string, 0, len(checks)+len(mandatoryKvFeatureGates)+1)
-	res = append(res, mandatoryKvFeatureGates...)
+
+	if fgs.DownwardMetrics != nil && !*fgs.DownwardMetrics {
```
tiraboschi, line 884: please implement this logic in `getFeatureGateChecks` and keep this as it is.

docs/cluster-configuration.md
```diff
@@ -130,6 +130,12 @@ or new features that are not enabled by default.
 To enable a feature, add its name to the `featureGates` list and set it to `true`. Missing or `false` feature gates
 disables the feature.
 
+### downwardMetrics Feature Gate
+Set the `downwardMetrics` feature gate in order to allow exposing a limited set of VM and host metrics to the guest.
+The format is compatible with [vhostmd](https://github.com/vhostmd/vhostmd).
+These metrics allow third-parties diagnosing issues.
+DownwardMetrics may be exposed to the guest through a `volume` or a `virtio-serial port`.
```
tiraboschi, line 137: Please document the upgrade strategy (in the past it was always on)

controllers/operands/kubevirt.go
```diff
@@ -128,8 +125,8 @@ var (
 		kvHotplugVolumesGate,
 		kvExpandDisksGate,
 		kvGPUGate,
+		kvDownwardMetrics,
```
nunnatsa, line 128: Should not be in this list. This list is the fg that hco always sets in kv.

controllers/operands/kubevirt.go
nunnatsa, line 884: Please revert. This function sould not be changed. Please follow one of the other optional featur gates. It should be in getFeatureGateChecks function.

nunnatsa (changes requested):
Thanks for the pr!

Addded a few inline comments.

In addition, default values are missing, for the new fg, the fg object and the spec. Follow one oc the other fg to get sll the places. Then, update the defaults e2e test.

tiraboschi:
> In addition, default values are missing, for the new fg, the fg object and the spec. Follow one oc the other fg to get sll the places. Then, update the defaults e2e test.

This was somehow intentional.
Now (for the past 6 releases) `downwardMetrics` was always enabled by HCO for Kubevirt.
We want to transition to situation where the cluster admin can configure it (potentially just for selected/trusted VMs/namespaces so I'm neither so sure that a FG is the best solution) but on the other side we do not want to do it in a single step to be sure we are not going to break VMs that are already relying on it during an upgrade.

api/v1beta1/hyperconverged_types.go
jcanocan: I agree. But as first step to go there, I think we should follow this path. WDYT?

controllers/operands/kubevirt.go
jcanocan: This change will force us to change some tests behavior which relies on the `hardCodeKvFgs`. Would you be fine with such change? Just follow the `hardCodeKvFgs` usages, it is widely used in the codebase. I implemented in this way because It creates minimal changes to the test codebase.

controllers/operands/kubevirt.go
jcanocan: See: https://github.com/kubevirt/hyperconverged-cluster-operator/pull/2823#discussion_r1505640247

api/v1beta1/hyperconverged_types.go
tiraboschi: DownwardMetrics is always on since version 1.5 (2.5 years ago) so I don't see any urgency to introduce a temporary API and then to deprecate it later on in order to move to a properly designed one.
If we already tolerated it for 2.5 years, I think we can wait for a few days/weeks more to get a proper API.

controllers/operands/kubevirt.go
jcanocan: Done.

controllers/operands/kubevirt.go
jcanocan: Done.

api/v1beta1/hyperconverged_types.go
jcanocan: We don't want to have a default value here right now.

docs/cluster-configuration.md
jcanocan: Done.

controllers/operands/kubevirt.go
```diff
@@ -751,6 +748,10 @@ func hcoConfig2KvConfig(hcoConfig hcov1beta1.HyperConvergedConfig, infrastructur
 func getFeatureGateChecks(featureGates *hcov1beta1.HyperConvergedFeatureGates) []string {
 	fgs := make([]string, 0, 2)
 
+	if featureGates.DownwardMetrics == nil || (featureGates.DownwardMetrics != nil && *featureGates.DownwardMetrics) {
```
nunnatsa, line 751: I think it's too verbose. Can be done by:
```suggestion
	if featureGates.DownwardMetrics == nil || *featureGates.DownwardMetrics {
```

jcanocan: Done.

nunnatsa (approved):

PR #2844, opened 2024-03-15
Make `downwardMetric` feature gate opt-in

jcanocan:
The feature gate `downwardMetric` is an opt-out feature gate. This is due to backward compatibility reasons. Before this feature gate was introduced, it was hardcoded. It makes this feature gate to be disabled by default, leaving this decision to the final user.
**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket

```

**Release note**:

```release-note
Disable by default the `downwardMetric` feature gate
```


api/v1beta1/hyperconverged_types.go
```diff
@@ -395,6 +395,8 @@ type VirtualMachineOptions struct {
 type HyperConvergedFeatureGates struct {
 	// Allow to expose a limited set of host metrics to guests.
 	// +optional
+	// +kubebuilder:default=false
```
nunnatsa, line 398: @jcanocan please also add the default value for the `FeatureGates` field (line 69), and for the `spec` field (line 824).

jcanocan: Done.

jcanocan:
> Looks good in general. Please also update the e2e defaults test to match this change. It currently fails in ci.

Thanks for the pointer. I wasn't aware of such tests. Done!

nunnatsa (approved):
I think it's OK. @tiraboschi - WDYT? will it work on update?

sarahbx:
Hi @nunnatsa , @tiraboschi,
This is recommended to mitigate https://access.redhat.com/security/cve/CVE-2024-31419 upstream for earlier v1 branches.
A more comprehensive solution to this access would be ideal, and I hope we see something in a future release, but currently any existing consumers of this feature on earlier v1 branches would have to disable this cluster-wide feature manually. While this is classified as 'low' impact, it is still a CVE originating from the kubevirt org without admin action.

tiraboschi:
This is compatibility breaking change so we need at least an intermediate release (1.11 here) to avoid silently breaking upstream users that are currently relying on that.
By the way this is just a partial mitigation since it could only be enabled or disabled at cluster scope with no more fine grained control: so, if at least one VM in the cluster needs it for a good reason, all the other VMs are still going to have it.
So this is still not the right solution.
I'd suggest to properly fix it instead it with a more fine grained API to limit it only to VMs/namespaces that really need it.

PR #2856, opened 2024-03-24
[operands]: Adding DaemonSet operand to deploy the swap agent

enp0s3:
Signed-off-by: Igor Bezukh [ibezukh@redhat.com](mailto:ibezukh@redhat.com)

What this PR does / why we need it:

The what

[wasp-agent ](https://github.com/fabiand/wasp-operator-prototype)allows burstable QoS pods to
use swap, thus allowing for higher VM workload density on the
worker nodes. The Agent is a script that delivered as DaemonSet.
The delivery is gated and its done directly by HCO operator.
The DS need privilege access to the node thus using the kubevirt-handler
service account.

The why

Swap support in kuberentes is currently beta1. In order to be enabled
natively in Openshift the swap needs to graduate to beta2 phase. The current
approach with wasp-agent aims to mimic the beta2 functionaliy.

Upgrade path

As soon as swap becomes beta2, the wasp-agent will be deprecated.
Users who would like to opt-in for higher density with swap will have
to enable the HigherDensityWithSwap FG in HCO. As soon as swap will
be natively supported in OCP, the cluster admin would be instructed
to validate proper kubelet configuration to allow transpart upgrade.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket

```

**Release note**:

```release-note
adding DaemonSet operand to deploy the swap agent
```


controllers/operands/wasp.go
```diff
@@ -0,0 +1,306 @@
...
+}
+
+type waspHooks struct {
+	cache *appsv1.DaemonSet
+}
+
+func (h *waspHooks) getFullCr(hc *hcov1beta1.HyperConverged) (client.Object, error) {
+	if h.cache == nil {
+		h.cache = NewWasp(hc)
+	}
```
nunnatsa, line 59: If you're using cache, please also implement the `reset()` method; see here for example: https://github.com/kubevirt/hyperconverged-cluster-operator/blob/5eb460e178f7c349f483e3586ff3694ec69cdf03/controllers/operands/cdi.go#L59-L61

nunnatsa (commented):
Added one last comment

controllers/operands/wasp.go
enp0s3: @nunnatsa Thanks! done

pkg/util/cluster.go
```diff
@@ -33,6 +33,7 @@ type ClusterInfo interface {
 	IsControlPlaneHighlyAvailable() bool
 	IsInfrastructureHighlyAvailable() bool
 	IsConsolePluginImageProvided() bool
+	IsWaspImageProvided() bool
```
nunnatsa, line 36: This should not be part of this interface, IMHO (`IsConsolePluginImageProvided` should not be here as well). The interface is about the cluster HCO is running in.

The question if an image is provided or not, does not belong here.

tests/func-tests/wasp_test.go
```diff
@@ -0,0 +1,98 @@
...
+	tests "github.com/kubevirt/hyperconverged-cluster-operator/tests/func-tests"
+	"kubevirt.io/client-go/kubecli"
+	"kubevirt.io/kubevirt/tests/flags"
+)
+
+const (
+	setWaspFGPatchTemplate = `[{"op": "replace", "path": "/spec/featureGates/enableHigherDensityWithSwap", "value": %t}]`
+)
+
+var _ = Describe("wasp-agent", Label("wasp"), Serial, Ordered, func() {
```
nunnatsa, line 22: Please add the new "OpenShift" label. See #2860 
```suggestion
var _ = Describe("wasp-agent", Label("wasp", "OpenShift"), Serial, Ordered, func() {
```

nunnatsa (commented):
Since we need to fix the CSV, please also see two new comments inline.

tests/func-tests/wasp_test.go
enp0s3: @nunnatsa Done

pkg/util/cluster.go
enp0s3: @nunnatsa Following our conversation, we decided to defer it to follow up PR

PR #2862, opened 2024-04-03
Fix the monitoring e2e test

nunnatsa:
The "KubeVirtCRModified alert should fired when there is a modification on a CR" e2e test fails frequently, because the test logic is trying to trigger the `KubeVirtCRModified` alert, by adding a new label to KubeVirt CR, while we now allowing adding custom lables.

This PR adds a fake feature gate to the KubeVirt CR in order to trigger the same alert. HCO fully reconciles the KV feature gates, so it should trigger the alert in this case.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
None
```

**Release note**:

```release-note
None
```

### DIFF
--- a/tests/func-tests/monitoring_test.go
+++ b/tests/func-tests/monitoring_test.go
@@ -95,14 +95,28 @@ var _ = Describe("[crit:high][vendor:cnv-qe@redhat.com][level:system]Monitoring"
 
 	It("KubeVirtCRModified alert should fired when there is a modification on a CR", func() {
 		By("Patching kubevirt object")
-		Eventually(func(g Gomega) map[string]string {
-			patch := []byte(`[{"op": "add", "path": "/metadata/labels/test-label", "value": "test-label-value"}]`)
+		const (
+			fakeFG = "fake-fg-for-testing"
+			query  = `kubevirt_hco_out_of_band_modifications_total{component_name="kubevirt/kubevirt-kubevirt-hyperconverged"}`
+		)
+
+		var valueBefore float64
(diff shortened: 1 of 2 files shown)


machadovilaca:
/lgtm

but I wonder why this didn't fail in the PR where the label addition was made possible

nunnatsa:
> /lgtm
> 
> but I wonder why this didn't fail in the PR where the label addition was made possible

I think it's flaky. Maybe some other test also changes KV, or even other CR. The order of the tests is random, so if there is a test that do that, the test will pass

machadovilaca:
> > /lgtm
> > but I wonder why this didn't fail in the PR where the label addition was made possible
> 
> I think it's flaky. Maybe some other test also changes KV, or even other CR. The order of the tests is random, so if there is a test that do that, the test will pass.
> 
> e.g. the `should have all the required labels in all the controlled resources` modifies a controlled label.

maybe we should improve the test to verify the out of band modifications metric value is increases after the CR change in the test

tiraboschi (approved):

nunnatsa:
hco-e2e-operator-sdk-sno-aws lane passed 
/override ci/prow/hco-e2e-operator-sdk-sno-azure

PR #2863, opened 2024-04-05
kubvirt: Fix aarch64 machine type and OVMF path

lyarwood:

As documented [1] downstream aarch64 guests require a different UEFI firmware to x86_64. This is provided by the edk2-aarch64 package with the resulting firmware placed under /usr/share/AAVMF.

This change updates the OVMFPath configured for use with such guests on aarch64 and also updates some tests to assert this is set.

This change also introduces arch specific versions of the machine type env variable. This avoids the same machine type being set for both x86_64 and aarch64. The original `MACHINETYPE` env variable is supported for a release to ensure any downstream consumers continue to work.

Finally we also stop setting the deprecated top level machineType [2] configurable in favour of these arch specific versions.

[1] https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html-single/configuring_and_managing_virtualization/index#how-virtualization-on-arm-64-differs-from-amd64-and-intel64_feature-support-and-limitations-in-rhel-9-virtualization
[2] https://kubevirt.io/api-reference/main/definitions.html#_v1_kubevirtconfiguration

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket

```

**Release note**:

```release-note
NONE
```

### DIFF
--- a/controllers/operands/kubevirt.go
+++ b/controllers/operands/kubevirt.go
@@ -30,9 +30,11 @@ import (
 
 // env vars
 const (
-	kvmEmulationEnvName = "KVM_EMULATION"
-	smbiosEnvName       = "SMBIOS"
-	machineTypeEnvName  = "MACHINETYPE"
+	kvmEmulationEnvName     = "KVM_EMULATION"
+	smbiosEnvName           = "SMBIOS"
+	machineTypeEnvName      = "MACHINETYPE"
+	amd64MachineTypeEnvName = "AMD64_MACHINETYPE"
+	arm64MachineTypeEnvName = "ARM64_MACHINETYPE"
 )
 
 const (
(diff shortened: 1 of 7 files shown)


lyarwood:
/hold

FWIW the handling of the machine type env variable is broken here, I'll look more on Monday but in short you can't reuse the same machine type for x86_64 and aarch64 guests.

nunnatsa:
> /hold
> 
> FWIW the handling of the machine type env variable is broken here, I'll look more on Monday but in short you can't reuse the same machine type for x86_64 and aarch64 guests.

@lyarwood maybe it's related to KV's `spec.configuration.machineType` been deprecated? I think we need to use ``spec.configuration.architectureConfiguration` instead.

lyarwood:
> > /hold
> > FWIW the handling of the machine type env variable is broken here, I'll look more on Monday but in short you can't reuse the same machine type for x86_64 and aarch64 guests.
> 
> @lyarwood maybe it's related to KV's `spec.configuration.machineType` been deprecated? I think we need to use `spec.configuration.architectureConfiguration.defaultArchitecture` instead.

Yes correct and we shouldn't be setting a single machine type for both architectures. I'll add some commits breaking the MACHINETYPE env into two later today.

controllers/operands/kubevirt_test.go
```diff
@@ -174,6 +174,8 @@ var _ = Describe("KubeVirt Operand", func() {
 
 		defer os.Unsetenv(smbiosEnvName)
 		defer os.Unsetenv(machineTypeEnvName)
+		defer os.Unsetenv(amd64MachineTypeEnvName)
+		defer os.Unsetenv(arm64MachineTypeEnvName)
```
nunnatsa, line 178: It's already 5 of them. WDYT about adding a function to unset all of them and all it from a single `defer`?

controllers/operands/kubevirt.go
```diff
@@ -458,21 +461,34 @@ func getKVConfig(hc *hcov1beta1.HyperConverged) (*kubevirtcorev1.KubeVirtConfigu
...
+			EmulatedMachines: strings.Split(DefaultAMD64EmulatedMachines, ","),
+		}
+	}
+
+	if armMachineType, ok := os.LookupEnv(arm64MachineTypeEnvName); ok {
+		if armMachineType = strings.TrimSpace(armMachineType); armMachineType != "" {
+			if config.ArchitectureConfiguration == nil {
+				config.ArchitectureConfiguration = &kubevirtcorev1.ArchConfiguration{}
+			}
+			config.ArchitectureConfiguration.Arm64 = &kubevirtcorev1.ArchSpecificConfiguration{
```
nunnatsa, line 488: Should we also set `config.architectureConfiguration.defaultArchitecture`?

lyarwood: I don't think that is required no, it defaults to the runtime arch during deployment and we only support homogeneous clusters downstream at the moment AFAIK.

lyarwood: @acardace apologies for the random ping but can you keep me honest here ^ I don't think there's any need for `config.architectureConfiguration.defaultArchitecture` to be defined by default at the moment.

controllers/operands/kubevirt.go
```diff
@@ -459,21 +461,34 @@ func getKVConfig(hc *hcov1beta1.HyperConverged) (*kubevirtcorev1.KubeVirtConfigu
 		}
 	}
 
-	if val, ok := os.LookupEnv(machineTypeEnvName); ok {
-		if val = strings.TrimSpace(val); val != "" {
-			config.MachineType = val
+	var amd64MachineType string
+	// Remain backwards compatibility with the original env variable for a release before removing
```
lyarwood, line 465: Urgh `Remain backwardly compatible` I'll correct if I need to respin.

nunnatsa (approved):
/approve

PR #2921, opened 2024-04-26
ssp: prpoagate tekton feature gate to SSP

akrejcir:
This PR propagates the value of tekton feature gate to SSP.
This is still needed to fix: https://issues.redhat.com/browse/CNV-33012

After the bug fix, user can intentionally remove all tekton resources by setting the feature gate to `nil` or `false`.

This change will make the behavior after update from previous version less surprising to the user.
It will still be possible to intentionally remove tekton resources without deleting SSP.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:
```jira-ticket
https://issues.redhat.com/browse/CNV-33012
```

**Release note**:
```release-note
None
```


docs/cluster-configuration.md
```diff
@@ -162,6 +162,12 @@ the [dataImportCronTemplates field](#configure-custom-golden-images), even if th
 
 **Default**: `true`
 
+### deployTektonTaskResources Feature Gate
+The `deployTektonTaskResources` feature gate is deprecated. If it is set to `false` on existing clusters,
+it will remove tasks and example pipelines. Setting in to `true` has no effect.
```
0xFelix, line 167:
```suggestion
it will remove tasks and example pipelines. Setting it to `true` has no effect.
```

akrejcir: done

0xFelix (approved):
/lgtm

nunnatsa:
@akrejcir - I'm not sure I understand the logic here.

If I understand correctly, this PR reuse the old FG in order to clean takton resources, so the user should actively set it to false. Not sure I understand why this is needed. Are we talking about the upgrade usecase, where the FG was set to true in the previous version? If yes, please add this to the PR description. Anyway, as the code is written now, there is no way not to set the FG. if you'll remove it, it will re-set to false, because this is the default value.

0xFelix:
It's a pointer so it should be possible to not set it, right?

This is intended as an optional way to clean up old Tekton tasks which may remain on the cluster but are deprecated and not required anymore.

akrejcir:
This change will make the behavior after update less surprising to the user.
In the old version, the feature gate behaved as:
- `true` - SSP creates tekton resources
- `false` - SSP removes created tekton resources.

In the new version with this change:
- `true` - SSP does not create tekton resources, only adds a deprecated annotation.
- `false` - SSP removes tekton resources that were created by the previous version.

Without this PR, user would have to manually remove old tekton resources, or uninstall SSP.

The feature gate in SSP is not a pointer, which is ok. Not setting it is the same as setting it to `false`.

I will change this PR, so that `nil` feature gate is interpreted as `false`.

nunnatsa:
> It's a pointer so it should be possible to not set it, right?
> 
> This is intended as an optional way to clean up old Tekton tasks which may remain on the cluster but are deprecated and not required anymore.

But we have CRD defaults there, setting this FG to be false by default. It will never be empty.

Also, to me it seems like an abuse of the API, and I really don't like it.

nunnatsa:
See lines 
* https://github.com/kubevirt/hyperconverged-cluster-operator/blob/8144382ca03156cadc4024c9ffe767503daaa08f/api/v1beta1/hyperconverged_types.go#L69
* https://github.com/kubevirt/hyperconverged-cluster-operator/blob/8144382ca03156cadc4024c9ffe767503daaa08f/api/v1beta1/hyperconverged_types.go#L425-L426
* https://github.com/kubevirt/hyperconverged-cluster-operator/blob/8144382ca03156cadc4024c9ffe767503daaa08f/api/v1beta1/hyperconverged_types.go#L832

akrejcir:
> But we have CRD defaults there, setting this FG to be false by default. It will never be empty.

It is ok if the FG is never empty. `false` is the same as `nil` for this FG.
 
> Also, to me it seems like an abuse of the API, and I really don't like it.

Why do you think it is an API abuse? In my opinion it is close to the previous functionality. The only difference is that `true` value does not do anything.
Do you maybe have a different idea how to do this? Or if you think it is a bad design, we can not implement it.

akrejcir:
We wanted this to go to 4.16, but it may be too late now. If 4.16 will be released without this PR, then adding this PR in later versions will introduce unwanted behavior.
In the current code without this PR, the SSP feature gate is always set to `false`. When it is released and users update their clusters, they will have the feature gate set to `false` and existing tekton resources will be marked as deprecated.
If we would merge this PR in a later version, and users would update, the feature gate would still be `false`, and unexpectedly for users, SSP would remove all the existing deprecated tekton resources.

That is not what we want. Tekton resources should not be unexpectedly removed.

/hold

akrejcir:
Closing, because it is too late.
We will add documentation to explain how to remove old tekton resources.

/close

PR #2945, opened 2024-05-06
Remove node label from cnv_abnormal

avlitman:
It's not possible to add the node label to telemetry, this pr removes it. 

**Jira Ticket**:

```jira-ticket: 
https://issues.redhat.com/browse/CNV-40387

```

**Release note**:

```release-note
none
```

### DIFF
--- a/pkg/monitoring/rules/recordingrules/operator.go
+++ b/pkg/monitoring/rules/recordingrules/operator.go
@@ -29,6 +29,6 @@ var operatorRecordingRules = []operatorrules.RecordingRule{
 			Help: "Monitors resources for potential problems",
 		},
 		MetricType: operatormetrics.GaugeType,
-		Expr:       intstr.FromString(`sum by (container, node, reason)(kubevirt_memory_delta_from_requested_bytes)`),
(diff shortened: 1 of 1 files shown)


pkg/monitoring/rules/recordingrules/operator.go
```diff
@@ -29,6 +29,6 @@ var operatorRecordingRules = []operatorrules.RecordingRule{
 			Help: "Monitors resources for potential problems",
 		},
 		MetricType: operatormetrics.GaugeType,
-		Expr:       intstr.FromString(`sum by (container, node, reason)(kubevirt_memory_delta_from_requested_bytes)`),
+		Expr:       intstr.FromString(`sum by (container, reason)(kubevirt_memory_delta_from_requested_bytes)`),
```
machadovilaca, line 32: fine by me, but why do we need to remove the node label? cardinality?

avlitman: Yes, cardinality, they say the node names may vary over time which may add too many series to telemeter

machadovilaca: makes sense

PR #2979, opened 2024-05-31
[release-1.12] Enable VM live update rollout strategy always

tiraboschi:
This PR enables the KubeVirt VM rollout strategy LiveUpdate, allowing changes to VM specs to dynamically propagate to their VMI.
Rollout strategy LiveUpdate is enabled using VMLiveUpdateFeatures Feature Gate and not a configuration knob.
This because, from Kubevirt side, the feature is still protected by a FG, therefore it is still considered not stable in all the contexts and can potentially 1) be dropped, 2) changed in a breaking manner.
So the FeatureGate is still there and no configuration knob is exposed.
Nevertheless, according to the results of our tests, we can consider it as suitable for this use case and so we can handle it as part of the HCO opinionated deployment.
Eventual breaking changes will be evaluated and handled on the HCO side when consuming a future version of Kubevirt.

This is just a manual cherry-pick of 2846

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-36370
```

**Release note**:

```release-note
KubeVirt VM rollout strategy LiveUpdate is now always enabled
```


tiraboschi:
closing for #2986

PR #2788, opened 2024-02-01
Bump AAQ to v1.1.4, and rename the component name

nunnatsa:
## What this PR does / why we need it
AAQ now called "application-aware-quota". One of the configuration field name also renamed.

This PR changes the import path of AAQ to the new path, bumps it to v1.1.4, and rename the component and the
`enableClusterAppsResourceQuota` field to
`allowApplicationAwareClusterResourceQuota`, to match the rename in AAQ.

**Reviewer Checklist**

> Reviewers are supposed to review the PR for every aspect below one by one. To check an item means the PR is either "OK" or "Not Applicable" in terms of that item. All items are supposed to be checked before merging a PR. 

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
- [ ] Troubleshooting Friendly

**Jira Ticket**:

```jira-ticket
None
```

**Release note**:

```release-note
Bump AAQ to v1.1.4
```

PR #3019, opened 2024-07-09
Make `applicationAwareConfig` fields optional

orenc1:
In order for the UI not to render a default value for `applicationAwareConfig`subfields for the `HyperConverged` custom resource, those should be set as optional as well.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-40208
```

**Release note**:

```release-note
Make applicationAwareConfig fields optional
```


nunnatsa:
@orenc1 I don't think this change will help. The problem is not in HCO, I think. As you can see, the CRD was not changed by your change, and the CRD is the input to the UI.

Also, it was never marked as as a required field in the CRD, so it's already optional.

orenc1:
> @orenc1 I don't think this change will help. The problem is not in HCO, I think. As you can see, the CRD was not changed by your change, and the CRD is the input to the UI.
> 
> Also, it was never marked as as a required field in the CRD, so it's already optional.

but `permittedHostDevices` and its subfields are also marked as optional, and they are not appearing by default in the UI. how it is possible?

nunnatsa:
> permittedHostDevices

I can guess it's because these fields are array (slices), and this one is boolean.

This is not an HCO bug

orenc1:
> > permittedHostDevices
> 
> I can guess it's because these fields are array (slices), and this one is boolean.
> 
> Anyway, this is not an HCO bug.

I think it's because we're setting default values.
I'm checking what happens when the default is not being set (PR is updated)

/hold

nunnatsa (changes requested):
Let's start over. The original implementation of activating aaq, was the existence of the aaq config object in the HyperConvered spec. 

Then, we found out that the ui, as result of a ui bug, always adds this object, activating aaq by default with no way to turn it off.


The bug is very clear: the ui adds optional objects if their fields have default values. This not specific to hco, but a general bug.


In order to workaround this bug, we already changed the original implementation: we added a new feature gate and now this feature gate is the enabler of the aaq. We no longer care if the ui adds the aaq config object, if the feature gate is false, and so we don't care anymore about this ui bug.


Bottom line, this pr is not needed.

nunnatsa:
See #2788

PR #3035, opened 2024-07-31
kubevirt-console-plugin: Use two replicas in case of highly available cluster

orenc1:
In case the cluster has an HighlyAvailable infrastructure, the kubevirt-console-plugin and kubevirt-console-proxy should be highly available as well, in order to avoid losing the Virtualization plugin when the hosting node is down. By running these pods with two replicas and PodAntiAffinity rule, we can ensure the plugin will remain available when a node is down.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-45427
```

**Release note**:

```release-note
Adjust kubevirt-console-plugin for high availability
```


nunnatsa (commented):
Looks good!

Please just add some tests to cover this change.

orenc1:
> Looks good!
> 
> Please just add some tests to cover this change.

Done, unit test cases have been added to cover the automatic PodAntiAffinity and replicas configuration.

nunnatsa (commented):
Added several inline comments. In addition, WDYT about adding two very short e2e tests, one with `Label(tests.SingleNodeLabel)` and the other with  `Label(tests.HighlyAvailableClusterLabel)`, to check how many pods are running and/or the deployment's `replicas` field?

controllers/operands/kubevirtConsolePlugin_test.go
```diff
@@ -686,6 +686,84 @@ var _ = Describe("Kubevirt Console Plugin", func() {
...
+				Entry("plugin deployment", hcoutil.AppComponentUIPlugin, NewKvUIPluginDeployment, newKvUIPluginDeploymentHandler),
+				Entry("proxy deployment", hcoutil.AppComponentUIProxy, NewKvUIProxyDeployment, newKvUIProxyDeploymentHandler),
+			)
+
+			DescribeTable("use one replica on SNO", func(appComponent hcoutil.AppComponent,
+				deploymentManifestor func(converged *hcov1beta1.HyperConverged) *appsv1.Deployment, handlerFunc GetHandler) {
+
+				hcoutil.GetClusterInfo = func() hcoutil.ClusterInfo {
+					return &commontestutils.ClusterInfoSNOMock{}
+				}
```
nunnatsa, line 737: Please restore `hcoutil.GetClusterInfo` - as above

controllers/operands/kubevirtConsolePlugin.go
```diff
@@ -195,12 +204,37 @@ func getKvUIDeployment(hc *hcov1beta1.HyperConverged, deploymentName string, ima
...
+						},
+						TopologyKey: corev1.LabelHostname,
+					},
+				},
+			},
+		}
+	}
+
+	return nil
+}
```
nunnatsa, line 236: not sure I want this function here. we only use it in the new tests. let's move it there

nunnatsa (commented):
Two more inline comments

controllers/operands/kubevirtConsolePlugin.go
orenc1: where should i move it to? `testUtils.go` ?

nunnatsa: to the only file that uses it

orenc1: could you please suggest where this function should be move into?
note that it isn't used by tests only.

orenc1: >to the only file that uses it

the only usage of that function is in the same file - `kubevirtConsolePlugin.go`

nunnatsa: sorry. misread

controllers/commontestutils/testUtils.go
```diff
@@ -141,6 +142,27 @@ func NewOtherNodePlacement() *sdkapi.NodePlacement {
...
+							},
+						},
+					},
+					TopologyKey: corev1.LabelHostname,
+				},
+			},
+		},
+	}
+}
+
```
nunnatsa, line 165: oh, this is the function I really want to comment about. sorry, these two functions share similar visual (:

So I would move **this one** to the file where it actually used, and make it also not exported.

orenc1: makes sense.
the function has been moved :+1:

PR #3045, opened 2024-08-11
DEBUG: vm e2e test fails many times

nunnatsa:
/hold


nunnatsa:
Fixed in #3055 

/close

PR #3057, opened 2024-08-18
Fix the monitoring flaky functest

nunnatsa:
The test modifies the KubeVirt CR and expects that the kubevirt_hco_out_of_band_modifications_total metric will increased by 1.
However, this test is sometimes fails, when the actual result is greater by 1 than the expected metric value.

The reason for that is that another test - labels test, also modifies the KubeVirt CR. In cases where the labels test runs first, sometimes prometheus takes some time to update the metric, and then, in the monitoring test, when reading the value before test, we're getting the value before it was modified by the labels test, and when reading the value after the KubeVirt modification, we get the metric values for both modifications.

This PR fixes the issue by using the CDI CR instead of the KubeVirt CR in the labels test, so there is only one modification for the KubeVirt CR during the functional test suite, and the test should always pass.

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-46588
```

**Release note**:

```release-note
None
```

### DIFF
--- a/tests/func-tests/monitoring_test.go
+++ b/tests/func-tests/monitoring_test.go
@@ -39,7 +39,7 @@ const (
 	criticalImpact
 )
 
-var _ = Describe("[crit:high][vendor:cnv-qe@redhat.com][level:system]Monitoring", Serial, Ordered, Label(tests.OpenshiftLabel), func() {
+var _ = Describe("[crit:high][vendor:cnv-qe@redhat.com][level:system]Monitoring", Serial, Ordered, Label(tests.OpenshiftLabel, "monitoring"), func() {
 	flag.Parse()
 
 	var (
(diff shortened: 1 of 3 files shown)


tests/func-tests/monitoring_test.go
```diff
@@ -39,7 +39,7 @@ const (
 	criticalImpact
 )
 
-var _ = Describe("[crit:high][vendor:cnv-qe@redhat.com][level:system]Monitoring", Serial, Ordered, Label(tests.OpenshiftLabel), func() {
+var _ = Describe("[crit:high][vendor:cnv-qe@redhat.com][level:system]Monitoring", Serial, Ordered, Label(tests.OpenshiftLabel, "monitoring"), func() {
```
machadovilaca, line 42: if we start using this for more tests, can we replace `"monitoring"` with something like `tests.MonitoringLabel` or so?

nunnatsa: Do we have additional monitoring tests? if yes, I'll do it in a followup PR.

machadovilaca: https://github.com/kubevirt/hyperconverged-cluster-operator/blob/015e6f76e2f95faafabb532811bdc85aa4fa10c0/tests/func-tests/observability_controller_test.go#L16 could belong to it I think

tiraboschi (approved):

PR #3063, opened 2024-08-22
Improve the monitoring functional test

nunnatsa:
The test now reads the kubevirt_hco_out_of_band_modifications_total metric before modifying the KubeVirt CR. When reading from prometheus, it can take some time for it to update from the HCO endpoints. That leads to flaky behavior, because the initial value may be wrong and not up-to-date.

This PR reads the initial value directly from the HCO pod, by adding a temporary route for the HCO prometheus endpoint, so it is always up-to-date.

The monitoring test can only run on OpenShift clusters. That allow us to use routers.

**Jira Ticket**:

```jira-ticket
None
```

**Release note**:

```release-note
None
```

### DIFF
--- a/tests/func-tests/monitoring_test.go
+++ b/tests/func-tests/monitoring_test.go
@@ -49,6 +49,7 @@ var _ = Describe("[crit:high][vendor:cnv-qe@redhat.com][level:system]Monitoring"
 		promClient                       promApiv1.API
 		prometheusRule                   monitoringv1.PrometheusRule
 		initialOperatorHealthMetricValue float64
+		tempRouteURL                     string
 	)
 
 	runbookClient.Timeout = time.Second * 3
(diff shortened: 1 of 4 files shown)


tests/func-tests/monitoring_test.go
```diff
@@ -63,6 +68,37 @@ var _ = Describe("[crit:high][vendor:cnv-qe@redhat.com][level:system]Monitoring"
 		prometheusRule = getPrometheusRule(ctx, restClient)
 
 		initialOperatorHealthMetricValue = getMetricValue(ctx, promClient, "kubevirt_hyperconverged_operator_health_status")
+
+		By("adding temporary route")
```
tiraboschi, line 72: Please (conditionally) do it only once for the whole test suite in the `BeforeSuite` node with a `DeferCleanup(...)` to clean it up at the end.

tests/func-tests/monitoring_test.go
```diff
@@ -283,3 +340,96 @@ func getPrometheusURL(ctx context.Context, cli rest.Interface) string {
...
+
+func getHCOMetric(ctx context.Context, url, query string) (float64, error) {
+	req, err := http.NewRequest(http.MethodGet, url, nil)
+	if err != nil {
+		return 0, fmt.Errorf("failed to create request: %w", err)
+	}
+
+	httpClient := http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
```
tiraboschi, line 410: can we extract the CAFile from the `rest.Config` object used to initialize the `client.Client` we are using to talk with the APIServer and trust this?

nunnatsa: done

tests/func-tests/monitoring_test.go
nunnatsa: done

tests/func-tests/monitoring_test.go
```diff
@@ -99,17 +113,22 @@ var _ = Describe("[crit:high][vendor:cnv-qe@redhat.com][level:system]Monitoring"
...
 
+		By(fmt.Sprintf("Reading the `%s` metric from HCO prometheus endpoint", query))
 		var valueBefore float64
 		Eventually(func(g Gomega, ctx context.Context) {
-			valueBefore = getMetricValue(ctx, promClient, query)
+			var err error
+			valueBefore, err = tests.GetHCOMetric(ctx, tempRouteURL, query)
+			g.Expect(err).NotTo(HaveOccurred())
 		}).WithTimeout(10 * time.Second).WithPolling(500 * time.Millisecond).WithContext(ctx).Should(Succeed())
+		GinkgoWriter.Printf("The metric value before the test is: %0.2f\n", valueBefore)
```
tiraboschi, line 129: Do we want to merge this?

tiraboschi: ok

tiraboschi (approved):

PR #3064, opened 2024-08-22
deploy: Fix script for external providers

oshoval:
When using external provider which is not OpenShift, make cluster-sync will fail.
Updating the condition.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
NONE
```

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/deploy.sh
+++ b/hack/deploy.sh
@@ -72,7 +72,7 @@ for namespace in ${namespaces[@]}; do
     fi
 done
 
-if [ "${CMD}" == "oc" ]; then
+if [ "${IS_OPENSHIFT}" == "true" ]; then
     # Switch project to kubevirt-hyperconverged
     oc project "${HCO_NAMESPACE}"
 else


tiraboschi (approved):

hack/deploy.sh
```diff
@@ -72,7 +72,7 @@ for namespace in ${namespaces[@]}; do
     fi
 done
 
-if [ "${CMD}" == "oc" ]; then
+if [ "${IS_OPENSHIFT}" == "true" ]; then
     # Switch project to kubevirt-hyperconverged
     oc project "${HCO_NAMESPACE}"
```
oshoval, line 77: Basically instead `oc` we can use here `${CMD}` (which can be either kubectl or oc)
but for now we are good, it just mean `oc` binary should exists for OpenShift clusters

PR #3087, opened 2024-09-02
API: Use new token generation SSP API and remove feature gate

akrejcir:
The token generation API was stabilized in the SSP, and feature gate was removed:
https://github.com/kubevirt/ssp-operator/pull/1018

This PR removes the same feature gate from HCO, and adds a new field in the .spec to enable this feature.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-45065
```

**Release note**:

```release-note
- The feature gate "deployVmConsoleProxy" was deprecated and is now ignored.
- Added a new field ".spec.enableTokenGenerationApi" to enable the token generation API.
```

### DIFF
--- a/controllers/operands/ssp.go
+++ b/controllers/operands/ssp.go
@@ -158,7 +158,9 @@ func NewSSP(hc *hcov1beta1.HyperConverged, opts ...string) (*sspv1beta2.SSP, []h
 	}
 
 	if hc.Spec.FeatureGates.DeployVMConsoleProxy != nil {
-		spec.FeatureGates.DeployVmConsoleProxy = *hc.Spec.FeatureGates.DeployVMConsoleProxy
+		spec.TokenGenerationService = &sspv1beta2.TokenGenerationService{
+			Enabled: *hc.Spec.FeatureGates.DeployVMConsoleProxy,
+		}
 	}
 
 	// Disable common-instancetypes deployment by SSP from 4.16, now handled by virt-operator
--- a/controllers/operands/ssp_test.go
+++ b/controllers/operands/ssp_test.go
@@ -196,7 +196,8 @@ var _ = Describe("SSP Operands", func() {
 			expectedResource, _, err := NewSSP(hco)
 			Expect(err).ToNot(HaveOccurred())
 
-			Expect(expectedResource.Spec.FeatureGates.DeployVmConsoleProxy).To(BeTrue())
+			Expect(expectedResource.Spec.TokenGenerationService).ToNot(BeNil())
+			Expect(expectedResource.Spec.TokenGenerationService.Enabled).To(BeTrue())
 		})
 
 		It("should create with deployCommonInstancetypes feature gate disabled", func() {
--- a/docs/cluster-configuration.md
+++ b/docs/cluster-configuration.md
@@ -166,9 +166,6 @@ the [dataImportCronTemplates field](#configure-custom-golden-images), even if th
 Set the `deployVmConsoleProxy` feature gate to true to allow SSP operator to deploy its resources. SSP operator will 
 deploy a proxy that provides an access to the VNC console of a KubeVirt Virtual Machine (VM).
 
-**Note**: Once `deployVmConsoleProxy` is set to true, SSP operator will not delete deployed resources if `deployVmConsoleProxy` is 
-reverted back to false.
-
 **Default**: `false`
 
 ### deployKubeSecondaryDNS Feature Gate


akrejcir:
> Thanks for the PR @akrejcir
> 
> What will happen on upgrade, if the `DeployVMConsoleProxy` FG is set?

Good point, I did not consider update.

Does HCO have functionality to run some logic only during update? It would be good, if we can copy the value of the feature gate to the new field in `.spec`. But only during update, because in the new version, the feature gate will be ignored.

nunnatsa:
> > Thanks for the PR @akrejcir
> > What will happen on upgrade, if the `DeployVMConsoleProxy` FG is set?
> 
> Good point, I did not consider update.
> 
> Does HCO have functionality to run some logic only during update? It would be good, if we can copy the value of the feature gate to the new field in `.spec`. But only during update, because in the new version, the feature gate will be ignored.

HCO does have code that only run during upgrade, but we need to avoid modifying the HyperConverged spec. Another alternative is the mutation webhook. @tiraboschi WDYT?

tiraboschi:
> > Good point, I did not consider update.
> > Does HCO have functionality to run some logic only during update? It would be good, if we can copy the value of the feature gate to the new field in `.spec`. But only during update, because in the new version, the feature gate will be ignored.
> 
> HCO does have code that only run during upgrade, but we need to avoid modifying the HyperConverged spec. Another alternative is the mutation webhook. @tiraboschi WDYT?

In `assets/upgradePatches.json` we are accumulating a set of jsonpatches to be executed during the upgrades (according to semver matching rules) on the spec stanza of the HCO CR.
`move` and `copy` operations are also supported. Please take a look at https://github.com/evanphx/json-patch/blob/master/v5/patch_test.go for examples.
I'd suggest to add there an upgradePatch to handle it during upgrades without the need for additional code.

akrejcir:
> > > ...
> > 
> > HCO does have code that only run during upgrade, but we need to avoid modifying the HyperConverged spec. Another alternative is the mutation webhook. @tiraboschi WDYT?
> 
> In `assets/upgradePatches.json` we are accumulating a set of jsonpatches to be executed during the upgrades (according to semver matching rules) on the spec stanza of the HCO CR. `move` and `copy` operations are also supported. Please take a look at https://github.com/evanphx/json-patch/blob/master/v5/patch_test.go for examples. I'd suggest to add there an upgradePatch to handle it during upgrades without the need for additional code.

Thanks, I will add a patch there. 
Which version is checked by the `semverRange` field? Is it the version **from** which we are upgrading, or **to** which we are upgrading? It was not clear from the code.

akrejcir:
I'm not sure why the sanity check is failing. `make sanity` locally succeeds.

tests/func-tests/defaults_test.go
```diff
@@ -70,7 +70,6 @@ var _ = Describe("Check Default values", Label("defaults"), Serial, func() {
 			DeployKubeSecondaryDNS:           ptr.To(false),
 			DeployKubevirtIpamController:     ptr.To(false),
 			DeployTektonTaskResources:        ptr.To(false),
-			DeployVMConsoleProxy:             ptr.To(false),
```
akrejcir, line 73: I think the functional tests fail, because of this change, but I'm not sure how to fix it.
I've tried running focused functional tests locally, but it outputs a long log, where it is hard to find relevant lines about test runs.

akrejcir: I've kept the default value of `deployVMProxy` feature gate. It should fix the failing test.

akrejcir:
Can you please take another look?

nunnatsa:
@akrejcir, @tiraboschi 

I'm not sure about this PR. 

Do we really need a new enabler field to replace the existing feature gate? The term "feature gate" is wrongly used in HCO anyway, and we need to treat the feature gates as "enablers". I think this change will just create mess. 

We want to introduce a new API, to fix the wrong feature gate naming anyway.

I suggest to just replace the default value to true.

nunnatsa:
And @akrejcir - sorry for the late response.

akrejcir:
> @akrejcir, @tiraboschi
> 
> I'm not sure about this PR.
> 
> Do we really need a new enabler field to replace the existing feature gate? The term "feature gate" is wrongly used in HCO anyway, and we need to treat the feature gates as "enablers". I think this change will just create mess.
> 
> We want to introduce a new API, to fix the wrong feature gate naming anyway.
> 
> I suggest to just replace the default value to true.

Thanks for the feedback. I don't have any opinion on HCO API change. 
I think it is up to you to decide what should change. Then, I will update this PR.

assets/upgradePatches.json
```diff
@@ -40,6 +40,21 @@
...
+          "op": "test",
+          "path": "/spec/featureGates/deployVmConsoleProxy",
+          "value": false
+        },
+        {
+          "op": "add",
+          "path": "/spec/enableTokenGenerationApi",
+          "value": false
+        }
+      ]
```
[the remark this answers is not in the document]
akrejcir: I wanted to do this update with minimal surprise for the user. 
The problem is, as you've written, that we don't know if user decided to disable the feature, or just left it default.
If it was intentionally disabled, it would be a surprise to enable it on update.

But maybe the risk is low, and we can enable it.

tiraboschi:
> @akrejcir, @tiraboschi
> 
> I'm not sure about this PR.
> 
> Do we really need a new enabler field to replace the existing feature gate? The term "feature gate" is wrongly used in HCO anyway, and we need to treat the feature gates as "enablers". I think this change will just create mess.
> 
> We want to introduce a new API, to fix the wrong feature gate naming anyway.

I agree with this:
the name of the `feature gate` stanza is misleading but it's already used for other configuration knobs.
For the sake of consistency, let's continue using it until we will be able to move all of them with the next version of the API.

> I suggest to just replace the default value to true.

We need to understand if we simply want to propose a new default for future fresh deployments or if we want also to "silently" change it for all the cluster on upgrades.

akrejcir:
I've removed most of the code from this PR.
Now it only uses the new SSP API without changing the HCO API.
I also did not change the default value for the feature gate, because I incorrectly assumed that we want to enable it by default.
In the acceptance criteria of the epic https://issues.redhat.com/browse/CNV-34204 , it says that vm-console-proxy should not be enabled by default.

dominikholler:
>  For the sake of consistency, let's continue using it until we will be able to move all of them with the next version of the API.

@tiraboschi is there already an upstream github issue or an downstream epic to track this?

tiraboschi:
> > For the sake of consistency, let's continue using it until we will be able to move all of them with the next version of the API.
> 
> @tiraboschi is there already an upstream github issue or an downstream epic to track this?

Yes: https://issues.redhat.com/browse/CNV-47166 for the design phase, currently targeted to v4.18.
and https://issues.redhat.com/browse/CNV-17501 for the actual implementation, targeted to v4.19 (this requires a proper support of conversion webhook mechanism in OLMv1 to avoid any disruption on upgrades).

PR #3104, opened 2024-09-25
unit-tests: Invoke unit tests using go test directly instead of ginkgo cli

jschintag:
Saves having to install the gingko cli for unit tests. Additionally, this brings HCO more in line with how the kubevirt/cdi invokes unit-tests with go.
Lastly invoking the tests with ginkgo on s390x causes it to randomly hang forever.

Please note if you compare the output of the coverprofiles, that golang since 1.23 also adds uncovered packages to the output.
This means that with this change, packages without tests will now be reported as 0%.

Switches out the ginkgo cli command for invoking unit-tests with the equivalent go command.
This fixes the test hanging on s390x on random points when invoking it via ginkgo.

**Reviewer Checklist**

- [x] PR Message
- [x] Commit Messages
- [x] How to test
- [x] Unit Tests
- [x] Functional Tests
- [x] User Documentation
- [x] Developer Documentation
- [x] Upgrade Scenario
- [x] Uninstallation Scenario
- [x] Backward Compatibility

**Jira Ticket**:

```jira-ticket
NONE
```

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/build-tests.sh
+++ b/hack/build-tests.sh
@@ -8,14 +8,13 @@ JOB_TYPE="${JOB_TYPE:-}"
 if [ "${JOB_TYPE}" == "travis" ]; then
     go get -v -t ./...
     go install github.com/mattn/goveralls@latest
-    go install github.com/onsi/ginkgo/v2/ginkgo@$(grep github.com/onsi/ginkgo go.mod | cut -d " " -f2)
     go mod vendor
     PKG_PACKAGE_PATH="./pkg/"
     CONTROLLERS_PACKAGE_PATH="./controllers/"
     mkdir -p coverprofiles
     # Workaround - run tests on webhooks first to prevent failure when running all the test in the following line.
-    ginkgo -r ${PKG_PACKAGE_PATH}webhooks
-    ginkgo -cover -output-dir=./coverprofiles -coverprofile=cover.coverprofile -r ${PKG_PACKAGE_PATH} -r ${CONTROLLERS_PACKAGE_PATH}
+    go test ${PKG_PACKAGE_PATH}webhooks/...
+    go test -v -outputdir=./coverprofiles -coverprofile=cover.coverprofile ${PKG_PACKAGE_PATH}... ${CONTROLLERS_PACKAGE_PATH}...
 else
     test_path="./tests/func-tests"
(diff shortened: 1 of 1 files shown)


hack/build-tests.sh
```diff
@@ -8,14 +8,13 @@ JOB_TYPE="${JOB_TYPE:-}"
...
-    go install github.com/onsi/ginkgo/v2/ginkgo@$(grep github.com/onsi/ginkgo go.mod | cut -d " " -f2)
     go mod vendor
     PKG_PACKAGE_PATH="./pkg/"
     CONTROLLERS_PACKAGE_PATH="./controllers/"
     mkdir -p coverprofiles
     # Workaround - run tests on webhooks first to prevent failure when running all the test in the following line.
-    ginkgo -r ${PKG_PACKAGE_PATH}webhooks
-    ginkgo -cover -output-dir=./coverprofiles -coverprofile=cover.coverprofile -r ${PKG_PACKAGE_PATH} -r ${CONTROLLERS_PACKAGE_PATH}
+    go test ${PKG_PACKAGE_PATH}webhooks/...
+    go test -outputdir=./coverprofiles -coverprofile=cover.coverprofile ${PKG_PACKAGE_PATH}... ${CONTROLLERS_PACKAGE_PATH}...
```
nunnatsa, line 17: I wonder if we'll want to add some logs here. The default ginkgo logs are a bit more richer than go-test, but still short. WDYT about adding the `-v` option?

jschintag: Sure i can add it. There are not that much tests, so it should not spam the log.

jschintag:
> In general, I think it's a good idea, but beside installing the ginkgo cli, there is no meaningful benefit here. The tests are not similar to KubeVirt and CDI, because they are using bazel.
> 
> Added one inline comment

Yes, the main benefit and reason behind this PR is, that i want to add a job for running the unit tests on s390x.
Unfortunately when running them with gingko, my experience so far is that they just randomly hang. To be more specific, they hang on a different part each time, but it always hangs indefinitely.
This does not happen when using go directly.

nunnatsa:
golang now compute the coverage of all the code, not only the packages that are tested. That causes a valid reduce of covarage.

/override coverage/coveralls

nunnatsa (approved):
/approve

PR #3105, opened 2024-09-25
fix: Allow disabling of common data import crons

akrejcir:
Some of the common data import crons can be disabled. Before this commit, the disabled data import crons were still added to SSP.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-49037
```

**Release note**:

```release-note
None
```

### DIFF
--- a/controllers/operands/ssp.go
+++ b/controllers/operands/ssp.go
@@ -284,6 +284,10 @@ func isDataImportCronTemplateEnabled(dict hcov1beta1.DataImportCronTemplate) boo
 
 func getCustomDicts(list []hcov1beta1.DataImportCronTemplateStatus, crDicts map[string]hcov1beta1.DataImportCronTemplate) []hcov1beta1.DataImportCronTemplateStatus {
 	for dictName, crDict := range crDicts {
+		if !isDataImportCronTemplateEnabled(crDict) {
+			continue
+		}
+
 		if _, isCommon := dataImportCronTemplateHardCodedMap[dictName]; !isCommon {
 			list = append(list, hcov1beta1.DataImportCronTemplateStatus{
 				DataImportCronTemplate: *crDict.DeepCopy(),
--- a/controllers/operands/ssp_test.go
+++ b/controllers/operands/ssp_test.go
@@ -764,6 +764,29 @@ var _ = Describe("SSP Operands", func() {
 					Expect(goldenImageList).To(ContainElements(*statusImage2Enabled, statusImage3, statusImage4))
 				})
 
+				It("should not add user DIC template if it is disabled", func() {
+					dataImportCronTemplateHardCodedMap = nil
+					hco := commontestutils.NewHco()
+					hco.Spec.FeatureGates.EnableCommonBootImageImport = ptr.To(true)
+
+					disabledUserImage := image1.DeepCopy()
+					disableDict(disabledUserImage)
+					enabledUserImage := image2.DeepCopy()
+					enableDict(enabledUserImage)
+
+					hco.Spec.DataImportCronTemplates = []hcov1beta1.DataImportCronTemplate{*disabledUserImage, *enabledUserImage}
+					goldenImageList, err := getDataImportCronTemplates(hco)
+					Expect(err).ToNot(HaveOccurred())
+					Expect(goldenImageList).To(HaveLen(1))
+
+					statusImageEnabled := hcov1beta1.DataImportCronTemplateStatus{
+						DataImportCronTemplate: *enabledUserImage,
+						Status:                 hcov1beta1.DataImportCronStatus{},
+					}
+
+					Expect(goldenImageList).To(ContainElements(statusImageEnabled))
+				})
+
 				It("Should reject if the CR list contain DIC templates with the same name, when there are also common DIC templates", func() {
 					dataImportCronTemplateHardCodedMap = map[string]hcov1beta1.DataImportCronTemplate{
 						image1.Name: image1,


controllers/operands/ssp.go
```diff
@@ -256,10 +258,6 @@ func getCommonDicts(list []hcov1beta1.DataImportCronTemplateStatus, crDicts map[
 		}
 
 		if crDict, found := crDicts[dictName]; found {
-			if !isDataImportCronTemplateEnabled(crDict) {
-				continue
-			}
```
lyarwood, line 261: This would disable user provided DICs with the annotation that don't correspond to an internal DIC. I don't think that's an issue but it is a change in behaviour, as we've seen these would previously still be created.

lyarwood (commented):
The approach looks valid, needs tests?

nunnatsa:
@akrejcir, @lyarwood I wonder how did you test this. What steps did you made, to get this bug. As far as i know, it should work, and it's tested.

nunnatsa:
@akrejcir, @lyarwood if you tried to add the disable annotation the yaml files and you expected that it will disable the DICT, then this is not how the feature was designed. The concept of "optional DICT" is a new concept and is not implemented.

I don't think this PR is the best implementation for this new feature. We need to think about it.

controllers/operands/ssp.go
```diff
@@ -277,6 +277,23 @@ func getCommonDicts(list []hcov1beta1.DataImportCronTemplateStatus, crDicts map[
...
+
+	if *original == nil {
+		*original = map[string]string{}
+	}
+
+	for key, value := range updates {
+		(*original)[key] = value
+	}
+}
+
```
nunnatsa, line 296: please check the standard library `maps` package. I believe there are some built-in solutions for this.

controllers/operands/ssp_test.go
```diff
@@ -756,14 +756,37 @@ var _ = Describe("SSP Operands", func() {
...
+					goldenImageList, err := getDataImportCronTemplates(hco)
+					Expect(err).ToNot(HaveOccurred())
+					Expect(goldenImageList).To(HaveLen(1))
+
+					statusImage2Enabled := statusImage2.DeepCopy()
+					enableDict(&statusImage2Enabled.DataImportCronTemplate)
+
+					Expect(goldenImageList).To(ContainElements(*statusImage2Enabled))
+				})
+
```
nunnatsa, line 789: I think we need to test different scenario instead of this: allow disabling a user defined DICT.

controllers/operands/ssp.go
```diff
@@ -241,6 +241,8 @@ func getDataImportCronTemplates(hc *hcov1beta1.HyperConverged) ([]hcov1beta1.Dat
 	}
 	dictList = getCustomDicts(dictList, crDicts)
 
+	dictList = filterEnabledDicts(dictList)
+
 	sort.Sort(dataImportTemplateSlice(dictList))
```
nunnatsa, line 246: I would keep the filtering within `getCommonDicts` and add it also to `getCustomDicts`, avoiding building the list and then filter it, but build a pre-filtered list instead. WDYT? IMHO, it will save some memory and improve performance.

controllers/operands/ssp.go
```diff
@@ -297,6 +314,22 @@ func getCustomDicts(list []hcov1beta1.DataImportCronTemplateStatus, crDicts map[
...
+	for _, dictStatus := range list {
+		if !isDataImportCronTemplateEnabled(dictStatus.DataImportCronTemplate) {
+			continue
+		}
+		result = append(result, dictStatus)
+	}
+
+	return result
+}
+
```
nunnatsa, line 332: It will be so much easy and nice code in golang 1.23... we'll get there :)

nunnatsa (commented):
Thanks @akrejcir. It looks good in general. Added a few inline comments.

controllers/operands/ssp.go
akrejcir: I've added a call to `maps.Copy()` but it does not create the map if it is `nil`.

controllers/operands/ssp.go
akrejcir: Yes, iterators will be nice.

controllers/operands/ssp.go
akrejcir: Done

controllers/operands/ssp.go
nunnatsa: I know, but then you only need to check this case, and then run `maps.Copy()` or `make(map[string]string` if it's nil

controllers/operands/ssp.go
```diff
@@ -256,34 +257,50 @@ func getCommonDicts(list []hcov1beta1.DataImportCronTemplateStatus, crDicts map[
...
+		if !isDataImportCronTemplateEnabled(targetDict.DataImportCronTemplate) {
+			continue
+		}
 		list = append(list, targetDict)
 	}
 
 	return list
 }
 
+func updateOrCreateMap[M1 ~map[K]V, M2 ~map[K]V, K comparable, V any](originalPtr *M1, updates M2) {
```
lyarwood, line 282: Why do we need any of this complexity when we know both callers just handle `map[string]string`?

controllers/operands/ssp_test.go
```diff
@@ -756,14 +756,60 @@ var _ = Describe("SSP Operands", func() {
...
-					Expect(goldenImageList).To(HaveCap(4))
 
 					statusImage2Enabled := statusImage2.DeepCopy()
+					enableDict(&statusImage2Enabled.DataImportCronTemplate)
 					statusImage2Enabled.Status.Modified = true
 
 					Expect(goldenImageList).To(ContainElements(*statusImage2Enabled, statusImage3, statusImage4))
 				})
 
+				It("Should not add a common DIC template if it is disabled in the file", func() {
```
lyarwood, line 767: Yeah I'm still not convinced by this behaviour, we shouldn't be marking internal DICTs as disabled and should just remove them so this just smells like something that could trip folks up with the process in the future.

akrejcir: Should I remove the test?

controllers/operands/ssp.go
akrejcir: removed

controllers/operands/ssp.go
akrejcir: done

controllers/operands/ssp_test.go
nunnatsa: Yes, please do. I don't think having such DICTs in the HCO image is a valid use-case.

akrejcir: ok, done

controllers/operands/ssp.go
```diff
@@ -256,21 +257,31 @@ func getCommonDicts(list []hcov1beta1.DataImportCronTemplateStatus, crDicts map[
...
+
 			targetDict.ObjectMeta.Namespace = crDict.Namespace
 			targetDict.Status.Modified = true
 		} else if ns := hc.Spec.CommonBootImageNamespace; ns != nil && len(*ns) > 0 {
 			targetDict.ObjectMeta.Namespace = *ns
 		}
 
+		if !isDataImportCronTemplateEnabled(targetDict.DataImportCronTemplate) {
+			continue
+		}
```
nunnatsa, line 284: If we say that we currently not supporting disabled common DICTs (from file), and I think we do, then we need to rollback these changes.

We only need to add filtering to the user defined DICTs.

akrejcir: done

lyarwood (commented):
/lgtm

nunnatsa:
Thanks @akrejcir !!

PR #3106, opened 2024-09-25
pkg/monitoring/metrics: add new alert for vms using outdated machine type

dasionov:

This PR introduces a new Prometheus alert for virtual machines (VMs) in the cluster that require a machine type update. For instance, VMs configured with the machine type set to RHEL 8 would be flagged as outdated. This is because the base image of the virt-launcher will transition to RHEL 10, which introduces breaking changes incompatible with older machine types.

Depends-On #3132, https://github.com/kubevirt/kubevirt/pull/13010

**Reviewer Checklist**

- [x] PR Message
- [x] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [x] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Release note**:
```release-note
None
```

### DIFF
--- a/pkg/monitoring/rules/alerts/operator_alerts.go
+++ b/pkg/monitoring/rules/alerts/operator_alerts.go
@@ -1,23 +1,31 @@
 package alerts
 
 import (
+	"fmt"
+	"os"
+	"strconv"
+	"strings"
+
 	promv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/utils/ptr"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 const (
-	outOfBandUpdateAlert          = "KubeVirtCRModified"
-	unsafeModificationAlert       = "UnsupportedHCOModification"
-	installationNotCompletedAlert = "HCOInstallationIncomplete"
-	singleStackIPv6Alert          = "SingleStackIPv6Unsupported"
-	MisconfiguredDeschedulerAlert = "HCOMisconfiguredDescheduler"
-	severityAlertLabelKey         = "severity"
-	healthImpactAlertLabelKey     = "operator_health_impact"
+	outOfBandUpdateAlert              = "KubeVirtCRModified"
+	unsafeModificationAlert           = "UnsupportedHCOModification"
+	installationNotCompletedAlert     = "HCOInstallationIncomplete"
+	singleStackIPv6Alert              = "SingleStackIPv6Unsupported"
+	MisconfiguredDeschedulerAlert     = "HCOMisconfiguredDescheduler"
+	VMOutdatedMachineTypeAlert        = "VMHasOutdatedMachineType"
+	minSupportedVirtLauncherOSVersion = 8
+	severityAlertLabelKey             = "severity"
+	healthImpactAlertLabelKey         = "operator_health_impact"
 )
 
 func operatorAlerts() []promv1.Rule {
-	return []promv1.Rule{
+	rules := []promv1.Rule{
 		{
 			Alert: outOfBandUpdateAlert,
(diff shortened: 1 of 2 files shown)


nunnatsa:
Is HCO the right place for th his metric? HCO does not know VMs at all and dhould not monitor them.

/hold

pkg/monitoring/rules/alerts/operator_alerts.go
```diff
@@ -80,4 +88,33 @@ func operatorAlerts() []promv1.Rule {
...
+			Annotations: map[string]string{
+				"description": "There are virtual machines using an outdated machine type that need to be patched.",
+				"summary":     "{{ $value }} virtual machines are using an outdated machine type.",
+			},
+			Labels: map[string]string{
+				severityAlertLabelKey:     "warning",
+				healthImpactAlertLabelKey: "none",
+			},
+		})
+	}
```
nunnatsa, line 117: I find the `return rules` lines confusing.

Let's pack all of this into a new function that will return rule and boolean for "rule created", and if it's true - append the new rule to the list.

dasionov: sounds good, done.

machadovilaca:
> Is HCO the right place for th his metric? HCO does not know VMs at all and dhould not monitor them.
> 
> /hold

and also I have some concerns if it makes sense to check for RHEL versions here

pkg/monitoring/rules/alerts/operator_alerts.go
```diff
@@ -1,23 +1,31 @@
...
-	MisconfiguredDeschedulerAlert = "HCOMisconfiguredDescheduler"
-	severityAlertLabelKey         = "severity"
-	healthImpactAlertLabelKey     = "operator_health_impact"
+	outOfBandUpdateAlert              = "KubeVirtCRModified"
+	unsafeModificationAlert           = "UnsupportedHCOModification"
+	installationNotCompletedAlert     = "HCOInstallationIncomplete"
+	singleStackIPv6Alert              = "SingleStackIPv6Unsupported"
+	MisconfiguredDeschedulerAlert     = "HCOMisconfiguredDescheduler"
+	VMOutdatedMachineTypeAlert        = "VMHasOutdatedMachineType"
+	minSupportedVirtLauncherOSVersion = 8
```
machadovilaca, line 22: nit: also mention that it's a RHEL version

pkg/monitoring/rules/alerts/operator_alerts.go
```diff
@@ -80,4 +88,51 @@ func operatorAlerts() []promv1.Rule {
...
+
+	if rule, created := createVMOutdatedMachineTypeRule(); created {
+		rules = append(rules, rule)
+	}
+
+	return rules
+}
+
+func createVMOutdatedMachineTypeRule() (promv1.Rule, bool) {
+	logger := logf.Log.WithName("operator-alerts")
```
machadovilaca, line 100: nit: make it a file variable

pkg/monitoring/rules/alerts/operator_alerts.go
```diff
@@ -80,4 +88,51 @@ func operatorAlerts() []promv1.Rule {
 			},
 		},
 	}
+
+	if rule, created := createVMOutdatedMachineTypeRule(); created {
```
machadovilaca, line 92: the alert should always exist, and the condition for firing should be in the alert expression

pkg/monitoring/rules/alerts/operator_alerts.go
```diff
@@ -80,4 +88,51 @@ func operatorAlerts() []promv1.Rule {
...
+		return promv1.Rule{}, false
+	}
+
+	virtLauncherOSVersion, err := strconv.Atoi(rhelVersion)
+	if err != nil {
+		logger.Error(err, "Error parsing VIRT_LAUNCHER_OS_VERSION")
+		return promv1.Rule{}, false
+	}
+
+	if virtLauncherOSVersion > minSupportedVirtLauncherOSVersion {
```
machadovilaca, line 112: can we create a new metric with the virt launcher version
with the os and version labels? and then use it in the alert expression

machadovilaca (changes requested):
please also add a release note

PR #3111, opened 2024-09-26
Fix the manual run of the bump bot, for specific branch

nunnatsa:
If running the bump bot on a specific branch, the bot will still run for all the release branches.

After this fix, only if running on main, it will run for all branches.

This change should be back ported to all the release branches in order to take affect.

```release-note
None
```


orenc1 (approved):
/approve

PR #3136, opened 2024-10-10
Manage Creation of UI configuration ConfigMaps by HCO

orenc1:
Currently, when kubevirt-hco is installed, the console-plugin configuration configmaps are created only when a cluster-admin opens the Virtualization tab for the first time. At that point two ConfigMaps are created in the installation namespace:
- kubevirt-user-settings
- kubevirt-ui-features

And also their RBACs to be reachable by regular users are created. Without login by an admin, the config maps are missing, and regular users cannot use the UI at that point.

This PR adds logic to create those configmaps and associated RBACs, if they are not already exist.
Also, remove the ConfigMaps from their previous location at the default namespace, used in v1.12.0 and below.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-49492
```

**Release note**:

```release-note
Manage Creation of UI configuration ConfigMaps by HCO
```

### DIFF
--- a/assets/upgradePatches.json
+++ b/assets/upgradePatches.json
@@ -162,6 +162,78 @@
         "name": "mtqs.mtq.kubevirt.io",
         "namespace": ""
       }
+    },
+    {
+      "semverRange": "<1.14.0",
+      "groupVersionKind": {
+        "group": "",
+        "version": "v1",
+        "kind": "ConfigMap"
+      },
+      "objectKey": {
+        "name": "kubevirt-ui-features",
+        "namespace": "default"
+      }
+    },
+    {
+      "semverRange": "<1.14.0",
+      "groupVersionKind": {
+        "group": "",
+        "version": "v1",
+        "kind": "ConfigMap"
+      },
+      "objectKey": {
+        "name": "kubevirt-user-settings",
+        "namespace": "default"
+      }
+    },
+    {
+      "semverRange": "<1.14.0",
+      "groupVersionKind": {
+        "group": "rbac.authorization.k8s.io",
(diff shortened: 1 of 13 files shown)


assets/upgradePatches.json
```diff
@@ -162,6 +162,18 @@
...
+      "semverRange": "<1.14.0",
+      "groupVersionKind": {
+        "group": "",
+        "version": "v1",
+        "kind": "ConfigMap"
+      },
+      "objectKey": {
+        "name": "kubevirt-ui-features",
+        "namespace": "default"
+      }
```
nunnatsa, line 176: why do we need this?

nunnatsa (commented):
One inline question

assets/upgradePatches.json
orenc1: on the other hand, I see I forgot to add the removal of the second CM (`kubevirt-user-settings`) and their Roles and RoleBindings. 6 objects in total. Do you think we should remove all of them? or just leave them as is?

nunnatsa: oh, I missed the `objectsToBeRemoved` - it was out  side of my screen... So yes, please remove everything.

nunnatsa (approved):
/approve

PR #3138, opened 2024-10-11
Skip observability controller tests if no SA TLS cert

machadovilaca:

Stop tests from failing if setups without SA TLS certificates

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
none
```

**Release note**:

```release-note
none
```

### DIFF
--- a/tests/func-tests/observability_controller_test.go
+++ b/tests/func-tests/observability_controller_test.go
@@ -2,6 +2,8 @@ package tests_test
 
 import (
 	"context"
+	"errors"
+	"os"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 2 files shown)


nunnatsa (commented):
@machadovilaca we try to avoid uding skip from inside the test. Instead, please add a label. The test should be running with the right labels, according to the specific environment, that is know to th he test user.

Instead of fix, add fail, to let the users know witch labels should be used.

See many examples in the func test code.

machadovilaca:
> @machadovilaca we try to avoid uding skip from inside the test. Instead, please add a label. The test should be running with the right labels, according to the specific environment, that is know to th he test user.
> 
> Instead of fix, add fail, to let the users know witch labels should be used.
> 
> See many examples in the func test code.

@nunnatsa updated
is this what you meant?

tests/func-tests/observability_controller_test.go
```diff
@@ -55,3 +68,13 @@ var _ = Describe("Observability Controller", Label(tests.OpenshiftLabel, "observ
 		})
 	})
 })
+
+func serviceAccountTlsCertPathExists() bool {
+	_, err := os.ReadFile(observability.ServiceAccountTlsCertPath)
```
nunnatsa, line 73: No need to actually read the file. you can use `os.Stat()` for that; e.g.
 ```suggestion
	_, err := os.Stat(observability.ServiceAccountTlsCertPath)
```

File: tests/func-tests/observability_controller_test.go [hunk 2]
```diff
@@ -55,3 +68,13 @@ var _ = Describe("Observability Controller", Label(tests.OpenshiftLabel, "observ
...
+	if errors.Is(err, os.ErrNotExist) {
+		return false
+	}
+	Expect(err).ToNot(HaveOccurred())
```
nunnatsa, line 77: Please avoid using gomega asserion in helper function (in new code...).

Please return the error, and asset it from the caller code; e.g. change the function signature to return bool and error.

nunnatsa (commented):
Added a few inline comments

tests/func-tests/observability_controller_test.go
machadovilaca: updated

tests/func-tests/observability_controller_test.go
machadovilaca: updated

nunnatsa (approved):
/approve

PR #3139, opened 2024-10-13
UDN: Add managedTap to Kubevirt CR

oshoval:

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket

```

**Release note**:

```release-note

```


controllers/operands/kubevirt_test.go
```diff
@@ -2165,7 +2165,8 @@ Version: 1.2.3`)
 							},
 						},
 					}
-					Expect(existingResource.Spec.Configuration.NetworkConfiguration.Binding[primaryUDNNetworkBindingName]).To(Equal(expectedInterfaceBindingPlugin))
+					Expect(existingResource.Spec.Configuration.NetworkConfiguration.Binding[primaryUDNNetworkBindingName]).To(Equal(expectedPasstInterfaceBindingPlugin))
+					Expect(existingResource.Spec.Configuration.NetworkConfiguration.Binding[managedTapUDNNetworkBindingName]).To(Equal(managedTapUserDefinedNetworkBinding()))
```
nunnatsa, line 2169: Please use the `HaveKeyWithValue` gomega matcher

oshoval: done

oshoval:
addressed comments, ptal

qinqon:
Going to create a new PR doing the opposite, so the userDefinedPrimaryNetworkBinding will point to managedTap, we can test later on passt by installing HCO using the json capababilities so we patch the binding and change from managedTap to passt, no need for new binding.

controllers/operands/kubevirt.go
```diff
@@ -47,7 +47,8 @@ const (
 )
 
 const (
-	primaryUDNNetworkBindingName = "passt"
+	primaryUDNNetworkBindingName    = "passt"
+	managedTapUDNNetworkBindingName = "managedTap"
```
qinqon, line 51: I was wrong, we should change the `primaryUDNNetworkBindingName` to `managedTap`, if we want to test passt we just use HCO json patching for kubevirt CR and create manually the NAD, at the end is a development workflow.

Also keeping the same name for the binding is important from the beginning so we don't have to modify VM/VMI to upgrade from managedTap to passt

qinqon (changes requested):

controllers/operands/kubevirt.go
oshoval: We can rename vars here but dont need to drop old passt support atm, 
the CR can hold both and it will allow smoother QE adaption
same for NAD on CNAO and all, cleaning will be on a follow-up according needs

for now we just need to think about the generic name (unless this also can be done in phases)
lets continue please the design that was started offline

oshoval:
TODO - need first to add `DynamicPodInterfaceNaming to the FG

oshoval:
rebased

oshoval:
changes https://github.com/kubevirt/hyperconverged-cluster-operator/pull/3139#issuecomment-2440037275

oshoval:
addded FG UT

controllers/operands/kubevirt.go
oshoval: switched the primary one to be managedTap
renamed it to have generic name `defaultPodNetworkBinding`

controllers/operands/kubevirt.go
```diff
@@ -47,7 +47,9 @@ const (
 )
 
 const (
-	primaryUDNNetworkBindingName = "passt"
+	primaryUDNNetworkBindingName = "defaultPodNetworkBinding"
+	managedTapNetworkBindingName = "managedTap" // kubevirt internal name
```
oshoval, line 51: once kubevirt is bumped on HCO, we can use the types.go one

controllers/operands/kubevirt_test.go
```diff
@@ -2142,6 +2142,45 @@ Version: 1.2.3`)
 					Expect(existingResource.Annotations).ToNot(HaveKey(kubevirtcorev1.EmulatorThreadCompleteToEvenParity))
 				})
 
+				It("should add the DynamicPodInterfaceNaming feature gate if PrimaryUserDefinedNetworkBinding is true in HyperConverged CR", func() {
+					hco.Spec.FeatureGates = hcov1beta1.HyperConvergedFeatureGates{
+						PrimaryUserDefinedNetworkBinding: ptr.To(true),
```
oshoval, line 2147: TODO - run locally, test it, check that the changes here are enough first

controllers/operands/kubevirt.go
```diff
@@ -162,13 +162,14 @@ var (
...
-	kvAutoResourceLimits     = "AutoResourceLimitsGate"
-	kvAlignCPUs              = "AlignCPUs"
+	kvDownwardMetrics           = "DownwardMetrics"
+	kvWithHostPassthroughCPU    = "WithHostPassthroughCPU"
+	kvRoot                      = "Root"
+	kvDisableMDevConfig         = "DisableMDEVConfiguration"
+	kvPersistentReservation     = "PersistentReservation"
+	kvAutoResourceLimits        = "AutoResourceLimitsGate"
+	kvAlignCPUs                 = "AlignCPUs"
+	kvDynamicPodInterfaceNaming = "DynamicPodInterfaceNaming"
```
orelmisan, line 172: @EdDev FYI.

EdDev: > If HCO PrimaryUserDefinedNetworkBinding FG is enabled,
> enable Kubevirt DynamicPodInterfaceNaming FG that is required
> for dynamic pod interface name detection.

It looks like the FG is used as a configuration knob.
The FG represents the feature graduation stage and therefore it cannot have dependencies like presented here.

If HCO is enabling a FG, it means it has guarantees it will never be removed, which is hard to commit to for Alpha or Beta features.

In this specific case, the dynamic pod iface name detection is not assured to be available in future Kubevirt releases. It needs to graduate to GA on Kubevirt and only then be permanently added to HCO as enabled. And it should not be dependent on any other feature lifecycle.

EdDev (changes requested):

controllers/operands/kubevirt.go
```diff
@@ -853,6 +864,10 @@ func getFeatureGateChecks(featureGates *hcov1beta1.HyperConvergedFeatureGates) [
 		fgs = append(fgs, kvAlignCPUs)
 	}
 
+	if featureGates.PrimaryUserDefinedNetworkBinding != nil && *featureGates.PrimaryUserDefinedNetworkBinding {
+		fgs = append(fgs, kvDynamicPodInterfaceNaming)
```
oshoval, line 868: @EdDev @orelmisan 
it enabled the FG only if HCO FG `PrimaryUserDefinedNetworkBinding` is enabled

controllers/operands/kubevirt.go
oshoval: look here https://github.com/kubevirt/hyperconverged-cluster-operator/pull/3139/files#r1836074173

controllers/operands/kubevirt.go
EdDev: And this is exactly what I commented on in the other [thread](https://github.com/kubevirt/hyperconverged-cluster-operator/pull/3139#discussion_r1836070837).

Such a dependency is IMO wrong.

oshoval: > If HCO is enabling a FG, it means it has guarantees it will never be removed, which is hard to commit to for Alpha or Beta features.

But it is enabled only if HCO optional FG `PrimaryUserDefinedNetworkBinding` is enabled,
which is also a FG that has graduation, doesn't it change the picture?

anyhow please decide if you want me to remove it,
it means that the user will need to patch the HCO with a json patch, which affects the kubevirt CR.

@maiqueb @qinqon 
fyi please

EdDev: I already voted and explained my vote in detail. You are welcome to challenge my reasoning.

maiqueb: so what do you suggest @EdDev ?

EdDev: > It needs to graduate to GA on Kubevirt and only then be permanently added to HCO as enabled. And it should not be dependent on any other feature lifecycle.

@maiqueb  this ^^

maiqueb: @oshoval I tend to agree here ...

Let's let @EdDev and his team handle this. 

Let us know when the feature is GA'd in KubeVirt. Until then, I guess, nothing to do.

oshoval:
removed the FG commit due to discussion
rebased

oshoval:
addressed https://github.com/kubevirt/hyperconverged-cluster-operator/pull/3139/files#r1818846303 
(used vendored kubevirt type directly)

oshoval:
sumup

we are blocked because we need to decide about the generic name, unless we agree to merge with the current name candid and update according needs / or to use non generic name for now (`managedTap`)

@EdDev fyi please

oshoval:
I think the above reason means FG should not block this PR,
so we just down to the naming, if we want to wait for generic name, have temp generic name, or have just specific name meanwhile (managedTap) and later adapt
this will allow to unblock this PR, and if it is tested already with a workaround for vmi.status IP, then it will be agnostic to the binding this way, and will able to start using managedTap

@qinqon @EdDev wdyt?

oshoval:
Updated PR desc, once we are aligned on the PR desc i will populate same to the commit message

controllers/operands/kubevirt_test.go
```diff
@@ -2059,7 +2059,8 @@ Version: 1.2.3`)
 							},
 						},
 					}
-					Expect(existingResource.Spec.Configuration.NetworkConfiguration.Binding[primaryUDNNetworkBindingName]).To(Equal(expectedInterfaceBindingPlugin))
+					Expect(existingResource.Spec.Configuration.NetworkConfiguration.Binding).To(HaveKeyWithValue(primaryUDNNetworkBindingName, primaryUserDefinedNetworkBinding()))
```
orelmisan, line 2062: Please consider instantiating an expected value for the `l2bridge`, similar to what was done for passt, as currently the unit test is not independent.

controllers/operands/kubevirt.go
```diff
@@ -47,7 +47,8 @@ const (
 )
 
 const (
-	primaryUDNNetworkBindingName = "passt"
+	primaryUDNNetworkBindingName = "l2bridge"
+	passtUDNNetworkBindingName   = "passt"
```
orelmisan, line 51: Please consider dropping the `UDN` prefix, as these are just two binding plugins, and using them does not guarantee a connection to UDN.
```suggestion
	l2bridgeBindingName = "l2bridge"
	passtBindingName   = "passt"
```

Same bellow.

orelmisan (commented):
Thank you for the PR @oshoval.
Please see the comments inline.

controllers/operands/kubevirt_test.go
oshoval: imo it is fine like that, the function is simple and doesn't even gets parameters
it allows to change just one place and i don't see it reduce safety in a way that make this change required

orelmisan: Another point is that is forces the reviewer of the test to look at the implementation details of `primaryUserDefinedNetworkBinding` in order to understand what is actually done.

controllers/operands/kubevirt.go
oshoval: True but we are triggering them by `primaryUserDefinedNetworkBinding` FG so it is for UDN
Moreover i prefer to do minimal changes on this PR, it is not its scope, especially cosmetics, 
once we do another cycle we can do whatever desired

controllers/operands/kubevirt_test.go
oshoval: otoh it hides details, and allow him to look just if he desire to

oshoval:
Thanks

@maiqueb @qinqon in case you have something to add please

@nunnatsa can you please take a look ?

oshoval:
/test ci/prow/hco-e2e-operator-sdk-aws 
/test ci/prow/hco-e2e-operator-sdk-azure
/test ci/prow/hco-e2e-operator-sdk-gcp 
/test ci/prow/hco-e2e-upgrade-prev-operator-sdk-azure 
/test ci/prow/okd-ci-index-hco-upgrade-operator-sdk-bundle

per request, but it doesnt seem at all this PR is related

nunnatsa:
hco-e2e-kv-smoke-gcp lane passed

PR #3144, opened 2024-10-21
Add OperatorConditionsUnhealthy alerts

machadovilaca:

Add alerts based on kubevirt_hco_system_health_status

The alert includes the reason of the condition that sets the health to a given status. For that, `kubevirt_hco_system_health_status` was updated to add a `reason` label, that has the value of the reason field in the underlying condition

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-48799
```

**Release note**:

```release-note
Add OperatorConditionsUnhealthy alerts
```

### DIFF
--- a/controllers/hyperconverged/hyperconverged_controller.go
+++ b/controllers/hyperconverged/hyperconverged_controller.go
@@ -1014,8 +1014,6 @@ func (r *ReconcileHyperConverged) updateConditions(req *common.HcoRequest) {
 		req.Instance.Status.SystemHealthStatus = systemHealthStatus
 		req.StatusDirty = true
 	}
-
-	metrics.SetHCOMetricSystemHealthStatus(getNumericalHealthStatus(systemHealthStatus))
 }
 
 func (r *ReconcileHyperConverged) setLabels(req *common.HcoRequest) {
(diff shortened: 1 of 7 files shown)


sradco:
@machadovilaca Hi, Please add the PR of the new alert to this PR description. Thank you.

sradco:
@machadovilaca In this PR you are adding a new alert but also updating the metrics with the reason. 
Please explain in the description how the reason is calculated.

sradco:
@machadovilaca  can you please add an example for the metrics with the new reason label?
And explain what happens if there are a few conditions that are not healthy.

sradco:
@avlitman please review

nunnatsa:
@sradco - it currently does not pass CI - including sanity test.

I think it needs at least #3153

nunnatsa:
@machadovilaca, I find the name of the alert confusing. The term `OperatorCondition` is used for something else in OLM, and it may misguide users.

sradco:
> @machadovilaca, I find the name of the alert confusing. The term `OperatorCondition` is used for something else in OLM, and it may misguide users.

@nunnatsa how would you call this alert? It is firing based on the HCO CR conditions.

nunnatsa:
> > @machadovilaca, I find the name of the alert confusing. The term `OperatorCondition` is used for something else in OLM, and it may misguide users.
> 
> @nunnatsa how would you call this alert? It is firing based on the HCO CR conditions.

Maybe `HCOOperatorConditionUnhealthy`?

avlitman:
@machadovilaca need to rebase (operatorrules should be operatorRegistry) and change alert name.
other than that looks good to me

sradco:
@nunnatsa @machadovilaca I would like to use a generic name that can be used by other operators.
That is why HCO is not in the alert name.

nunnatsa:
@machadovilaca - the "should fired when KubeDescheduler is installed and not properly configured for KubeVirt" e2e test is failing.

The reason for that is the addition of the reason label, while in `HCOPrometheusClient::GetHCOMetric()`, we only remove the query text from the result metric line, and then trying to convert text into int, but we still have the label in the text so the conversion fails.

I suggest to change tests/func-tests/hco_prometheus_route.go:

```go
var metricValueRegex = regexp.MustCompile(` (\d+(?:\.\d+)?)$`) // <== add this

func (hcoCli HCOPrometheusClient) GetHCOMetric(ctx context.Context, query string) (float64, error) {
	req, err := http.NewRequest(http.MethodGet, hcoCli.url, nil)
	if err != nil {
		return 0, fmt.Errorf("failed to create request: %w", err)
	}

	resp, err := hcoCli.cli.Do(req.WithContext(ctx))
	if err != nil {
		return 0, err
	}
	defer resp.Body.Close()
	if resp.StatusCode != http.StatusOK {
		return 0, fmt.Errorf("failed to read the temp route status: %s", resp.Status)
	}

	scanner := bufio.NewScanner(resp.Body)
	for scanner.Scan() {
		line := scanner.Text()
		if strings.HasPrefix(line, query) {
			groups := metricValueRegex.FindStringSubmatch(line)  // <== add this
			if len(groups) != 2 { // <== add this
				return 0, fmt.Errorf("no numeric value in: %q", line) // <== add this
			}
			
			res, err := strconv.ParseFloat(groups[1], 64) // <== change this
			if err != nil {
				return 0, fmt.Errorf("error converting %s to int: %v\n", line, err)
			}
			return res, nil
		}
	}
	return 0, nil
}
```

machadovilaca:
@nunnatsa since the exposition format is always in the format

`$METRIC{$LABELS} $VALUE $TIMESTAMP`

what do you think about:

```
			parts := strings.Fields(line)
			if len(parts) < 2 {
				return 0, fmt.Errorf("metric line does not contain a value")
			}
			res, err := strconv.ParseFloat(strings.TrimSpace(parts[1]), 64)
```
?

would work for every possible allowed format

ref: https://github.com/prometheus/docs/blob/main/content/docs/instrumenting/exposition_formats.md

nunnatsa:
your change makes sense. What I don't understand is how it's worked until now, and still working for other metrics, because we never had issues with the timestamp with the old code of
```go
res, err := strconv.ParseFloat(strings.TrimSpace(strings.TrimPrefix(line, query)), 64)
```

nunnatsa:
hco-e2e-upgrade-prev-operator-sdk-sno-azure lane passed
/override ci/prow/hco-e2e-upgrade-prev-operator-sdk-sno-aws

hco-e2e-operator-sdk-sno-azure lane passed
/override ci/prow/hco-e2e-operator-sdk-sno-aws

PR #3170, opened 2024-11-19
Reset kubevirt_hco_system_health_status on update

machadovilaca:

kubevirt_hco_system_health_status old error/warning values with different reasons still show after conditions are update. This PR resets the vector on each update. 

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-51403
```

**Release note**:

```release-note
Reset kubevirt_hco_system_health_status on update
```

/cc @sradco @avlitman

### DIFF
--- a/controllers/hyperconverged/hyperconverged_controller_test.go
+++ b/controllers/hyperconverged/hyperconverged_controller_test.go
@@ -192,7 +192,7 @@ var _ = Describe("HyperconvergedController", func() {
 					Message: "Initializing HyperConverged cluster",
 				})))
 
-				verifySystemHealthStatusError(foundResource)
+				verifySystemHealthStatusError(foundResource, reconcileInit)
 
 				expectedFeatureGates := []string{
 					"CPUManager",
@@ -314,7 +314,7 @@ var _ = Describe("HyperconvergedController", func() {
 					Message: "Initializing HyperConverged cluster",
 				})))
 
-				verifySystemHealthStatusError(foundResource)
+				verifySystemHealthStatusError(foundResource, reconcileInit)
 
 				res, err = r.Reconcile(context.TODO(), request)
 				Expect(err).ToNot(HaveOccurred())
@@ -401,7 +401,7 @@ var _ = Describe("HyperconvergedController", func() {
 					Message: reconcileCompletedMessage,
 				})))
 
-				verifySystemHealthStatusError(foundResource)
+				verifySystemHealthStatusError(foundResource, "SSPConditions")
 
 				Expect(foundResource.Status.RelatedObjects).To(HaveLen(21))
 				expectedRef := corev1.ObjectReference{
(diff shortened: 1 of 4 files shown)


controllers/hyperconverged/hyperconverged_controller_test.go
```diff
@@ -401,7 +401,7 @@ var _ = Describe("HyperconvergedController", func() {
 					Message: reconcileCompletedMessage,
 				})))
 
-				verifySystemHealthStatusError(foundResource)
+				verifySystemHealthStatusError(foundResource, "SSPConditions")
```
avlitman, line 404: why here we have "" and in lines 195/317 we don't have?

pkg/monitoring/metrics/operator_metrics_test.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/kubevirt/hyperconverged-cluster-operator/pkg/monitoring/metrics"
+)
+
+var _ = Describe("Operator Metrics", func() {
+	Context("kubevirt_hco_system_health_status", func() {
```
avlitman, line 11: Context("kubevirt_hco_system_health_status" metric , func() {

pkg/monitoring/metrics/operator_metrics_test.go
```diff
@@ -0,0 +1,28 @@
...
+		It("should set system error reason correctly", func() {
```
avlitman, line 12: maybe I would split it to:
`It("should report error status when a system error reason is set", func() {
`
and `It("should update the system error reason correctly", func() {`

pkg/monitoring/metrics/operator_metrics.go
```diff
@@ -137,7 +141,7 @@ func GetHCOMetricSystemHealthStatus(reason string) (float64, error) {
 	value := dto.Gauge.GetValue()
 
 	if err != nil {
-		return 0, err
+		return SystemHealthStatusUnknown, err
```
avlitman, line 144: so 0 is only unknown? can it be 0 also in other use cases?

controllers/hyperconverged/hyperconverged_controller_test.go
machadovilaca: it is only used once, that's why it uses no variable

pkg/monitoring/metrics/operator_metrics.go
machadovilaca: not in this metric

PR #3171, opened 2024-11-20
POC: block operator upgrade when detecting outdated vms

dasionov:
This PR introduces the logic necessary for blocking hco-operator upgrade when we detect that there are outdated vms in the cluster

step 1: list all virt-controller pods
step 2: loop over the pods and perform http request to the /metrics endpoint to extract all the metrics data
step 3: use regex to make sure no one is using rhel8 machine types
step 4: set upgradeable false when we do detect rhel8 machines

the logic right now is simple and straight forward, and it can be further improved

**Release note**:
```release-note
None
```

### DIFF
--- a/controllers/hyperconverged/hyperconverged_controller.go
+++ b/controllers/hyperconverged/hyperconverged_controller.go
@@ -1,13 +1,19 @@
 package hyperconverged
 
 import (
+	"bufio"
 	"cmp"
 	"context"
+	"crypto/tls"
 	"encoding/json"
 	"fmt"
+	"net/http"
 	"os"
 	"reflect"
+	"regexp"
 	"slices"
+	"sync"
+	"time"
 
 	"github.com/blang/semver/v4"
 	jsonpatch "github.com/evanphx/json-patch/v5"
@@ -100,6 +106,8 @@ var JSONPatchAnnotationNames = []string{
 	common.JSONPatchCNAOAnnotationName,
 	common.JSONPatchSSPAnnotationName,
 }
+
+var rhel8Regex = regexp.MustCompile(`.*rhel8.*`)
 
 // RegisterReconciler creates a new HyperConverged Reconciler and registers it into manager.
 func RegisterReconciler(mgr manager.Manager, ci hcoutil.ClusterInfo, upgradeableCond hcoutil.Condition) error {
@@ -336,6 +344,11 @@ func (r *ReconcileHyperConverged) Reconcile(ctx context.Context, request reconci
 		return result, err
 	}
 
+	err = r.evaluateUpgradeEligibility(hcoRequest)
+	if err != nil {
+		hcoRequest.Logger.Error(err, "Failed to evaluate upgrade eligibility", "err", err)
+		return reconcile.Result{}, err
+	}
 	if err = r.setOperatorUpgradeableStatus(hcoRequest); err != nil {
 		return reconcile.Result{}, err
 	}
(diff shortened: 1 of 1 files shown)


controllers/hyperconverged/hyperconverged_controller.go
```diff
@@ -1330,6 +1340,66 @@ func (r *ReconcileHyperConverged) deleteObj(req *common.HcoRequest, obj client.O
...
+		metricsURL := fmt.Sprintf("https://%s:%d/metrics", pod.Status.PodIP, 8443)
+
+		resp, err := httpClient.Get(metricsURL)
+		if err != nil {
+			req.Logger.Info("Failed to query metrics from pod", "pod", pod.Name, "error", err)
+			continue
+		}
+
+		if resp.StatusCode != http.StatusOK {
+			req.Logger.Info("Metrics endpoint returned non-200 status", "pod", pod.Name, "status", resp.StatusCode)
```
nunnatsa, line 1379: Not sure about the term "non-200" in an error message.

WDYT about "Metrics endpoint returned error" ?

controllers/hyperconverged/hyperconverged_controller.go
```diff
@@ -1330,6 +1340,66 @@ func (r *ReconcileHyperConverged) deleteObj(req *common.HcoRequest, obj client.O
...
+			resp.Body.Close()
```
nunnatsa, line 1380: use defer instead

controllers/hyperconverged/hyperconverged_controller.go
```diff
@@ -1330,6 +1340,66 @@ func (r *ReconcileHyperConverged) deleteObj(req *common.HcoRequest, obj client.O
...
+			continue
+		}
+
+		body, err := io.ReadAll(resp.Body)
+		resp.Body.Close()
```
nunnatsa, line 1385: use defer instead

controllers/hyperconverged/hyperconverged_controller.go
```diff
@@ -1330,6 +1340,66 @@ func (r *ReconcileHyperConverged) deleteObj(req *common.HcoRequest, obj client.O
...
+			continue
+		}
+
+		if resp.StatusCode != http.StatusOK {
+			req.Logger.Info("Metrics endpoint returned non-200 status", "pod", pod.Name, "status", resp.StatusCode)
+			resp.Body.Close()
+			continue
+		}
+
+		body, err := io.ReadAll(resp.Body)
```
nunnatsa, line 1384: I think it's better to use a scanner. The body can be big.

You can do something like
```go
scanner := bufio.NewScanner(resp.Body)
...
for scanner.Scan() {
	if rhel8Regex.MatchString(scanner.Text()) {
		req.Logger.Info("Detected outdated machine type in metrics", "pod", pod.Name, "matched", rhel8Regex.FindString(virtMetricsData))
		req.Upgradeable = false
		return nil
	}
}
```

nunnatsa (commented):
I know it's a POC, but still have several comments inline.

controllers/hyperconverged/hyperconverged_controller.go
```diff
@@ -1330,6 +1340,66 @@ func (r *ReconcileHyperConverged) deleteObj(req *common.HcoRequest, obj client.O
...
+
+	if err := r.client.List(req.Ctx, podList, listOpts...); err != nil {
+		req.Logger.Info("Failed to list virt-controller pods", "namespace", req.Namespace, "error", err)
+		return fmt.Errorf("failed to list virt-controller pods: %w", err)
+	}
+
+	if len(podList.Items) == 0 {
+		req.Logger.Info("No virt-controller pods found", "namespace", req.Namespace)
+		return nil
+	}
```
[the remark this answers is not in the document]
dasionov: we don't have a service for virt-controller as far as i know, only for virt-api

controllers/hyperconverged/hyperconverged_controller.go
```diff
@@ -1330,6 +1340,66 @@ func (r *ReconcileHyperConverged) deleteObj(req *common.HcoRequest, obj client.O
...
+
+		body, err := io.ReadAll(resp.Body)
+		resp.Body.Close()
+		if err != nil {
+			req.Logger.Info("Failed to read metrics response from pod", "pod", pod.Name, "error", err)
+			continue
+		}
+
+		virtMetricsData := string(body)
+		rhel8Regex := regexp.MustCompile(`.*rhel8.*`)
```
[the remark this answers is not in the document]
dasionov: the regex will be created each reconcile loop then, that is ok?

controllers/hyperconverged/hyperconverged_controller.go
dasionov: i did use that at first, but got a warning said that it's not safe to use defer in a loop, and it might cause a resources leek

controllers/hyperconverged/hyperconverged_controller.go
dasionov: fixed

controllers/hyperconverged/hyperconverged_controller.go
dasionov: sounds good

controllers/hyperconverged/hyperconverged_controller.go
```diff
@@ -1330,6 +1340,66 @@ func (r *ReconcileHyperConverged) deleteObj(req *common.HcoRequest, obj client.O
 	return removed, nil
 }
 
+func (r *ReconcileHyperConverged) EvaluateUpgradeEligibility(req *common.HcoRequest) error {
```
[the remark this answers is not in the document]
dasionov: fixed

controllers/hyperconverged/hyperconverged_controller.go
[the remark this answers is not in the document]
nunnatsa: if it will be defined in the package scope, out of the method, then no. It will only be initialized once, on boot.

controllers/hyperconverged/hyperconverged_controller.go
nunnatsa: right. so we can move the loop content into a new method. That will work.

nunnatsa:
Another comment I forgot to write: The current HTTP client code does not use any type of timeout. Use a context (derived from req.Context) with 3-seconds timeout. Sending HTTP request w/o timeout may be risky.

dasionov:
/close
we can close this, since https://github.com/kubevirt/kubevirt/pull/13690 is good enough solution for the use-case.

PR #3173, opened 2024-11-21
Fix nightly test: use operator-sdk v1.37, instead of latest

nunnatsa:

operator-sdk v1.38 does not support kubebuilder's go v3 in the PROJECT file, and so it fails to run in this project, even to just install OLM.

Until we'll move to kubebuilder's go v4, we need to use older operator-sdk.

This PR changes the nightly test to use oprator-sdk v1.37.0, instead of taking the latest version.

**Release note**:

```release-note
None
```


orenc1:
need to track the transition to `go.kubebuilder.io/v4` in:
https://github.com/kubevirt/hyperconverged-cluster-operator/blob/main/PROJECT
/lgtm
/approve

PR #3186, opened 2024-12-01
Don't rely on InfraStructureTopology for infra HA

orenc1:
There are some cases in which the number of worker nodes are changing throughout the lifecycle of the cluster, and the `InfrastructureTopology` in the Infrastructure resource is statically set at cluster installation type and it is not getting updated dynamically. For example, an SNO cluster that is being added a new worker node. In that case, the infrastructure topology should be updated to `HighlyAvailable` and don't remain `SingleReplica`.
Instead, we should count the worker nodes, same as we do in case of a k8s cluster.

In addition, fixing a potential bug were we used only the `node-role.kubernetes.io/master` label to find and count the masters/control-plane nodes.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-50027
```

**Release note**:

```release-note
NONE
```

### DIFF
--- a/controllers/nodes/nodes_controller.go
+++ b/controllers/nodes/nodes_controller.go
@@ -1,0 +1,127 @@
+package nodes
+
+import (
+	"context"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+
+	hcov1beta1 "github.com/kubevirt/hyperconverged-cluster-operator/api/v1beta1"
+
+	operatorhandler "github.com/operator-framework/operator-lib/handler"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	hcoutil "github.com/kubevirt/hyperconverged-cluster-operator/pkg/util"
+)
+
+var (
+	log = logf.Log.WithName("controller_nodes")
+)
+
+// RegisterReconciler creates a new Nodes Reconciler and registers it into manager.
+func RegisterReconciler(mgr manager.Manager) error {
+	return add(mgr, newReconciler(mgr))
+}
+
+// newReconciler returns a new reconcile.Reconciler
+func newReconciler(mgr manager.Manager) reconcile.Reconciler {
+	r := &ReconcileNodeCounter{
+		client: mgr.GetClient(),
+	}
+
+	return r
+}
+
+// add adds a new Controller to mgr with r as the reconcile.Reconciler
+func add(mgr manager.Manager, r reconcile.Reconciler) error {
+	// Create a new controller
+	c, err := controller.New("nodes-controller", mgr, controller.Options{Reconciler: r})
+	if err != nil {
+		return err
+	}
+
+	// Watch for changes to the cluster's nodes
+	err = c.Watch(
+		source.Kind(
+			mgr.GetCache(), client.Object(&corev1.Node{}),
+			&operatorhandler.InstrumentedEnqueueRequestForObject[client.Object]{},
+			nodeCountChangePredicate{},
+		))
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Custom predicate to detect changes in node count
+type nodeCountChangePredicate struct {
+	predicate.Funcs
+}
+
+func (nodeCountChangePredicate) Update(e event.UpdateEvent) bool {
+	return false
+}
+
+func (nodeCountChangePredicate) Create(e event.CreateEvent) bool {
+	// node is added
+	return true
+}
+
+func (nodeCountChangePredicate) Delete(e event.DeleteEvent) bool {
+	// node is removed
+	return true
+}
+
+func (nodeCountChangePredicate) Generic(e event.GenericEvent) bool {
+	return false
+}
+
+// ReconcileNodeCounter reconciles the nodes count
+type ReconcileNodeCounter struct {
+	// This client, initialized using mgr.Client() above, is a split client
+	// that reads objects from the cache and writes to the apiserver
+	client              client.Client
+	HyperConvergedQueue workqueue.TypedRateLimitingInterface[reconcile.Request]
+}
+
+// Reconcile updates the nodes count on ClusterInfo singleton
+func (r *ReconcileNodeCounter) Reconcile(ctx context.Context, _ reconcile.Request) (reconcile.Result, error) {
+	log.Info("Triggered by a node count change")
+	clusterInfo := hcoutil.GetClusterInfo()
+	err := clusterInfo.SetHighAvailabilityMode(ctx, r.client)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	hco := &hcov1beta1.HyperConverged{}
+	namespace, err := hcoutil.GetOperatorNamespaceFromEnv()
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	hcoKey := types.NamespacedName{
+		Name:      hcoutil.HyperConvergedName,
+		Namespace: namespace,
+	}
+	err = r.client.Get(ctx, hcoKey, hco)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if hco.Status.InfrastructureHighlyAvailable != clusterInfo.IsInfrastructureHighlyAvailable() {
+		hco.Status.InfrastructureHighlyAvailable = clusterInfo.IsInfrastructureHighlyAvailable()
+		err = r.client.Status().Update(ctx, hco)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+	}
+	return reconcile.Result{}, nil
+}
--- a/pkg/util/cluster.go
+++ b/pkg/util/cluster.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"os"
 	"slices"
+	"sync/atomic"
 
 	"github.com/go-logr/logr"
 	openshiftconfigv1 "github.com/openshift/api/config/v1"
@@ -16,8 +17,6 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/client-go/discovery"
 	"k8s.io/utils/net"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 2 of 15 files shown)


controllers/nodes/nodes_controller.go
```diff
@@ -0,0 +1,124 @@
...
+	// that reads objects from the cache and writes to the apiserver
+	client              client.Client
+	HyperConvergedQueue workqueue.TypedRateLimitingInterface[reconcile.Request]
+}
+
+// Reconcile updates the nodes count on ClusterInfo singleton
+func (r *ReconcileNodeCounter) Reconcile(ctx context.Context, _ reconcile.Request) (reconcile.Result, error) {
+	log.Info("Triggered by node count change")
+	logger := logf.Log.WithName("nodes-controller")
+	clusterInfo := hcoutil.GetClusterInfo()
```
nunnatsa, line 98: If we're reading and writing from several go-routines, then we must protect the data. 

But clusterInfo interface is too large for that, and will force us to protect each and every read method. So I think we must take out `IsControlPlaneHighlyAvailable` and `IsInfrastructureHighlyAvailable` out of the clusterInfo interface, together with their related data, into a new type/interface, and protect them.

controllers/nodes/nodes_controller.go
```diff
@@ -0,0 +1,124 @@
...
+	err := clusterInfo.Init(ctx, r.client, logger)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	hco := &hcov1beta1.HyperConverged{}
```
nunnatsa, line 104: Why do we need this?

Also, the update will do nothing, as there is no change in theHyperConverged CR

pkg/util/cluster.go
```diff
@@ -58,12 +59,21 @@ type ClusterInfoImp struct {
 	monitoringAvailable           bool
 	deschedulerAvailable          bool
 	singlestackipv6               bool
+	masterNodesCount              int
```
nunnatsa, line 62: These two new fields can be a method variable. I don't think we need them in the type.

pkg/util/cluster.go
```diff
@@ -58,12 +59,21 @@ type ClusterInfoImp struct {
...
+	workerNodesCount              int
 	domain                        string
 	baseDomain                    string
 	ownResources                  *OwnResources
 	logger                        logr.Logger
 }
 
+type NodesType string
+
+const (
```
nunnatsa, line 72: no needed, I think

pkg/util/cluster.go
```diff
@@ -426,3 +427,48 @@ func isValidCipherName(str string) bool {
...
+
+func getNodesCount(cl client.Client, nodesType NodesType) (int, error) {
+	if nodesType == ControlPlaneNodes {
+		masterReq, err := labels.NewRequirement("node-role.kubernetes.io/master", selection.Exists, nil)
+		if err != nil {
+			return -1, err
+		}
+		controlplaneReq, err := labels.NewRequirement("node-role.kubernetes.io/control-plane", selection.Exists, nil)
+		if err != nil {
+			return -1, err
```
nunnatsa, line 439: return zero-value (0)

pkg/util/cluster.go
```diff
@@ -426,3 +427,48 @@ func isValidCipherName(str string) bool {
...
+		}
+		masterNodeList := &corev1.NodeList{}
+		cpNodeList := &corev1.NodeList{}
+		masterSelector := labels.NewSelector().Add(*masterReq)
+		masterLabelSelector := client.MatchingLabelsSelector{Selector: masterSelector}
+		err = cl.List(context.TODO(), masterNodeList, masterLabelSelector)
+		if err != nil {
+			return -1, err
```
nunnatsa, line 447: same

pkg/util/cluster.go
```diff
@@ -426,3 +427,48 @@ func isValidCipherName(str string) bool {
 		slices.Contains(openshiftconfigv1.TLSProfiles[openshiftconfigv1.TLSProfileIntermediateType].Ciphers, str) ||
 		slices.Contains(openshiftconfigv1.TLSProfiles[openshiftconfigv1.TLSProfileModernType].Ciphers, str)
 }
+
+func getNodesCount(cl client.Client, nodesType NodesType) (int, error) {
+	if nodesType == ControlPlaneNodes {
```
nunnatsa, line 432: can we read the whole node list and count the items according to their annotations, instead of several reads?

pkg/util/cluster.go
```diff
@@ -426,3 +427,48 @@ func isValidCipherName(str string) bool {
...
+		workerReq, err := labels.NewRequirement("node-role.kubernetes.io/worker", selection.Exists, nil)
+		if err != nil {
+			return -1, err
+		}
+		workerNodeList := &corev1.NodeList{}
+		workerSelector := labels.NewSelector().Add(*workerReq)
+		labelSelector := client.MatchingLabelsSelector{Selector: workerSelector}
+		err = cl.List(context.TODO(), workerNodeList, labelSelector)
+		if err != nil {
+			return -1, err
```
nunnatsa, line 468: return zero-value

pkg/util/cluster.go
```diff
@@ -426,3 +427,48 @@ func isValidCipherName(str string) bool {
...
+		}
+		return len(workerNodeList.Items), nil
+	}
+
+	return -1, fmt.Errorf("unrecognized node type: %s", nodesType)
```
nunnatsa, line 473: return zero-value

pkg/util/cluster.go
```diff
@@ -426,3 +427,48 @@ func isValidCipherName(str string) bool {
...
+		if err != nil {
+			return -1, err
+		}
+		// Adhere to both old (node-role.kubernetes.io/master) and new (node-role.kubernetes.io/control-plane) conventions
+		return len(masterNodeList.Items) + len(cpNodeList.Items), nil
+	}
+	if nodesType == WorkerNodes {
+		workerReq, err := labels.NewRequirement("node-role.kubernetes.io/worker", selection.Exists, nil)
+		if err != nil {
+			return -1, err
```
nunnatsa, line 461: return zero-value

pkg/util/cluster.go
```diff
@@ -426,3 +427,48 @@ func isValidCipherName(str string) bool {
...
+		masterLabelSelector := client.MatchingLabelsSelector{Selector: masterSelector}
+		err = cl.List(context.TODO(), masterNodeList, masterLabelSelector)
+		if err != nil {
+			return -1, err
+		}
+		cpSelector := labels.NewSelector().Add(*controlplaneReq)
+		cpLabelSelector := client.MatchingLabelsSelector{Selector: cpSelector}
+		err = cl.List(context.TODO(), cpNodeList, cpLabelSelector)
+		if err != nil {
+			return -1, err
```
nunnatsa, line 453: return zero-value

nunnatsa (commented):
Added several inline comments, but my general comment is about the ClusterInfo type, and thread-safty.

I think we must split the high-available info & methods from this type/interface into a new type, and then protect it with RWMutex.

@orenc1, WDYT?

pkg/util/cluster.go
```diff
@@ -426,3 +424,47 @@ func isValidCipherName(str string) bool {
 		slices.Contains(openshiftconfigv1.TLSProfiles[openshiftconfigv1.TLSProfileIntermediateType].Ciphers, str) ||
 		slices.Contains(openshiftconfigv1.TLSProfiles[openshiftconfigv1.TLSProfileModernType].Ciphers, str)
 }
+
+func getNodesCount(cl client.Client, nodesType NodesType) (int, error) {
```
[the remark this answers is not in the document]
orenc1: yes, at the beginning the majority of the function was shared for the two cases.
but then it turned out there is no logical OR when selecting nodes based on two labels where either of them exists, so i needed to make two queries.

pkg/util/cluster.go
```diff
@@ -426,3 +424,47 @@ func isValidCipherName(str string) bool {
...
+	if nodesType == ControlPlaneNodes {
+		masterReq, err := labels.NewRequirement("node-role.kubernetes.io/master", selection.Exists, nil)
+		if err != nil {
+			return -1, err
```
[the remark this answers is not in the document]
orenc1: ok.
anyway in case of an error, the returned int is not being read.

pkg/util/cluster.go
orenc1: according to the node-role labels, yes. i'll update.

pkg/util/cluster.go
orenc1: this is not needed at all eventually.

pkg/util/cluster.go
orenc1: this function was refactored. returning 0 on error.

pkg/util/cluster.go
orenc1: done.

pkg/util/cluster.go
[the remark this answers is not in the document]
orenc1: function refactored. now returning both workers and masters count in a single API call.

controllers/nodes/nodes_controller.go
```diff
@@ -0,0 +1,128 @@
...
+	client              client.Client
+	HyperConvergedQueue workqueue.TypedRateLimitingInterface[reconcile.Request]
+}
+
+// Reconcile updates the nodes count on ClusterInfo singleton
+func (r *ReconcileNodeCounter) Reconcile(ctx context.Context, _ reconcile.Request) (reconcile.Result, error) {
+	log.Info("Triggered by node count change")
+	logger := logf.Log.WithName("nodes-controller")
+	clusterInfo := hcoutil.GetClusterInfo()
+	err := clusterInfo.Init(ctx, r.client, logger)
```
nunnatsa, line 101: I think clusterInfo.Init() does too much and update too many fields.

I do think we should take the high availability info out of the clusterInfo, but even if we don't, let's at least add a dedicated setter for it.

pkg/util/cluster.go
```diff
@@ -123,33 +121,15 @@ func (c *ClusterInfoImp) Init(ctx context.Context, cl client.Client, logger logr
...
-	workerLabelSelector := client.MatchingLabelsSelector{Selector: workerSelector}
-	err = cl.List(context.TODO(), workerNodeList, workerLabelSelector)
-	if err != nil {
-		return err
-	}
-
-	c.controlPlaneHighlyAvailable = len(masterNodeList.Items) >= 3
-	c.infrastructureHighlyAvailable = len(workerNodeList.Items) >= 2
+	c.controlPlaneHighlyAvailable = masterNodeCount >= 3
+	c.infrastructureHighlyAvailable = workerNodeCount >= 2
```
nunnatsa, line 132: If we allow to get async access to this fields, we MUST protect them.

Please add a RWMutext, and lock each time we read or write these fields.

alternatively, use `atomic.Bool` from the standard library, instead of bool for the `controlPlaneHighlyAvailable` and `infrastructureHighlyAvailable` fields.

nunnatsa (commented):
Please handle async locks - see inline coments.

pkg/util/cluster_test.go
```diff
@@ -407,11 +407,35 @@ var _ = Describe("test clusterInfo", func() {
...
+							"node-role.kubernetes.io/control-plane": "",
+						},
+					},
+				}
+				nodesArray = append(nodesArray, masterNode)
+			}
+			for i := range numWorkerNodes {
+				workerNode := &corev1.Node{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "worker" + fmt.Sprint(i),
```
nunnatsa, line 425: same

nunnatsa (approved):
looks good!

Added two nit comments. Feel free to ignore.

nunnatsa (approved):
/approve

PR #3188, opened 2024-12-03
Fix health impact values in health status metrics

machadovilaca:

E2E tests using hardcoded old values for operator health metric. Changed to use the values defined in the metrics package

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-52319
```

**Release note**:

```release-note
Fix health impact values in health status metrics
```

### DIFF
--- a/hack/prom-rule-ci/prom-rules-tests.yaml
+++ b/hack/prom-rule-ci/prom-rules-tests.yaml
@@ -610,7 +610,7 @@ tests:
   input_series:
   - series: 'kubevirt_hco_system_health_status'
     # time:  0     1     2     3     4     5     6     7     8     9     10     11
-    values: "0     0     0     0     1     1     1     1     2     2      2      2"
+    values: "1     1     1     1     2     2     2     2     3     3      3      3"
   - series: 'ALERTS{kubernetes_operator_part_of="kubevirt", alertstate="firing", operator_health_impact="warning"}'
     # time:  0     1     2     3     4     5     6     7     8     9     10     11
     values: "1     1   stale stale   1     1   stale stale   1     1    stale  stale"
--- a/pkg/monitoring/rules/recordingrules/operator.go
+++ b/pkg/monitoring/rules/recordingrules/operator.go
@@ -1,9 +1,19 @@
 package recordingrules
 
 import (
+	"fmt"
+
 	"github.com/machadovilaca/operator-observability/pkg/operatormetrics"
 	"github.com/machadovilaca/operator-observability/pkg/operatorrules"
 	"k8s.io/apimachinery/pkg/util/intstr"
+
+	"github.com/kubevirt/hyperconverged-cluster-operator/pkg/monitoring/metrics"
+)
+
+const (
+	NoImpact float64 = iota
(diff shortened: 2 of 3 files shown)


pkg/monitoring/rules/recordingrules/operator.go
```diff
@@ -32,3 +42,19 @@ var operatorRecordingRules = []operatorrules.RecordingRule{
...
+	)
+
+	warningExpr := fmt.Sprintf(
+		`(vector(%d) and on() ((kubevirt_hco_system_health_status==%d) or (count(ALERTS{kubernetes_operator_part_of="kubevirt", alertstate="firing", operator_health_impact="warning"})>0)))`,
+		int64(WarningImpact), int64(metrics.SystemHealthStatusWarning),
+	)
+
+	healthyExpr := fmt.Sprintf("vector(%d)", int64(NoImpact))
+
+	return intstr.FromString("label_replace(" + criticalExpr + " or " + warningExpr + " or " + healthyExpr + `,"name","kubevirt-hyperconverged","","")`)
```
nunnatsa, line 59: nit:

not so important because it happens only once, but I think that, performance wise,  there are too many concatenations here. Please check if you can use `strings.Builder` here.

machadovilaca:
retesting to check for flakiness
/retest

pkg/monitoring/rules/recordingrules/operator.go
machadovilaca:
```
	var expr strings.Builder
	expr.WriteString(addHCOLabelPrefix)
	expr.WriteString(criticalExpr)
	expr.WriteString(orDivider)
	expr.WriteString(warningExpr)
	expr.WriteString(orDivider)
	expr.WriteString(healthyExpr)
	expr.WriteString(addHCOLabelSuffix)

	return intstr.FromString(expr.String())
```

It would look something like this, right?
I would suggest accepting the performance penalty (as you said happens once) for improved readability that the current approach offers

let me know your thoughts

machadovilaca:
@nunnatsa

nunnatsa (approved):
/approve

hack/prom-rule-ci/prom-rules-tests.yaml
```diff
@@ -610,7 +610,7 @@ tests:
   input_series:
   - series: 'kubevirt_hco_system_health_status'
     # time:  0     1     2     3     4     5     6     7     8     9     10     11
-    values: "0     0     0     0     1     1     1     1     2     2      2      2"
+    values: "1     1     1     1     2     2     2     2     3     3      3      3"
```
sradco, line 613: @machadovilaca There should no be a value 3.
Values of this metric should be healthy (0), warning (1), or error (2).

PR #3189, opened 2024-12-04
cnao, ipam-extensions: deploy IPAMController by default

maiqueb:
This PR removes the `deployKubevirtIpamController` feature gate, and deploys the ipam-extensions component by default.

**Reviewer Checklist**

- [X] PR Message
- [X] Commit Messages
- [X] How to test
- [X] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [X] Upgrade Scenario
- [] Uninstallation Scenario
- [X] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-52397
```

**Release note**:

```release-note
Remove the `deployKubevirtIpamController` feature gate, and install the IPAMController component by default via CNAO.
```


nunnatsa (commented):
@maiqueb - thanks for the PR!

Looks good in general, and it also save some time for me because I was about to move this field out of the `spec.featureGates` field...

Please run `make build-manifests` to re-generate the CRD. Please fix the one linter comment in controllers/operands/networkAddons_test.go.

I just wonder if we can just remove a field from the API w/o deprecating it first. I think it's ok. @tiraboschi - WDYT?

maiqueb:
> @maiqueb - thanks for the PR!
> 
> Looks good in general, and it also save some time for me because I was about to move this field out of the `spec.featureGates` field...
> 
> Please run `make build-manifests` to re-generate the CRD. Please fix the one linter comment in controllers/operands/networkAddons_test.go.
> 
> I just wonder if we can just remove a field from the API w/o deprecating it first. I think it's ok. @tiraboschi - WDYT?

FWIW, this feature was never actually GA'ed in OpenShift.

nunnatsa:
Functional test is failing because the name if the ipam pod is wrong. see here: https://github.com/kubevirt/hyperconverged-cluster-operator/blob/0bd61aa62819cf3a2af803a6eb0fcb08e5bb3bf8/tests/func-tests/node_placement_test.go#L171C5-L171C19

The name should  be `"ipam-virt-workloads"`. But it would be better to add it to the `expectedWorkloadsPods` map and remove the condition in lines 170-172.

nunnatsa (approved):

api/v1beta1/hyperconverged_types.go
```diff
@@ -430,13 +430,6 @@ type HyperConvergedFeatureGates struct {
 	// +default=false
 	DeployKubeSecondaryDNS *bool `json:"deployKubeSecondaryDNS,omitempty"`
 
-	// Deploy KubevirtIpamController by CNAO.
-	// Allows having persistent IPs for Kubevirt user defined networks.
-	// +optional
-	// +kubebuilder:default=false
-	// +default=false
-	DeployKubevirtIpamController *bool `json:"deployKubevirtIpamController,omitempty"`
-
```
nunnatsa, line 439: According to this https://kubernetes.io/docs/reference/using-api/deprecation-policy/#deprecating-parts-of-the-api, we can't just drop API fields. We'll do it in v1 of the API.

All the rest of the code is great, e.g. removing the default value etc.

Please just deprecate it, but don't remove it from the API

```go
    // Deprecated: this field is ignored and will be removed in the next version of the API.
	DeployKubevirtIpamController *bool `json:"deployKubevirtIpamController,omitempty"`

```

nunnatsa (changes requested):
Please do not remove the `spec.featureGates.deployKubevirtIpamController` field, but deprecate it. see inline comment.

We can remove it from the CR (not from the CRD) on upgrade.

Please add the following to DeployKubevirtIpamController, under `hcoCRPatchList`:
```json
    {
      "semverRange": "<1.14.0",
      "jsonPatch": [
        {
          "op": "remove",
          "path": "/spec/featureGates/deployKubevirtIpamController"
        }
      ],
      "jsonPatchApplyOptions": {
        "allowMissingPathOnRemove": true
      }
    }
```

api/v1beta1/hyperconverged_types.go
maiqueb: Should I remove it from the docs ? @nunnatsa

nunnatsa: I think it's ok to remove it from the doc. it is ignored anyway. 

Also, there is a mistake in my previous comment. the `semverRange` in the json I wrote should be `<=1.4.0`, not `<1.14.0`

maiqueb: Done. Thank you.

nunnatsa (approved):

nunnatsa:
hco-e2e-kv-smoke-gcp lane passed

nunnatsa:
hco-e2e-consecutive-operator-sdk-upgrades-aws lane passed
/override ci/prow/hco-e2e-consecutive-operator-sdk-upgrades-azure

tiraboschi (approved):

PR #3197, opened 2024-12-08
Remove deprecated DataVolume garbage collection doc

arnongilboa:
After several releases with CDI DataVolume garbage collection disabled by default, we decided to [deprecate](https://github.com/kubevirt/containerized-data-importer/pull/3552) it, as unfortunately it violates fundamental principle of Kubernetes. CR should not be auto-deleted when it completes its role (Job with TTLSecondsAfterFinished is an exception), and once CR was created we can assume it is there until explicitly deleted. In addition, CR should keep idempotency, so the same CR manifest can be applied multiple times, as long as it is a valid update (e.g. DataVolume validation webhook does not allow updating the spec).

**Jira Ticket**:
```jira-ticket
https://issues.redhat.com/browse/CNV-30806
```

**Release note**:
```release-note
NONE
```

### DIFF
--- a/docs/cluster-configuration.md
+++ b/docs/cluster-configuration.md
@@ -1341,28 +1341,6 @@ metadata:
         }
       ]
 ```
-##### Disable DataVolume garbage collection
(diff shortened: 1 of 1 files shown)


nunnatsa (approved):
/approve

PR #3199, opened 2024-12-09
Fix the docgen tool

nunnatsa:
### What this PR does / why we need it
rewrite parts of the docgen tools to fix:
* fix old, broken K8s links (was 1.17) to match the api version from go.mod
* stop using the deprecated ast.Package type
* better error handling (exit on error)
* only one pass on the ast
* use go-template instead of fmt.Print to generate the doc file

### Release note

```release-note
None
```

### DIFF
--- a/tools/docgen/main.go
+++ b/tools/docgen/main.go
@@ -20,134 +20,167 @@ package main
 
 import (
 	"bytes"
+	"embed"
+	"errors"
 	"fmt"
 	"go/ast"
-	"go/doc"
 	"go/parser"
 	"go/token"
+	"io"
 	"os"
 	"reflect"
+	"slices"
 	"strings"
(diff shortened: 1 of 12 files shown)


tools/docgen/main.go
```diff
@@ -20,171 +20,192 @@ package main
...
+	initial, err := getInitialInfo(paths)
+	if err != nil {
+		return nil, err
+	}
 
-	for _, t := range types {
-		strukt := t[0]
-		if len(t) > 1 {
-			fmt.Printf("\n## %s\n\n%s\n\n", strukt.Name, strukt.Doc)
+	for _, strukt := range initial {
```
orenc1, line 90: what does `strukt` mean?

PR #3205, opened 2024-12-12
Fix issues with high availability detection

nunnatsa:
* make the status field a pointer to bool, to distinglish between false and field not set.
* On mutation webhook, fix the eviction strategy if SNO becam HA.
* fix an issue where the status field is not set if no node is added or deleted.
* fix unit test (assert old HCO instead of HCO from client)
* fix issue in the node controller, to ignore error if HCO is not present.

**Jira Ticket**:

```jira-ticket
none
```

**Release note**:

```release-note
None
```


orenc1 (approved):
/approve
/thanks

PR #3206, opened 2024-12-12
Fix the nightly test

nunnatsa:

For some reason, the nightly test is failed to deploy because the HyperConverged never becomes available.

CNAO is now depend on the cert-manager.

Fixed by also deploy the cert-manager.
**Release note**:

```release-note
None
```


orenc1 (approved):
/approve

PR #3208, opened 2024-12-15
DEV (kubevirtci): allow redploying the operator/webhook

nunnatsa:
When developing a complex feature, or debugging a bug, we sometimes need to rebuild the operator or the webhook several times until we're happy with it. But currently, we need to re-deploy the whole project, and sometimes even redeploy the cluster, just to fix something in the code.

This PR adds two new Makefile targets:
```shell
make cluster-redeploy-operator
make cluster-redeploy-webhook
```

It will rebuild the image, and restart the operator or the webhook pod, with a new image that contains the latest changes.

**Release note**:

```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -132,6 +132,12 @@ cluster-down:
 cluster-sync:
 	IMAGE_REGISTRY=$(IMAGE_REGISTRY) REGISTRY_NAMESPACE=$(REGISTRY_NAMESPACE) ./cluster/sync.sh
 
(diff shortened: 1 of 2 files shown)


avlitman:
nit: @nunnatsa can you add more details why we need this to the pr description?

nunnatsa:
> nit: @nunnatsa can you add more details why we need this to the pr description?

Good point. Thanks!

Done.

PR #3218, opened 2024-12-19
[Tools] Remove deprecated flags

nunnatsa:

The `--mtq-version` and `--mtq-csv` flags of the csv-mergrer and the manifest-templater tools, are deprecated and are not in use.

This PR removes these flags.

**Jira Ticket**:

```jira-ticket
None
```

**Release note**:

```release-note
None
```


avlitman:
I think we need to add some release note about it since it's a removal of a flag no?

PR #3219, opened 2024-12-19
Handle kubevirt control plane NodePlacement on hypershift clusters

orenc1:
Since kubevirt v1.3, the kubevirt control plane components (e.g. virt-controller and virt-api) are getting schdeuled by default on the control-plane/master nodes, not regular worker nodes.
In hypershift (Hosted Control Planes) clusters, there are no control plane nodes at all (the control plane is hosted as pods on the management cluster). Therefore, in case there are no control plane nodes, we're setting the kubevirt's node placement as an empty struct to avoid any predefined scheduling/affinity rules that might prevent the pods to be scheduled on the cluster.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-52266
```

**Release note**:

```release-note
Avoid scheduling kubevirt components on control-plane nodes if there are no control-plane nodes in the cluster.
```

### DIFF
--- a/controllers/operands/kubevirt.go
+++ b/controllers/operands/kubevirt.go
@@ -282,7 +282,8 @@ func NewKubeVirt(hc *hcov1beta1.HyperConverged, opts ...string) (*kubevirtcorev1
 
 	kvCertConfig := hcoCertConfig2KvCertificateRotateStrategy(hc.Spec.CertConfig)
 
-	infrastructureHighlyAvailable := hcoutil.GetClusterInfo().IsInfrastructureHighlyAvailable()
+	controlPlaneHighlyAvailable := hcoutil.GetClusterInfo().IsControlPlaneHighlyAvailable()
+	controlPlaneNodeExists := hcoutil.GetClusterInfo().IsControlPlaneNodeExists()
 
 	uninstallStrategy := kubevirtcorev1.KubeVirtUninstallStrategyBlockUninstallIfWorkloadsExist
 	if hc.Spec.UninstallStrategy == hcov1beta1.HyperConvergedUninstallStrategyRemoveWorkloads {
@@ -291,8 +292,8 @@ func NewKubeVirt(hc *hcov1beta1.HyperConverged, opts ...string) (*kubevirtcorev1
 
 	spec := kubevirtcorev1.KubeVirtSpec{
 		UninstallStrategy:           uninstallStrategy,
-		Infra:                       hcoConfig2KvConfig(hc.Spec.Infra, infrastructureHighlyAvailable),
-		Workloads:                   hcoConfig2KvConfig(hc.Spec.Workloads, true),
+		Infra:                       hcoConfig2KvConfig(hc.Spec.Infra, controlPlaneHighlyAvailable, controlPlaneNodeExists),
+		Workloads:                   hcoConfig2KvConfig(hc.Spec.Workloads, true, true),
 		Configuration:               *config,
 		CertificateRotationStrategy: *kvCertConfig,
 		WorkloadUpdateStrategy:      hcWorkloadUpdateStrategyToKv(&hc.Spec.WorkloadUpdateStrategy),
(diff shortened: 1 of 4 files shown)


controllers/operands/kubevirt_test.go
```diff
@@ -3015,14 +3015,14 @@ Version: 1.2.3`)
...
-				It("should not set replica with SingleReplica ControlPlane but HighAvailable Infrastructure ", func() {
+				It("should set replica=1 with SingleReplica ControlPlane and HighAvailable Infrastructure ", func() {
 					hcoutil.GetClusterInfo = func() hcoutil.ClusterInfo {
 						return &commontestutils.ClusterInfoSRCPHAIMock{}
 					}
 					kv, err := NewKubeVirt(hco)
 					Expect(err).ToNot(HaveOccurred())
 					Expect(kv.Spec.Infra).To(Not(BeNil()))
-					Expect(kv.Spec.Infra.Replicas).To(BeNil())
+					Expect(*kv.Spec.Infra.Replicas).To(Equal(uint8(1)))
```
nunnatsa, line 3025: nit - prefer using the `HaveValue` matcher, over getting the pointer value directly, because this is also check the `!= nil` case.

controllers/operands/kubevirt_test.go
```diff
@@ -3059,13 +3059,15 @@ Version: 1.2.3`)
...
+				It("should set replica=1 with SingleReplica ControlPlane but HighAvailable Infrastructure ", func() {
 					hcoutil.GetClusterInfo = func() hcoutil.ClusterInfo {
 						return &commontestutils.ClusterInfoSRCPHAIMock{}
 					}
 					kv, err := NewKubeVirt(hco)
 					Expect(err).ToNot(HaveOccurred())
-					Expect(kv.Spec.Infra).To(BeNil())
+					Expect(kv.Spec.Infra).To(Not(BeNil()))
+					Expect(kv.Spec.Infra.Replicas).To(Not(BeNil()))
+					Expect(*kv.Spec.Infra.Replicas).To(Equal(uint8(1)))
```
nunnatsa, line 3070: same

controllers/operands/kubevirt_test.go
```diff
@@ -3098,13 +3100,15 @@ Version: 1.2.3`)
...
```
nunnatsa, line 3111: same

controllers/operands/kubevirt_test.go
```diff
@@ -3143,14 +3147,14 @@ Version: 1.2.3`)
...
-				It("should not set replica with SingleReplica ControlPlane but HighAvailable Infrastructure ", func() {
+				It("should set replica=1 with SingleReplica ControlPlane and HighAvailable Infrastructure ", func() {
 					hcoutil.GetClusterInfo = func() hcoutil.ClusterInfo {
 						return &commontestutils.ClusterInfoSRCPHAIMock{}
 					}
 					kv, err := NewKubeVirt(hco)
 					Expect(err).ToNot(HaveOccurred())
 					Expect(kv.Spec.Infra).To(Not(BeNil()))
-					Expect(kv.Spec.Infra.Replicas).To(BeNil())
+					Expect(*kv.Spec.Infra.Replicas).To(Equal(uint8(1)))
```
nunnatsa, line 3157: same

nunnatsa (approved):
Added few nit comments inline. I don't mind to fix it in a future PR (after branching out release 1.14).

We really need to think about taking the high availability info, out of the cluster info. The clusterInfo interface is too big, and any change in it causes many changes all over the code, mostly in tests. 

But again - we can do that in a future PR, after branching out 1.14

controllers/operands/kubevirt.go
```diff
@@ -802,16 +803,24 @@ func NewKubeVirtWithNameOnly(hc *hcov1beta1.HyperConverged, opts ...string) *kub
...
 		return nil
 	}
 
 	kvConfig := &kubevirtcorev1.ComponentConfig{}
-	if !infrastructureHighlyAvailable {
+	if !controlPlaneHighlyAvailable {
 		kvConfig.Replicas = ptr.To[uint8](1)
 	}
 
+	// In case there are no worker nodes, we're setting an empty struct for NodePlacement
```
dankenigsberg, line 816: Comments are a code smell. In this case, the English says one thing, and the code says another.
It is better to have no comment. When you feel that comments are required, it is typically a sign that you should create a small peroperly-named helper function that should be used here. e.g `func kvNodePlacement(hcoConfig, controlPlaneNodeExists)`

orenc1: Since it wasn't so obvious, I felt it's reasonable to add a bit explanation for the rationale behind this condition, so future code readers will know exactly why we have this logic.

dankenigsberg: I think that future readers are no worse off, since the English comment is wrong.

orenc1: oh, I now got what you meant. Thanks for the catch. Fixed the comment in #3231

PR #3228, opened 2024-12-22
wip: Fix the monitoring test

nunnatsa:

The "KubeVirtCRModified alert should fired when there is a modification on a CR" functional test became flacky.

The test changes the KubeVirt CR, and then checks several metrics and alerts. When checking the "kubevirt_hyperconverged_operator_health_status" alert, the test sometimes fails, as KubeVirt get into non stable state for a short while for every chenage in the KubeVirt CR, making the health alret to be in "critical" status, instead of "warning".

This PR fixes the issue by modifiying the CDI CR instead of the KubeVirt CR, to get the same effect.

**Jira Ticket**:

```jira-ticket
None
```

**Release note**:

```release-note
None
```

### DIFF
--- a/tests/func-tests/hco_prometheus_route.go
+++ b/tests/func-tests/hco_prometheus_route.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"context"
 	"fmt"
+	"github.com/onsi/ginkgo/v2"
 	"net/http"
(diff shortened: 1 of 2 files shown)


nunnatsa:
Closed in favor of #3242

PR #3242, opened 2024-12-23
Fix hco system health metric values

avlitman:
**What this PR does / why we need it**:
kubevirt_hco_system_health_status values changed to value+1,
this pr: 

- remove unknown from the values options
- metric values is changeed back the values to be 0 for healthy, 1 for warning and 2 for error. 
- remove reason label
- change alert name from OperatorConditionsUnhealth to HCOOperatorConditionsUnhealth.

also tests and related recording rules and alerts updated accordingly.

updated metric (before was 1):
Screenshot from 2024-12-23 20-06-01

**Reviewer Checklist**

> Reviewers are supposed to review the PR for every aspect below one by one. To check an item means the PR is either "OK" or "Not Applicable" in terms of that item. All items are supposed to be checked before merging a PR. 

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
- [ ] Troubleshooting Friendly

**Jira Ticket**:

```jira-ticket
jira-ticket: https://issues.redhat.com/browse/CNV-52735

```

**Release note**:

```release-note
none

```

PR #3245, opened 2024-12-24
Add a workaround for a flaky test

nunnatsa:

The `"console should reach kubevirt-plugin manifests"` functional test
became flaky. The reason for that may be a bug in the console code,
that sometimes return partial and corrupted json response.

This PR reads the json in a retry mechanism to give the console plugin
pod a chance to return a valid response.

**Jira Ticket**:

```jira-ticket
none
```

**Release note**:

```release-note
none
```

### DIFF
--- a/tests/func-tests/console_plugin_test.go
+++ b/tests/func-tests/console_plugin_test.go
@@ -65,29 +65,35 @@ var _ = Describe("kubevirt console plugin", Label(tests.OpenshiftLabel, "console
 		pluginServiceName := kubevirtPlugin.Spec.Backend.Service.Name
 		pluginServicePort := kubevirtPlugin.Spec.Backend.Service.Port
 
-		consolePods := &corev1.PodList{}
-		Expect(cli.List(ctx, consolePods, client.MatchingLabels{
-			"app":       "console",
-			"component": "ui",
-		}, client.InNamespace(openshiftConsoleNamespace))).To(Succeed())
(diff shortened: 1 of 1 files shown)


tests/func-tests/console_plugin_test.go
```diff
@@ -65,29 +65,35 @@ var _ = Describe("kubevirt console plugin", Label(tests.OpenshiftLabel, "console
...
+			"name": "hyperconverged-cluster-operator",
+		}, client.InNamespace(tests.InstallNamespace))).To(Succeed())
+
+		Expect(hcoPods.Items).ToNot(BeEmpty())
+
+		testConsolePod := hcoPods.Items[0]
+		command := fmt.Sprintf(`curl -ks https://%s:%d/plugin-manifest.json`,
+			pluginServiceName, pluginServicePort)
+
+		Eventually(func(g Gomega, ctx context.Context) string {
```
avlitman, line 79: maybe let's add here a comment with the console bug number? that this Eventually can be removed once it's fixed.

avlitman:
in the pr description:
> will be useful anyway

can you explain why? since we use Eventually now? is there additional reason?

tests/func-tests/console_plugin_test.go
nunnatsa: I prefer to not add comments

nunnatsa:
Closed in favor of #3242 3242

PR #3248, opened 2024-12-26
Improve the upgrade json patch mechanism

nunnatsa:

Fix the dependencies of the release-notes tool
The tool uses an internal package, with own go.mod file with no good reason. This PR removes this go.mod file, to simplify the release-notes dependencies. Also, did some cosmetic fixes regarding text output, to make the code more clear.

Improve the upgrade json patch by
1. move it to its own package 
2. made some refactoring for better performance
3. the code is now more testable
4. move the initialization of the upgradePatch object to boot, instead of in the reconcile loop, in order to fail fast on error
5. To implement 3, add a json string of the HyperConverged CR with default values.
6. To implement 4, add a tiny tool to go-generate the go file with the json string, and call it at build time

Remove two old upgrade mechanism:
1. dropped the CRD remover. this is not needed anymore. the v2v CRD was dropped many versions ago.
2. dropped the code that remove the v1alpha1 API version from the HC CRD, as it was done many versions ago.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
None
```

**Release note**:

```release-note
Droped unused code
```


nunnatsa:
Redo. Closing

PR #3249, opened 2024-12-30
Fix smoke test on ocp-4.18

nunnatsa:
When trying to bump the ocp version to 4.18, the smoke test fails.

The cause is that for some unknown reason, the
`quay.io/kubevirt/fedora-with-test-tooling-container-disk:latest` image can't be pulled on 4.18. This image tag is misleading and it's actually very old image.

This PR is trying to fix the issue by setting the image tag to the KV release tag.

**Jira Ticket**:

```jira-ticket
None
```

**Release note**:

```release-note
None
```

### DIFF
--- a/hack/kv-smoke-tests.sh
+++ b/hack/kv-smoke-tests.sh
@@ -249,7 +249,9 @@ ${TESTS_BINARY} \
     -ginkgo.flake-attempts=3 \
     -oc-path="$(which oc)" \
     -kubectl-path="$(which oc)" \
+    -container-tag="${KUBEVIRT_VERSION}" \
     -utility-container-prefix=quay.io/kubevirt \
+    -utility-container-tag="${KUBEVIRT_VERSION}" \
     -test.timeout=3h \
     -ginkgo.timeout=3h \
     -artifacts=${ARTIFACT_DIR}/kubevirt_dump


hack/kv-smoke-tests.sh
```diff
@@ -250,6 +250,7 @@ ${TESTS_BINARY} \
     -oc-path="$(which oc)" \
     -kubectl-path="$(which oc)" \
     -utility-container-prefix=quay.io/kubevirt \
+    -utility-container-tag="${KUBEVIRT_VERSION}" \
```
akalenyu, line 253: Aren't you missing `--container-tag=` too?

akalenyu (commented):
Nice catch!

hack/kv-smoke-tests.sh
nunnatsa: I don't think so.

akalenyu: I know for fact you need it at least for some storage tests
https://github.com/kubevirt/kubevirt/blob/0702cc5fc82deffa4534be8b4ee2b1ec7bf0f927/tests/storage/events.go#L119

nunnatsa: My concern is that it will be to wide change that will introduce non-existing images. This way I can limit it. I hope it's enough...

nunnatsa: ok. let's try.

nunnatsa: done

PR #3255, opened 2024-12-31
bump golang to v1.23

nunnatsa:
**Jira Ticket**:

```jira-ticket
None
```

**Release note**:

```release-note
None
```


nunnatsa:
Needs: https://github.com/openshift/release/pull/60238

.github/workflows/build-push-images.yaml
```diff
@@ -29,7 +29,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v5
         with:
-          go-version: "1.22"
+          go-version-file: go.mod
```
avlitman, line 32: just a question, so now instead of giving the version we direct to go.mod for the version? which I see is 1.23.2?

nunnatsa: yes. And that way we don't need to edit the github action yamls each time we change the go version

PR #3265, opened 2025-01-08
Fix the override-bot

nunnatsa:
1. The override bot sometimes fail, if it runs on ubunto 24.04, and then the python version is not supported.

   Fixed it by bumping the python version to 3.13

2. The override bot sometimes works, but not performing override. The reason for that is that the statuses github API only returns the first 30 statuses. Somtimes the succeeded lane status is not in the first page, and then the code can't find it.

   Fixed it by reading all the pages of the statuses API.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
None
```

**Release note**:

```release-note
None
```

### DIFF
--- a/.github/workflows/override-bot.yaml
+++ b/.github/workflows/override-bot.yaml
@@ -2,6 +2,9 @@ name: Override Bot
 on:
   issue_comment:
     types: [created]
+  schedule:
+    - cron:  '0/10 * * * *'
+  workflow_dispatch:
 jobs:
   override:
     name: Check for redundant CI lanes for override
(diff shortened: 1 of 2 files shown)


orenc1:
thanks for the fix.
could you please also add:
```
  schedule:
    - cron: '*/10 * * * *'    
```
under `on` (below `issue_comment`) ?

orenc1 (approved):
/approve

PR #3269, opened 2025-01-09
Fix the override bot

nunnatsa:

The bot now can run from a comment in the PR, from a schedul, or
manually from github ui.

But the action condition requires a PR number to not be empty. That
makes the schedule and the manual triggers to be ignored and skipped.

This PR fixes the condition to allow the other 2 triggers.

In addition, this PR also improves the bot by:
1. add Authorization header to the HTTP calls to github, to prevent 401
   error as results of too many requests.
2. move some loops to lambdas, for better performance

**Jira Ticket**:

```jira-ticket
None
```

**Release note**:

```release-note
None
```

### DIFF
--- a/.github/workflows/override-bot.yaml
+++ b/.github/workflows/override-bot.yaml
@@ -8,7 +8,8 @@ on:
 jobs:
   override:
     name: Check for redundant CI lanes for override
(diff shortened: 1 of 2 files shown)


orenc1 (approved):
/approve

PR #3281, opened 2025-01-29
Allow customizing the CLI download link

nunnatsa:

Sometimes, the user must not use the default CLI download links. For example, if the domain is internal, but the download link need to be internal.

openshift allows modifying the openshift CLI download link by adding a component route to the Ingress cluster CR spec.

openshift suggest to add additional component routes by first add them to the ingress status by the operator, and then let the user to add them to the ingress spec if needed. the operator is responsible to modify the route and the download link, if the user modified the url.

This PR adds a new controller to HCO, to watch the ingress, add the required component to the status, and modify the route and the download links, if the user modified the url in the ingress spec.

The user should add a route component named `"virt-downloads"`, with the namesapce where HCO is deployed.

For example:
```yaml
spec:
  componentRoutes:
  - hostname: my-dl-link.apps.nahshon-cluster.cnvqedevops.qe.gcp.devcluster.openshift.com
    name: virt-downloads
```

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-55666
```

**Release note**:

```release-note
Allow customizing the CLI download link
```

### DIFF
--- a/cmd/cmdcommon/cmdcommon.go
+++ b/cmd/cmdcommon/cmdcommon.go
@@ -8,6 +8,7 @@ import (
 	"net/http/pprof"
 	"os"
 	"runtime"
+	"slices"
 
 	apiruntime "k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
@@ -144,7 +145,7 @@ func (h HcCmdHelper) checkNameSpace() {
 		communityHubNamespace,
 		communityHubTargetNamespace,
 	}
-	if !StringInSlice(actualNS, nsAllowList) {
+	if !slices.Contains(nsAllowList, actualNS) {
 		err := fmt.Errorf("%s is running in different namespace than expected", h.Name)
 		msg := fmt.Sprintf("Please re-deploy this %s into %v namespace", h.Name, requiredNS)
 		h.ExitOnError(err, msg, "Expected.Namespace", requiredNS, "Deployed.Namespace", actualNS)
@@ -164,13 +165,4 @@ func updateFlagSet(flags ...*flag.FlagSet) {
 	for _, f := range flags {
 		pflag.CommandLine.AddGoFlagSet(f)
 	}
-}
-
-func StringInSlice(a string, list []string) bool {
-	for _, b := range list {
-		if b == a {
-			return true
-		}
-	}
-	return false
 }
--- a/cmd/cmdcommon/cmdcommon.go
+++ b/cmd/cmdcommon/cmdcommon.go
@@ -10,15 +10,14 @@ import (
 	"runtime"
 	"slices"
 
+	"github.com/go-logr/logr"
+	"github.com/spf13/pflag"
 	apiruntime "k8s.io/apimachinery/pkg/runtime"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 
 	hcoutil "github.com/kubevirt/hyperconverged-cluster-operator/pkg/util"
-
-	"github.com/go-logr/logr"
-	"github.com/spf13/pflag"
-	logf "sigs.k8s.io/controller-runtime/pkg/log"
-	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 )
 
 // list of namespace allowed for HCO installations (for tests)
@@ -122,7 +121,7 @@ func (h HcCmdHelper) printVersion() {
 
 func (h HcCmdHelper) checkNameSpace() {
 	// Get the namespace that we should be deployed in.
-	requiredNS, err := hcoutil.GetOperatorNamespaceFromEnv()
+	requiredNS, err := getOperatorNamespaceFromEnv()
 	h.ExitOnError(err, "Failed to get namespace from the environment")
 
 	// Get the namespace we are currently deployed in.
@@ -165,4 +164,13 @@ func updateFlagSet(flags ...*flag.FlagSet) {
 	for _, f := range flags {
 		pflag.CommandLine.AddGoFlagSet(f)
 	}
+}
+
+func getOperatorNamespaceFromEnv() (string, error) {
+	namespace := os.Getenv(hcoutil.OperatorNamespaceEnv)
+	if len(namespace) == 0 {
+		return "", fmt.Errorf("%s unset or empty in environment", hcoutil.OperatorNamespaceEnv)
+	}
+
+	return namespace, nil
 }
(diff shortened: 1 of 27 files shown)


cmd/cmdcommon/cmdcommon.go
```diff
@@ -166,3 +166,11 @@ func updateFlagSet(flags ...*flag.FlagSet) {
 		pflag.CommandLine.AddGoFlagSet(f)
 	}
 }
+
+func getOperatorNamespaceFromEnv() (string, error) {
+	if namespace, ok := os.LookupEnv(hcoutil.OperatorNamespaceEnv); ok {
+		return namespace, nil
+	}
+
+	return "", fmt.Errorf("%s unset or empty in environment", hcoutil.OperatorNamespaceEnv)
```
orenc1, line 175: if `hcoutil.OperatorNamespaceEnv` is set but empty, it will return an empty string without an error.

nunnatsa: Anyway, this is checked on boot (cmd/*/main.go). The boot fails if the env var is not set.

cmd/cmdcommon/cmdcommon.go
```diff
@@ -166,3 +166,11 @@ func updateFlagSet(flags ...*flag.FlagSet) {
 		pflag.CommandLine.AddGoFlagSet(f)
 	}
 }
+
+func getOperatorNamespaceFromEnv() (string, error) {
```
orenc1, line 170: why this function is defined both here and in `pkg/util/util.go` ?

pkg/components/components.go
```diff
@@ -565,9 +560,19 @@ func GetClusterPermissions() []rbacv1.PolicyRule {
...
+		{
+			APIGroups: stringListToSlice(configOpenshiftIO),
+			Resources: stringListToSlice("ingresses"),
+			Verbs:     stringListToSlice("get", "list", "watch"),
+		},
+		{
+			APIGroups: stringListToSlice(configOpenshiftIO),
+			Resources: stringListToSlice("ingresses/status"),
+			Verbs:     stringListToSlice("update"),
+		},
```
tiraboschi, line 575: Not strictly needed, but maybe implementing a Validating Admission Policy (GA in k8s 1.30) to have an external mechanism as a safety guard to enforce that HCO can only touch a specific subset of the status could be a nice to have addition (even in a subsequent PR).

cmd/cmdcommon/cmdcommon.go
nunnatsa: The pkg/util version does not return error

cmd/cmdcommon/cmdcommon.go
nunnatsa: I can change that. But this is the same as what was don in the pkg/util before this change, so there is no behavior change here. it's exactly the same logic.

pkg/components/components.go
nunnatsa: I will look into it. Thanks

cmd/cmdcommon/cmdcommon.go
nunnatsa: @orenc1 - done

controllers/ingresscluster/tls.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+	if now.Before(certificate.NotBefore) {
+		return fmt.Errorf("custom TLS certificate is not valid yet")
+	}
+	return nil
+}
+
+func verifyPrivateKey(customKey []byte) error {
+	block, rest := pem.Decode(customKey)
+	fmt.Println(rest)
```
orenc1, line 31: maybe you can use a proper logger for this log?

pkg/downloadhost/domainname.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+const (
+	CLIDownloadsServiceName = "hyperconverged-cluster-cli-download"
+)
+
+type CLIDownloadHost struct {
+	DefaultHost configv1.Hostname
+	CurrentHost configv1.Hostname
+	CRT         string
```
orenc1, line 16: tiny nit. i think this name to store the certificate would be a better fit
```suggestion
	Cert         string
```

nunnatsa: fixed

controllers/ingresscluster/tls.go
nunnatsa: removed.

pkg/components/components.go
nunnatsa: Not sure it's the right mechanism for that. It's *may* be possible, but I think it's too complicated and easy to write a working CEL expression to do that.

controllers/ingresscluster/ingress-cluster-controller.go
```diff
@@ -0,0 +1,294 @@
...
+				Status:  metav1.ConditionTrue,
+				Reason:  "WrongConfiguration",
+				Message: dhErr.Error(),
+			})
+		} else {
+			meta.SetStatusCondition(&clusterIngress.Status.ComponentRoutes[routeInd].Conditions, metav1.Condition{
+				Type:   "Degraded",
+				Status: metav1.ConditionFalse,
+				Reason: "AsExpected",
+			})
```
orenc1, line 117: I think `message` is a required field in a condition.
If the final result will be an empty string in the message field, i think you should put "All is well".

pkg/components/components.go
```diff
@@ -585,6 +590,11 @@ func GetClusterPermissions() []rbacv1.PolicyRule {
 		},
 		roleWithAllPermissions("coordination.k8s.io", stringListToSlice("leases")),
 		roleWithAllPermissions("route.openshift.io", stringListToSlice("routes")),
+		{
+			APIGroups: stringListToSlice("route.openshift.io"),
+			Resources: stringListToSlice("routes/custom-host"),
+			Verbs:     stringListToSlice("get", "create", "update", "patch"),
```
orenc1, line 596: we already have `get` permission on the Route resources, which allows reading all of the route data, so no need `get` also specifically for the `routes/custom-host` subresource.

controllers/ingresscluster/ingress-cluster-controller.go
```diff
@@ -0,0 +1,299 @@
...
+				Type:    "Degraded",
+				Status:  metav1.ConditionTrue,
+				Reason:  "WrongConfiguration",
+				Message: dhErr.Error(),
+			})
+		} else {
+			meta.SetStatusCondition(&clusterIngress.Status.ComponentRoutes[routeInd].Conditions, metav1.Condition{
+				Type:    "Degraded",
+				Status:  metav1.ConditionFalse,
+				Reason:  "AsExpected",
```
orenc1, line 118: maybe `"AsExpected"` should also be defined as a const? it is used in 3 places.

orenc1 (approved):
/approve
/lgtm

PR #3292, opened 2025-02-04
Re-enable the VM console logs

nunnatsa:

After the CRI-O bug was solved, we can safly enable the logs again.

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-41605
```

**Release note**:

```release-note
Re-enable the VM console logs
```


nunnatsa:
> Please align also `disableSerialConsoleLog is a boolean, logging of serial console is disabled by default (disableSerialConsoleLog=true)` in `docs/cluster-configuration.md` which is not autogenered on code changes.

Good catch. Thanks @tiraboschi 

Fixed.

PR #3303, opened 2025-02-10
Require Bearer token for /metrics endpoint

machadovilaca:

The metrics endpoint is unprotected. This PR adds auth to the endpoint.

Screenshot from 2025-02-20 15-52-52

Alternatives tried:

- Using RBAC role to validate the caller to the endpoint. In Openshift, the Prometheus pod does not have a service account, and they suggested in the documentation to use basic auth or a token configured in the ServiceMonitor.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-56071
```

**Release note**:

```release-note
Require Bearer token for /metrics endpoint
```


pkg/jwt/bearer_token.go
```diff
@@ -0,0 +1,54 @@
+package jwt
```
nunnatsa, line 1: Please rename. This is clashes with `"github.com/golang-jwt/jwt/v5"`, and is confusing.

pkg/jwt/bearer_token.go
```diff
@@ -0,0 +1,54 @@
...
+	tokenString, err := token.SignedString(secretKey)
+	if err != nil {
+		return "", fmt.Errorf("error signing token: %v", err)
+	}
+
+	return tokenString, nil
+}
+
+func ValidateToken(tokenString string) (bool, error) {
+	token, err := jwt.Parse(tokenString, func(token *jwt.Token) (interface{}, error) {
```
nunnatsa, line 32: please replace `interface{}` ==> `any`

pkg/jwt/http.go
```diff
@@ -0,0 +1,37 @@
...
+			valid, err := ValidateToken(token)
+			if err != nil {
+				http.Error(w, "Unauthorized", http.StatusUnauthorized)
+				return
+			}
+
+			if !valid {
+				http.Error(w, "Unauthorized", http.StatusUnauthorized)
+				return
+			}
```
nunnatsa, line 32: can be unified:
```suggestion
			if err != nil || !valid{
				http.Error(w, "Unauthorized", http.StatusUnauthorized)
				return
			}
```

controllers/alerts/secret.go
```diff
@@ -0,0 +1,104 @@
...
+	modified := false
+
+	logger.Info("creating bearer token")
+	token, err := jwt.CreateToken()
+	if err != nil {
+		logger.Info("failed to create bearer token", "message", err)
+		token = ""
+	}
+
+	logger.Info("bearer token created", "token", token)
```
nunnatsa, line 56: Please avoid printing out the token. It may be sensitive

controllers/alerts/secret.go
```diff
@@ -0,0 +1,104 @@
+package alerts
```
nunnatsa, line 1: I'm not sure I understand why we need to store the token in a secret. First, this is a security risk - not sure we need this exposure. In addition, the only usage (I found) is for e2e tests, and we don't use it for production.

So unless I miss something, I would drop this secret.

nunnatsa (commented):
Added some inline comments

pkg/jwt/bearer_token.go
machadovilaca: changed to `authorization`

pkg/jwt/bearer_token.go
machadovilaca: done

pkg/jwt/http.go
machadovilaca: done

controllers/alerts/secret.go
machadovilaca: ofc

controllers/alerts/secret.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	return found, modified, nil
+}
+
+func NewSecret(namespace string, owner metav1.OwnerReference) *corev1.Secret {
+	token, err := authorization.CreateToken()
+	if err != nil {
+		logger.Info("failed to create bearer token", "message", err)
+		token = ""
```
nunnatsa, line 79: Should we create the secret w/o the token?

pkg/authorization/bearer_token_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+var _ = Describe("BearerToken", func() {
+	BeforeEach(func() {
+		tmpFile, err := os.CreateTemp("", "fake-token")
+		Expect(err).ToNot(HaveOccurred())
+
+		_, err = tmpFile.Write([]byte("test-secret-key"))
+		Expect(err).ToNot(HaveOccurred())
+
+		os.Setenv("KUBERNETES_SERVICE_TOKEN_PATH", tmpFile.Name())
```
nunnatsa, line 20: Please use the `tokenPathEnvVar` const, or if you don't want to expose it, redefine it here.

pkg/components/components.go
```diff
@@ -612,6 +612,11 @@ func GetClusterPermissions() []rbacv1.PolicyRule {
 			Resources: stringListToSlice("alertmanagers", "alertmanagers/api"),
 			Verbs:     stringListToSlice("get", "list", "create", "delete"),
 		},
+		{
+			APIGroups: stringListToSlice(""),
+			Resources: stringListToSlice("secrets"),
+			Verbs:     stringListToSlice("get", "list", "create", "update", "watch"),
+		},
```
nunnatsa, line 619: please add the missing permissions to the RBAC object in line 516, instead of this new object.

nunnatsa (commented):
Thanks @machadovilaca 

Added a few comments

controllers/alerts/secret.go
machadovilaca: I didn't want to throw an error and crash the whole pod
nor not create the secret at all, and disallow the access to the endpoint

I thought about generating a secret and keeping it in memory, maybe that's better, wdyt?

@nunnatsa

pkg/authorization/bearer_token_test.go
machadovilaca: done

pkg/components/components.go
machadovilaca: done

controllers/alerts/secret.go
nunnatsa: I don't think we need to kill the pod. Just try again in the next reconcile loop. Will ot work?

machadovilaca: I wonder if there is just some configuration where the `serviceaccount/token` is not populated in the pod, it will not exist in any further reconcilations

pkg/authorization/bearer_token.go
```diff
@@ -0,0 +1,75 @@
...
+	if err != nil {
+		return "", fmt.Errorf("error signing token: %v", err)
+	}
+
+	return tokenString, nil
+}
+
+func ValidateToken(tokenString string) (bool, error) {
+	token, err := jwt.Parse(tokenString, func(token *jwt.Token) (any, error) {
+		return getSecretKey()
```
nunnatsa, line 37: Could the secret key file changed while the pod is running? if not, then reading from a file is expensive, and I think we should read it on boot, or at least do it only once, and keep the secret in memory (e.g. by using sync.Once).

pkg/authorization/bearer_token.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/golang-jwt/jwt/v5"
+)
+
+const (
+	TokenPathEnvVar = "KUBERNETES_SERVICE_TOKEN_PATH"
+
+	defaultTokenPath = "/var/run/secrets/kubernetes.io/serviceaccount/token"
+)
+
+var inMemoryToken *[]byte
```
nunnatsa, line 17: no need to use a pointer to slice. non initiated slice is already nil.

machadovilaca: done

pkg/authorization/bearer_token.go
machadovilaca: since we were already keeping the generated token in memory, maybe we can also keep this one, and just check if it is populated already

nunnatsa: yes, exactly.

PR #3308, opened 2025-02-18
Fix issues with build multi-arch images

nunnatsa:
When building image with different architecture than the build machine architecture, any binary that runs as part of the build will fail (even just creating a user).

Another issue is that github action uses ubuntu, that its podman version is 4.x, and cannot build manifest from local images, as possible in podman 5.x.

docker/podman needs qemu installed on the build machine, to allow running binaries in diffferent architecture as part of the image build process.

This PR installs qemu in the relevant  github actions, and pushes the images before building the manifest, to match the podman 4.x patter.

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-55711
```

**Release note**:

```release-note
None
```

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -7,7 +7,9 @@ ARG TARGETARCH
 
 RUN GOOS=${TARGETOS} GOARCH=${TARGETARCH} make build-operator build-csv-merger
 
-FROM registry.access.redhat.com/ubi9/ubi-minimal
+FROM --platform=${TARGETPLATFORM} registry.access.redhat.com/ubi9/ubi-minimal
+ARG TARGETPLATFORM
+
 ENV OPERATOR=/usr/local/bin/hyperconverged-cluster-operator \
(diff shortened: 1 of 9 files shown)


build/Dockerfile
```diff
@@ -33,4 +35,4 @@ ARG git_sha=NONE
 LABEL multi.GIT_URL=${git_url} \
       multi.GIT_SHA=${git_sha} \
       app=hyperconverged-cluster-operator \
-      arch=${ARCH}
+      golang.build.platform=${TARGETPLATFORM}
```
orenc1, line 38: `arch` label is not needed for HCO images?

nunnatsa: this is a label I added in the previous PR, for us, for debug. not a requirement. just rename it. We also get the `architecture` label from the base images.

orenc1 (approved):
/approve

PR #3312, opened 2025-02-20
Fix issue in AWS lanes

nunnatsa:

If the worker is with ARM architecture, CI failes to build the index image, as it explicitly uses AMD64 base image.

This PR change the base image to be a multi-arch manifest, to allow ARM build.

Also, fix upgrade-issue in AWS:
Many upgrade test are failing in AWS, when the VM launcher image is not replaced after upgrade.

The nodes in AWS does not have the same labels, and KV can't find a node live-migrate the VM.

This commit disables the VM's "gds-no" CPU feature, so the VM should now find avaliable node.

**Jira Ticket**:

```jira-ticket
None
```

**Release note**:

```release-note
None
```


nunnatsa:
Close in favor of #3314

PR #3313, opened 2025-02-21
Fix another issues in multi-arch images

nunnatsa:
Follow-up of PR #3308 - need to push the images before creating the manifest, also for the index image.

Also, fix nightly test by installing qemu-user-static, to allow running binaries of one arch, in image build of another arch.

**Release note**:

```release-note
None
```

### DIFF
--- a/automation/nightly/test-nightly-build.sh
+++ b/automation/nightly/test-nightly-build.sh
@@ -18,9 +18,12 @@ function dump() {
 
 # Get golang
 $CRI_BIN login --username "$(cat "${QUAY_USER}")" --password-stdin quay.io < "${QUAY_PASSWORD}"
-wget -q https://dl.google.com/go/go1.22.6.linux-amd64.tar.gz
+wget -q https://dl.google.com/go/go1.23.5.linux-amd64.tar.gz
 tar -C /usr/local -xf go*.tar.gz
 export PATH=/usr/local/go/bin:$PATH
+
+# add qemu-user-static
+$CRI_BIN run --rm --privileged docker.io/multiarch/qemu-user-static --reset -p yes
 
 # get latest KubeVirt commit
 latest_kubevirt=$(curl -sL https://storage.googleapis.com/kubevirt-prow/devel/nightly/release/kubevirt/kubevirt/latest)
--- a/hack/build-index-image.sh
+++ b/hack/build-index-image.sh
@@ -84,8 +84,10 @@ function create_file_based_catalog() {
 
   IMAGES=
   for arch in ${ARCHITECTURES}; do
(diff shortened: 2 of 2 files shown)


automation/nightly/test-nightly-build.sh
```diff
@@ -18,10 +18,12 @@ function dump() {
 
 # Get golang
 $CRI_BIN login --username "$(cat "${QUAY_USER}")" --password-stdin quay.io < "${QUAY_PASSWORD}"
-wget -q https://dl.google.com/go/go1.22.6.linux-amd64.tar.gz
+wget -q https://dl.google.com/go/go1.23.5.linux-amd64.tar.gz
 tar -C /usr/local -xf go*.tar.gz
 export PATH=/usr/local/go/bin:$PATH
 
+dnf install -y qemu-user-static
```
orenc1, line 25: this is not a dockerfile. are you sure we have `dnf` where this script is running?

nunnatsa: replaced with:
```shell
$CRI_BIN run --rm --privileged docker.io/multiarch/qemu-user-static --reset -p yes
```

automation/nightly/test-nightly-build.sh
```diff
@@ -18,10 +18,13 @@ function dump() {
 
 # Get golang
 $CRI_BIN login --username "$(cat "${QUAY_USER}")" --password-stdin quay.io < "${QUAY_PASSWORD}"
-wget -q https://dl.google.com/go/go1.22.6.linux-amd64.tar.gz
+wget -q https://dl.google.com/go/go1.23.5.linux-amd64.tar.gz
 tar -C /usr/local -xf go*.tar.gz
 export PATH=/usr/local/go/bin:$PATH
 
+# add qemu-user-static
+$CRI_BIN run --rm --privileged docker.io/multiarch/qemu-user-static --reset -p yes
```
orenc1, line 26:
```suggestion
$CRI_BIN run --rm --privileged docker.io/multiarch/qemu-user-static --reset --persistent yes
```
for readability.

also, is this image available in quay.io as well?
because we can hit the docker.io rate limit

nunnatsa: This is running once, each night, so I'm not so worry about rate limit. but checking about quai.io

orenc1: but other jobs from kubevirt prow might pull images from docker hub, and they are originated from the same IP.

nunnatsa: This is (1) not very common; and (b) if it's the same IP, then the image is already pulled locally.

orenc1 (approved):
/approve

PR #3315, opened 2025-02-26
Fix the nightly build

nunnatsa:

Push the virt-artifact-server image to the right registry.

Simplify the image naming to ease the change. Use the right environment variable names to be used correctly by the Makefile.

**Release note**:

```release-note
None
```


orenc1 (approved):
/approve

PR #3316, opened 2025-02-27
Don't use the default SA in the artifact-server

nunnatsa:
* Create new ServiceAccount for the artifact-server, with no permissions.
* Set the ServiceAccountName to the new SA, in the pod template
* Set the `automountServiceAccountToken` field to false.

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-57139
```

**Release note**:

```release-note
Don't use the default SA in the artifact-server
```


nunnatsa:
rebased to fix a conflict

PR #3325, opened 2025-03-03
Fix the publish github action

nunnatsa:
The "Publish Tagged Version to Community Operators" github action fails because when setting up golang, we want to use golang version from the the go.mod file, but this is checked out only in the next step.

Fix by replacing the order of the step, so we first checking out the code and only then setting up golang.

**Release note**:

```release-note
None
```


orenc1 (approved):
/approve

PR #3329, opened 2025-03-06
Remove old api version from the SSP CRD on upgrade

orenc1:
Since version 1.10, SSP moved to a new api version - `v1beta2`.
In latest release, SSP completely removed the old API `v1beta1` from their CRD.
This might result in an upgrade issue with OLM, if the cluster has initially started with version 1.9 or below.
This PR removes the old SSP api version v1beta1 from the storedVersions list in the SSP CRD status, to allow a smooth upgrade to 1.14.0.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-57082
```

**Release note**:

```release-note
NONE
```

### DIFF
--- a/controllers/hyperconverged/hyperconverged_controller.go
+++ b/controllers/hyperconverged/hyperconverged_controller.go
@@ -21,6 +21,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	schedulingv1 "k8s.io/api/scheduling/v1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	apimetav1 "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -397,6 +398,12 @@ func (r *ReconcileHyperConverged) handleUpgrade(req *common.HcoRequest) (*reconc
 	modified, err := r.migrateBeforeUpgrade(req)
 	if err != nil {
 		return &reconcile.Result{Requeue: true}, err
+	}
+
+	const sspCrd = "ssps.ssp.kubevirt.io"
+	const sspApiVersionToRemove = "v1beta1"
+	if err = r.removeOldApifromCrd(req, sspCrd, sspApiVersionToRemove); err != nil {
+		return &reconcile.Result{Requeue: false}, err
 	}
 
 	if modified {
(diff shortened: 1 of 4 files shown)


controllers/hyperconverged/hyperconverged_controller.go
```diff
@@ -1041,6 +1048,33 @@ func (r *ReconcileHyperConverged) migrateBeforeUpgrade(req *common.HcoRequest) (
...
+	if !slices.Contains(crd.Status.StoredVersions, removedApiVersion) {
+		return nil
+	}
+	var newStoredVersions []string
+	for _, ver := range crd.Status.StoredVersions {
+		if ver == removedApiVersion {
+			continue
+		}
+		newStoredVersions = append(newStoredVersions, ver)
+	}
```
nunnatsa, line 1067: I think we can use methods from the slices package, instead of this loop.; e.g.
```suggestion
	i := slices.Index(crd.Status.StoredVersions, removedApiVersion)
	if i == -1 {
		return nil
	}
	
	slices.Delete(crd.Status.StoredVersions, i, i+1)
```

I think it's cleaner. WDYT?

nunnatsa: But actually, I think we should try to use json patch here. So maybe the we only need `slices.Index`: if it `-1`, we can we're fine. if not, we can build a json patch based on this index.

controllers/hyperconverged/hyperconverged_controller.go
```diff
@@ -1041,6 +1048,33 @@ func (r *ReconcileHyperConverged) migrateBeforeUpgrade(req *common.HcoRequest) (
...
+
+	crd.Status.StoredVersions = newStoredVersions
+	if err := r.client.Status().Update(req.Ctx, &crd); err != nil {
+		req.Logger.Error(err, fmt.Sprintf("failed to remove %v from status of CRD %v", removedApiVersion, crdName))
+		return err
+	}
```
nunnatsa, line 1073: Let's try to do it with json patch

nunnatsa (commented):
Added a few comment. I think it's makes sense. I would consider to add a boolean (in the reconciler? not sure where) to not read and parse the CRD again, if it wasn't change.

We're not caching the CRD, so reading it over and over again may be pricey.

nunnatsa (commented):
Looks good!

Added one comment inline.

controllers/hyperconverged/hyperconverged_controller.go
```diff
@@ -1041,6 +1048,37 @@ func (r *ReconcileHyperConverged) migrateBeforeUpgrade(req *common.HcoRequest) (
...
+	var crd apiextensionsv1.CustomResourceDefinition
+	if err := r.client.Get(req.Ctx, types.NamespacedName{Name: crdName}, &crd); err != nil {
+		req.Logger.Error(err, fmt.Sprintf("unable to get %s CRD", crdName))
+		return err
+	}
+
+	if !slices.Contains(crd.Status.StoredVersions, removedApiVersion) {
+		return nil
+	}
+
```
nunnatsa, line 1061: We don't need this. We're doing it in line 1063

controllers/hyperconverged/hyperconverged_controller.go
```diff
@@ -1041,6 +1048,37 @@ func (r *ReconcileHyperConverged) migrateBeforeUpgrade(req *common.HcoRequest) (
...
+	versionIndex := slices.Index(crd.Status.StoredVersions, removedApiVersion)
+	if versionIndex == -1 {
+		return nil
+	}
+
+	patchBytes := []byte(fmt.Sprintf(`[
+					{"op":"test", "path":"/status/storedVersions/%d", "value": "%s"},
+					{"op":"remove", "path":"/status/storedVersions/%d"}]`,
+		versionIndex, removedApiVersion, versionIndex))
```
nunnatsa, line 1070: nit: we can use indexes, to not use the same parameters again; e.g.
```suggestion
	patchBytes := []byte(fmt.Sprintf(`[
					{"op":"test", "path":"/status/storedVersions/%[1]d", "value": "%[2]s"},
					{"op":"remove", "path":"/status/storedVersions/%[1]d"}]`,
		versionIndex, removedApiVersion))
```

nunnatsa (commented):
Added two inline comment. 

I wonder if we can write the e2e test in golang.

orenc1:
this fix can't work on main and 1.14. need to do it at 1.13 when `v1beta1` still exists on the SSP CRD.  
closing in favor of #3332

PR #3341, opened 2025-03-13
KubeVirtCRModified e2e: Remove prometheus check for metric value

avlitman:
In KubeVirtCRModified alert e2e test we check the metric value in hco no need to check also prometheus

**Release note**:

```release-note
none
```


machadovilaca:
@avlitman please update the PR title

avlitman:
> @avlitman please update the PR title (and the commit title)

Right thanks, is it ok now?

nunnatsa (commented):
/approve

PR #3360, opened 2025-03-27
Fix issue in upgrade tests

nunnatsa:

We have an inherent issue in our upgrade tests. The test is taking the unstable bundles from quay. We build these bundles each time we merge a PR.

All the images in the unstable bundles are listed with their digests, rather than their tag, in order to support disconnected environment.

If this process is failed in the middle, e.g. when the operator image was already built and pushed to quay.io, but the bundle image failed, then the bundle is actually broken: it is an old version of the bundle, pointing to images that are no longer there, because the operator images are was already replaced.

This PR fixes this issue by first pushing the images with a temporary tag, and only if everything went well, retugs the images with the unstable tag.

As a side effect, the old images are not been GCed and we can investigate issues if needed.

Here is how it will look like in quay.io:
image

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
None
```

**Release note**:

```release-note
None
```

### DIFF
--- a/.github/workflows/build-push-images.yaml
+++ b/.github/workflows/build-push-images.yaml
@@ -41,8 +41,7 @@ jobs:
       - name: set unstable tag
         # This workflow always pushes unstable tags, both from main and release branches.
         run: |
-          echo "IMAGE_TAG=${CSV_VERSION}-unstable" >> $GITHUB_ENV
-          echo "UNSTABLE=UNSTABLE" >> $GITHUB_ENV
+          echo "IMAGE_TAG=$(date -u +%Y%m%d-%H%M%S) >> $GITHUB_ENV
 
       - name: Build and Push Applications Images
         env:
@@ -71,4 +70,10 @@ jobs:
       - name: Build and Push the Index Image
         run: |
           export OPM=$(pwd)/linux-amd64-opm
-          ./hack/build-index-image.sh latest ${{ env.UNSTABLE }}
+          ./hack/build-index-image.sh latest UNSTABLE
+      - name: Re-tag and Push all the Images
+        env:
+          IMAGE_TAG: ${{ env.IMAGE_TAG }}
+          CSV_VERSION: ${{ env.CSV_VERSION }}
+        run: |
+          IMAGE_TAG="${IMAGE_TAG}" NEW_TAG="${CSV_VERSION}-unstable" make retag-push-all-images
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,8 @@ OPERATOR_IMAGE     ?= $(REGISTRY_NAMESPACE)/hyperconverged-cluster-operator
 WEBHOOK_IMAGE      ?= $(REGISTRY_NAMESPACE)/hyperconverged-cluster-webhook
 FUNC_TEST_IMAGE    ?= $(REGISTRY_NAMESPACE)/hyperconverged-cluster-functest
 VIRT_ARTIFACTS_SERVER ?= $(REGISTRY_NAMESPACE)/virt-artifacts-server
+BUNDLE_IMAGE       ?= $(REGISTRY_NAMESPACE)/hyperconverged-cluster-bundle
+INDEX_IMAGE        ?= $(REGISTRY_NAMESPACE)/hyperconverged-cluster-index
 LDFLAGS            ?= -w -s
 GOLANDCI_LINT_VERSION ?= v1.64.4
 HCO_BUMP_LEVEL ?= minor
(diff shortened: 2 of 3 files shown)


hack/retag-multi-arch-images.sh
```diff
@@ -0,0 +1,33 @@
...
+
+if [[ "${MULTIARCH}" == "true" ]]; then
+  for arch in ${ARCHITECTURES}; do
+    NEW_IMAGE="${NEW_IMAGE_REG}:${NEW_TAG}-${arch}"
+    . "hack/cri-bin.sh" && ${CRI_BIN} tag "${IMAGE_REG}:${CURRENT_TAG}-${arch}" "${NEW_IMAGE}"
+    . "hack/cri-bin.sh" && ${CRI_BIN} push "${NEW_IMAGE}"
+  done
+fi
+
+# retug the manifest
```
orenc1, line 30:
```suggestion
# retag the manifest
```

.github/workflows/build-push-images.yaml
```diff
@@ -41,8 +41,7 @@ jobs:
       - name: set unstable tag
         # This workflow always pushes unstable tags, both from main and release branches.
         run: |
-          echo "IMAGE_TAG=${CSV_VERSION}-unstable" >> $GITHUB_ENV
-          echo "UNSTABLE=UNSTABLE" >> $GITHUB_ENV
+          echo "IMAGE_TAGE=$(date -u +%Y%m%d-%H%M%S) >> $GITHUB_ENV
```
orenc1, line 44:
```suggestion
          echo "IMAGE_TAG=$(date -u +%Y%m%d-%H%M%S) >> $GITHUB_ENV
```

Makefile
```diff
@@ -145,6 +147,14 @@ container-push-functest:
 container-push-artifacts-server:
 	. "hack/cri-bin.sh" && $$CRI_BIN manifest push $(IMAGE_REGISTRY)/$(VIRT_ARTIFACTS_SERVER):$(IMAGE_TAG)
 
+retag-push-all-images:
+	IMAGE_REG=$(IMAGE_REGISTRY)/$(OPERATOR_IMAGE) MULTIARCH=true CURRENT_TAG=$(IMAGE_TAG) NEW_TAG=$(NEW_TAG) ./hack/retag-multi-arch-images.sh
+	IMAGE_REG=$(IMAGE_REGISTRY)/$(WEBHOOK_IMAGE) MULTIARCH=true CURRENT_TAG=$(IMAGE_TAG) NEW_TAG=$(NEW_TAG) ./hack/retag-multi-arch-images.sh
+	IMAGE_REG=$(IMAGE_REGISTRY)/$(FUNC_TEST_IMAGE) MULTIARCH=true CURRENT_TAG=$(IMAGE_TAG) NEW_TAG=$(NEW_TAG) ./hack/retag-multi-arch-images.sh
+	IMAGE_REG=$(IMAGE_REGISTRY)/$(VIRT_ARTIFACTS_SERVER) MULTIARCH=true CURRENT_TAG=$(IMAGE_TAG) NEW_TAG=$(NEW_TAG) ./hack/retag-multi-arch-images.sh
+	IMAGE_REG=$(IMAGE_REGISTRY)/$(BUNDLE_IMAGE) CURRENT_TAG=$(IMAGE_TAG) NEW_TAG=$(NEW_TAG) ./hack/retag-multi-arch-images.sh
+	IMAGE_REG=$(IMAGE_REGISTRY)/$(INDEX_IMAGE) MULTIARCH=true CURRENT_TAG=$(IMAGE_TAG) NEW_TAG=$(NEW_TAG) ./hack/retag-multi-arch-images.sh
```
orenc1, line 156: shouldn't the first var we're passing to the retag script be named `IMAGE_NAME` ?
`IMAGE_REG` implies only the registry name (e.g. quay.io)

hack/retag-multi-arch-images.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+
+ARCHITECTURES="amd64 arm64 s390x"
+
+if [[ -z ${IMAGE_REG} ]]; then
+  echo "IMAGE_NAME must be defined"
```
orenc1, line 6: mismatch condition and error message

Makefile
nunnatsa: I don't think so. First, image registry refer to the the image in general (not a specific image tag) - not only the server. e.g. `quay.io/kubevirt/hyperconverged-cluster-operator`. Image name is the full name, with the tag.

orenc1: i think you were describing the term "image repository", which it the image name without the tag(s).
image registry is the registry server, as defined in line 6

nunnatsa:
sanity test fails because coverall.io is down for maintenance, and the sanity action fails to send the report. 

Since this PR has nothing to do with anything tested in sanity, trying to override the failure.

/override "Sanity Checks / Sanity Checks"

orenc1 (approved):
/approve

PR #3379, opened 2025-04-03
Fix a bug when setting a CPU quantity without a type

nunnatsa:

When setting, for example, the
`spec.resourceRequirements.storageWorkloads.limit` field to `"1.5"`, the client set the value in CDI to `"1500m"`. Then HCO will always find this value as not matches to the required value of `"1.5"`, and will constantly try to update CDI.

This PR fixes this issue by turning the required objects to json and back to objects. The json marshaling performs the same change in the object. Now when HCO will compare the required value with the existing one, both will be at the newer format, with the quantity type, and HCO will understand that no change was done, and it will not try to update the CR again.

Also. This PR adds a mutex protection to the cache in the operand handlers.

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-58521
```

**Release note**:

```release-note
Fix a bug when setting a CPU quantity without a type
```

### DIFF
--- a/controllers/operands/cdi_test.go
+++ b/controllers/operands/cdi_test.go
@@ -5,12 +5,10 @@ import (
 	"maps"
 	"time"
 
-	openshiftconfigv1 "github.com/openshift/api/config/v1"
-	conditionsv1 "github.com/openshift/custom-resource-status/conditions/v1"
-
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-
+	openshiftconfigv1 "github.com/openshift/api/config/v1"
+	conditionsv1 "github.com/openshift/custom-resource-status/conditions/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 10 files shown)


controllers/operands/cdi_test.go
```diff
@@ -5,12 +5,10 @@ import (
...
 	"time"
 
-	openshiftconfigv1 "github.com/openshift/api/config/v1"
-	conditionsv1 "github.com/openshift/custom-resource-status/conditions/v1"
-
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-
+	openshiftconfigv1 "github.com/openshift/api/config/v1"
+	conditionsv1 "github.com/openshift/custom-resource-status/conditions/v1"
```
avlitman, line 11: add an empty line after line 11

nunnatsa: why? no need for new line here

PR #3380, opened 2025-04-03
mutator: default evictionStrategy to None on ARM64 clusters

dasionov:

Ensure that newly created HyperConverged resources default to EvictionStrategyNone when all worker nodes in the cluster are ARM64. This prevents unexpected live migrations on architectures where it may not be supported. The mutation only applies if the eviction strategy is unset, allowing users to override it later.

**Reviewer Checklist**
- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:
```jira-ticket
https://issues.redhat.com/browse/CNV-58613
```

**Release note**:
```release-note
none
```


pkg/webhooks/mutator/hyperConvergedMutator.go
```diff
@@ -106,20 +107,42 @@ func (hcm *HyperConvergedMutator) mutateHyperConverged(_ context.Context, req ad
...
+func mutateEvictionStrategy(ctx context.Context, cli client.Client, hc *hcov1beta1.HyperConverged, patches []jsonpatch.JsonPatchOperation) []jsonpatch.JsonPatchOperation {
 	if hc.Status.InfrastructureHighlyAvailable == nil || hc.Spec.EvictionStrategy != nil { // New HyperConverged CR
 		return patches
 	}
 
-	var value = kubevirtcorev1.EvictionStrategyNone
-	if *hc.Status.InfrastructureHighlyAvailable {
-		value = kubevirtcorev1.EvictionStrategyLiveMigrate
+	workerNodes := &corev1.NodeList{}
+	err := cli.List(ctx, workerNodes, client.MatchingLabels{"node-role.kubernetes.io/worker": ""})
```
nunnatsa, line 116: not sure this label is enough. I case the workloads node-placement is set, I think we will need to use it instead.

@orenc1 - WDYT?

pkg/webhooks/mutator/hyperConvergedMutator.go
```diff
@@ -106,20 +107,42 @@ func (hcm *HyperConvergedMutator) mutateHyperConverged(_ context.Context, req ad
...
+
+	var evictionStrategy kubevirtcorev1.EvictionStrategy
+
+	switch {
+	case allArm64:
+		evictionStrategy = kubevirtcorev1.EvictionStrategyNone
+	case hc.Status.InfrastructureHighlyAvailable != nil && *hc.Status.InfrastructureHighlyAvailable:
+		evictionStrategy = kubevirtcorev1.EvictionStrategyLiveMigrate
+	default:
+		evictionStrategy = kubevirtcorev1.EvictionStrategyNone
```
nunnatsa, line 139: nit (not sure it's better): we only have two cases here. I wonder if this we;ll be more clear:
```suggestion
    evictionStrategy := kubevirtcorev1.EvictionStrategyNone
    if !allArm64 && hc.Status.InfrastructureHighlyAvailable != nil && *hc.Status.InfrastructureHighlyAvailable {
       evictionStrategy = kubevirtcorev1.EvictionStrategyLiveMigrate
    }
```

nunnatsa (commented):
Added some inline comments. you can ignore the nit one if you find it not heelping.

as a general comment, I would prefer to reuse the node controller, but this will probably require some API change in the status field, and we'll need to do that anyway to support heterogeneous cluster in the future, and this is not designed yet. so for now we'll have to do that this way.

pkg/webhooks/mutator/hyperConvergedMutator.go
dasionov: Thanks, it's quite nice actually.

dasionov:
I also Adjusted the unit tests

pkg/webhooks/mutator/hyperConvergedMutator.go
```diff
@@ -106,20 +107,36 @@ func (hcm *HyperConvergedMutator) mutateHyperConverged(_ context.Context, req ad
...
-func mutateEvictionStrategy(hc *hcov1beta1.HyperConverged, patches []jsonpatch.JsonPatchOperation) []jsonpatch.JsonPatchOperation {
+func mutateEvictionStrategy(ctx context.Context, cli client.Client, hc *hcov1beta1.HyperConverged, patches []jsonpatch.JsonPatchOperation) []jsonpatch.JsonPatchOperation {
 	if hc.Status.InfrastructureHighlyAvailable == nil || hc.Spec.EvictionStrategy != nil { // New HyperConverged CR
 		return patches
 	}
 
-	var value = kubevirtcorev1.EvictionStrategyNone
-	if *hc.Status.InfrastructureHighlyAvailable {
-		value = kubevirtcorev1.EvictionStrategyLiveMigrate
+	workerNodes := &corev1.NodeList{}
```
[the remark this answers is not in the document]
nunnatsa: > This makes a webhook depend on external resources, which is something we try to avoid as much as possible...

You are correct. I don't like it either. Also, we do have a node controller to watch the nodes. But this controller is running in the operator, on different pod. We will fix this behavior when implementing the multi-arch cluster support. For now, I can't see better option.

> Also this will happen just once and ignore new nodes added later.

This is code only sets the default value of the `spec.evictionStrategy` field, so it must only run once, and that's ok, assuming that this is the case (setting the field on creation of the HyperConverged PR) for ARM clusters, as it wasn't supported until now.

pkg/webhooks/mutator/hyperConvergedMutator_test.go
```diff
@@ -264,9 +265,17 @@ var _ = Describe("test HyperConverged mutator", func() {
...
+				err := cli.List(context.TODO(), nodeList, client.MatchingLabels{"node-role.kubernetes.io/worker": ""})
+				Expect(err).ToNot(HaveOccurred())
+				for i := range nodeList.Items {
+					node := &nodeList.Items[i]
+					Expect(cli.Delete(context.TODO(), node)).To(Succeed())
+				}
 			})
 
-			DescribeTable("check EvictionStrategy default", func(SNO bool, strategy *kubevirtcorev1.EvictionStrategy, patches []jsonpatch.JsonPatchOperation) {
+			DescribeTable("check EvictionStrategy default", func(SNO bool, strategy *kubevirtcorev1.EvictionStrategy, patches []jsonpatch.JsonPatchOperation, workerNodeArches []string) {
```
nunnatsa, line 278: This test is a unit test. Adding and removing nodes using the client is overkill. We can just create a new fake client with the required nodes for the specific entry.

e.g.
```suggestion
			DescribeTable("check EvictionStrategy default", func(SNO bool, strategy *kubevirtcorev1.EvictionStrategy, patches []jsonpatch.JsonPatchOperation, workerNodeArches []string) {
			
				cr.Status.InfrastructureHighlyAvailable = ptr.To(!SNO) // <<<=== this change is not related to this PR, but please apply it

				cr.Spec.EvictionStrategy = strategy
				cli = commontestutils.InitClient(createFakeNodes(workerNodeArches))
				mutator = initHCMutator(s, cli)
```

pkg/webhooks/mutator/hyperConvergedMutator_test.go
```diff
@@ -264,9 +265,17 @@ var _ = Describe("test HyperConverged mutator", func() {
...
 			AfterEach(func() {
 				util.GetClusterInfo = getClusterInfo
+
+				nodeList := &corev1.NodeList{}
+				err := cli.List(context.TODO(), nodeList, client.MatchingLabels{"node-role.kubernetes.io/worker": ""})
+				Expect(err).ToNot(HaveOccurred())
+				for i := range nodeList.Items {
+					node := &nodeList.Items[i]
+					Expect(cli.Delete(context.TODO(), node)).To(Succeed())
+				}
```
nunnatsa, line 275: This is not needed, if we'll create the cli inside the test. please drop this code.

pkg/webhooks/mutator/hyperConvergedMutator_test.go
```diff
@@ -465,3 +522,28 @@ func initHCMutator(s *runtime.Scheme, testClient client.Client) *HyperConvergedM
...
+				Labels: map[string]string{
+					"node-role.kubernetes.io/worker": "",
+					"kubernetes.io/arch":             arch,
+				},
+			},
+		}
+		nodes = append(nodes, node)
+	}
+	return nodes
+}
```
nunnatsa, line 549: This is overkill and not needed, if we'll create the cli inside the test. please drop this function.

pkg/webhooks/mutator/hyperConvergedMutator_test.go
```diff
@@ -264,9 +265,17 @@ var _ = Describe("test HyperConverged mutator", func() {
 
 			AfterEach(func() {
 				util.GetClusterInfo = getClusterInfo
+
```
nunnatsa, line 268: As a follow up of #3205, please remove line 264 and the whole `AfterEach` method, as this is not needed anymore (after removing the new code from `AfterEach`).

nunnatsa (commented):
Thanks for the PR @dasionov! 

We can simplify the unit tests; see the inline comments.

pkg/webhooks/mutator/hyperConvergedMutator_test.go
dasionov: but i want to have fake node objects.. so if we are going to use 

`cli = commontestutils.InitClient(createFakeNodes(workerNodeArches))` we still needs this specific function am i wrong? 
the rest will be removed no doubt.

pkg/webhooks/mutator/hyperConvergedMutator_test.go
dasionov: removed this part

pkg/webhooks/mutator/hyperConvergedMutator_test.go
dasionov: done

pkg/webhooks/mutator/hyperConvergedMutator_test.go
nunnatsa: You are right. I meant the `addFakeNodes` function that you already removed.

PR #3383, opened 2025-04-07
Use the official coveralls github action

nunnatsa:
Until now, we're using the shogo82148/actions-goveralls github action to send the coverage information to coverall.io.

It's appears that there is an official github action for that, and we want to use it to be in sync with coveralls.io, and get better support.

**Jira Ticket**:

```jira-ticket
None
```

**Release note**:

```release-note
None
```


nunnatsa:
I want to see it working before merging it. Currently coveralls.io is down, so we can't check if the change is actually working

/hold

PR #3384, opened 2025-04-07
re-enable the revive linter rules

nunnatsa:
Apparently, when adding a specific rule setting to revive setting, it cancels all the other rules, leaving only the ones under `rervive.rules`.

In PR 2803, we configured revive's dot-import rule, so from then, revive only checks this single rule.

This commit re-enable all the default revive rules, according to https://github.com/mgechev/revive/blob/62f0b012e5f927a733e4ee41c5e27091d8c57c71/revive.toml#L1

Disable any rule that currently introduces new linter warnings. They will be enabled one by one with their finding fixed, in the following commits.

The following rules are removed as they won't implemented:
* enforce-map-style - not agree with the rule logic
* exported - too much work to implement, not worth it
* package-comments - too much work to implement, not worth it

**Jira Ticket**:

```jira-ticket
None
```

**Release note**:

```release-note
None
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -11,11 +11,56 @@ linters:
       force-expect-to: true
     revive:
       rules:
+        - name: bare-return
+          disabled: true
+        - name: blank-imports
+        - name: context-as-argument
+        - name: context-keys-type
         - name: dot-imports
           arguments:
             - allowedPackages:
                 - github.com/onsi/ginkgo/v2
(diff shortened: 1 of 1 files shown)


nunnatsa:
> can you group the disabled rules together? it would be easier to understand which ones are already enforced and which ones are disabled

I see what you mean. But I want them to be sorted, mostly because my plan is to enable them one by one.

PR #3390, opened 2025-04-08
Add missing mutex lock for aaq.reset()

nunnatsa:

Follow up PR #3379: the cache in the aaqHooks is now protected by a mutex. But we didn't lock the mutex in the reset() method.

This PR adds the missing lock.

**Jira Ticket**:
```jira-ticket
None
```

**Release note**:
```release-note
None
```


orenc1 (approved):
/approve

PR #3474, opened 2025-04-10
Bump golang dependencies for k8s v0.32.3

tiraboschi:
**What this PR does / why we need it**:
Bump various golang dependencies:
- github.com/evanphx/json-patch/v5 v5.9.0 => v5.9.11
- github.com/machadovilaca/operator-observability v0.0.24 => v0.0.26
- github.com/onsi/ginkgo/v2 v2.20.2 => v2.23.4
- github.com/onsi/gomega v1.34.2 => v1.37.0
- github.com/machadovilaca/operator-observability v0.0.24 => v0.0.26
- github.com/onsi/ginkgo/v2 v2.20.2 => v2.23.4
- github.com/onsi/gomega v1.34.2 => v1.37.0
- github.com/openshift/library-go v0.0.0-20240830130947-d9523164b328 => v0.0.0-20250402180609-ce2ba53fb2a4
- github.com/operator-framework/api v0.27.0 => v0.30.0
- github.com/operator-framework/operator-lib v0.15.0 => v0.17.0
- github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring v0.76.0 => v0.81.0
- github.com/prometheus/client_golang v1.20.2 => v1.22.0
- github.com/prometheus/common v0.57.0 => v0.63.0
- github.com/samber/lo v1.47.0 => v1.49.1
- github.com/spf13/pflag v1.0.5 => v1.0.6
- golang.org/x/mod v0.20.0 => v0.24.0
- golang.org/x/sync v0.12.0 => v0.13.0
- golang.org/x/tools v0.24.0 => v0.32.0
- gomodules.xyz/jsonpatch/v2 v2.4.0 => v2.4.0
- k8s.io/* v0.31.0 => v0.32.3
- k8s.io/kube-openapi v0.0.0-20240827152857-f7e401e7b4c2 => v0.0.0-20250318190949-c8a335a9a2ff
- sigs.k8s.io/controller-runtime v0.19.3
- sigs.k8s.io/controller-tools v0.16.2

**Reviewer Checklist**
- No real code changes
- References to minTLSVersion configuration got fixed in Openshift APIs with https://github.com/openshift/api/pull/1679 and we are now consuming it. All the changes are in textual descriptions
that are now finally consistent with the actual APIs.
- TLS 1.3 ciphers were disables in the past on openshift due to golang limitations. They got finally enabled with https://github.com/openshift/library-go/pull/1850 and we are now consuming it due to the dependency bump. Let's align tests expectations.

[...]

PR #3476, opened 2025-04-10
Check descheduler according to new profile

tiraboschi:
On the descheduler we now have a KubeVirt
dedicated profile named KubeVirtRelieveAndMigrate.
Enforce its presence if the descheduler is in use.

**Reviewer Checklist**

- [x] PR Message
- [x] Commit Messages
- [ ] How to test
- [x] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [x] Upgrade Scenario
- [ ] Uninstallation Scenario
- [x] Backward Compatibility

**Jira Ticket**:

```jira-ticket
NONE
```

**Release note**:

```release-note
Check descheduler configuration according to its new KubeVirt dedicated profile
```


tiraboschi:
/hold
wait for https://github.com/kubevirt/hyperconverged-cluster-operator/pull/3474 first

pkg/util/cluster.go
```diff
@@ -353,11 +353,17 @@ func (c *ClusterInfoImp) RefreshDeschedulerCR(ctx context.Context, cl client.Cli
...
+		// the dev prefix
+		futureRelieveAndMigrateStable := deschedulerv1.DeschedulerProfile("KubeVirtRelieveAndMigrate")
+
+		if slices.Contains(instance.Spec.Profiles, deschedulerv1.RelieveAndMigrate) || slices.Contains(instance.Spec.Profiles, futureRelieveAndMigrateStable) {
+			misconfiguredDescheduler = false
 		} else {
-			misconfiguredDescheduler = !instance.Spec.ProfileCustomizations.DevEnableEvictionsInBackground
+			misconfiguredDescheduler = true
 		}
+
```
nunnatsa, line 366: I think it should be simplify to:
```go
misconfiguredDescheduler = !slices.Contains(instance.Spec.Profiles, deschedulerv1.RelieveAndMigrate) &&
                           !slices.Contains(instance.Spec.Profiles, futureRelieveAndMigrateStable)
```

pkg/util/cluster.go
```diff
@@ -353,11 +353,17 @@ func (c *ClusterInfoImp) RefreshDeschedulerCR(ctx context.Context, cl client.Cli
 			}
 			return err
 		}
-		if instance.Spec.ProfileCustomizations == nil {
-			misconfiguredDescheduler = true
+
+		// TODO: remove this once deschedulerv1.RelieveAndMigrate will graduate loosing
+		// the dev prefix
```
nunnatsa, line 358: The TODO is not clear to me. When it says "remove ***this***", what **exactly** "this" refers to?

nunnatsa (commented):
Thanks for the PR @tiraboschi !!

Added a few inline comments, but the more meaningful comment is about the general flow of setting the `kubevirt_hco_misconfigured_descheduler` metric.

I can't understand why the logic is split from the controller into clusterInfo. The only user of this logic is the Descheduler controller. 

Also, adding more and more functions to the ClusterInfo interface is a bug practice. This interface is too large anyway, making it hard to implement it (e.g. for injection to tests), and to maintain it.

It would be great if we could move the code to the controller, and simplify it, even in a follow up PR.

nunnatsa (commented):
Thanks for the PR @tiraboschi !!

Added a few inline comments, but the more meaningful comment is about the general flow of setting the `kubevirt_hco_misconfigured_descheduler` metric.

I can't understand why the logic is split from the controller into clusterInfo. The only user of this logic is the Descheduler controller. 

Also, adding more and more functions to the ClusterInfo interface is a bug practice. This interface is too large anyway, making it hard to implement it (e.g. for injection to tests), and to maintain it.

It would be great if we could move the code to the controller, and simplify it, even in a follow up PR.

pkg/util/cluster.go
```diff
@@ -353,11 +353,17 @@ func (c *ClusterInfoImp) RefreshDeschedulerCR(ctx context.Context, cl client.Cli
...
+		futureRelieveAndMigrateStable := deschedulerv1.DeschedulerProfile("KubeVirtRelieveAndMigrate")
```
nunnatsa, line 359: Shouldn't it be 
```suggestion
		futureRelieveAndMigrateStable := deschedulerv1.DeschedulerProfile("DevKubeVirtRelieveAndMigrate")
```
?

This is the value in https://github.com/kubevirt/hyperconverged-cluster-operator/blob/703fa5acb0cdb2d280cf1a6a911e64252f49dbbc/vendor/github.com/openshift/cluster-kube-descheduler-operator/pkg/apis/descheduler/v1/types_descheduler.go#L230

nunnatsa: OK, so it's appears that `deschedulerv1.RelieveAndMigrate` is `"DevKubeVirtRelieveAndMigrate"`.

resolving

nunnatsa (approved):
/approve

PR #3500, opened 2025-04-28
Add retry for multi-arch image pushes

nunnatsa:
Some automations somatimes failed with temporary server error (502), when pushing images.

This PR adds a retry mechanism when pushing the images, to make these automations more stable.
**Release note**:

```release-note
None
```


orenc1 (approved):
/approve

PR #3501, opened 2025-04-28
Retry push images on the retag script

nunnatsa:
Follow up of #3500 

The action may still fail for server error 502, when re-tagging the images

This PR add a retry mechanism also to the `hack/retag-multi-arch-images.sh` script, to fix the issue.

Also, this PR adds a common list of supported architectures, to be used by the scripts, instead of duplicate the list in each script.

**Release note**:

```release-note
None
```

### DIFF
--- a/hack/architecture.sh
+++ b/hack/architecture.sh
@@ -1,0 +1,1 @@
+ARCHITECTURES="amd64 arm64 s390x"
--- a/hack/retag-multi-arch-images.sh
+++ b/hack/retag-multi-arch-images.sh
@@ -25,11 +25,11 @@ if [[ "${MULTIARCH}" == "true" ]]; then
   for arch in ${ARCHITECTURES}; do
     NEW_IMAGE="${NEW_IMAGE_REPO}:${NEW_TAG}-${arch}"
     ${CRI_BIN} tag "${IMAGE_REPO}:${CURRENT_TAG}-${arch}" "${NEW_IMAGE}"
-    ${CRI_BIN} push "${NEW_IMAGE}"
+    ./hack/retry.sh 3 10 "${CRI_BIN} push ${NEW_IMAGE}"
   done
 fi
 
 # retag the manifest
 NEW_IMAGE="${NEW_IMAGE_REPO}:${NEW_TAG}"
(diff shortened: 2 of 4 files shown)


hack/architecture.sh
```diff
@@ -0,0 +1 @@
+ARCHITECTURES="amd64 arm64 s390x"
```
avlitman, line 1: add new line at the end?

hack/retag-multi-arch-images.sh
```diff
@@ -19,15 +19,17 @@ if [[ -z ${NEW_TAG} ]]; then
...
 
+. ./hack/cri-bin.sh && export CRI_BIN=${CRI_BIN}
+
 if [[ "${MULTIARCH}" == "true" ]]; then
   for arch in ${ARCHITECTURES}; do
     NEW_IMAGE="${NEW_IMAGE_REPO}:${NEW_TAG}-${arch}"
-    . "hack/cri-bin.sh" && ${CRI_BIN} tag "${IMAGE_REPO}:${CURRENT_TAG}-${arch}" "${NEW_IMAGE}"
-    . "hack/cri-bin.sh" && ${CRI_BIN} push "${NEW_IMAGE}"
+    ${CRI_BIN} tag "${IMAGE_REPO}:${CURRENT_TAG}-${arch}" "${NEW_IMAGE}"
+    ./hack/retry.sh 3 10 "${CRI_BIN} push ${NEW_IMAGE}"
```
avlitman, line 28: what does `3 10` means?

hack/architecture.sh
nunnatsa: Because?

hack/retag-multi-arch-images.sh
nunnatsa: It's in the ./hack/retry.sh script. The first parameter is the number of retries. The 2nd is the number of seconds to wait for the next attempt, in case of failure. The 3rd parameter is the command to run.

https://github.com/kubevirt/hyperconverged-cluster-operator/blob/366678ebec7bf6dbc207db72293f70a37fc744bd/hack/retry.sh#L8-L11

PR #3522, opened 2025-05-21
Use latest nightly CNAO for HCO nightly build

nunnatsa:
To find integration issues early, use latest nightly CNAO for HCO nightly test and build, in addition to KubeVirt and CDI that are already part of the nightly build.

**Jira Ticket**:
```jira-ticket
https://issues.redhat.com/browse/CNV-62158
```

**Release note**:
```release-note
None
```

### DIFF
--- a/automation/nightly/test-nightly-build.sh
+++ b/automation/nightly/test-nightly-build.sh
@@ -37,24 +37,44 @@ latest_cdi_image=$(curl -sL "https://storage.googleapis.com/kubevirt-prow/devel/
 IFS=: read -r cdi_image cdi_tag <<< "${latest_cdi_image}"
 latest_cdi_commit=$(curl -sL "https://storage.googleapis.com/kubevirt-prow/devel/nightly/release/kubevirt/containerized-data-importer/${latest_cdi}/commit")
 
+latest_cnao=$(curl -sL https://storage.googleapis.com/kubevirt-prow/devel/nightly/release/kubevirt/cluster-network-addons-operator/latest)
+latest_cnao_image=$(curl -sL "https://storage.googleapis.com/kubevirt-prow/devel/nightly/release/kubevirt/cluster-network-addons-operator/${latest_cnao}/operator.yaml" | grep "image:" | grep "cluster-network-addons-operator" | sed -E "s|^ +image: (.*)$|\1|")
+IFS=: read -r cnao_image cnao_tag <<< "${latest_cnao_image}"
+latest_cnao_commit=$(curl -sL "https://storage.googleapis.com/kubevirt-prow/devel/nightly/release/kubevirt/cluster-network-addons-operator/${latest_cnao}/commit")
+
 # Update HCO dependencies
 go mod tidy
 go mod vendor
-rm -rf kubevirt cdi
+rm -rf kubevirt cdi cnao
 
 # Get latest kubevirt
 git clone https://github.com/kubevirt/kubevirt.git
-(cd kubevirt; git checkout "${latest_kubevirt_commit}")
+(
+  cd kubevirt
+  git checkout "${latest_kubevirt_commit}"
+ )
 
 # Get latest CDI
 git clone https://github.com/kubevirt/containerized-data-importer.git cdi
-(cd cdi; git checkout "${latest_cdi_commit}")
+(
+  cd cdi
+  git checkout "${latest_cdi_commit}"
+)
+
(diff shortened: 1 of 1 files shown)


automation/nightly/test-nightly-build.sh
```diff
@@ -37,25 +37,45 @@ latest_cdi_image=$(curl -sL "https://storage.googleapis.com/kubevirt-prow/devel/
...
+  git checkout "${latest_kubevirt_commit}"
+ )
 
 # Get latest CDI
 git clone https://github.com/kubevirt/containerized-data-importer.git cdi
-(cd cdi; git checkout "${latest_cdi_commit}")
+(
+  cd cdi
+  git checkout "${latest_cdi_commit}"
+)
```
RamLavi, line 62: nit: this `)` is indented differently than the others added

automation/nightly/test-nightly-build.sh
```diff
@@ -67,9 +87,11 @@ make build-push-multi-arch-operator-image build-push-multi-arch-webhook-image bu
 
 # Update image digests
 sed -i "s#quay.io/kubevirt/virt-#${kv_image/-*/-}#" deploy/images.csv
+sed -i "s#quay.io/kubevirt/cluster-network-addons-operator#${cnao_image}#" deploy/images.csv
+
 sed -i "s#^KUBEVIRT_VERSION=.*#KUBEVIRT_VERSION=\"${kv_tag}\"#" hack/config
 sed -i "s#^CDI_VERSION=.*#CDI_VERSION=\"${cdi_tag}\"#" hack/config
-(cd ./tools/digester && go build .)
```
RamLavi, line 72: can you explain what this line does and why is this removed?

RamLavi (commented):
Looks good, added nit and a question

automation/nightly/test-nightly-build.sh
nunnatsa: yes: it compiles a tool that finds the image digests from image tags. but the it already compiled within the `automation/digester/update_images.sh` script, so the line is actually redundant.

automation/nightly/test-nightly-build.sh
nunnatsa: fixed.

automation/nightly/test-nightly-build.sh
```diff
@@ -37,25 +37,45 @@ latest_cdi_image=$(curl -sL "https://storage.googleapis.com/kubevirt-prow/devel/
 IFS=: read -r cdi_image cdi_tag <<< "${latest_cdi_image}"
 latest_cdi_commit=$(curl -sL "https://storage.googleapis.com/kubevirt-prow/devel/nightly/release/kubevirt/containerized-data-importer/${latest_cdi}/commit")
 
+latest_cnao=$(curl -sL https://storage.googleapis.com/kubevirt-prow/devel/nightly/release/kubevirt/cluster-network-addons-operator/latest)
+latest_cnao_image=$(curl -sL "https://storage.googleapis.com/kubevirt-prow/devel/nightly/release/kubevirt/cluster-network-addons-operator/${latest_cnao}/operator.yaml" | grep "image:" | grep "cluster-network-addons-operator" | sed -E "s|^ +image: (.*)$|\1|")
```
orenc1, line 41: nit:
i think it would be more "clean" to extract the CNAO image by using the yaml structure, instead of trying to catch it with grep and sed:
```suggestion
latest_cnao_image=$(curl -sL "https://storage.googleapis.com/kubevirt-prow/devel/nightly/release/kubevirt/cluster-network-addons-operator/${latest_cnao}/operator.yaml" | yq -r 'select(.kind=="Deployment").spec.template.spec.containers[0].image')
```
however, for this we need `yq`, which shouldn't be difficult to get from https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64, IMHO

nunnatsa: This is a very good suggestion, and it much simpler. However, yq is not installed by default in the automation and I don't want to maintain its installation just for this line.

orenc1: ok.
please note we can also use it for lines 30 and 36 above, that's 3 usages

orenc1 (approved):
/approve

automation/nightly/test-nightly-build.sh
RamLavi: you can use the dockerized: https://hub.docker.com/r/linuxserver/yq
note I didn't try this specific image, but it looks promising

PR #3533, opened 2025-05-26
Remove SA cert usage from observability func-tests

machadovilaca:

Depending on the system from which the func-tests are executed , the service account TLS certificate might be available (executed from an external platform).

This PR replaces the SA certificate with `HTTPClientFor` call with `tests.GetClientConfig`

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
none
```

**Release note**:

```release-note
Remove SA cert usage from observability func-tests
```


tests/func-tests/observability_controller_test.go
```diff
@@ -55,11 +86,53 @@ var _ = Describe("Observability Controller", Label(tests.OpenshiftLabel, testNam
...
+		req.Header.Add("Authorization", "Bearer "+bearerToken)
+	}
+
+	resp, err := httpClient.Do(req)
+	if err != nil {
+		return false
+	}
+	defer resp.Body.Close()
+
+	return resp.StatusCode < 500
```
orenc1, line 114: if the request returns HTTP status code of 4xx, why it is considered that the alert manager is accessible?
probably the endpoint is not available, or you're not authorized or other reasons.
I think we want the response to be at 2xx, no?

machadovilaca: @orenc1 updated

PR #3554, opened 2025-06-08
Fix issues with ingress-controller

orenc1:
This PR fixes two bugs at the `ingress-cluster-controller` of HCO:
1. If the reconciliation completed before `HyperConverged` CR is created, and then the CR is created, the ingress-controller reconciliation is not triggered and thus the `virt-downloads` ComponentRoute is not added to the `ingresses.config.openshift.io` CR. This causes a test failure of `should allow download URL customization`. The fix is to add a watch of this controller for the HyperConverged CR, on creation and deletion.
2. When the HyperConverged CR is deleted, the virt-downloads ComponentRoute is deleted from the `ingresses.config.openshift.io` status, and then the ingress-controller tries to update conditions on non-existent entry, resulting in panic in the reconciliation loop. The fix is to not set conditions if the CR was deleted.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-62752
```

**Release note**:

```release-note
NONE
```

### DIFF
--- a/controllers/ingresscluster/ingress-cluster-controller.go
+++ b/controllers/ingresscluster/ingress-cluster-controller.go
@@ -68,14 +68,35 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
-	return c.Watch(source.Kind(mgr.GetCache(), client.Object(&configv1.Ingress{
+	if err = c.Watch(source.Kind(mgr.GetCache(), client.Object(&configv1.Ingress{
 		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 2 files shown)


nunnatsa (approved):
/approve

PR #3556, opened 2025-06-08
Fix non-stand-alone unit tests

nunnatsa:

Unit tests must be stand alone, pass when run as a single test and must not be depend on other previous tests.

This is not the case in the `controllers/hyperconverged`, `controllers/operands` and the `controllers/common` packages, where several tests are failing when running individually (e.g. with `FIt`), or when running tests with random order.

This PR moves all the required settings to the top level of the test file, or makes other changes, to make all the test independent.

**Release note**:
```release-note
None
```

### DIFF
--- a/controllers/operands/ssp_test.go
+++ b/controllers/operands/ssp_test.go
@@ -36,6 +36,15 @@ var _ = Describe("SSP Operands", func() {
 	var (
 		testFilesLocation = getTestFilesLocation() + "/dataImportCronTemplates"
 	)
+
+	origDICTMap := dataImportCronTemplateHardCodedMap
+	BeforeEach(func() {
+		dataImportCronTemplateHardCodedMap = nil
+		DeferCleanup(func() {
+			dataImportCronTemplateHardCodedMap = origDICTMap
+		})
+	})
+
 	Context("SSP", func() {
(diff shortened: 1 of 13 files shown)


controllers/operands/ssp_test.go
```diff
@@ -748,40 +676,36 @@ var _ = Describe("SSP Operands", func() {
...
 
-					disabledUserImage := image1.DeepCopy()
-					disableDict(disabledUserImage)
-					enabledUserImage := image2.DeepCopy()
-					enableDict(enabledUserImage)
+					disableDict(&image1)
+					enableDict(&image2)
 
-					hco.Spec.DataImportCronTemplates = []hcov1beta1.DataImportCronTemplate{*disabledUserImage, *enabledUserImage}
+					hco.Spec.DataImportCronTemplates = []hcov1beta1.DataImportCronTemplate{image1, image2}
```
orenc1, line 702: didn't `disabledUserImage`, `enabledUserImage` and `image3Modified` variable names had a special meaning before?

nunnatsa: It's just that now, when the  image1-4 variables are re-created at `BeforeEach`, the custom images (created by DeepCopy and then modified), are not needed anymore.

PR #3562, opened 2025-06-10
Add the missing `ports` nob to the operator and the webhook pods

nunnatsa:

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-63421
```

**Release note**:

```release-note
Add the missing `ports` nob to the operator and the webhook pods
```

### DIFF
--- a/controllers/operands/cliDownload.go
+++ b/controllers/operands/cliDownload.go
@@ -139,7 +139,7 @@ func NewCliDownloadsService(hc *hcov1beta1.HyperConverged) *corev1.Service {
 			},
 			Ports: []corev1.ServicePort{
 				{
-					Name:       strconv.Itoa(util.CliDownloadsServerPort),
+					Name:       strconv.Itoa(int(util.CliDownloadsServerPort)),
 					Port:       util.CliDownloadsServerPort,
 					Protocol:   corev1.ProtocolTCP,
 					TargetPort: intstr.FromInt32(util.CliDownloadsServerPort),
--- a/pkg/components/components.go
+++ b/pkg/components/components.go
@@ -190,6 +190,9 @@ func GetDeploymentSpecOperator(params *DeploymentOperatorParams) appsv1.Deployme
 						},
 						SecurityContext:          GetStdContainerSecurityContext(),
 						TerminationMessagePolicy: corev1.TerminationMessageFallbackToLogsOnError,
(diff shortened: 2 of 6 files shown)


controllers/operands/cliDownload.go
```diff
@@ -139,7 +139,7 @@ func NewCliDownloadsService(hc *hcov1beta1.HyperConverged) *corev1.Service {
 			},
 			Ports: []corev1.ServicePort{
 				{
-					Name:       strconv.Itoa(util.CliDownloadsServerPort),
+					Name:       strconv.Itoa(int(util.CliDownloadsServerPort)),
```
avlitman, line 142: why we need the int() also here?

nunnatsa: The PR changes `util.CliDownloadsServerPort` to be `int32` instead if `int`, because this is the port field type. But here we need casting.

pkg/components/components.go
```diff
@@ -334,12 +337,12 @@ func GetDeploymentSpecCliDownloads(params *DeploymentOperatorParams) appsv1.Depl
...
 								Protocol:      corev1.ProtocolTCP,
-								ContainerPort: int32(8080),
+								ContainerPort: util.CliDownloadsServerPort,
 							},
 						},
 						SecurityContext:          GetStdContainerSecurityContext(),
-						ReadinessProbe:           getReadinessProbe("/health", 8080),
-						LivenessProbe:            getLivenessProbe("/health", 8080),
+						ReadinessProbe:           getReadinessProbe("/health", util.CliDownloadsServerPort),
+						LivenessProbe:            getLivenessProbe("/health", util.CliDownloadsServerPort),
```
orenc1, line 345: are you sure `int32` here is ok? previously it was 8080 without any casting, i.e. `int`

nunnatsa: I think it's safer. the k8s port field type is `int32`, and also, this is the type of the port parameter of the `getReadinessProbe` function:
```go
func getLivenessProbe(endpoint string, port int32) *corev1.Probe {
	return &corev1.Probe{
		ProbeHandler: corev1.ProbeHandler{
		...
```

nunnatsa: I think go knows how to use the right type when using literals.

PR #3566, opened 2025-06-11
Split the CRD overlap validation from csv-merger to a new tool

nunnatsa:

The CRD overlap validation is used only in build-manifest, and not in any other place. It should not be part of the csv-merger, and not part of the operator image.

This PR take this functionality out of the csv-merger, and introduces the new crd-overlap-validator tool, to be used from the build-manifests script.

**Release note**:
```release-note
None
```


orenc1 (commented):
/lgtm
/approve
/thanks

PR #3568, opened 2025-06-15
Support NetworkPolicies in the bundle image

nunnatsa:

* Add the `manifest-splitter` tool, to allow multiple types of K8s manifests in other operators csv-gen
* Use the new `manifest-splitter` tool in the build-manifests script, so the other operator would able to generate NetworkPolicy manifests, to get into the bundle image.
* Add the new `network-policy-gen` tool to generate a set of network policies for HCO and other components.
**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-63363
https://issues.redhat.com/browse/CNV-63366
https://issues.redhat.com/browse/CNV-63382
```

**Release note**:

```release-note
Allow adding network policies to the bundle image
```

### DIFF
--- a/controllers/alerts/metrics_test.go
+++ b/controllers/alerts/metrics_test.go
@@ -14,7 +14,9 @@ import (
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/utils/ptr"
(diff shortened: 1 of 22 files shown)


controllers/alerts/metrics_test.go
```diff
@@ -108,6 +110,56 @@ var _ = Describe("alert tests", func() {
...
+			pod := ci.GetPod()
+			if pod.Labels == nil {
+				pod.Labels = make(map[string]string)
+			}
+			pod.Labels[hcoutil.AllowEgressToDNSAndAPIServerLabel] = "true"
+			pod.Labels[hcoutil.AllowIngressToMetricsEndpointLabel] = "true"
+
+			DeferCleanup(func() {
+				delete(pod.Labels, hcoutil.AllowEgressToDNSAndAPIServerLabel)
+				delete(pod.Labels, hcoutil.AllowIngressToMetricsEndpointLabel)
```
machadovilaca, line 133: i may be missing something, the pod is a local object, we don't patch the resource after the test, why are we deleting the those labels, and where is that applied?

controllers/alerts/network_policy.go
```diff
@@ -0,0 +1,118 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	hcoutil "github.com/kubevirt/hyperconverged-cluster-operator/pkg/util"
+)
+
+const (
+	policyName = "hco-allow-egress-to-alert-manager"
+
+	alertmanagerPort = 9095 // the HTTP call goes to port 9094, but for some internal implementation of the AlertManager,
+	// the egress rule must allow port 9095.
```
machadovilaca: nit, personal preference, ignore if you want

```suggestion
        // the HTTP call goes to port 9094, but for some internal implementation of the AlertManager,
        // the egress rule must allow port 9095.
	alertmanagerPort = 9095
```

controllers/alerts/metrics_test.go
nunnatsa: it's not actually local. It a pointer to the real pod object in the test utils.

controllers/alerts/metrics_test.go
```diff
@@ -108,6 +110,56 @@ var _ = Describe("alert tests", func() {
 			Expect(ee.CheckEvents(expectedEvents)).To(BeTrue())
 		})
 
+		It("should not create network policy if the pod is without the np labels", func() {
```
avlitman, line 113: np is a known term?

PR #3569, opened 2025-06-16
Add two new labels for NetworkPolicy

nunnatsa:

Optionally add two new labels to the operator and the webhook pods, in the CSV.

The `hco.kubevirt.io/allow-access-cluster-services` label is used to allow egress to the DNS and the API-Server.

The `hco.kubevirt.io/allow-prometheus-access` label is used to allow ingress to the metrics port of the pod.

**Jira Ticket**:
```jira-ticket
https://issues.redhat.com/browse/CNV-63367
```

**Release note**:
```release-note
CSV: Add two new labels for NetworkPolicy
```

### DIFF
--- a/pkg/components/components.go
+++ b/pkg/components/components.go
@@ -51,6 +51,10 @@ const (
 
 	kubevirtProjectName = "KubeVirt project"
 	rbacVersionV1       = "rbac.authorization.k8s.io/v1"
+
+	// labels to apply network policies:
+	allowEgressToDNSAndAPIServerLabel  = "hco.kubevirt.io/allow-access-cluster-services"
+	allowIngressToMetricsEndpointLabel = "hco.kubevirt.io/allow-prometheus-access"
 )
 
 var deploymentType = metav1.TypeMeta{
@@ -83,6 +87,7 @@ type DeploymentOperatorParams struct {
 	MtqVersion             string
 	AaqVersion             string
 	Env                    []corev1.EnvVar
+	AddNetworkPolicyLabels bool
 }
 
 func GetDeploymentOperator(params *DeploymentOperatorParams) appsv1.Deployment {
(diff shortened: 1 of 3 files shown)


pkg/components/components.go
```diff
@@ -51,6 +51,10 @@ const (
 
 	kubevirtProjectName = "KubeVirt project"
 	rbacVersionV1       = "rbac.authorization.k8s.io/v1"
+
+	// labels to apply network policies:
+	allowEgressToDNSAndAPIServerLabel  = "hco.kubevirt.io/allow-access-cluster-services"
```
orenc1, line 56: maybe the label should be renamed like:
```suggestion
	allowEgressToDNSAndAPIServerLabel  = "hco.kubevirt.io/allow-cluster-services-access"
```
so it will be with the same format as the prometheus one?

tools/csv-merger/csv-merger.go
```diff
@@ -123,8 +123,9 @@ var (
 		"Comma separated list of CSVs that can be skipped (read replaced) by this version")
 	olmSkipRange = flag.String("olm-skip-range", "",
 		"Semver range expression for CSVs that can be skipped (read replaced) by this version")
-	mgImage        = flag.String("mg-image", "quay.io/kubevirt/must-gather", "Operator suggested must-gather image")
-	testImagesNVRs = flag.String("test-images-nvrs", "", "Test Images NVRs")
+	mgImage             = flag.String("mg-image", "quay.io/kubevirt/must-gather", "Operator suggested must-gather image")
+	testImagesNVRs      = flag.String("test-images-nvrs", "", "Test Images NVRs")
+	dumpNetworkPolicies = flag.Bool("dump-network-policies", false, "Dump network policy yamls to stdout")
```
orenc1, line 128:
```suggestion
	dumpNetworkPolicies = flag.Bool("dump-network-policies", false, "Dump network policies yamls to stdout")
```
if we're creating several network policies

pkg/components/components.go
nunnatsa: Too late for that. This is the agreed name between all the teams. We'll leave it as it is now.

tools/csv-merger/csv-merger.go
nunnatsa: I think that the original format is the right one. We should only use plural for the last noun, which is `yamls` in this case.

orenc1 (approved):

PR #3570, opened 2025-06-17
Fix the publish-community-operators github action

nunnatsa:

Since we're now creating multi-arch images during the release, the publish-community-operators github action fails when trying to build the images for the next patch release, because the host is out of space.

This PR split the action into two jobs, one to release the community version, and the second, that is depend on the first one, opens a new PR to prepare the next patch version.
**Release note**:
```release-note
None
```


orenc1 (approved):
/approve

PR #3594, opened 2025-06-23
Monitor the cluster architectures

nunnatsa:
### API changes: 
1. add the new nodeInfo field to the HyperConverged's status
2. add conditions field to the DICT status type

### Logic changes:
1. monitor the cluster node's architectures
2. Set the new nodeInfo field in the HC status
3. webhook: validate the nodePlace Affinity field
4. Filter out unsupported architectures from DICT annotation
5. when there is not supported architecture for a DICT, add the `Deployed=False` condition to it status

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-61339
```

**Release note**:
```release-note
API changes: 
1. add the new nodeInfo field to the HyperConverged's status
2. add conditions field to the DICT status type

Logic changes:
1. monitor the cluster node's architectures
2. Set the new nodeInfo field in the HC status
3. webhook: validate the nodePlace Affinity field
4. Filter out unsupported architectures from DICT annotation
5. when there is not supported architecture for a DICT, add the `Deployed=False` condition to it status
```


avlitman:
Did you take into consideration restarts? 

/lgtm

nunnatsa:
> Did you take into consideration restarts?
> 
> /lgtm

yes. see cmd/hyperconverged-cluster-operator/main.go, line 130

nunnatsa:
Changes: rebase

nunnatsa:
Closing and re opening to try fix ci.

nunnatsa:
/hold

After some manual tests, the following need to be fixed:
1. [BUG] When modifying the `"ssp.kubevirt.io/dict.architectures"` annotation in a pre-defined (common) DICT, in the HyperConverged spec, the change is ignored and the modified annotation is not copied to the SSP CR.
2. The behavior of `"ssp.kubevirt.io/dict.architectures"` annotation in the HyperConverged status DICTs is inconsistent:
     if the FG is not enabled, the original value of annotation is used. If the FG is enabled and the annotation contains cluster-supported architecture(s), the final value is set, and if no architecture is match, the value is again the original one. This is confusing. The behavior should be to use the same value as set in the SSP DICT.
3. To still have the original annotation value (e.g. for debug), we'll add new field to the DICT status in HyperConverged status, to reflect it.

controllers/operands/ssp.go
```diff
@@ -147,18 +156,18 @@ func NewSSP(hc *hcov1beta1.HyperConverged, opts ...string) (*sspv1beta3.SSP, []h
...
 	}
 
-	var dataImportCronTemplates []hcov1beta1.DataImportCronTemplate
-	for _, dictStatus := range dataImportCronStatuses {
-		dataImportCronTemplates = append(dataImportCronTemplates, dictStatus.DataImportCronTemplate)
-	}
+	//var dataImportCronTemplates []hcov1beta1.DataImportCronTemplate
+	//for _, dictStatus := range dataImportCronStatuses {
+	//	dataImportCronTemplates = append(dataImportCronTemplates, dictStatus.DataImportCronTemplate)
+	//}
```
nunnatsa: remove dead code

controllers/operands/ssp.go
```diff
@@ -347,7 +378,36 @@ func (d dataImportTemplateSlice) Len() int           { return len(d) }
...
+		}
+	}
+	hcoDictStatus.Annotations[MultiArchDICTAnnotation] = sspArchsAnnotation
+	hcoDictStatus.Status.OriginalSupportedArchitectures = hcoArchsAnnotation
+}
+
+func hcoDictToSSP(hcoDictStatus hcov1beta1.DataImportCronTemplateStatus, multiArchEnabled bool) (sspv1beta3.DataImportCronTemplate, bool) {
+	hcoDict := hcoDictStatus.DataImportCronTemplate
+	var sspArchsAnnotation string
+	if multiArchEnabled && meta.IsStatusConditionFalse(hcoDictStatus.Status.Conditions, "Deployed") {
```
nunnatsa: use the `dictConditionDeployedType` const

api/v1beta1/hyperconverged_types.go
```diff
@@ -798,6 +808,14 @@ type DataImportCronTemplateStatus struct {
 	Status DataImportCronStatus `json:"status,omitempty"`
 }
 
+// NodeInfoStatus holds information about the cluster nodes
+type NodeInfoStatus struct {
+	// WorkloadsArchitectures is a distinct list of the CPU architectures of the workloads nodes in the cluster.
+	WorkloadsArchitectures []string `json:"workloadsArchitectures,omitempty"`
+	// ControllerNodeArchitecture is a distinct list of the CPU architecture of the control-plane nodes.
+	ControllerNodeArchitecture []string `json:"controllerNodeArchitecture,omitempty"`
```
orenc1: "controller node" is not a well-defined term. maybe use `ControlPlaneNodesArchitecture` instead?

api/v1beta1/hyperconverged_types.go
```diff
@@ -798,6 +808,14 @@ type DataImportCronTemplateStatus struct {
 	Status DataImportCronStatus `json:"status,omitempty"`
 }
 
+// NodeInfoStatus holds information about the cluster nodes
+type NodeInfoStatus struct {
+	// WorkloadsArchitectures is a distinct list of the CPU architectures of the workloads nodes in the cluster.
+	WorkloadsArchitectures []string `json:"workloadsArchitectures,omitempty"`
```
orenc1, line 814: if you're referring to the cluster nodes' roles, i think it would be better to use the term `workerNodesArchitecture`

api/v1beta1/hyperconverged_types.go
nunnatsa: good catch. Had to be `ControlPlaneNodesArchitecture`. Silly typo. Thanks!

api/v1beta1/hyperconverged_types.go
nunnatsa: This is intentional. HCO does not necessary checks the worker node. If the `spec.workloads.nodePlacement` field is populated, then HCO checks the relevant nodes instead of the worker nodes.

api/v1beta1/hyperconverged_types.go
nunnatsa: Fixed

controllers/nodes/nodes_controller.go
```diff
@@ -52,84 +62,81 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+	nodeInfoChanged, err := nodeinfo.HandleNodeChanges(ctx, r, hc, logger)
 	if err != nil {
-		if errors.IsNotFound(err) {
-			return reconcile.Result{}, nil
-		}
 		return reconcile.Result{}, err
 	}
 
-	if !hco.DeletionTimestamp.IsZero() {
+	if hc == nil || !hc.DeletionTimestamp.IsZero() {
```
orenc1, line 115: if `hc` is nil at that point, i think `nodeinfo.HandleNodeChanges(ctx, r, hc, logger)`  would return an error.

nunnatsa: hc == nil is handled. Look at the `isWorkloadNodeFunc` function in pkg/internal/nodeinfo/nodeinfo.go. This function choose the function to filter workloads nodes. so if hc is nil, it will return a function that only uses the node's worker label. And if it's not nil, it checks if hc's `spec.workloads.nodePlacement` field is populated.

nunnatsa: In other words, this code is running before the hc cr is created, and after it is deleted.

controllers/nodes/predicates.go
```diff
@@ -0,0 +1,68 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+
+	"github.com/kubevirt/hyperconverged-cluster-operator/api/v1beta1"
+)
+
+// Custom predicate to detect changes in node count
+type nodeCountChangePredicate predicate.TypedFuncs[*corev1.Node]
+
+func (nodeCountChangePredicate) Update(e event.TypedUpdateEvent[*corev1.Node]) bool {
+	return !maps.Equal(e.ObjectOld.GetLabels(), e.ObjectNew.GetLabels())
```
orenc1, line 18: why reconcile if the labels of a node are changed?
we rely on the node's `.status.nodeInfo` to get its arch, not labels.
what change in labels can affect the outcome of the nodes controller reconciliation?

controllers/nodes/predicates.go
```diff
@@ -0,0 +1,68 @@
...
+
+	if !reflect.DeepEqual(e.ObjectNew.Spec.Workloads, e.ObjectOld.Spec.Workloads) {
+		// If the Workloads spec not changed, we want to reconcile
+		return true
+	}
+
+	if !reflect.DeepEqual(e.ObjectNew.Spec.Infra, e.ObjectOld.Spec.Infra) {
+		// If the Infra spec not changed, we want to reconcile
+		return true
+	}
```
orenc1, line 57: this logic is not so clear to me.
it means that there will be a reconciliation only in case where the HCO CR change was not about node placement. why is that?

pkg/internal/nodeinfo/nodeinfo.go
```diff
@@ -27,3 +31,71 @@ func getNodes(ctx context.Context, cl client.Client) ([]corev1.Node, error) {
...
+		return func(node corev1.Node) bool {
+			matches, err := workloadMatcher.Match(&node)
+			if err != nil { // should not happen, because the validation webhook checks it, but just in case
+				return false
+			}
+			return matches
+		}
+	}
+
+	return isWorkerNode
```
orenc1, line 100:
```suggestion
	return isWorkloadNode
```

pkg/upgradepatch/hc.cr.json
```diff
@@ -53,5 +53,7 @@
     "deployVmConsoleProxy": false,
     "enableApplicationAwareQuota": false
   },
-  "status": {}
+  "status": {
+    "nodeInfo": {}
+  }
```
orenc1, line 58: why we need to specifically mention this stanza in the status? auto-generated?

pkg/webhooks/validator/validator.go
```diff
@@ -481,6 +506,32 @@ func (wh *WebhookHandler) validateDeprecatedFeatureGates(hc *v1beta1.HyperConver
...
+		if err := validateAffinity(hc.Spec.Infra.NodePlacement.Affinity); err != nil {
+			return fmt.Errorf("invalid infra node placement affinity: %v", err.Error())
+		}
+	}
+
+	return nil
+}
+
+func validateAffinity(affinity *corev1.Affinity) error {
+	if affinity == nil || affinity.NodeAffinity == nil || affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution == nil {
```
orenc1, line 526: users can use the other affinity rule - `preferredDuringSchedulingIgnoredDuringExecution`, why we won't accept it?

controllers/nodes/predicates.go
nunnatsa: Few things:

First, this is also for high availability, not only to collect architectures.

Then both in HA and arch, we are collecting data according to the node labels, either by the standard k8s worker/control-plane/master labels, or any other label that is part of the node placement configurations in the HyperConverged CR.

And BTW, this specific code wasn't changed, but only moved to a new file for better code organization.

controllers/nodes/predicates.go
nunnatsa: Yes, the node placement is the only part in the HyperConverged CR that is relevant for collecting node architectures by their function (workloads/infra).

We can actually drop the infra part as we don't look at this field. let's do it in a follow up PR.

pkg/internal/nodeinfo/nodeinfo.go
nunnatsa: This line is correct. The `isWorkloadNodeFunc` function returns a function that checks if a node is a workloads node. If the HyperConverged field is empty, we select the worker nodes as the workloads nodes. So in this case we return the `isWorkerNode` function that is also used for the HA use-case, so it defined elsewhere.

pkg/upgradepatch/hc.cr.json
nunnatsa: auto generated, yes.

pkg/webhooks/validator/validator.go
nunnatsa: There was a reason for that. Need to refresh my memory. Let's review it later.

PR #3596, opened 2025-06-24
Enable SecureExecution feature gate by default

jschintag:

Enable the `SecureExecution` feature gate in kubevirt. This allows Clusters to use Secure Execution VMs on IBM Z

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket

```

**Release note**:

```release-note
Enable SecureExecution Feature for s390x
```

### DIFF
--- a/controllers/hyperconverged/hyperconverged_controller_test.go
+++ b/controllers/hyperconverged/hyperconverged_controller_test.go
@@ -220,6 +220,7 @@ var _ = Describe("HyperconvergedController", func() {
 					"KubevirtSeccompProfile",
 					"VMPersistentState",
 					"InstancetypeReferencePolicy",
+					"SecureExecution",
 				}
 				// Get the KV
 				kvList := &kubevirtcorev1.KubeVirtList{}
--- a/controllers/operands/kubevirt.go
+++ b/controllers/operands/kubevirt.go
@@ -95,6 +95,9 @@ const (
 
 	// enables the instance type reference policy feature
 	kvInstancetypeReferencePolicy = "InstancetypeReferencePolicy"
+
+	// Allow running IBM Z Secure Execution VMs
+	kvSecureExecution = "SecureExecution"
 )
 
 const (
@@ -113,6 +116,7 @@ var (
 		kvKubevirtSeccompProfile,
 		kvVMPersistentState,
 		kvInstancetypeReferencePolicy,
+		kvSecureExecution,
 	}
 
 	// holds a list of mandatory KubeVirt feature gates. Some of them are the hard coded feature gates and some of


nunnatsa:
Thanks for the PR @jschintag 

This FG is still alpha in KubeVirt 1.6.0. We can't enable alpha FGs by default.

/hold

jschintag:
> Thanks for the PR @jschintag
> 
> This FG is still alpha in KubeVirt 1.6.0. We can't enable alpha FGs by default.

If it is not possible by default, then this means i have to add it as a FeatureGate to HCO right?

nunnatsa:
> If it is not possible by default, then this means i have to add it as a FeatureGate to HCO right?

Again, too soon.

The FGs in HCO are named fields. it was maybe a mistake, but it is what it is. 

So adding an alpha FG to the HCO API is not something we can approve. For alpha FG, please use the json patch annotation: https://github.com/kubevirt/hyperconverged-cluster-operator/blob/main/docs/cluster-configuration.md#jsonpatch-annotations

jschintag:
> > If it is not possible by default, then this means i have to add it as a FeatureGate to HCO right?
> 
> Again, too soon.
> 
> The FGs in HCO are named fields. it was maybe a mistake, but it is what it is.
> 
> So adding an alpha FG to the HCO API is not something we can approve. For alpha FG, please use the json patch annotation: https://github.com/kubevirt/hyperconverged-cluster-operator/blob/main/docs/cluster-configuration.md#jsonpatch-annotations

Okay, thank you for the explanation

jschintag:
@nunnatsa I have rebased this PR on top of main, as Secure Execution is now in Beta Stage for 1.7.
PTAL

controllers/handlers/kubevirt.go
```diff
@@ -114,6 +117,7 @@ var (
 		kvHostDevicesGate,
 		kvVMExportGate,
 		kvKubevirtSeccompProfile,
+		kvSecureExecution,
 	}
```
nunnatsa, line 125: Is there any point to set this feature gate on clusters with no s390x nodes?

See below for alternative implementation.

controllers/handlers/kubevirt.go
nunnatsa, line 889: WDYT about this implementation, instead of hard code the FG?

```go
	if slices.Contains(nodeinfo.GetWorkloadsArchitectures(), "s390x") {
		fgs = append(fgs, kvSecureExecution)
	}
```

Two comments about this implementation: 
1. Please move the `kvSecureExecution` constant from the kubevirt hard coded FGs, to a new const group.
2. Please add a new constant for the `"s390x"` string. Preferred in the `nodeinfo` package.

jschintag:
Thank you for the review @nunnatsa. Sure i can wait till 1.7 release.
I will implement your feedback tomorrow

controllers/handlers/kubevirt.go
jschintag: There is no point, but also no cost.

controllers/handlers/kubevirt.go
jschintag: I have implemented it this way and added a unit-test for it.
@nunnatsa PTAL

nunnatsa:
After PR #3864 was merged, the KubeVirt `SecureExecution` is on beta, and we can start set it to true by default.

/unhold.

nunnatsa:
/unhold

with no dot...

controllers/commontestutils/nodeinfomocks.go
```diff
@@ -58,3 +62,13 @@ func SRCPHAINodeInfoMock() {
 		return false
 	}
 }
+
+// SpecificArchitectureNodeInfoMocks mocks a specific architecture for the nodes
+func SpecificArchitectureNodeInfoMocks(arch string) {
```
nunnatsa: Please change to list of archs, and rename the function accordingly.

Please don't mix control-plane and workloads architecture. Either split to two functions, or only do it for workloads, as control-plane arch won't be in use.

jschintag: I have split it to 2 functions, just in case someone in the future needs the ControlPlane one.
They take a list of strings and ensure it will never return nil.

controllers/commontestutils/nodeinfomocks.go
```diff
@@ -58,3 +62,23 @@ func SRCPHAINodeInfoMock() {
 		return false
 	}
 }
+
+// ControlPlaneArchitecturesMock mocks the architecures for ControlPlane nodes
+func ControlPlaneArchitecturesMock(arch ...string) {
+	if arch == nil {
+		arch = []string{}
+	}
```
nunnatsa: Not needed. nil is a valid return value.

controllers/commontestutils/nodeinfomocks.go
```diff
@@ -58,3 +62,23 @@ func SRCPHAINodeInfoMock() {
...
+	nodeinfo.GetControlPlaneArchitectures = func() []string {
+		return arch
+	}
+}
+
+// WorkloadseArchitecturesMock mocks the architecures for compute nodes
+func WorkloadsArchitecturesMock(arch ...string) {
+	if arch == nil {
+		arch = []string{}
+	}
```
nunnatsa: Same

nunnatsa (commented):
Thanks for fixing the comment @jschintag! Added two last nit comments.

controllers/commontestutils/nodeinfomocks.go
jschintag: Ok, i dropped that part.

nunnatsa (approved):
/approve

PR #3034, opened 2024-07-31
Register user-defined-network (UDN) binding on kubevirt CR

RamLavi:
**What this PR does / why we need it**:
This PR adds support to user-defined-network binding to the kubevirt CR. 
This is part of the process described on [kubevirt user guide](https://kubevirt.io/user-guide/network/network_binding_plugins/#definition-flow).
The behavior added is currently dev-preview, and as such is protected behind a FG: `PrimaryUserDefinedNetworkBinding`, that is set to `false` by default.

**Reviewer Checklist**

> Reviewers are supposed to review the PR for every aspect below one by one. To check an item means the PR is either "OK" or "Not Applicable" in terms of that item. All items are supposed to be checked before merging a PR. 

- [x] PR Message
- [x] Commit Messages
- [x] How to test
- [x] Unit Tests
- [x] Functional Tests
- [x] User Documentation
- [x] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
- [ ] Troubleshooting Friendly

**Jira Ticket**:

```jira-ticket

```

**Release note**:

```release-note
NONE
```

PR #3600, opened 2025-06-29
network, passt: Introduce deployPasstNetworkBinding annotation

oshoval:

Introduce HCO CR annotation `deployPasstNetworkBinding=true`.
The annotation toggles updating Kubevirt CR with Passt network binding.
Follow-up PR will deploy the required NAD, CNI DaemonSet, SCC and ServiceAccount.
Only then, Passt will be usable. 
Therefore this PR doesn't include updated docs yet.
PRs were split to ease review.

Basically it resurrects https://github.com/kubevirt/hyperconverged-cluster-operator/pull/3034 with the required changes.

`kubectl annotate hco kubevirt-hyperconverged -n kubevirt-hyperconverged deployPasstNetworkBinding=true --overwrite`

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket

```

**Release note**:

```release-note
None
```


controllers/operands/kubevirt.go
```diff
@@ -528,14 +536,20 @@ func getKVConfig(hc *hcov1beta1.HyperConverged) (*kubevirtcorev1.KubeVirtConfigu
...
 	networkBindings := maps.Clone(hcoNetworkBindings)
 
 	if networkBindings == nil {
 		networkBindings = make(map[string]kubevirtcorev1.InterfaceBindingPlugin)
 	}
 
 	networkBindings[primaryUDNNetworkBindingName] = primaryUserDefinedNetworkBinding()
+
+	if hcoFeatureGates.PasstNetworkBinding != nil && *hcoFeatureGates.PasstNetworkBinding {
+		passtImage, _ := os.LookupEnv(hcoutil.PasstImageEnvV)
```
nunnatsa: If you don't care about the 2nd return value, please use `os.Getenv` instead.

controllers/operands/kubevirt.go
```diff
@@ -1005,3 +1022,17 @@ func hcoCertConfig2KvCertificateRotateStrategy(hcoCertConfig hcov1beta1.HyperCon
...
+		NetworkAttachmentDefinition: path.Join(passtNetworkBindingNADNamespace, passtNetworkBindingNADName),
+		SidecarImage:                sidecarImage,
+		Migration:                   &kubevirtcorev1.InterfaceBindingMigration{},
+		ComputeResourceOverhead: &kubevirtcorev1.ResourceRequirementsWithoutClaims{
+			Requests: corev1.ResourceList{
+				corev1.ResourceMemory: resource.MustParse(bindingComputeMemoryOverhead),
+			},
+		},
+	}
+}
```
nunnatsa: Even though this method is not used too often, still there is no point to re-parse a constant value over and over again.

Please consider changing to something like:
```suggestion
const bindingComputeMemoryOverhead = "500Mi"

var passtResourceMemory = resource.MustParse(bindingComputeMemoryOverhead)

func passtNetworkBinding(sidecarImage string) kubevirtcorev1.InterfaceBindingPlugin {
	return kubevirtcorev1.InterfaceBindingPlugin{
		NetworkAttachmentDefinition: path.Join(passtNetworkBindingNADNamespace, passtNetworkBindingNADName),
		SidecarImage:                sidecarImage,
		Migration:                   &kubevirtcorev1.InterfaceBindingMigration{},
		ComputeResourceOverhead: &kubevirtcorev1.ResourceRequirementsWithoutClaims{
			Requests: corev1.ResourceList{
				corev1.ResourceMemory: passtResourceMemory,
			},
		},
	}
}
```

nunnatsa: Similar comment about
```go
path.Join(passtNetworkBindingNADNamespace, passtNetworkBindingNADName)
```
this value should be a constant, as all its components are known at build time.

WDTY about this?
```go
const networkAttachmentDefinition = passtNetworkBindingNADNamespace + "/" + passtNetworkBindingNADName

...

	return kubevirtcorev1.InterfaceBindingPlugin{
		NetworkAttachmentDefinition: networkAttachmentDefinition,
```

nunnatsa: nit (optional):

Looking on this function some more, it's seems that the only variable here is the image name. But actually, even the image name is defined by an environment variable, and so it also not going to be changed at runtime.

Please consider to pre-compute it on boot time, or only once (e.g. using `sync.Once`).

pkg/util/util.go
```diff
@@ -52,6 +52,13 @@ func GetOperatorNamespaceFromEnv() string {
 	return os.Getenv(OperatorNamespaceEnv)
 }
 
+func CheckPasstImagesEnvExists() error {
+	if _, passtImageVarExists := os.LookupEnv(PasstImageEnvV); !passtImageVarExists {
+		return fmt.Errorf("%s env var not found", PasstImageEnvV)
+	}
+	return nil
+}
+
```
nunnatsa: Why do we need this "util" function? first, it increase the util package, which is something we need to avoid. Then, it only called once. And also, it just a wrapper of a standard library function, with no meaningful additional value.

cmd/hyperconverged-cluster-operator/main.go
```diff
@@ -239,6 +239,9 @@ func main() {
 	err = metrics.SetupMetrics()
 	cmdHelper.ExitOnError(err, "failed to setup metrics: %v")
 
+	err = hcoutil.CheckPasstImagesEnvExists()
+	cmdHelper.ExitOnError(err, "failed to retrieve passt env vars")
```
nunnatsa: Are we really going to kill HCO if the image is not provided? How critical it is? Can we just not implement the passt feature in this case?

nunnatsa: And if it does critical, please be more robust with the error message (env var name); for example:

`the PASST_SIDECAR_IMAGE environment variable is not set`

or maybe 

`the PASST_SIDECAR_IMAGE environment variable must be set`

Something like this.

oshoval:
> Not sure we can add the feature gate as a named field, when it's not even in beta. The problem of API pollution, mostly for feature gates, is a real pain in HCO.

If it helps, on beta the plan is to remove the whole thing, (and make it integral part of kubevirt instead plugin),
please suggest what is the alternative if possible, when discussed locally we concluded the alternatives are problematic as well aren't they ?

cmd/hyperconverged-cluster-operator/main.go
oshoval: i think it is critical, we can't ignore passt deploy-ability, 
and the value is there anyhow usually unless a bug happens isnt it ?
(btw it was also considered critical on the PR which this PR is based on)

will update the error (if we converge to keep it)

EDIT - i wonder, aint it over protective to test the env var exists ?
they aren't optional, and do exists on both upstream and downstream,
unless this is tested for all env var, maybe worth to remove it ?
if env var doesnt exists, something really wrong happened
please correct me if i am wrong

nunnatsa:
> It it helps, on beta the plan is to remove the whole thing, please suggest what is the alternative if possible, when discussed locally we concluded the alternatives are problematic as well aren't they ?

What do you mean by "the whole thing"? the image? the binding in the KubeVirt CR? anything else? what is going to be removed?

Anyway, if this is the case, then the issue is much worst. We'll stay with unused field in the API, just to solve a very temporary issue. This is a big no-no.

oshoval:
> What do you mean by "the whole thing"? the image? the binding in the KubeVirt CR? anything else? what is going to be removed?

All = Kubevirt CR updating, NAD, DS, SCC, ServiceAccount, FG, the 2 images (sidecar and cni binary image)
Since all those are required because it is a network binding plugin, and the plan (not final, just as far as i understand)
is to turn it as part of kubevirt core instead a network plugin, so it wont be needed.

cc @EdDev @nirdothan @orelmisan 
fyi please 
(also see the hold above due to that)

EdDev:
> What do you mean by "the whole thing"? the image? the binding in the KubeVirt CR? anything else? what is going to be removed?
> 
> Anyway, if this is the case, then the issue is much worst. We'll stay with unused field in the API, just to solve a very temporary issue. This is a big no-no.

If the Alpha is successful, the idea is to move the feature into core kubevirt as a core network binding.
But this is not assured to happen and it is not clear how long it will take (in terms of versions).

Therefore, the API "pollution" is related to how you manage FGs in HCO, but I do not think it needs to impact gating experimental features in general. If you want to change the FG method in HCO, we could be your first users.

pkg/util/util.go
oshoval: on 3601 we will add another line
https://github.com/kubevirt/hyperconverged-cluster-operator/pull/3601/files#diff-0fa449830b5dc2b274b7c95a58b5071edc23f1e44af84a99c1fa9e600767829fR59

fwiw it was also requested more or less on
https://github.com/kubevirt/hyperconverged-cluster-operator/pull/3034/commits/c6d449d755a441a78d0d5e6970bf23b313b42522#r1713620518
having a named function allows to give it meaning

Please let me know what you prefer
(just flattening it, instead function might bloat the caller)
maybe move this function to main.go where it is called? 

please see last paragraph of https://github.com/kubevirt/hyperconverged-cluster-operator/pull/3600#discussion_r2174951326
as it might help to get of this function as a whole

oshoval:
addressed / answered comments
thanks

controllers/operands/kubevirt.go
oshoval: Done

btw https://github.com/kubevirt/hyperconverged-cluster-operator/blob/9dd37595acc436f3e454ab9f650ca7f142726e71/tests/func-tests/client.go#L48
here it seems the sync.Once happens again and again iiuc

oshoval:
As discussed, updated to use annotation instead FG

controllers/operands/kubevirt.go
nunnatsa: yes, you are right. good catch.

oshoval:
Please look the 4th commit here
https://github.com/kubevirt/hyperconverged-cluster-operator/pull/3614

(the rest are just cherry-pick of this PR)
It consolidates passt related logic to its own package, it will help on the follow-up https://github.com/kubevirt/hyperconverged-cluster-operator/pull/3601

So i plan to move that commit into 3601, is it fine ?
This way we can add the new functions that we present (i.e create daemonSet etc)
to that new passt package

another alternative is to add that commit as part of this PR (3600)

For now i will simply put it as part of 3601 so i can continue 3601 and freeze state of 3600 more or less,
unless you say otherwise

oshoval:
rebased to solve conflict

cmd/hyperconverged-cluster-operator/main.go
nunnatsa: this is usually to prevent null pointer exceptions and for fast-fail. I think it's more safe and provides better error messages.

controllers/handlers/kubevirt.go
```diff
@@ -1017,3 +1048,23 @@ func getLabelPatch(dest, src map[string]string) ([]byte, error) {
...
+		Migration:                   &kubevirtcorev1.InterfaceBindingMigration{},
+		ComputeResourceOverhead: &kubevirtcorev1.ResourceRequirementsWithoutClaims{
+			Requests: corev1.ResourceList{
+				corev1.ResourceMemory: passtResourceMemory,
+			},
+		},
+	}
+}
+
+func passtImageOnceFn() string {
```
nunnatsa: nit - naming: Not sure why we need the `Fn` suffix, and `Once` describing the "algorithm" within the function, that is not relevant to the user/reader of the function. I would simply call it `getPasstImage` or something like that.

controllers/handlers/kubevirt_test.go
```diff
@@ -1894,6 +1896,15 @@ Version: 1.2.3`)
 				})
 			})
 
+			BeforeEach(func() {
```
nunnatsa: there is another `BeforeEach`, just above this one. Please combine them.

controllers/handlers/kubevirt_test.go
```diff
@@ -1964,6 +1975,51 @@ Version: 1.2.3`)
 					Expect(existingResource.Annotations).ToNot(HaveKey(kubevirtcorev1.EmulatorThreadCompleteToEvenParity))
 				})
 
+				It("should add the Passt Binding to Kubevirt CR if PasstNetworkBinding is true in HyperConverged CR", func() {
+					hco.Annotations[deployPasstNetworkBindingAnnotation] = "true"
+					hco.Spec.NetworkBinding = nil
+
+					existingResource, err := NewKubeVirt(hco)
```
nunnatsa: I know, there are tons of `existingResource` vars all over the tests. But these are leftovers. Please call it `kv` to be more clear about what it is (+ few more instances below).

controllers/handlers/kubevirt_test.go
```diff
@@ -1964,6 +1975,51 @@ Version: 1.2.3`)
...
+					Expect(err).ToNot(HaveOccurred())
+
+					Expect(existingResource.Spec.Configuration.NetworkConfiguration).NotTo(BeNil())
+					Expect(existingResource.Spec.Configuration.NetworkConfiguration.Binding).NotTo(BeNil())
+					expectedInterfaceBindingPlugin := kubevirtcorev1.InterfaceBindingPlugin{
```
nunnatsa: maybe add an helper function to generate it? I think the test will be more readable.

controllers/handlers/kubevirt_test.go
```diff
@@ -1964,6 +1975,51 @@ Version: 1.2.3`)
...
+
+				It("should not add the Passt Network Binding to Kubevirt CR if PasstNetworkBinding is not set in HyperConverged CR", func() {
+					delete(hco.Annotations, deployPasstNetworkBindingAnnotation)
+					hco.Spec.NetworkBinding = nil
+
+					existingResource, err := NewKubeVirt(hco)
+					Expect(err).ToNot(HaveOccurred())
+
+					Expect(existingResource.Spec.Configuration.NetworkConfiguration).NotTo(BeNil())
+					Expect(existingResource.Spec.Configuration.NetworkConfiguration.Binding).ToNot(ContainElement("PasstIPStackMigration"))
```
nunnatsa: Since `Binding` is a map, please use the `HaveKey` matcher instead of `ContainElement` (it's for slices, mostly).

pkg/util/util.go
nunnatsa: Please just don't put it in utils. It is used only once, so please move it to where it is called.

controllers/handlers/kubevirt.go
```diff
@@ -834,7 +860,7 @@ func hcoConfig2KvConfig(
 	return kvConfig
 }
 
-func getFeatureGateChecks(featureGates *hcov1beta1.HyperConvergedFeatureGates) []string {
+func getFeatureGateChecks(featureGates *hcov1beta1.HyperConvergedFeatureGates, annotations map[string]string) []string {
```
nunnatsa, line 861: Optional: pass the whole HC CR, to simplify the function signature.

controllers/handlers/kubevirt.go
```diff
@@ -422,7 +441,7 @@ func getKVConfig(hc *hcov1beta1.HyperConverged) (*kubevirtcorev1.KubeVirtConfigu
 
 	seccompConfig := getKVSeccompConfig()
 
-	networkBindings := getNetworkBindings(hc.Spec.NetworkBinding)
+	networkBindings := getNetworkBindings(hc.Spec.NetworkBinding, hc.Annotations)
```
nunnatsa, line 444: Optional: pass the whole HC CR, to simplify the function signature. Also relevant for `getKvFeatureGateList` and `getFeatureGateChecks`.

controllers/handlers/kubevirt.go
```diff
@@ -970,8 +1001,8 @@ func getMandatoryKvFeatureGates(isKVMEmulation bool) []string {
 }
 
 // get list of feature gates or KV FG list
-func getKvFeatureGateList(fgs *hcov1beta1.HyperConvergedFeatureGates) []string {
-	checks := getFeatureGateChecks(fgs)
+func getKvFeatureGateList(fgs *hcov1beta1.HyperConvergedFeatureGates, annotations map[string]string) []string {
```
nunnatsa, line 1001: Optional: pass the whole HC CR, to simplify the function signature.

nunnatsa (commented):
Thanks for the fixes @oshoval.

Added some more inline comments.

controllers/handlers/kubevirt.go
oshoval: According clean code, we usually pass the minimum required
this makes the function care exactly about what needed
we have only 2 params, otherwise maybe it fits
(in any case, since it is optional i prefer not, unless you do of course)

controllers/handlers/kubevirt.go
oshoval: same as https://github.com/kubevirt/hyperconverged-cluster-operator/pull/3600#discussion_r2194411194 i think

controllers/handlers/kubevirt.go
oshoval: same as https://github.com/kubevirt/hyperconverged-cluster-operator/pull/3600#discussion_r2194411194

oshoval:
Thanks
Addressed / answered comments
Also fixed wrong binding name (it had the FG value instead), and used already existing consts where possible

controllers/handlers/kubevirt.go
nunnatsa: As said - it's optional. Up to you.

oshoval:
fixed undefined func

controllers/handlers/kubevirt.go
```diff
@@ -51,14 +51,32 @@ const (
...
 
-const primaryUDNNetworkBindingName = "l2bridge"
+const (
+	primaryUDNNetworkBindingName = "l2bridge"
+
+	deployPasstNetworkBindingAnnotation = "deployPasstNetworkBinding"
+
+	passtBindingName                = "passt"
+	passtNetworkBindingNADName      = "primary-udn-kubevirt-binding"
+	passtNetworkBindingNADNamespace = "default"
```
orelmisan, line 61: Is it possible to make it configurable?

oshoval:
fixed lint

controllers/handlers/kubevirt.go
oshoval: yes but on a follow-up (via HCO CR)
also worth to check each case if really needed

for example the namespace we can derive from isOpenShift and to check if the namespace is the expected one that OVN whitelist then we can use that one instead having it configurable

atm i prefer to focus on what we had, and to strive to make it deployable

oshoval:
rebased

controllers/handlers/kubevirt.go
```diff
@@ -850,6 +876,11 @@ func getFeatureGateChecks(featureGates *hcov1beta1.HyperConvergedFeatureGates) [
 		fgs = append(fgs, kvAlignCPUs)
 	}
 
+	value, ok := annotations[deployPasstNetworkBindingAnnotation]
+	if ok && value == "true" {
+		fgs = append(fgs, kvPasstIPStackMigration)
+	}
+
```
nunnatsa: nit: since the zero value of string is an empty string, and since we get the zero value from a map if the key is missing, we don't need to explicitly check the existence of the key in this case. It's enough to only check the value; e.g.
```suggestion
	if value := annotations[deployPasstNetworkBindingAnnotation]; value == "true" {
		fgs = append(fgs, kvPasstIPStackMigration)
	}
```

oshoval: done

oshoval:
changed https://github.com/kubevirt/hyperconverged-cluster-operator/pull/3600#discussion_r2195112268

controllers/handlers/kubevirt.go
oshoval: `if annotations[deployPasstNetworkBindingAnnotation] == "true"` also works, since we dont need the `value`
will change on next push - done

nunnatsa:
/lgtm

But we need to wait for KubeVirt v1.6, with the new feature gate that is used here in the PR.
/hold

controllers/handlers/kubevirt.go
```diff
@@ -51,14 +51,32 @@ const (
...
 )
 
-const primaryUDNNetworkBindingName = "l2bridge"
+const (
+	primaryUDNNetworkBindingName = "l2bridge"
+
+	deployPasstNetworkBindingAnnotation = "deployPasstNetworkBinding"
+
+	passtBindingName                = "passt"
+	passtNetworkBindingNADName      = "primary-udn-kubevirt-binding"
```
nirdothan, line 60: We need to think carefully about this name. Basically I would advise against choosing this name at this early stage of the project. However if you think that it will be complex to change in the future then maybe it is a reasonable name. 
Hopefully we will soon not need a NAD anyway,

controllers/handlers/kubevirt.go
nirdothan: Does KubeVirt have access to NADs on  `default` ns in D/S systems?

controllers/handlers/kubevirt.go
```diff
@@ -51,14 +51,32 @@ const (
...
+	passtNetworkBindingNADNamespace = "default"
+	networkAttachmentDefinition     = passtNetworkBindingNADNamespace + "/" + passtNetworkBindingNADName
+
+	bindingComputeMemoryOverhead = "500Mi"
```
nirdothan: Please reduce to "250Mi", following memory footprint improvements in `passt`.

controllers/handlers/kubevirt_test.go
```diff
@@ -4376,3 +4424,16 @@ Version: 1.2.3`)
...
 })
+
+func passtInterfaceBindingPlugin(networkAttachmentDefinition, sidecarImage string) kubevirtcorev1.InterfaceBindingPlugin {
+	return kubevirtcorev1.InterfaceBindingPlugin{
+		NetworkAttachmentDefinition: networkAttachmentDefinition,
+		SidecarImage:                sidecarImage,
+		Migration:                   &kubevirtcorev1.InterfaceBindingMigration{},
+		ComputeResourceOverhead: &kubevirtcorev1.ResourceRequirementsWithoutClaims{
+			Requests: corev1.ResourceList{
+				corev1.ResourceMemory: resource.MustParse("500Mi"),
```
nirdothan: As I indicated above, this is now "250Mi".

controllers/handlers/kubevirt.go
oshoval: This is the name we already used in the past (when deployed as part of ipam)
After the first phase we can consider whatever needed, atm it isn't crucial imho.

controllers/handlers/kubevirt.go
oshoval: Yes, this was the solution that worked on the past already (and the same that you tried on upstream and downstream lately)

oshoval:
changed to 250Mi, and few cosmetics

oshoval:
rebased to solve conflict

@nirdothan please consider to lgtm at least wrt to the part you asked
anyhow it is not blocking as far as i understand

nirdothan:
>

> rebased to solve conflict
> 
> @nirdothan please consider to lgtm at least wrt to the part you asked anyhow it is not blocking as far as i understand

/LGTM for what it's worth, I'm not an HCO expert

oshoval:
Rebased to solve conflict (and ran the required make targets of course)

oshoval:
can we retest please ? or need first https://github.com/kubevirt/hyperconverged-cluster-operator/pull/3635

nunnatsa:
We need #3635 to pass the upgrade tests

oshoval:
/retest
(3635 was merged)

nunnatsa:
I'm pretty sure it will fail again. I think you should rebase, to adopt the fix from 3635

oshoval:
Prow always rebases, unlike git actions, this why just test works, 
unless ci/prow has different behavior

Looking on the job yaml we can see the latest HEAD, so we should be good.

https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/kubevirt_hyperconverged-cluster-operator/3600/pull-ci-kubevirt-hyperconverged-cluster-operator-main-hco-e2e-upgrade-operator-sdk-aws/1947194536895713280
```
  refs:
    base_link: https://github.com/kubevirt/hyperconverged-cluster-operator/commit/bf069574fab4877cd3c29e68ca94f137a9a83bb6
    base_ref: main
    base_sha: bf069574fab4877cd3c29e68ca94f137a9a83bb6
    org: kubevirt
```
the hash is `Fix the upgrade test (#3635)`

nunnatsa:
> Prow always rebases, unlike git actions, this why just test works, unless ci/prow has different behavior
> 
> Looking on the job yaml we can see the latest HEAD, so we should be good.
> 
> https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/kubevirt_hyperconverged-cluster-operator/3600/pull-ci-kubevirt-hyperconverged-cluster-operator-main-hco-e2e-upgrade-operator-sdk-aws/1947194536895713280
> 
> ```
>   refs:
>     base_link: https://github.com/kubevirt/hyperconverged-cluster-operator/commit/bf069574fab4877cd3c29e68ca94f137a9a83bb6
>     base_ref: main
>     base_sha: bf069574fab4877cd3c29e68ca94f137a9a83bb6
>     org: kubevirt
> ```
> 
> the hash is `Fix the upgrade test (#3635)`

:crossed_fingers:

EdDev:
> Prow always rebases, unlike git actions, this why just test works,

GH action `checkup` also does that. I would say they do everything better just because millions are using them, unlike prow which is a niche.

oshoval:
last time i checked there wasnt auto rebase on GH jobs
there are plugins for that / blogs about it
(need one that is integrated into the job itself)

unless it was fixed inline since

EdDev:
> last time i checked there wasnt auto rebase on GH jobs there are plugins for that / blogs about it (need one that is integrated into the job itself)
> 
> unless it was fixed inline since

Last message in this subject from my side.
You can read https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request to see that the context is the target branch HEAD (i.e. anything that is already merged), on top of it the `checkout` action is running, i.e. the end result is the merge result between the HEAD of the target with the HEAD of the PR branch.

Sorry for hijacking this PR for this discussion, signing out.

---
EDIT: Maybe I should have also linked https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target as the text about this mess is there.

PR #3601, opened 2025-06-29
network,passt: Deploy Passt required objects

oshoval:

In continuation of https://github.com/kubevirt/hyperconverged-cluster-operator/pull/3600
Which updates the Kuebvirt CR with Passt binding, once `deployPasstNetworkBinding=true` annotation is set (HCO CR),
This PR deploys the following items:
1. DaemonSet that deploys the CNI on the nodes (deploys according affinity).
2. NetworkAttachmentDefinition.

OpenShift only:

3. Service Account (connects SCC to the DaemonSet).
4. SCC.

Tested manually with kind + OVN-k8s, UDN that use Passt, and on KCLI OpenShift.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket

```

**Release note**:

```release-note
None
```


deploy/images.csv
```diff
@@ -9,6 +9,7 @@ KUBEVIRT_LIBGUESTFS_TOOLS_IMAGE,quay.io/kubevirt/libguestfs-tools,KUBEVIRT_VERSI
 KUBEVIRT_EXPORTPROXY_IMAGE,quay.io/kubevirt/virt-exportproxy,KUBEVIRT_VERSION,1017c0e7097bf309e7a2e8542656f76c3ee1365ff03d01f563dbe108a98153b1
 KUBEVIRT_EXPORSERVER_IMAGE,quay.io/kubevirt/virt-exportserver,KUBEVIRT_VERSION,10d35daa56ec1981bbbcf898f5200909e044b3ae34a3f790ca82fedc906c52d9
 NETWORK_PASST_BINDING_IMAGE,quay.io/kubevirt/network-passt-binding,KUBEVIRT_VERSION,e6e01ca7adedb6215caa7284c8e38ee5afa9ff9033180907d6027941b2477f0a
+NETWORK_PASST_BINDING_CNI_IMAGE,quay.io/oshoval/network-passt-binding-cni,KUBEVIRT_VERSION,237dfdc49476a2477665faf2b74262ed8821f31a6da261eb78883eaa2134716c
```
oshoval: dont care about the hacks right now please, it is because there isnt release yet with it
there are more hacks / todos, hence it is wip and partial, i.e it doesnt have all the unit tests / doesnt deploy all required objs yet
it is just so we can sync about the direction so far please
thanks

nunnatsa: This is tricky. If we'll bump KubeVirt, it won't work. we need to add it elsewhere. or use another temporary environment variable than `KUBEVIRT_VERSION`.

oshoval: it is in kubevirt, it is a pre merge hack, just so i will be unblocked developing it
please disregard it for now
it will be on `quay.io/kubevirt/network-passt-binding-cni`
once there is a new version 1.6
exactly as was done for `quay.io/kubevirt/network-passt-binding` (line above)
on https://github.com/kubevirt/hyperconverged-cluster-operator/pull/3034

we wont merge it before the image is available

EDIT - updated with official image

hack/generate.sh
```diff
@@ -26,12 +26,12 @@ defaulter-gen \
 	--go-header-file "${PROJECT_ROOT}/hack/boilerplate.go.txt" \
 	"${PACKAGE}/${API_FOLDER}/${API_VERSION}"
 
-openapi-gen \
-	--output-file zz_generated.openapi.go \
-	--go-header-file "${PROJECT_ROOT}/hack/boilerplate.go.txt" \
-	--output-dir api/v1beta1/ \
-	--output-pkg github.com/kubevirt/hyperconverged-cluster-operator/api/v1beta1 \
-	"${PACKAGE}/${API_FOLDER}/${API_VERSION}"
+#openapi-gen \
```
oshoval: just hack because it doesnt work locally for me
wont be merged

EDIT - Removed

oshoval:
Hi @nunnatsa @orenc1 

Please let us know if we should continue this approach once you can ?
This is just service account example out of the 4 required objects (additional are SCC, DaemonSet, NAD)
If you prefer meeting or whatever can help please let me know

Of course we still also need to converge about https://github.com/kubevirt/hyperconverged-cluster-operator/pull/3600
but that PR (3600) is needed this way or another (wrt FG), orthogonal to how we deploy rest of the objects.

Thanks

controllers/operands/kubevirtConsolePlugin.go
```diff
@@ -281,6 +281,17 @@ func NewKvUIPluginSvc(hc *hcov1beta1.HyperConverged) *corev1.Service {
 	}
 }
 
+// TODO move to a new fileu
+func NewPasstBindingCNISA(hc *hcov1beta1.HyperConverged) *corev1.ServiceAccount {
```
oshoval, line 285: this (and the rest of the objects we deploy for passt)
will be moved to a new file
`controllers/operands/passt.go` which has the required function to generate the objects

oshoval:
aligned to updated https://github.com/kubevirt/hyperconverged-cluster-operator/pull/3600 (annotation instead of FG)

oshoval:
Rebased on top of [3600](https://github.com/kubevirt/hyperconverged-cluster-operator/pull/3600
and on top of main including latest refactors

oshoval:
/hold 

until we remove the hacks (see PR desc)

controllers/handlers/passt/passt.go
```diff
@@ -0,0 +1,349 @@
...
+		},
+	}
+}
+
+// NewPasstBindingCNIDaemonSet creates a DaemonSet for the passt binding CNI
+func NewPasstBindingCNIDaemonSet(hc *hcov1beta1.HyperConverged, isOpenShift bool) *appsv1.DaemonSet {
+	maxUnavailable := intstr.FromString("10%")
+
+	hostpath := "/opt/cni/bin"
+	if isOpenShift {
```
nunnatsa, line 87: This is not going to be changed on runtime. You can set the right valur on boot, e.g. in the handler constructor. We don't need to pass the isOpenshift argument to the function.

controllers/handlers/passt/passt.go
```diff
@@ -0,0 +1,349 @@
...
+						Image: os.Getenv(hcoutil.PasstCNIImageEnvV),
+						Command: []string{
+							"/bin/sh",
+							"-ce",
+						},
+						Args: []string{
+							`ls -la "/cni/kubevirt-passt-binding"
+cp -f "/cni/kubevirt-passt-binding" "/opt/cni/bin"
+echo "passt binding CNI plugin installation complete..sleep infinity"
+sleep 2147483647`,
```
nunnatsa, line 128: ????

controllers/handlers/passt/passt.go
```diff
@@ -0,0 +1,349 @@
...
+						},
+						Resources: corev1.ResourceRequirements{
+							Requests: corev1.ResourceList{
+								corev1.ResourceCPU:    resource.MustParse("10m"),
```
nunnatsa, line 132: Can be fone once, as a global variable. No point to re-parse constant value again and again.

controllers/handlers/passt/passt.go
```diff
@@ -0,0 +1,349 @@
...
+								corev1.ResourceMemory: resource.MustParse("15Mi"),
```
nunnatsa, line 133: Same

controllers/handlers/passt/passt.go
```diff
@@ -0,0 +1,349 @@
...
+					},
+				},
+			},
+		},
+	}
+
+	daemonSet := NewPasstBindingCNIDaemonSetWithNameOnly(hc)
+	daemonSet.Spec = spec
+
+	infrastructureHighlyAvailable := nodeinfo.IsInfrastructureHighlyAvailable()
```
nunnatsa: Redundant variable. Just call the function

controllers/handlers/passt/passt.go
```diff
@@ -0,0 +1,349 @@
...
+	affinity := getPodAntiAffinity(daemonSet.Labels[hcoutil.AppLabelComponent], infrastructureHighlyAvailable)
+
+	if hc.Spec.Infra.NodePlacement != nil {
+		if hc.Spec.Infra.NodePlacement.NodeSelector != nil {
+			daemonSet.Spec.Template.Spec.NodeSelector = maps.Clone(hc.Spec.Infra.NodePlacement.NodeSelector)
+		} else {
```
nunnatsa: Not needed. nil is the zero value of a pointer

controllers/handlers/passt/passt.go
```diff
@@ -0,0 +1,349 @@
...
+			daemonSet.Spec.Template.Spec.Affinity = hc.Spec.Infra.NodePlacement.Affinity.DeepCopy()
+		} else {
+			daemonSet.Spec.Template.Spec.Affinity = affinity
+		}
+
+		if hc.Spec.Infra.NodePlacement.Tolerations != nil {
+			daemonSet.Spec.Template.Spec.Tolerations = make([]corev1.Toleration, len(hc.Spec.Infra.NodePlacement.Tolerations))
+			copy(daemonSet.Spec.Template.Spec.Tolerations, hc.Spec.Infra.NodePlacement.Tolerations)
+		} else {
+			daemonSet.Spec.Template.Spec.Tolerations = nil
```
nunnatsa: Same

controllers/handlers/passt/passt.go
```diff
@@ -0,0 +1,349 @@
...
+		}
+	} else {
+		daemonSet.Spec.Template.Spec.NodeSelector = nil
+		daemonSet.Spec.Template.Spec.Affinity = affinity
+		daemonSet.Spec.Template.Spec.Tolerations = nil
```
nunnatsa: Same

nunnatsa (commented):
Thanks @oshoval 

Not a full review. Just a few comments about things that cought my eye

controllers/handlers/passt/passt.go
oshoval: I believe you point on the `sleep 2147483647`
since we run on upstream busybox that doesnt have `sleep infinity` we use this max int32,

This is a demonSet that once it deployed the binary doesnt need to do anything more, so it just sleep forever.
This what was done by CNAO until today so i just imitated.

Theoretical alternative is to exit 0 and, use `restartPolicy: OnFailure`, and then it will look as `Completed` instead `Running` (which can confuse scripts, code and people looking for Running).
I prefer to not try to optimize it and just do what we did on CNAO.


One thing we can fix on follow-up is that killing this pod takes time, because sleep doesnt trap SIGTERM,
we can sleep & and wait, with trap which does catch the signal, but didn't try it yet.

controllers/handlers/passt/passt.go
oshoval: True but please consider the following:
1. It is just a boolean, and might be more readable.
2. There is cache for the new daemonSet so we do it just once.
3. If it is boolean, we can also toggle ServiceAccount addon to the daemonSet for non OpenShift, removing one unrequired item on upstream (of course to install the ServiceAccount only for OpenShift as well).
It is nicer to have a condition on bool instead string.

Btw we can also use `hcoutil.GetClusterInfo().IsOpenshift()` but then testing is bit more complex.
(alternatives on PR desc)

Please let me know what you prefer.

oshoval:
rebased only so far to solve conflicts
(we will have lots of those due to the timestamps in the manifests)

controllers/handlers/passt/passt.go
oshoval: we have cache for daemonSet, it happens anyhow once
would it be nicer to have it inline then ?
instead creating a global variable and using it

btw the fact we have cache, means that if affinity is changed, it wont affect the daemonSet
It might be fine, just raising for awareness

controllers/handlers/passt/passt.go
oshoval: https://github.com/kubevirt/hyperconverged-cluster-operator/pull/3601#discussion_r2212623644

controllers/handlers/passt/passt.go
oshoval: if you mean to remove ` != nil` from the boolean, then we can't, it wont compile
`if hc.Spec.Infra.NodePlacement {`

`./passt.go:177:5: non-boolean condition in if statement`

or you mean to remove (it is not a pointer, but `map[string]string`) but here you are right, i imitated other examples of HCO, it is nit more explicit, but anyhow on it

```
 else {
			daemonSet.Spec.Template.Spec.NodeSelector = nil
		}
```

oshoval:
Thanks
answered / addressed the marked ones

controllers/handlers/passt/passt.go
nunnatsa: If it only happens once, the ok.

controllers/handlers/passt/passt.go
nunnatsa: I think it's a strange behavior. Completed is a valid and well known k8s pod status and I can't find any issue with it. Sleeping just to never exit is a debug practice, not a prodution one, IMHO.

controllers/handlers/passt/passt.go
nunnatsa: > Btw we can also use hcoutil.GetClusterInfo().IsOpenshift() but then testing is bit more complex.

Yes, this is exactly what you need to do. Testing is not an issue. There are pre-prepared ClusterInfo mocks you can use for testing (see tons of examples all over). If you mean to functional tests, then you can use a label to run the test only on openshift. Again, many exaples in the func test code.

oshoval: general question - so far i implemented only unit tests, i dont think we need functional tests for opt-in feature, we will test it e2e by QE, unit tests gives nice coverage
is it fine ?

Will adapt meanwhile to use it - done (also added ut)

controllers/handlers/passt/passt.go
oshoval: Tried to use `restartPolicy: OnFailure` but it isn't legit for daemonSet
only 'Always'

```
{"level":"error","ts":"2025-07-17T10:01:18Z","logger":"controller_hyperconverged","msg":"failed to Ensure an operand","Request.Namespace":"kubevirt-hyperconverged","Request.Name":"kubevirt-hyperconverged","error":"DaemonSet.apps \"passt-binding-cni\" is invalid: spec.template.spec.restartPolicy: Unsupported value: \"OnFailure\": supported values: \"Always\"",....
```
DaemonSet doesnt support it, because the definition of them is they always run,
and since we dont want it do die and restart in endless loop, and to have it on every node, this is the practice we always did (having ds with sleep)

oshoval:
addressed https://github.com/kubevirt/hyperconverged-cluster-operator/pull/3601#discussion_r2212781258
(used ClusterInfo instead passing IsOpenShift bool + added unit tests for it)
also done that ServiceAccount will be installed only on OCP where it is actually needed (to connect the SCC with the DS via the ServiceAccount)

oshoval:
rebased on top of main / 3600
removed CNI image hack

oshoval:
Removed temporary wa for openapi-gen (it doesnt work on some machines)

oshoval:
/hold cancel

the hold i put because we waited for release is resolved + rebased (hacks removed)
we still depend on 3600 to be merged before this one

oshoval:
Do we need to add HCO labels, so Network policies will be created ?

controllers/operandhandler/operandHandler.go
```diff
@@ -220,6 +221,7 @@ func (h *OperandHandler) EnsureDeleted(req *common.HcoRequest) error {
 		passt.NewPasstBindingCNISA(req.Instance),
 		passt.NewPasstBindingCNIDaemonSetWithNameOnly(req.Instance),
 		passt.NewPasstBindingCNINetworkAttachmentDefinition(req.Instance),
+		passt.NewPasstBindingCNISecurityContextConstraints(req.Instance),
```
oshoval, line 224: it is safe to have it here on upstream where we dont have SCC right ?
we do `securityv1.Install` and if it isn't found it simply returns NoMatchErr
as far as i understand

oshoval:
Rebased since 3600 was merged

oshoval:
Rebased
Since there are lots of merge conflicts due to the createdAt timestamp

Created this initial script that helps (there are more ways to solve / imporve it from here of course)
it can save precious time (of course it should be used only if the conflicts are just `createdAt` / ones that will be auto fixed by the `make`)

```
#!/bin/bash -e

git checkout --ours  .
git add -u
make generate build-manifests
git add -u
export GIT_EDITOR=true
git rebase --continue
```

controllers/handlers/passt/passt.go
nunnatsa: OK. it's really ugly but if this is the only option, then ok

controllers/handlers/passt/passt.go
```diff
@@ -0,0 +1,345 @@
...
+		return fmt.Errorf("the %s environment variable must be set", hcoutil.PasstImageEnvV)
+	}
+	if _, passtCNIImageVarExists := os.LookupEnv(hcoutil.PasstCNIImageEnvV); !passtCNIImageVarExists {
+		return fmt.Errorf("the %s environment variable must be set", hcoutil.PasstCNIImageEnvV)
+	}
+	return nil
+}
+
+// NetworkBinding creates an InterfaceBindingPlugin for passt network binding
+func NetworkBinding(sidecarImage string) kubevirtcorev1.InterfaceBindingPlugin {
```
nunnatsa: Why do we need to pass the image as parameter? the function only called once. The same is for the GetImage function - it is only called once, to pass the parameter to this function. We can just call GetImage from within the function.

In addition, the content of the function is actually static. It can be a variable instead of a function. May the function can just return the var.

controllers/handlers/passt/passt.go
```diff
@@ -0,0 +1,345 @@
...
+	}
+}
+
+// GetImage gets the passt image from environment variable
+func GetImage() string {
+	passtImageOnce.Do(func() {
+		passtImage = os.Getenv(hcoutil.PasstImageEnvV)
+	})
+	return passtImage
+}
```
nunnatsa: not sure we need this function at all. we can just read the env var when defining the passtImage variable, in line 43. Then we won't need the syncOnce either.

controllers/handlers/passt/passt.go
```diff
@@ -0,0 +1,345 @@
...
+			RollingUpdate: &appsv1.RollingUpdateDaemonSet{
+				MaxUnavailable: &maxUnavailable,
+			},
+		},
+		Template: corev1.PodTemplateSpec{
+			ObjectMeta: metav1.ObjectMeta{
+				Labels: map[string]string{
+					"name": "passt-binding-cni",
+					"tier": "node",
+					"app":  "passt-binding-cni",
```
nunnatsa: please use consts

controllers/handlers/passt/passt.go
```diff
@@ -0,0 +1,345 @@
...
+				},
+				Annotations: map[string]string{
+					"description": "passt-binding-cni installs 'passt binding' CNI on cluster nodes",
```
nunnatsa: same

controllers/handlers/passt/passt.go
```diff
@@ -0,0 +1,345 @@
...
+	return createPasstConditionalHandler(
+		operands.NewSecurityContextConstraintsHandler(Client, Scheme, NewPasstBindingCNISecurityContextConstraints),
+		func(hc *hcov1beta1.HyperConverged) client.Object {
+			return NewPasstBindingCNISecurityContextConstraints(hc)
+		},
+	)
+}
+
+// createPasstConditionalHandler creates a conditional handler that checks for passt deployment annotation
+func createPasstConditionalHandler(handler *operands.GenericOperand, objectCreator func(hc *hcov1beta1.HyperConverged) client.Object) operands.Operand {
```
nunnatsa, line 325: nit: Not sure why this wrap is needed. 

we can define the condition function as regular function, and send it to `operands.NewConditionalHandler` as a parameter. But I guess it's just a matter of style. feel free to ignore.

controllers/operandhandler/operandHandler.go
nunnatsa: You can wrap it with isOpenshift. This will solve the issue.

controllers/operands/daemonSetHandler.go
```diff
@@ -0,0 +1,109 @@
+package operands
```
nunnatsa, line 1: The whole type feels for me like a pre-mature optimization, or generalization. We only have one DaemonSet now.

controllers/operands/securityContextConstraintsHandler.go
```diff
@@ -0,0 +1,89 @@
+package operands
```
nunnatsa, line 1: same

controllers/operands/networkAttachmentDefinitionHandler.go
```diff
@@ -0,0 +1,64 @@
+package operands
```
nunnatsa, line 1: same (?)

controllers/operands/serviceAccountHandler.go
```diff
@@ -0,0 +1,63 @@
+package operands
```
nunnatsa, line 1: same

pkg/components/components.go
```diff
@@ -641,6 +646,26 @@ func GetClusterPermissions() []rbacv1.PolicyRule {
...
+		{
+			APIGroups: stringListToSlice(""),
+			Resources: stringListToSlice("serviceaccounts"),
+			Verbs:     stringListToSlice("get", "list", "watch", "create", "update", "delete"),
+		},
+		{
+			APIGroups: stringListToSlice("apps"),
+			Resources: stringListToSlice("daemonsets"),
+			Verbs:     stringListToSlice("get", "list", "watch", "create", "update", "delete"),
+		},
```
nunnatsa: can be added to the existing "deployment" rule

pkg/util/cluster.go
```diff
@@ -31,6 +31,7 @@ type ClusterInfo interface {
 	IsConsolePluginImageProvided() bool
 	IsMonitoringAvailable() bool
 	IsDeschedulerAvailable() bool
+	IsNADAvailable() bool
```
nunnatsa, line 34: Please avoid extending this huge interface. We're trying to reduce its size and split it. If you don't find and easy way then ok for now.

nunnatsa (commented):
Thanks @oshoval 

Added some inline comments.

nunnatsa:
> conflicts

the build-manifests already ignores the createdAt field if it is the only different.

pkg/util/cluster.go
oshoval: I don't have another idea at the moment, since we are short in time, would appreciate it we can consider it on follow-up please

unless we wrap caller with isOpenShift instead, wdyt ?
(NAD CRD anyhow doesnt exist on upstream at this point, only on OCP, unless we will always deploy the CRD, but it isnt crucial, info about that in last commit message)

controllers/operands/daemonSetHandler.go
oshoval: but we might have more that use it, if for example we move other operands from CNAO to HCO
it also keeps the handlers agnostic of passt, and seperate packages to the handler and the user

what you suggest ? to make it part of passt package ? 
imho the current separation is nicer (even that it is single use) and will worth it if more are using it

controllers/operands/securityContextConstraintsHandler.go
oshoval: lets continue please at https://github.com/kubevirt/hyperconverged-cluster-operator/pull/3601#discussion_r2224355434

controllers/operands/networkAttachmentDefinitionHandler.go
oshoval: lets continue please at https://github.com/kubevirt/hyperconverged-cluster-operator/pull/3601#discussion_r2224355434

controllers/operands/serviceAccountHandler.go
oshoval: lets continue please at https://github.com/kubevirt/hyperconverged-cluster-operator/pull/3601#discussion_r2224355434

controllers/operandhandler/operandHandler.go
oshoval: I am not sure there is an issue at all, just wanted to make sure
If there is no known issue, i think it is better to not wrap it

oshoval:
> > conflicts
> 
> the build-manifests already ignores the createdAt field if it is the only different.

but git conflict doesnt, as github shows now again

controllers/operands/daemonSetHandler.go
nunnatsa: I won't fight over it. Just want to make sure it's not over engineering

pkg/util/cluster.go
nunnatsa: I'll take of it in followup PRs, sometime in the future. We must split this interface (already started with `nodeInfo`). So, ok by me.

nunnatsa:
> but git conflict doesnt, as github shows now again

yes, but this is not so common. right now we have 2 other big efforts (network policies and multi-arch support), that requires big changes. That why we have a frequent changes in the CSV. This is not usually the status.

Really sorry it delay us. but will get through it.

controllers/handlers/passt/passt.go
oshoval: Called `GetImage` directly from `NetworkBinding` (will push soon (done), and we can iterate if you meant more things)

not sure i understood the last paragraph, do you suggest to just inline the body of NetworkBinding where it is called ? i think it is nicer to have it in passt file, instead inline on kubevirt file, but whatever you choose

fwiw `NetworkBinding` is called only at the kubevirt handler (via cache), 
but more times indirectly via `NewKubeVirt` (on `pkg/webhooks/validator/validator.go`)

oshoval:
rebased only

oshoval:
> > but git conflict doesnt, as github shows now again
> 
> yes, but this is not so common. right now we have 2 other big efforts (network policies and multi-arch support), that requires big changes. That why we have a frequent changes in the CSV. This is not usually the status.
> 
> Really sorry it delay us. but will get through it.

np, with the new scripts it saves me lots of time, personally i like stuff that make cycles shorter

oshoval:
addressed / answered comments
will check the failures

thanks

oshoval:
fixed sanity / lint

btw do you know by heart please what command exactly renders the two csvs and deploy/cluster_role.yaml ?
it will be faster to run those instead the whole `make sanity / make generate build-manifests` that i am using now

nunnatsa:
> fixed sanity / lint
> 
> btw do you know by heart please what command exactly renders the two csvs and deploy/cluster_role.yaml ? it will be faster to run those instead the whole `make sanity / make generate build-manifests` that i am using now

You'll need `make build-manifests` for that

controllers/handlers/passt/passt.go
```diff
@@ -0,0 +1,334 @@
...
+		return fmt.Errorf("the %s environment variable must be set", hcoutil.PasstCNIImageEnvV)
+	}
+	return nil
+}
+
+// NetworkBinding creates an InterfaceBindingPlugin for passt network binding
+func NetworkBinding() kubevirtcorev1.InterfaceBindingPlugin {
+	return kubevirtcorev1.InterfaceBindingPlugin{
+		NetworkAttachmentDefinition: NetworkAttachmentDefinition,
+		SidecarImage:                os.Getenv(hcoutil.PasstImageEnvV),
```
nunnatsa, line 59: nit: (what I was trying to explain in previous comment. I guess I wasn't clear enough). Not critical.

WDYT about having the (removed) `passtImage` var, and to initialize it on definition (where it was originally defined); e.g.

```go
var passtImage = os.Getenv(hcoutil.PasstImageEnvV)
```

then use it here:
```suggestion
		SidecarImage:                passtImage,
```

controllers/handlers/passt/passt.go
```diff
@@ -0,0 +1,334 @@
...
+			Name:      passtCNIObjectName,
+			Namespace: hc.Namespace,
+			Labels:    hcoutil.GetLabels(hcoutil.HyperConvergedName, hcoutil.AppComponentNetwork),
+		},
+	}
+}
+
+// NewPasstBindingCNIDaemonSet creates a DaemonSet for the passt binding CNI
+func NewPasstBindingCNIDaemonSet(hc *hcov1beta1.HyperConverged) *appsv1.DaemonSet {
+	maxUnavailable := intstr.FromString("10%")
```
nunnatsa, line 82: This is a compute of a constant value. can be done once, by defining the variable outside of the function.

nunnatsa: Unless it happens once, then please ignore.

oshoval: happens once (via cache)

controllers/handlers/passt/passt.go
oshoval: hmm, i am less fond of those global things

oshoval:
> > fixed sanity / lint
> > btw do you know by heart please what command exactly renders the two csvs and deploy/cluster_role.yaml ? it will be faster to run those instead the whole `make sanity / make generate build-manifests` that i am using now
> 
> You'll need `make build-manifests` for that

thanks
btw instead of pulling images as it does, maybe it can use skopeo and be faster ? if it just need hash / check if image exists

oshoval:
rebased to solve conflicts

PR #3615, opened 2025-07-07
Refactor operands

nunnatsa:
split the operands package to three new packages, to get more organized code base, and to allow adding new operands in new packages, to be owned by another teams.

This PR splits the operands package into 3 packages:
* the OperandHandler logic moved to the new `operatorhandler` package
* the resource specific handler logic moved to the new `handlers` package
* all the operands/handlers infrastructure logic reminds in the `operands` package

**Release note**:

```release-note
None
```


controllers/operandhandler/operandHandler_test.go
```diff
@@ -461,4 +493,105 @@ var _ = Describe("Test operandHandler", func() {
...
+			Expect(eventEmitter.CheckEvents(expectedEvents)).To(BeFalse())
+		})
+
+	})
 })
+
+const (
+	pkgDirectory = "controllers/operandhandler"
+	testFilesLoc = "testFiles"
+)
```
machadovilaca: move to file top

controllers/operandhandler/operandHandler_test.go
```diff
@@ -461,4 +493,105 @@ var _ = Describe("Test operandHandler", func() {
 			})
 		})
 	})
+
+	Context("test imageStream deletion", func() {
```
machadovilaca, line 502: why did we move this from "controllers/handlers/imageStream_test.go"

controllers/operandhandler/operandHandler_test.go
```diff
@@ -461,4 +493,105 @@ var _ = Describe("Test operandHandler", func() {
...
+	if strings.HasSuffix(wd, pkgDirectory) {
+		return testFilesLoc
+	}
+	return path.Join(pkgDirectory, testFilesLoc)
+}
+
+func TestOperators(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "OperatorHandler Suite")
+}
```
machadovilaca: we also usually have this before the test Describe, no?

controllers/handlers/imageStream.go
```diff
@@ -74,59 +78,56 @@ func (iso imageStreamOperand) ensure(req *common.HcoRequest) *EnsureResult {
...
+		iso.hooks.required.Namespace = iso.hooks.originalNS
 	}
 	return nil
 }
 
-func (iso imageStreamOperand) deleteImageStream(req *common.HcoRequest) *EnsureResult {
-	h := iso.operand.hooks.(*isHooks)
-	_, err := util.EnsureDeleted(req.Ctx, iso.operand.Client, h.required, req.Instance.Name, req.Logger, false, true, false)
+func (iso imageStreamOperand) deleteImageStream(req *common.HcoRequest) *operands.EnsureResult {
+	//h := iso.operand.hooks.(*isHooks)
```
machadovilaca: if not needed, i think we can remove

controllers/operandhandler/operandHandler_test.go
nunnatsa: I prefer to leave it where we're using it, so the reader can see the whole thing in the same screen, or close.

Moving up is a bit "C". But I won't fight about it.

controllers/operandhandler/operandHandler_test.go
nunnatsa: These two tests checks also the events we emit when deleting an images stream. The code that does that is in the Operator controller.

controllers/operandhandler/operandHandler_test.go
nunnatsa: No rules about it, AFAIK. Different tests do different things. I can move if it makes more sense to you.

controllers/operandhandler/operandHandler_test.go
nunnatsa: OK. moved it into the function.

controllers/operandhandler/operandHandler_test.go
nunnatsa: Moved

PR #3619, opened 2025-07-09
build: Use quay instead docker.io

oshoval:
Avoid rate limits.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket

```

**Release note**:

```release-note
None
```

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,4 +1,4 @@
-FROM --platform=${BUILDPLATFORM} docker.io/golang:1.24.3 AS builder
+FROM --platform=${BUILDPLATFORM} quay.io/projectquay/golang:1.24 AS builder
 
 WORKDIR /go/src/github.com/kubevirt/hyperconverged-cluster-operator/
 COPY . .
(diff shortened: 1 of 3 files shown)


nunnatsa:
This change is a very like a good idea. Thanks @oshoval !

One concern I have is that all the tags in this repository are only major.minor. There is no good way to know the exact golang version, if we need to fix it for CVEs and so on.

Even when running podman inspect, there is no label with the exact version. The only thing I've got is the "created by" on one of the layers:
```shell
"created_by": "|1 GO_VERSION=1.24.5 /bin/sh -c ...
```

Anyway, the PR itself is great

/lgtm
/approve

oshoval:
Thanks

btw if you add env var like this, it will auto download the desired version it seems
tried manually, maybe worth to add it in the Dockerfile (take version according go.mod)
```
[root@b8f4c95bf35b go]# export GOTOOLCHAIN=go1.24.3
[root@b8f4c95bf35b go]# go version
go: downloading go1.24.3 (linux/amd64)
go version go1.24.3 linux/amd64
```

PR #3620, opened 2025-07-09
API: Rename two fields with wrong names

nunnatsa:

Since we didn't release a version with the new
fields we can still fix their names.

The following two fields was renamed:
* `status.nodeInfo.controlPlaneNodesArchitecture` - three issues with this field:
1. it contains the word "Nodes" where the `status.nodeInfo.workloadsArchitectures` does not
2. it in singular form
3. it does not match to the corresponding field in the SSP spec. The field name was changed to `status.nodeInfo.controlPlaneArchitectures`
* `spec.featureGates.enableMultiArchCommonBootImageImport`
   - the "Common" part of the name is wrong and misleading. The field name will be changed to `spec.featureGates.enableMultiArchBootImageImport`.

**Jira Ticket**:
```jira-ticket
https://issues.redhat.com/browse/CNV-64974
```

**Release note**:

```release-note
None
```


tests/func-tests/defaults_test.go
```diff
@@ -89,7 +89,7 @@ var _ = Describe("Check Default values", Label("defaults"), Serial, func() {
 			Entry("when removing /spec/featureGates/deployKubeSecondaryDNS", "/spec/featureGates/deployKubeSecondaryDNS"),
 			Entry("when removing /spec/featureGates/persistentReservation", "/spec/featureGates/persistentReservation"),
 			Entry("when removing /spec/featureGates/alignCPUs", "/spec/featureGates/alignCPUs"),
-			Entry("when removing /spec/featureGates/enableMultiArchCommonBootImageImport", "/spec/featureGates/enableMultiArchCommonBootImageImport"),
+			Entry("when removing /spec/featureGates/enableMultiArchBootImageImport", "/spec/featureGates/enableMultiArchCommonBootImageImport"),
```
avlitman: Here is supposed to be with common?

nunnatsa: No. We should never use the old name

avlitman: I mean at the end of the line it's still the old name

nunnatsa: Good catch! thanks @avlitman - fixed.

PR #3628, opened 2025-07-15
passt: Deploy NAD on non default ns

oshoval:
On OpenShift deploy NAD on openshift-cnv, instead on default namespace
Because CNO deploys multus with access to openshift-cnv.

It is cleaner.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket

```

**Release note**:

```release-note
None
```

### DIFF
--- a/controllers/handlers/passt/passt.go
+++ b/controllers/handlers/passt/passt.go
@@ -32,10 +32,7 @@ const (
 
 	passtCNIObjectName = "passt-binding-cni"
 
-	networkBindingNADName       = "primary-udn-kubevirt-binding"
-	networkBindingNADNamespace  = "default"
-	NetworkAttachmentDefinition = networkBindingNADNamespace + "/" + networkBindingNADName
-
+	networkBindingNADName        = "primary-udn-kubevirt-binding"
 	bindingComputeMemoryOverhead = "250Mi"
 )
 
@@ -55,7 +52,7 @@ func CheckPasstImagesEnvExists() error {
 // NetworkBinding creates an InterfaceBindingPlugin for passt network binding
 func NetworkBinding() kubevirtcorev1.InterfaceBindingPlugin {
 	return kubevirtcorev1.InterfaceBindingPlugin{
-		NetworkAttachmentDefinition: NetworkAttachmentDefinition,
+		NetworkAttachmentDefinition: getPasstNADNamespace() + "/" + networkBindingNADName,
 		SidecarImage:                os.Getenv(hcoutil.PasstImageEnvV),
 		Migration:                   &kubevirtcorev1.InterfaceBindingMigration{},
 		ComputeResourceOverhead: &kubevirtcorev1.ResourceRequirementsWithoutClaims{
@@ -64,6 +61,13 @@ func NetworkBinding() kubevirtcorev1.InterfaceBindingPlugin {
 			},
 		},
 	}
+}
+
+func getPasstNADNamespace() string {
+	if hcoutil.GetClusterInfo().IsOpenshift() {
+		return "openshift-cnv"
+	}
+	return "default"
 }
 
 // NewPasstBindingCNISA creates a ServiceAccount for the passt binding CNI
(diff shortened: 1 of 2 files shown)


controllers/handlers/passt/passt.go
```diff
@@ -66,6 +67,17 @@ func NetworkBinding() kubevirtcorev1.InterfaceBindingPlugin {
 	}
 }
 
+// getPasstNADNamespace determines the namespace for the NAD based on platform and HCO namespace.
+// On OpenShift, if HCO is deployed in openshift-cnv, the NAD is deployed there because
+// CNO deploys multus with access to that namespace. Otherwise, deploys to default namespace.
+func getPasstNADNamespace(hcoNamespace string) string {
```
oshoval: @orelmisan @nirdothan @EdDev 

If we focus on this function only,
Please let me know if you are fine with the logic

Motivation is to create the NAD on "openshift-cnv" ns on OpenShift only, instead creating it on "default" namespace.

The logic is checking we are on openshift, and if hco is deployed on "openshift-cnv",
because if it is not deployed there, we can't know for sure it is the classic bundle, with CNO that allow-list "openshift-cnv".
Unless we want to simplify it, and just check isOpenShift, if this all what we care about,
(under assumption, that the default CNO is used, and it allow list "openshift-cnv" as global ns).

It is instead adding a variable to HCO CR, that will says where the NAD should be created.
Less meta parameters and configuration is better imho.

Thanks

nirdothan: > The logic is checking we are on openshift, and if hco is deployed on "openshift-cnv"

I'm in favor of this ^^

nunnatsa: Just to remind you that we have the community releases, that may installed on OKD or OpenShift. In this case, the deployment will be on the `kubevirt-hyperconverged` namespace. @oshoval, don't you want to support this usecase as well?

oshoval: it depends if they still use offical CNO, that allow-list "openshift-cnv"
(CNO wont allow list other ns, nor needed, as just the NAD will be there, not HCO / passt itself, as far as i understand)
this is indeed the alternative, which is less strict but more robust

it means just to alter the condition to 

```
	if hcoutil.GetClusterInfo().IsOpenshift() {
		return "openshift-cnv"
	}
```

fine by me, if netcore accept it as well

controllers/operands/networkAttachmentDefinitionHandler.go
```diff
@@ -14,18 +14,19 @@ import (
 	"github.com/kubevirt/hyperconverged-cluster-operator/pkg/util"
 )
 
-type newNetworkAttachmentDefinitionFunc func(hc *hcov1beta1.HyperConverged) *netattdefv1.NetworkAttachmentDefinition
+type newNetworkAttachmentDefinitionFunc func(hc *hcov1beta1.HyperConverged, namespace string) *netattdefv1.NetworkAttachmentDefinition
```
nunnatsa: Maybe I miss something, but I don't understand why do we need this mess with this new function type. It can be dramatically simplified. The whole thing is just too complex.

controllers/handlers/passt/passt.go
```diff
@@ -32,11 +34,10 @@ const (
...
 	passtCNIObjectName = "passt-binding-cni"
 
-	networkBindingNADName       = "primary-udn-kubevirt-binding"
-	networkBindingNADNamespace  = "default"
-	NetworkAttachmentDefinition = networkBindingNADNamespace + "/" + networkBindingNADName
-
+	networkBindingNADName        = "primary-udn-kubevirt-binding"
 	bindingComputeMemoryOverhead = "250Mi"
+
+	OpenShiftCNVNamespace = "openshift-cnv"
```
nunnatsa: As someone else said: "DON'T". Please remove this string from this PR, including from the PR title.

oshoval:
(wip - please dont review yet)

controllers/handlers/passt/passt.go
oshoval: done like latest comment

controllers/operands/networkAttachmentDefinitionHandler.go
oshoval: it was all because we needed the namespace, but now we dont, it is way simpler indeed
done

oshoval:
addressed comments

last push - rebase

oshoval:
closing for now

PR #3635, opened 2025-07-20
Fix the upgrade test

nunnatsa:

In upgrade test, we are trying to run a command inside a VMI, using `virtctl ssh` this is now failing after bumping virtctl to 1.6.x, because there is a breaking change in `virtctl` that is no longer accept the vmi name, but also requires the type (`vmi/vmi-name`, intead of just `vmi-name`).

This PR fixes the issue by adding the missing type to the virtctl ssh command.

**Jira Ticket**:

```jira-ticket
None
```

**Release note**:
```release-note
None
```

### DIFF
--- a/hack/check-uptime.sh
+++ b/hack/check-uptime.sh
@@ -5,7 +5,7 @@ function check_uptime() {
   timeout=$2
 
   for _ in $(seq 1 $retries); do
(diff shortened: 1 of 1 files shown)


oshoval:
Thanks !
/lgtm

PR #3661, opened 2025-07-27
csv-merger tool: dump network policies also with CRD

nunnatsa:

The `--dump-network-policies` flag also works when the `--outputMode` is set to `CRDs`.

The meaning is that we can dump the Network policies either with the CSV, as done until now, or with the CRD.

In addition, added the new `NPs` mode to the csv-merger, to only drop the network policies, and not the CRD or the CSV. This is going to be used for CI.

Also, fixed a bug in where the network policies are without namespace. This is not so bad, because OLM overrides it anyway with the installed namespace. but if will want to deploy them manually or without olm, this
is needed.

This PR also removes the `temp_manifests.yaml` that was accidentally added to git, and made sure it won't be added again.

**Jira Ticket**:
```jira-ticket
https://issues.redhat.com/browse/CNV-63363
```

**Release note**:

```release-note
csv-merger tool: dump network policies also with CRD
```


tools/csv-merger/csv-merger.go
```diff
@@ -148,20 +149,20 @@ func main() {
 
 	switch *outputMode {
 	case CRDMode:
-		panicOnError(util.MarshallObject(components.GetOperatorCRD(*apiSources), os.Stdout))
+		crd, err := components.GetOperatorCRD(*apiSources)
+		panicOnError(err, "failed to generate CRD for")
```
orenc1: shouldn't the error message specify for which component the CRD was failed being generated?
if we panic on error here, the next line won't be executed.

nunnatsa: I think the word "for" (probably added by some ai and I missed it) should be dropped because it's confusing.

This is our CRD, not any other component, and it's ok to fail here - this is by design and also what we always did here. I just prefer to exit from main and not from a package (bad practice). That why I modified `components.GetOperatorCRD` to return also error, and now the code checks it explicitly.

tools/csv-merger/csv-merger.go
```diff
@@ -478,13 +479,10 @@ func addRelatedImage(images []csvv1alpha1.RelatedImage, image string) []csvv1alp
...
 	if err != nil {
-		moreInfo := ""
-		if len(info) > 0 {
-			moreInfo = strings.Join(info, " ")
-		}
+		moreInfo := strings.Join(info, " ")
 
 		log.Println("Error!", err, moreInfo)
-		panic(err)
+		os.Exit(1)
```
orenc1: If we're no longer panicking on error, maybe it would be better to rename `panicOnError` to something like `exitOnErrror` ?
btw, with panic we're printing the stack trace for easier debug, and run deferred functions, and in os.Exit we're just terminating immediately. why we don't want to use panic any more?
also, in panic the exit code was 2, and here you're changing it to 1.

tools/manifest-templator/manifest-templator.go
```diff
@@ -82,7 +82,7 @@ var (
 func check(err error) {
 	if err != nil {
 		log.Println("error: ", err)
-		panic(err)
+		os.Exit(1)
```
orenc1: since we have here a defer function above the caller, the last comment is more significant

tools/manifest-templator/manifest-templator.go
```diff
@@ -438,10 +438,13 @@ func writeOperatorCR() {
...
 
 func writeOperatorCRD() {
+	crd, err := components.GetOperatorCRD(*apiSources)
+	check(err)
+
 	operatorCrd, err := os.Create(path.Join(*deployDir, "crds/hco.crd.yaml"))
 	check(err)
 	defer operatorCrd.Close()
-	check(util.MarshallObject(components.GetOperatorCRD(*apiSources), operatorCrd))
+	check(util.MarshallObject(crd, operatorCrd))
```
orenc1: now the distinction between `crd` and `operatorCrd` is not clear on first glance.
i would suggest to rename `operatorCrd` to `crdFile`, to indicate it's a file.

orenc1 (approved):
/approve

PR #3662, opened 2025-07-27
passt: Add e2e

oshoval:
Add Passt e2e test.
Checks only what we deploy both upstream and downstream (i.e daemonSet and NAD).

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket

```

**Release note**:

```release-note
None
```


tests/func-tests/passt_test.go
```diff
@@ -0,0 +1,169 @@
...
+			enablePasstDeployment(ctx, k8scli)
+
+			By("checking that the passt DaemonSet is created and ready")
+			Eventually(func(g Gomega, ctx context.Context) {
+				ds := &appsv1.DaemonSet{}
+				err := k8scli.Get(ctx, types.NamespacedName{
+					Name:      passtCNIObjectName,
+					Namespace: tests.InstallNamespace,
+				}, ds)
+				g.Expect(err).ToNot(HaveOccurred())
```
nunnatsa: nit:

```suggestion
				g.Expect(k8scli.Get(ctx, types.NamespacedName{
					Name:      passtCNIObjectName,
					Namespace: tests.InstallNamespace,
				}, ds)).To(Succeed())
```

tests/func-tests/passt_test.go
```diff
@@ -0,0 +1,169 @@
...
+	tests.FlagParse()
+	var (
+		k8scli client.Client
+	)
+
+	BeforeEach(func(ctx context.Context) {
+		k8scli = tests.GetControllerRuntimeClient()
+
+		err := netattdefv1.AddToScheme(k8scli.Scheme())
+		Expect(err).ToNot(HaveOccurred())
```
nunnatsa: nit: 
```suggestion
		Expect(netattdefv1.AddToScheme(k8scli.Scheme())).To(Succeed())
```

tests/func-tests/passt_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+			By("checking that the passt NetworkAttachmentDefinition is created")
+			Eventually(func(g Gomega, ctx context.Context) {
+				nad := &netattdefv1.NetworkAttachmentDefinition{}
+				err := k8scli.Get(ctx, types.NamespacedName{
+					Name:      networkBindingNADName,
+					Namespace: networkBindingNADNamespace,
+				}, nad)
+				g.Expect(err).ToNot(HaveOccurred())
+				g.Expect(nad.Spec.Config).ToNot(BeEmpty())
```
nunnatsa, line 69: Is this enough?

tests/func-tests/passt_test.go
```diff
@@ -0,0 +1,169 @@
...
+				Should(Succeed())
+
+			By("checking that the passt NetworkAttachmentDefinition is created")
+			Eventually(func(g Gomega, ctx context.Context) {
+				nad := &netattdefv1.NetworkAttachmentDefinition{}
+				err := k8scli.Get(ctx, types.NamespacedName{
+					Name:      networkBindingNADName,
+					Namespace: networkBindingNADNamespace,
+				}, nad)
+				g.Expect(err).ToNot(HaveOccurred())
```
nunnatsa: nit:
```suggestion
				g.Expect(k8scli.Get(ctx, types.NamespacedName{
					Name:      networkBindingNADName,
					Namespace: networkBindingNADNamespace,
				}, nad)).To(Succeed())
```

tests/func-tests/passt_test.go
```diff
@@ -0,0 +1,169 @@
...
+			enablePasstDeployment(ctx, k8scli)
+
+			By("waiting for resources to be created")
+			Eventually(func(g Gomega, ctx context.Context) {
+				ds := &appsv1.DaemonSet{}
+				err := k8scli.Get(ctx, types.NamespacedName{
+					Name:      passtCNIObjectName,
+					Namespace: tests.InstallNamespace,
+				}, ds)
+				g.Expect(err).ToNot(HaveOccurred())
```
nunnatsa: nit: same

tests/func-tests/passt_test.go
```diff
@@ -0,0 +1,169 @@
...
+				}, nad)
+				g.Expect(err).ToNot(HaveOccurred())
+				g.Expect(nad.Spec.Config).ToNot(BeEmpty())
+			}).WithTimeout(2 * time.Minute).
+				WithPolling(5 * time.Second).
+				WithContext(ctx).
+				Should(Succeed())
+		})
+
+		It("should remove passt resources when annotation is disabled", func(ctx context.Context) {
```
nunnatsa, line 76: Maybe we can add it as steps to the prev test, to avoid duplications?

tests/func-tests/passt_test.go
```diff
@@ -0,0 +1,169 @@
...
+		}
+
+		if value == "false" {
+			delete(hco.Annotations, passt.DeployPasstNetworkBindingAnnotation)
+		} else {
+			hco.Annotations[passt.DeployPasstNetworkBindingAnnotation] = value
+		}
+
+		_, err := tests.UpdateHCO(ctx, cli, hco)
+		return err
```
nunnatsa: Please use patch instead of update

nunnatsa (commented):
Thank for this PR @oshoval! Added some inline comment.

tests/func-tests/passt_test.go
oshoval: yes, we just need to see it is deployed, no need details

tests/func-tests/passt_test.go
oshoval: aint a test need to do one thing ? 
anyhow whatever you prefer
it will just make the test do more / be longer, less maintainable
those tests are pretty fast, they dont spin vm (20s for both)

oshoval:
thanks, addressed comments

tests/func-tests/passt_test.go
nunnatsa: ok

tests/func-tests/passt_test.go
```diff
@@ -0,0 +1,162 @@
...
+		_ = tests.PatchHCO(ctx, cli, patchBytes)
+	} else {
+		patchBytes := []byte(fmt.Sprintf(`[{"op": "replace", "path": "/metadata/annotations/hco.kubevirt.io~1deployPasstNetworkBinding", "value": "%s"}]`, value))
+		err := tests.PatchHCO(ctx, cli, patchBytes)
+		if err != nil {
+			ensurePatch := []byte(fmt.Sprintf(`[
+				{"op": "add", "path": "/metadata/annotations", "value": {}},
+				{"op": "replace", "path": "/metadata/annotations/hco.kubevirt.io~1deployPasstNetworkBinding", "value": "%s"}
+			]`, value))
+			Expect(tests.PatchHCO(ctx, cli, ensurePatch)).To(Succeed())
```
nunnatsa: Please avoid errors in helpers. Return error and check it when calling this function.

tests/func-tests/passt_test.go
```diff
@@ -0,0 +1,162 @@
...
+		if value == "false" {
+			g.Expect(hco.Annotations).ToNot(HaveKey(passt.DeployPasstNetworkBindingAnnotation))
+		} else {
+			g.Expect(hco.Annotations).To(HaveKeyWithValue(passt.DeployPasstNetworkBindingAnnotation, value))
+		}
+	}).WithTimeout(30 * time.Second).
+		WithPolling(time.Second).
+		WithContext(ctx).
+		Should(Succeed())
+}
```
nunnatsa: not sure we need this check, if we exit without error.

tests/func-tests/passt_test.go
nunnatsa: I think we can do it nicer using merge path instead of json patch. for that, you'll need to add the following to tests/func-tests/util.go
```go
// PatchMergeHCO updates the HCO CR using a DynamicClient, it can return errors on failures
func PatchMergeHCO(ctx context.Context, cli client.Client, patchBytes []byte) error {
	patch := client.RawPatch(types.StrategicMergePatchType, patchBytes)
	hco := &v1beta1.HyperConverged{
			Name:      hcoutil.HyperConvergedName,
			Namespace: InstallNamespace,
		},
	}

	return cli.Patch(ctx, hco, patch)
}
```

Then, you could simplify the code here with a single path, with merge path like:
(*) as a bonus, we won't need to escape the '/', so we could use the constant
```go
patchBytes := []byte(fmt.Sprintf(`{"metadata":{"annotations":{%q: "true"}}}`, passt.DeployPasstNetworkBindingAnnotation))

return tests.PatchMergeHCO(ctx, cli, patchBytes)
```
WDYT?

tests/func-tests/passt_test.go
```diff
@@ -0,0 +1,162 @@
...
+func disablePasstDeployment(ctx context.Context, cli client.Client) {
+	By("disabling the passt deployment annotation")
+	setPasstDeploymentAnnotation(ctx, cli, "false")
+}
+
+func setPasstDeploymentAnnotation(ctx context.Context, cli client.Client, value string) {
+	if value == "false" {
+		patchBytes := []byte(`[{"op": "remove", "path": "/metadata/annotations/hco.kubevirt.io~1deployPasstNetworkBinding"}]`)
+		// Try to remove, ignore errors if path doesn't exist
+		_ = tests.PatchHCO(ctx, cli, patchBytes)
```
nunnatsa: We can use the merge patch here as well, to allow multiple calls:
```go
patchBytes := []byte(fmt.Sprintf(`{"metadata":{"annotations":{%q: null}}}`, passt.DeployPasstNetworkBindingAnnotation))

return tests.PatchMergeHCO(ctx, cli, patchBytes)
```

tests/func-tests/passt_test.go
oshoval: nice, on it, note please that HCO CRD doesnt support `StrategicMergePatchType`, so using `MergePatchType`

tests/func-tests/passt_test.go
oshoval: tried (removed all over), the test doesnt spin trying if we remove it

oshoval:
addressed comments

tests/func-tests/passt_test.go
oshoval: refactored in general so it was removed on the way

tests/func-tests/passt_test.go
```diff
@@ -0,0 +1,159 @@
...
+			Expect(setPasstDeploymentAnnotation(ctx, k8scli, "false")).To(Succeed())
+
+			By("checking that the passt DaemonSet is removed")
+			Eventually(func(g Gomega, ctx context.Context) {
+				ds := &appsv1.DaemonSet{}
+				err := k8scli.Get(ctx, types.NamespacedName{
+					Name:      passtCNIObjectName,
+					Namespace: tests.InstallNamespace,
+				}, ds)
+				g.Expect(errors.IsNotFound(err)).To(BeTrue())
```
nunnatsa: Please change:
```suggestion
				g.Expect(err).To(MatchError(errors.IsNotFound, "error should be NotFound"))
```

The `BeTrue()` matcher error message is very crypt.

tests/func-tests/passt_test.go
```diff
@@ -0,0 +1,159 @@
...
+	})
+})
+
+func setPasstDeploymentAnnotation(ctx context.Context, cli client.Client, value string) error {
+	var annotationValue string
+	if value == "false" {
+		annotationValue = "null"
+	} else {
+		annotationValue = fmt.Sprintf(`"%s"`, value)
+	}
```
nunnatsa: Let's use boolean value. It feels more natural to use boolean in condtions. 

Also, if the value is true, then we don't need to compute it on runtime. we know the answer for this

WDYT about:
```suggestion
func setPasstDeploymentAnnotation(ctx context.Context, cli client.Client, value bool) error {
	var annotationValue string
	if value {
		annotationValue = "true"
	} else {
		annotationValue = "null"
	}
```

tests/func-tests/passt_test.go
```diff
@@ -0,0 +1,159 @@
...
+		if value == "false" {
+			g.Expect(hco.Annotations).ToNot(HaveKey(passt.DeployPasstNetworkBindingAnnotation))
+		} else {
+			g.Expect(hco.Annotations).To(HaveKeyWithValue(passt.DeployPasstNetworkBindingAnnotation, value))
+		}
+	}).WithTimeout(30 * time.Second).
+		WithPolling(time.Second).
+		WithContext(ctx).
+		Should(Succeed())
+
```
nunnatsa: Still not sure why we need it

oshoval: https://github.com/kubevirt/hyperconverged-cluster-operator/pull/3662#discussion_r2235728625

it doesnt spin waiting without it, and test will do nothing and just pass

nunnatsa: As far as I understand, this id not a test but a helper function. Do I miss something?

oshoval: still we use it, so we cant allow it to not spin until it gets the result, it is unsafe
we saw it is problematic pattern, there is no reason to code it wrong, it doesnt matter it is not a test, it is still a code snippet that we use


without it, it isnt eventually, it is just one timer try, the Should is the termination that tells it what it should reach, hence loop

nunnatsa: > still we use it, so we cant allow it to not spin until it gets the result

But we do get result. if `tests.PatchMergeHCO()` returns error, then it didn't happen.

Anyway, I won't fight over it.

I'm just trying to get rid of the `if value == "false"`s - I think that if you pass a boolean to a function, and then you ask about it, maybe you need two functions.

oshoval: wont this be enough ?
https://github.com/kubevirt/hyperconverged-cluster-operator/pull/3662#discussion_r2236739850

nunnatsa: I'm not following

oshoval: sorry i might mixed, i will see what i can do about this specific condition, cleanest will be just to split the function it seems indeed

oshoval:
addressed comments, some became outdated because split to 2 functions as you suggested, 
so lets please review the newest version

If we remove the last `GetHCO` part (it is just extra precaution), we can declare the `patchBytes` as const arguments and reduce duplication, i believe this might you wanted to strive to,
EDIT  - will push it soon

tests/func-tests/passt_test.go
oshoval: regarding the latest changes
see please https://github.com/kubevirt/hyperconverged-cluster-operator/pull/3662#issuecomment-3131866957

oshoval:
Thanks,
if you prefer i removed the functions totally as described [here](https://github.com/kubevirt/hyperconverged-cluster-operator/pull/3662#issuecomment-3131866957), you can see it here https://termbin.com/cplz
Basically just calling
`Expect(tests.PatchMergeHCO(ctx, k8scli, []byte(setPasstAnnotationPatch))).To(Succeed())`
where the bytes are declared as const (and one for unset)

tests/func-tests/passt_test.go
```diff
@@ -0,0 +1,172 @@
...
+	AfterEach(func(ctx context.Context) {
+		By("disabling the passt deployment annotation")
+		Expect(unsetPasstDeploymentAnnotation(ctx, k8scli)).To(Succeed())
+	})
+
+	When("deployPasstNetworkBinding annotation is set to true", func() {
+		It("should create the passt DaemonSet and NetworkAttachmentDefinition", func(ctx context.Context) {
+			By("enabling the passt deployment annotation")
+			Expect(setPasstDeploymentAnnotation(ctx, k8scli)).To(Succeed())
```
nunnatsa: @oshoval ,This should be probably get into Eventually. It fails on e2e.

Please do the same for all set/unsent.

/hold

oshoval:
addressed comment

oshoval:
fixed lint

PR #3664, opened 2025-07-28
csv-merger and manifest-templator: dump pre-built CRD

nunnatsa:

Both the `csv-merger` and the `manifest-templator` tools can generate the HyperConverged CRD. However, the way they are implemented, requires them to run from inside the the source directory, and they are fails to produce the CRD if running elsewhere, e.g. from a container.

This PR is all about fixing this, by including the CRD yaml in the tools binary, and so they can just write it to stdout, without the need to actually generate it.

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-63364
```

**Release note**:

```release-note
None
```


orenc1 (approved):
/approve

PR #3672, opened 2025-07-30
network addons: Override ipam-ext default network NAD namespace

ormergi:
kubevirt-ipam-controller reads OVN-Kubernetes default network NAD. 
When running on OCP, the NAD exist in diffrent namespace then ipam-ext default.

Override the NAD target namespace.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket

```

**Release note**:

```release-note
Override kubevirt-ipam-controllers default network NAD namespace when running on Openshift
```


controllers/handlers/networkAddons.go
```diff
@@ -139,12 +139,16 @@ func setDeployOvsAnnotation(req *common.HcoRequest, found *networkaddonsv1.Netwo
 }
 
 func NewNetworkAddons(hc *hcov1beta1.HyperConverged) (*networkaddonsv1.NetworkAddonsConfig, error) {
+	ipam := &networkaddonsshared.KubevirtIpamController{}
+	if util.GetClusterInfo().IsOpenshift() {
+		ipam = &networkaddonsshared.KubevirtIpamController{DefaultNetworkNADNamespace: "openshift-ovn-kubernetes"}
```
nunnatsa: Can we just populate the field, instead of creating a new instance? e.g.
```suggestion
		ipam.DefaultNetworkNADNamespace = "openshift-ovn-kubernetes"
```

controllers/handlers/networkAddons_test.go
```diff
@@ -697,6 +697,17 @@ var _ = Describe("CNA Operand", func() {
 			)
 		})
 
+		It("when running on openshift, it should override kubernetes-ipam-controller default network NAD namespace", func() {
+			getClusterInfo := hcoutil.GetClusterInfo
+			hcoutil.GetClusterInfo = func() hcoutil.ClusterInfo { return &commontestutils.ClusterInfoMock{} }
+			defer func() { hcoutil.GetClusterInfo = getClusterInfo }()
```
nunnatsa, line 703: Please use ginkgo's `DeferCleanup` instead of a regular defer. I think it's safer (not sure, though).

ormergi: In this case there is no complex tests container or test setup involving BeforeEach (or similar Before*), using the standard defer here is actually more predictable then Ginkgo's. WDYT?

controllers/handlers/networkAddons.go
ormergi: Done

controllers/handlers/networkAddons_test.go
nunnatsa: I won't hold the PR for that...

PR #3673, opened 2025-07-30
Fix multi-arch image build

nunnatsa:
Previous commit broke the multi-arch image build, by adding the `generate` make target as a condition to building the operator. This is not working in multi-arch image building.

Removed the wrong condition to fix the issue.

**Release note**:

```release-note
None
```


nunnatsa:
This is a critical fix. Merging now.

PR #3674, opened 2025-07-31
Customize the `ssp.kubevirt.io/dict.architectures` DICT annotation

nunnatsa:

On a multi-architecture cluster, if the user customized a common (pre-defined) DICT, we should use the same value for the `enableMultiArchBootImageImport` DICT annotation, as in the original DICT, because this annotation describes the image itself. However, if the the image itself was customized, then we have no knowledge about its supported architectures, and e must trust the customized annotation.

This PR changes the behavior of DICT customization:

If the HyperConverged CR contains a customized common DICT, and the `enableMultiArchBootImageImport` feature gate is set to `true`, customize the DICT `ssp.kubevirt.io/dict.architectures` annotation, as follow:

if DICT registry was not customized, use the original common DICT annotation in the result DICT, 
or if it's missing from the common DICT, remove it from the result DICT.

if DICT registry was customized, use the customized DICT annotation in the result DICT, or if it's
missing from the customized DICT, remove it from the result DITC.

**Release note**:

```release-note
Fix the behavior regarding the `ssp.kubevirt.io/dict.architectures` DICT annotation
```

### DIFF
--- a/controllers/handlers/ssp.go
+++ b/controllers/handlers/ssp.go
@@ -284,6 +284,7 @@ func getDataImportCronTemplates(hc *hcov1beta1.HyperConverged) ([]hcov1beta1.Dat
 }
 
 func getCommonDicts(list []hcov1beta1.DataImportCronTemplateStatus, crDicts map[string]hcov1beta1.DataImportCronTemplate, hc *hcov1beta1.HyperConverged) []hcov1beta1.DataImportCronTemplateStatus {
+	enableMultiArchBootImageImport := ptr.Deref(hc.Spec.FeatureGates.EnableMultiArchBootImageImport, false)
 	for dictName, commonDict := range dataImportCronTemplateHardCodedMap {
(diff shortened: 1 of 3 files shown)


controllers/handlers/ssp.go
```diff
@@ -316,20 +317,52 @@ func customizeCommonDICT(targetDict *hcov1beta1.DataImportCronTemplateStatus, cr
...
+			targetDict.Annotations = maps.Clone(crDictAnnotations)
+		} else {
+			maps.Copy(targetDict.Annotations, crDictAnnotations)
+		}
+	}
+	if enableMultiArchBootImageImport && registryModified {
+		_, ext := crDictAnnotations[MultiArchDICTAnnotation]
+		if !ext {
+			delete(targetDict.Annotations, MultiArchDICTAnnotation)
+		}
```
machadovilaca, line 362: nit: we should avoid deep nesting

PR #3680, opened 2025-08-01
Adding ClusterRole and ClusterRoleBinding handlers

enp0s3:
Signed-off-by: Igor Bezukh <ibezukh@redhat.com>

I've decided to split the work of adding the wasp-agent deployment into few separate PRs.
This PR code prepares the functionality to deploy cluster roles for the wasp-agent SA, 
to that the wasp-agent pods will be able to list and watch all the pods in the cluster.

The handlers are simple, without cache and mutex, because for my case the HCO API won't infuence
the configuration of the Roles, so there is no need to implement the cache and to recalculate the desired configuration
when HCO CR is being changed.

**What this PR does / why we need it**:
enhance the HCO handler functionality so that it will be able to prepare the relevant roles and role-bindings for operands that need cluster-wide access to kube resources.

**Reviewer Checklist**

> Reviewers are supposed to review the PR for every aspect below one by one. To check an item means the PR is either "OK" or "Not Applicable" in terms of that item. All items are supposed to be checked before merging a PR. 

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
- [ ] Troubleshooting Friendly

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-57781
```

**Release note**:

```release-note
NONE
```

PR #3682, opened 2025-08-03
Deploy wasp-agent using HCO

enp0s3:

wasp-agent is a node-local agent that is responsible to mimic
the kubernetes swap feature in out-of-tree manner. It will be used
until we will develop an operational approach to use kuberentes swap

wasp-agent is composed of the following components
- Daemonset that deploys the node-local agent on the worker nodes
- Service account
- Cluster role and cluster role binding
- Security Context Constraints

wasp-agent deployment is conditional, the deployment is being
triggered as far as the memory overcommit percentage is higher
than 100%. In that case in order to make the memory overcommit safer
wasp-agent allows memory tiering with swap so that workload idle
memory would be able to swap instead of OOMKilled.

wasp-agent needs access to pods in all namespaces, therefore
it is required to use the cluster role.

since wasp-agent mimics the kubelet functionality, it is required
to use the privileged SCC in order to be able to function properly on
the node (configuring the cgroups, talk to CRI-O, etc.)

wasp-agent needs to watch all the running pods in the cluster,
therefore it uses the allow-access-cluster-services network
policy in order to communicate with the API server.

wasp-agent currently supported only on OpenShift.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-57781
```

**Release note**:

```release-note
Deploy wasp-agent using HCO
```

### DIFF
--- a/controllers/handlers/wasp-agent/daemonset.go
+++ b/controllers/handlers/wasp-agent/daemonset.go
@@ -1,0 +1,200 @@
+package wasp_agent
+
+import (
+	"maps"
+	"os"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+
+	hcov1beta1 "github.com/kubevirt/hyperconverged-cluster-operator/api/v1beta1"
+	"github.com/kubevirt/hyperconverged-cluster-operator/controllers/operands"
+	"github.com/kubevirt/hyperconverged-cluster-operator/pkg/nodeinfo"
+	hcoutil "github.com/kubevirt/hyperconverged-cluster-operator/pkg/util"
+)
+
+const (
+	clusterRoleName             = "wasp-cluster"
+	verbosity                   = "1"
+	AppComponentWaspAgent       = "wasp-agent"
+	waspAgentServiceAccountName = "wasp"
+	waspAgentSCCName            = "wasp"
+	NoOverCommitPercentage      = 100
+)
+
+var (
+	resourceCPU    = resource.MustParse("100m")
+	resourceMemory = resource.MustParse("50M")
+)
+
+func NewWaspAgentDaemonSetHandler(Client client.Client, Scheme *runtime.Scheme) operands.Operand {
+	return operands.NewConditionalHandler(
+		operands.NewDaemonSetHandler(Client, Scheme, newWaspAgentDaemonSet),
+		shouldDeployWaspAgent,
+		func(hc *hcov1beta1.HyperConverged) client.Object {
+			return NewWaspAgentWithNameOnly(hc)
+		},
+	)
+}
+
+func NewWaspAgentWithNameOnly(hc *hcov1beta1.HyperConverged) *appsv1.DaemonSet {
+	return &appsv1.DaemonSet{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      AppComponentWaspAgent,
+			Labels:    operands.GetLabels(hc, AppComponentWaspAgent),
+			Namespace: hc.Namespace,
+		},
+	}
+}
+
+func newWaspAgentDaemonSet(hc *hcov1beta1.HyperConverged) *appsv1.DaemonSet {
+	waspImage, _ := os.LookupEnv(hcoutil.WaspAgentImageEnvV)
+
+	podLabels := operands.GetLabels(hc, AppComponentWaspAgent)
+	podLabels[hcoutil.AllowEgressToDNSAndAPIServerLabel] = "true"
+	podLabels["name"] = AppComponentWaspAgent
+
+	container := corev1.Container{
+		Name:            AppComponentWaspAgent,
+		Image:           waspImage,
+		ImagePullPolicy: corev1.PullIfNotPresent,
+		Resources: corev1.ResourceRequirements{
+			Requests: corev1.ResourceList{
+				corev1.ResourceCPU:    resourceCPU,
+				corev1.ResourceMemory: resourceMemory,
+			},
+		},
+		SecurityContext: &corev1.SecurityContext{
+			Privileged: ptr.To(true),
+		},
+		VolumeMounts: []corev1.VolumeMount{
+			{
+				Name:      "host",
+				MountPath: "/host",
+			},
+			{
+				Name:      "rootfs",
+				MountPath: "/rootfs",
+			},
+		},
+	}
+	container.Env = createDaemonSetEnvVar()
+
+	spec := appsv1.DaemonSetSpec{
+		Selector: &metav1.LabelSelector{
+			MatchLabels: map[string]string{
+				"name": AppComponentWaspAgent,
+			},
+		},
+		Template: corev1.PodTemplateSpec{
+			ObjectMeta: metav1.ObjectMeta{
+				Labels: podLabels,
+			},
+			Spec: corev1.PodSpec{
+				ServiceAccountName:            waspAgentServiceAccountName,
+				HostPID:                       true,
+				HostUsers:                     ptr.To(true),
+				TerminationGracePeriodSeconds: ptr.To[int64](5),
+				Containers:                    []corev1.Container{container},
+				Volumes: []corev1.Volume{
+					{
+						Name: "host",
+						VolumeSource: corev1.VolumeSource{
+							HostPath: &corev1.HostPathVolumeSource{
+								Path: "/",
+							},
+						},
+					},
+					{
+						Name: "rootfs",
+						VolumeSource: corev1.VolumeSource{
+							HostPath: &corev1.HostPathVolumeSource{
+								Path: "/",
+							},
+						},
+					},
+				},
+				PriorityClassName: "system-node-critical",
+			},
+		},
+	}
+
+	ds := NewWaspAgentWithNameOnly(hc)
+	ds.Spec = spec
+
+	if hc.Spec.Infra.NodePlacement != nil {
+		if hc.Spec.Infra.NodePlacement.NodeSelector != nil {
+			ds.Spec.Template.Spec.NodeSelector = maps.Clone(hc.Spec.Infra.NodePlacement.NodeSelector)
+		}
+
+		if hc.Spec.Infra.NodePlacement.Affinity != nil {
+			ds.Spec.Template.Spec.Affinity = hc.Spec.Infra.NodePlacement.Affinity.DeepCopy()
+		}
+
+		if hc.Spec.Infra.NodePlacement.Tolerations != nil {
+			ds.Spec.Template.Spec.Tolerations = make([]corev1.Toleration, len(hc.Spec.Infra.NodePlacement.Tolerations))
+			copy(ds.Spec.Template.Spec.Tolerations, hc.Spec.Infra.NodePlacement.Tolerations)
+		}
+	} else {
+		affinity := getPodAntiAffinity(ds.Labels[hcoutil.AppLabelComponent], nodeinfo.IsInfrastructureHighlyAvailable())
+		ds.Spec.Template.Spec.Affinity = affinity
+	}
+
+	return ds
+}
+
+func createDaemonSetEnvVar() []corev1.EnvVar {
+	return []corev1.EnvVar{
+		{
+			Name:  "VERBOSITY",
+			Value: verbosity,
+		},
+		{
+			Name: "NODE_NAME",
+			ValueFrom: &corev1.EnvVarSource{
+				FieldRef: &corev1.ObjectFieldSelector{
+					FieldPath: "spec.nodeName",
+				},
+			},
+		},
+	}
+}
+
+func shouldDeployWaspAgent(hc *hcov1beta1.HyperConverged) bool {
+	overcommitPercentage := hc.Spec.HigherWorkloadDensity.MemoryOvercommitPercentage
+	return overcommitPercentage > NoOverCommitPercentage
+}
+
+func getPodAntiAffinity(componentLabel string, infrastructureHighlyAvailable bool) *corev1.Affinity {
+	if infrastructureHighlyAvailable {
+		return &corev1.Affinity{
+			PodAntiAffinity: &corev1.PodAntiAffinity{
+				PreferredDuringSchedulingIgnoredDuringExecution: []corev1.WeightedPodAffinityTerm{
+					{
+						Weight: 90,
+						PodAffinityTerm: corev1.PodAffinityTerm{
+							LabelSelector: &metav1.LabelSelector{
+								MatchExpressions: []metav1.LabelSelectorRequirement{
+									{
+										Key:      hcoutil.AppLabelComponent,
+										Operator: metav1.LabelSelectorOpIn,
+										Values:   []string{componentLabel},
+									},
+								},
+							},
+							TopologyKey: corev1.LabelHostname,
+						},
+					},
+				},
+			},
+		}
+	}
+
+	return nil
+}
--- a/controllers/handlers/wasp-agent/rbac.go
+++ b/controllers/handlers/wasp-agent/rbac.go
@@ -1,0 +1,104 @@
+package wasp_agent
+
+import (
+	log "github.com/go-logr/logr"
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	hcov1beta1 "github.com/kubevirt/hyperconverged-cluster-operator/api/v1beta1"
+	"github.com/kubevirt/hyperconverged-cluster-operator/controllers/operands"
+)
+
+func NewWaspAgentClusterRoleBindingHandler(
+	_ log.Logger, Client client.Client, Scheme *runtime.Scheme, hc *hcov1beta1.HyperConverged) (operands.Operand, error) {
+	return operands.NewConditionalHandler(
+		operands.NewClusterRoleBindingHandler(Client, Scheme, newWaspAgentClusterRoleBinding(hc)),
+		shouldDeployWaspAgent,
+		func(hc *hcov1beta1.HyperConverged) client.Object {
+			return newWaspAgentClusterRoleBindingWithNameOnly(hc)
+		},
+	), nil
+}
+
+func newWaspAgentClusterRoleBinding(hc *hcov1beta1.HyperConverged) *rbacv1.ClusterRoleBinding {
+	crb := newWaspAgentClusterRoleBindingWithNameOnly(hc)
+
+	crb.RoleRef = rbacv1.RoleRef{
+		Kind:     "ClusterRole",
+		Name:     clusterRoleName,
+		APIGroup: "rbac.authorization.k8s.io",
+	}
+
+	crb.Subjects = []rbacv1.Subject{
+		{
+			Kind:      "ServiceAccount",
+			Name:      waspAgentServiceAccountName,
+			Namespace: hc.Namespace,
+		},
+	}
+
+	return crb
+}
+
+func newWaspAgentClusterRoleBindingWithNameOnly(hc *hcov1beta1.HyperConverged) *rbacv1.ClusterRoleBinding {
+	return &rbacv1.ClusterRoleBinding{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: rbacv1.SchemeGroupVersion.String(),
+			Kind:       "ClusterRoleBinding",
+		},
+		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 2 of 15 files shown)


controllers/handlers/wasp-agent/rbac.go
```diff
@@ -0,0 +1,113 @@
...
+)
+
+func NewClusterRoleBindingHandler(Client client.Client, Scheme *runtime.Scheme) operands.Operand {
+	return operands.NewConditionalHandler(
+		operands.NewClusterRoleBindingHandler(Client, Scheme, newClusterRoleBinding),
+		shouldDeployWaspAgent,
+		func(hc *hcov1beta1.HyperConverged) client.Object {
+			return newClusterRoleBindingWithNameOnly(hc)
+		},
+	)
```
nunnatsa: Why do we need this function? Can't we just pass a RoleBinding object?

nunnatsa: e.g.
```suggestion
func NewClusterRoleBindingHandler(Client client.Client, Scheme *runtime.Scheme) operands.Operand {
    crb := newClusterRoleBindingWithNameOnly(hc)
	return operands.NewConditionalHandler(
		operands.NewClusterRoleBindingHandler(Client, Scheme, newClusterRoleBinding),
		shouldDeployWaspAgent,
		crb,
	)
```

controllers/handlers/wasp-agent/rbac.go
```diff
@@ -0,0 +1,113 @@
...
+				Namespace: hc.Namespace,
+			},
+		},
+	}
+}
+
+func newClusterRoleBindingWithNameOnly(hc *hcov1beta1.HyperConverged) *rbacv1.ClusterRoleBinding {
+	return &rbacv1.ClusterRoleBinding{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "rbac.authorization.k8s.io/v1",
```
nunnatsa: If we can take the original const from rbacv1? i.e.
```suggestion
			APIVersion: rbacv1.SchemeGroupVersion.String(),
```

controllers/handlers/wasp-agent/rbac.go
```diff
@@ -0,0 +1,113 @@
...
+		func(hc *hcov1beta1.HyperConverged) client.Object {
+			return newClusterRoleBindingWithNameOnly(hc)
+		},
+	)
+}
+
+func newClusterRoleBinding(hc *hcov1beta1.HyperConverged) *rbacv1.ClusterRoleBinding {
+	return &rbacv1.ClusterRoleBinding{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "rbac.authorization.k8s.io/v1",
```
nunnatsa: Can we use `rbacv1.SchemeGroupVersion.String()` instead?

controllers/handlers/wasp-agent/rbac.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+}
+
+func NewClusterRoleHandler(Client client.Client, Scheme *runtime.Scheme) operands.Operand {
+	return operands.NewConditionalHandler(
+		operands.NewClusterRoleHandler(Client, Scheme, newClusterRole),
+		shouldDeployWaspAgent,
+		func(hc *hcov1beta1.HyperConverged) client.Object {
+			return newClusterRoleWithNameOnly(hc)
+		},
```
nunnatsa: Same comment about the function. This is not needed.

controllers/handlers/wasp-agent/scc.go
```diff
@@ -0,0 +1,76 @@
...
+	hcov1beta1 "github.com/kubevirt/hyperconverged-cluster-operator/api/v1beta1"
+)
+
+func NewSecurityContextConstraintHandler(Client client.Client, Scheme *runtime.Scheme) operands.Operand {
+	return operands.NewConditionalHandler(
+		operands.NewSecurityContextConstraintsHandler(Client, Scheme, newSCC),
+		shouldDeployWaspAgent,
+		func(hc *hcov1beta1.HyperConverged) client.Object {
+			return newSCCWithNameOnly(hc)
+		},
```
nunnatsa: same

controllers/operandhandler/operandHandler.go
```diff
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"time"
 
+	wasp_agent "github.com/kubevirt/hyperconverged-cluster-operator/controllers/handlers/wasp-agent"
```
nunnatsa: should be `waspagent`

controllers/operandhandler/operandHandler.go
```diff
@@ -56,6 +58,11 @@ func NewOperandHandler(client client.Client, scheme *runtime.Scheme, ci hcoutil.
 		handlers.NewAAQHandler(client, scheme),
 		passt.NewPasstDaemonSetHandler(client, scheme),
 		passt.NewPasstNetworkAttachmentDefinitionHandler(client, scheme),
+		wasp_agent.NewServiceAccountHandler(client, scheme),
+		wasp_agent.NewClusterRoleHandler(client, scheme),
+		wasp_agent.NewClusterRoleBindingHandler(client, scheme),
+		wasp_agent.NewSecurityContextConstraintHandler(client, scheme),
+		wasp_agent.NewDaemonSetHandler(client, scheme),
 	}
```
nunnatsa: Should it always be deployed, or only on openshift?

controllers/handlers/wasp-agent/rbac.go
nunnatsa: ok got it. sorry

controllers/handlers/wasp-agent/rbac.go
nunnatsa: got it. ok.

hack/config
```diff
@@ -8,6 +8,7 @@ TTO_VERSION="v0.5.0"
 HPPO_VERSION="v0.22.0"
 HPP_VERSION="v0.22.0"
 AAQ_VERSION="v1.5.0"
+WASP_AGENT_VERSION="latest"
```
nunnatsa: Not sure it we'll work. We'll need real releases.

nunnatsa (changes requested):
Added some inline comment.

I mostly concern about the new image `latest` tag. This will cause issues.

We'll need real released, non-floating tag.

enp0s3:
> Added some inline comment.
> 
> I mostly concern about the new image `latest` tag. This will cause issues.
> 
> We'll need real released, non-floating tag.

@nunnatsa Thank you! Fixed. Using fixed version now.

controllers/handlers/wasp-agent/daemonset.go
```diff
@@ -0,0 +1,201 @@
...
+				},
+				PriorityClassName: "system-node-critical",
+			},
+		},
+	}
+
+	ds := NewWaspAgentWithNameOnly(hc)
+	ds.Spec = spec
+
+	affinity := getPodAntiAffinity(ds.Labels[hcoutil.AppLabelComponent], nodeinfo.IsInfrastructureHighlyAvailable())
```
nunnatsa: I think it should move done, just before line 147

controllers/handlers/wasp-agent/rbac.go
```diff
@@ -0,0 +1,113 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	hcov1beta1 "github.com/kubevirt/hyperconverged-cluster-operator/api/v1beta1"
+	"github.com/kubevirt/hyperconverged-cluster-operator/controllers/operands"
+)
+
+func NewClusterRoleBindingHandler(Client client.Client, Scheme *runtime.Scheme) operands.Operand {
+	return operands.NewConditionalHandler(
+		operands.NewClusterRoleBindingHandler(Client, Scheme, newClusterRoleBinding),
```
nunnatsa: This maybe a comment for #3680, but IIUC, the `newClusterRoleBinding` function will generate the same `ClusterRoleBinding` object again and again. There is nothing that will cause it to change any value in it.

That why I think the `clusterRoleHooks` should contain an object, and not a factory function. There is no need to call this function more than once.

i.e. we need to change operands/rbac.go like this:
```go
func NewClusterRoleBindingHandler(Client client.Client, Scheme *runtime.Scheme, cr *rbacv1.ClusterRoleBinding) *GenericOperand {
	return NewGenericOperand(Client, Scheme, "ClusterRoleBinding", &clusterRoleBindingHooks{required: cr}, false)
}

type clusterRoleBindingHooks struct {
	required *rbacv1.ClusterRoleBinding
}

func (h clusterRoleBindingHooks) GetFullCr(_ *hcov1beta1.HyperConverged) (client.Object, error) {
	return h.required.DeepCopy(), nil
}
```

And use it here like this:
```go
func NewClusterRoleBindingHandler(Client client.Client, Scheme *runtime.Scheme, hc *hcov1beta1.HyperConverged) operands.Operand {
	return operands.NewConditionalHandler(
		operands.NewClusterRoleBindingHandler(Client, Scheme, newClusterRoleBinding(hc)),
```

controllers/handlers/wasp-agent/rbac.go
```diff
@@ -0,0 +1,113 @@
...
+		shouldDeployWaspAgent,
+		func(hc *hcov1beta1.HyperConverged) client.Object {
+			return newClusterRoleBindingWithNameOnly(hc)
+		},
+	)
+}
+
+func newClusterRoleBinding(hc *hcov1beta1.HyperConverged) *rbacv1.ClusterRoleBinding {
```
nunnatsa: nit: Maybe we want more specific name here? e.g. `newWaspAgentClusterRoleBinding`

controllers/handlers/wasp-agent/rbac.go
```diff
@@ -0,0 +1,113 @@
...
+		},
+		Subjects: []rbacv1.Subject{
+			{
+				Kind:      "ServiceAccount",
+				Name:      waspAgentServiceAccountName,
+				Namespace: hc.Namespace,
+			},
+		},
+	}
+}
```
nunnatsa: we can reuse `newClusterRoleBindingWithNameOnly` here; for example:
```suggestion
	crb := newClusterRoleBindingWithNameOnly(hc)
	
	crb.RoleRef = rbacv1.RoleRef{
		Kind:     "ClusterRole",
		Name:     clusterRoleName,
		APIGroup: "rbac.authorization.k8s.io",
	}
	
	crb.Subjects = []rbacv1.Subject{
		{
			Kind:      "ServiceAccount",
			Name:      waspAgentServiceAccountName,
			Namespace: hc.Namespace,
		},
	}
	
	return crb
}
```
WDYT?

controllers/handlers/wasp-agent/rbac.go
```diff
@@ -0,0 +1,113 @@
...
+
+func newClusterRoleBindingWithNameOnly(hc *hcov1beta1.HyperConverged) *rbacv1.ClusterRoleBinding {
+	return &rbacv1.ClusterRoleBinding{
```
nunnatsa, line 46: nit: please consider a specific name.

controllers/handlers/wasp-agent/rbac.go
```diff
@@ -0,0 +1,113 @@
...
+	return operands.NewConditionalHandler(
+		operands.NewClusterRoleHandler(Client, Scheme, newClusterRole),
+		shouldDeployWaspAgent,
+		func(hc *hcov1beta1.HyperConverged) client.Object {
+			return newClusterRoleWithNameOnly(hc)
+		},
+	)
+}
+
+func newClusterRole(hc *hcov1beta1.HyperConverged) *rbacv1.ClusterRole {
```
nunnatsa: nit: please consider a specific name

controllers/handlers/wasp-agent/rbac.go
```diff
@@ -0,0 +1,113 @@
...
+	return &rbacv1.ClusterRole{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: rbacv1.SchemeGroupVersion.String(),
+			Kind:       "ClusterRole",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:   clusterRoleName,
+			Labels: operands.GetLabels(hc, AppComponentWaspAgent),
```
nunnatsa: Please reuse `newClusterRoleWithNameOnly`

controllers/handlers/wasp-agent/rbac.go
```diff
@@ -0,0 +1,113 @@
...
+		},
+	}
+}
+
+func NewClusterRoleHandler(Client client.Client, Scheme *runtime.Scheme) operands.Operand {
+	return operands.NewConditionalHandler(
+		operands.NewClusterRoleHandler(Client, Scheme, newClusterRole),
+		shouldDeployWaspAgent,
```
nunnatsa, line 62: Same comment about the passing a function here, as in `NewClusterRoleBindingHandler`. It can be an object rather than a function.

controllers/handlers/wasp-agent/scc.go
```diff
@@ -0,0 +1,76 @@
+package wasp_agent
+
+import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/kubevirt/hyperconverged-cluster-operator/controllers/operands"
```
nunnatsa: nit: please move down

controllers/handlers/wasp-agent/scc.go
```diff
@@ -0,0 +1,76 @@
...
+
+	securityv1 "github.com/openshift/api/security/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	hcov1beta1 "github.com/kubevirt/hyperconverged-cluster-operator/api/v1beta1"
+)
+
+func NewSecurityContextConstraintHandler(Client client.Client, Scheme *runtime.Scheme) operands.Operand {
```
nunnatsa: nit: Please use a specific name. We can use `SCC` to shorten it; e.g.

`NewWaspAgentSCCHandler`

controllers/handlers/wasp-agent/scc.go
```diff
@@ -0,0 +1,76 @@
...
+	return operands.NewConditionalHandler(
+		operands.NewSecurityContextConstraintsHandler(Client, Scheme, newSCC),
+		shouldDeployWaspAgent,
+		func(hc *hcov1beta1.HyperConverged) client.Object {
+			return newSCCWithNameOnly(hc)
+		},
+	)
+}
+
+func newSCC(hc *hcov1beta1.HyperConverged) *securityv1.SecurityContextConstraints {
```
nunnatsa: nit: please use a specific name.

controllers/handlers/wasp-agent/scc.go
```diff
@@ -0,0 +1,76 @@
...
+		AllowedUnsafeSysctls: []string{
+			"*",
+		},
+		SeccompProfiles: []string{
+			"*",
+		},
+	}
+}
+
+func newSCCWithNameOnly(hc *hcov1beta1.HyperConverged) *securityv1.SecurityContextConstraints {
```
nunnatsa: same

controllers/handlers/wasp-agent/scc.go
```diff
@@ -0,0 +1,76 @@
...
+		},
+	)
+}
+
+func newSCC(hc *hcov1beta1.HyperConverged) *securityv1.SecurityContextConstraints {
+	return &securityv1.SecurityContextConstraints{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:   waspAgentSCCName,
+			Labels: operands.GetLabels(hc, AppComponentWaspAgent),
+		},
```
nunnatsa: Please reuse `newSCCWithNameOnly`

controllers/handlers/wasp-agent/scc.go
```diff
@@ -0,0 +1,76 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	hcov1beta1 "github.com/kubevirt/hyperconverged-cluster-operator/api/v1beta1"
+)
+
+func NewSecurityContextConstraintHandler(Client client.Client, Scheme *runtime.Scheme) operands.Operand {
+	return operands.NewConditionalHandler(
+		operands.NewSecurityContextConstraintsHandler(Client, Scheme, newSCC),
+		shouldDeployWaspAgent,
```
nunnatsa, line 20: Here too, there is no reason to use a function for the creation of the required scc.

controllers/handlers/wasp-agent/service_account.go
```diff
@@ -0,0 +1,32 @@
...
+	return operands.NewConditionalHandler(
+		operands.NewServiceAccountHandler(Client, Scheme, newSA),
+		shouldDeployWaspAgent,
+		func(hc *hcov1beta1.HyperConverged) client.Object {
+			return newSA(hc)
+		},
+	)
+}
+
+func newSA(hc *hcov1beta1.HyperConverged) *corev1.ServiceAccount {
```
nunnatsa: nit: specific name.

controllers/handlers/wasp-agent/service_account.go
```diff
@@ -0,0 +1,32 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/kubevirt/hyperconverged-cluster-operator/controllers/operands"
+
+	hcov1beta1 "github.com/kubevirt/hyperconverged-cluster-operator/api/v1beta1"
+)
+
+func NewServiceAccountHandler(Client client.Client, Scheme *runtime.Scheme) operands.Operand {
+	return operands.NewConditionalHandler(
+		operands.NewServiceAccountHandler(Client, Scheme, newSA),
```
nunnatsa: Please pass the pre-computed object, instead of a function

controllers/operandhandler/operandHandler.go
```diff
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"time"
 
+	waspagent "github.com/kubevirt/hyperconverged-cluster-operator/controllers/handlers/wasp-agent"
```
nunnatsa: nit: please move down

tests/func-tests/wasp_agent_test.go
```diff
@@ -0,0 +1,158 @@
...
+		modifiedHco := tests.GetHCO(ctx, cli)
+		if originalHigherDensity != nil {
+			if modifiedHco.Spec.HigherWorkloadDensity == nil {
+				modifiedHco.Spec.HigherWorkloadDensity =
+					&hcov1beta1.HigherWorkloadDensityConfiguration{}
+			}
+			modifiedHco.Spec.HigherWorkloadDensity.MemoryOvercommitPercentage = originalHigherDensity.MemoryOvercommitPercentage
+		}
+
+		tests.UpdateHCORetry(ctx, cli, modifiedHco)
```
nunnatsa: Please check if we can patch instead of update.

tests/func-tests/wasp_agent_test.go
```diff
@@ -0,0 +1,158 @@
...
+	)
+
+	BeforeAll(func(ctx context.Context) {
+		cli = tests.GetControllerRuntimeClient()
+		tests.FailIfNotOpenShift(ctx, cli, "wasp-agent")
+		originalHco := tests.GetHCO(ctx, cli)
+		if originalHco.Spec.HigherWorkloadDensity != nil {
+			originalHigherDensity = &hcov1beta1.HigherWorkloadDensityConfiguration{
+				MemoryOvercommitPercentage: originalHco.Spec.HigherWorkloadDensity.MemoryOvercommitPercentage,
+			}
```
nunnatsa: Should we patch the HyperConverged CR with these changes?

enp0s3: @nunnatsa My assumption was that we need to restore the HCO CR original values as they were before this test, because I don't know what is the "baseline" that is being used for the e2e.

tests/func-tests/wasp_agent_test.go
enp0s3: Fixed.

controllers/operandhandler/operandHandler.go
enp0s3: Fixed

controllers/handlers/wasp-agent/rbac.go
enp0s3: @nunnatsa Sure, fixed.

enp0s3:
@nunnatsa Thank you for the review. I've addressed the comments.

tests/func-tests/wasp_agent_test.go
nunnatsa: but the the code populate the field... so is it the original state?

enp0s3: @nunnatsa I think so, I am running it in the `BeforeAll` it should be the value before I start to modify it in the suite

nunnatsa (approved):
Great work @enp0s3 , Thanks

/approve

PR #3702, opened 2025-08-11
Set default value for KV's defaultRuntimeClass field

nunnatsa:
Fix bug: The default KubeVirt's defaultRuntimeClass should be the priority class managed by HCO. Currently, if this configuration is missing in the HyperConverged CR, HCO does not set the KV field.

This PR fixes the issue, so if the HyperConverged
`spec.defaultRuntimeClass` field is empty, HCO will set the KV field to `kubevirt-cluster-critical`.

**Jira Ticket**:
```jira-ticket
https://issues.redhat.com/browse/CNV-62721
```

**Release note**:
```release-note
Set default value for KV's defaultRuntimeClass field
```

### DIFF
--- a/controllers/handlers/kubevirt.go
+++ b/controllers/handlers/kubevirt.go
@@ -495,9 +495,7 @@ func getKVConfig(hc *hcov1beta1.HyperConverged) (*kubevirtcorev1.KubeVirtConfigu
 		config.CPUModel = *hc.Spec.DefaultCPUModel
 	}
 
-	if hc.Spec.DefaultRuntimeClass != nil {
-		config.DefaultRuntimeClass = *hc.Spec.DefaultRuntimeClass
-	}
+	config.DefaultRuntimeClass = getKvPriorityClassName(hc)
 
 	if hc.Spec.VMStateStorageClass != nil {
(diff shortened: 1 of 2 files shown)


controllers/handlers/kubevirt.go
```diff
@@ -1036,3 +1034,10 @@ func getLabelPatch(dest, src map[string]string) ([]byte, error) {
 
 	return json.Marshal(patches)
 }
+
+func getKvPriorityClassName(hc *hcov1beta1.HyperConverged) string {
+	if hc.Spec.DefaultRuntimeClass != nil && *hc.Spec.DefaultRuntimeClass != "" {
+		return *hc.Spec.DefaultRuntimeClass
+	}
+	return kvPriorityClass
```
tiraboschi, line 1042: This is probably too much.
By the way the idea of setting by default an higher priority class for VMs as part of the opinionated deployment is still debated.

nunnatsa: > This is probably too much. By the way the idea of setting by default an higher priority class for VMs as part of the opinionated deployment is still debated.

/hold

nunnatsa:
Won't do. Closing.

PR #3708, opened 2025-08-14
Fix the Webhook's metrics endpoint

nunnatsa:

PR #3303 introduced an authorization protection for the metrics endpoints in the operator and the webhook pods.

The HCO logic creates the Bearer token for the operator, but not for the webhook, as the operator does not have access to the webhook secrets, and it can't produce the bearer token for it.

The meaning of this is that the metrics endpoint of the webhook became not accessible, because there is no way to get the bearer token. And the meaning of that, is that Prometheus can't get any metric information from the webhook.

This PR fixes the issue by adding a new controller to the webhook pod, to create the required resources (a service, a bearer token secret, and a ServiceMonitor). Now Prometheus can fetch the bearer token and use it, and webhook metrics are visible.

For example, here is an OCP screenshot with a successful querying of the webhook pod
image

> Note: Some of the unit tests in the new `controllers/webhooks/bearer-token-controller` package were written with the help of the `cursor` IDE.

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-65875
```

**Release note**:

```release-note
Fix a bug where the webhook's metrics endpoint was not accessible.
```


controllers/alerts/reconciler.go
```diff
@@ -29,42 +29,50 @@ type MetricReconciler interface {
...
 
+type GetReconcilersFunc func(hcoutil.ClusterInfo, string, metav1.OwnerReference) []MetricReconciler
+
 var logger = logf.Log.WithName("hyperconverged-operator-monitoring-reconciler")
 
 func NewMonitoringReconciler(ci hcoutil.ClusterInfo, cl client.Client, ee hcoutil.EventEmitter, scheme *runtime.Scheme) *MonitoringReconciler {
+	return CreateMonitoringReconciler(ci, cl, ee, scheme, true, getReconcilers)
+}
+
+func CreateMonitoringReconciler(ci hcoutil.ClusterInfo, cl client.Client, ee hcoutil.EventEmitter, scheme *runtime.Scheme, shouldReconcileNS bool, getReconcilersFn GetReconcilersFunc) *MonitoringReconciler {
```
machadovilaca, line 49: nit:
i know it doesn't sound the best
but I think usually we keep the New wording but add the `With${Option}` suffix to the name of the funcion, something like:

```suggestion
func NewMonitoringReconcilerWithReconcilers(ci hcoutil.ClusterInfo, cl client.Client, ee hcoutil.EventEmitter, scheme *runtime.Scheme, shouldReconcileNS bool, getReconcilersFn GetReconcilersFunc) *MonitoringReconciler {
```

controllers/webhooks/bearer-token-controller/resources_test.go
machadovilaca, line 1: nit:
I know they would be small, but I would still prefer to see each test associated with its file
"newServiceMonitor configures selector, endpoint, auth, and TLS" -> serviceMonitor.go -> serviceMonitor_test.go
"newSecret builds the expected Secret object" -> secret.go -> secret_test.go

controllers/webhooks/bearer-token-controller/bearer_token_controller.go
```diff
@@ -0,0 +1,141 @@
...
+	reqLogger := getLogger(ctx, request)
+
+	reqLogger.Info("Reconciling WH Bearer Token")
+	req := common.NewHcoRequest(ctx, request, reqLogger, false, true)
+
+	if err := r.metricReconciler.Reconcile(req, true); err != nil {
+		return reconcile.Result{RequeueAfter: time.Millisecond * 100}, err
+	}
+
+	return reconcile.Result{RequeueAfter: 5 * time.Minute}, nil // this will cause the reconciler to run every 5 minutes
```
machadovilaca: 5 minutes seems low for the kind of resource, maybe move the `5 * time.Minute` to a const, easy to change if needed

also we can use it in the respective test

controllers/alerts/service.go
```diff
@@ -15,38 +15,41 @@ import (
...
 const (
-	operatorPortName    = "http-metrics"
-	defaultOperatorName = "hyperconverged-cluster-operator"
-	operatorNameEnv     = "OPERATOR_NAME"
-	metricsSuffix       = "-operator-metrics"
-	serviceName         = hcoutil.HyperConvergedName + metricsSuffix
+	OperatorPortName = "http-metrics"
+	OperatorNameEnv  = "OPERATOR_NAME"
+	metricsSuffix    = "-operator-metrics"
+	serviceName      = hcoutil.HyperConvergedName + metricsSuffix
```
machadovilaca: nit

```suggestion
	OperatorPortName = "http-metrics"
	OperatorNameEnv  = "OPERATOR_NAME"

	metricsSuffix = "-operator-metrics"
	serviceName   = hcoutil.HyperConvergedName + metricsSuffix
```

controllers/alerts/reconciler.go
nunnatsa: Seems too long and too cumbersome to me... WDYT? any other idea?

controllers/webhooks/bearer-token-controller/bearer_token_controller.go
nunnatsa: I think 5 minutes is enough. The usecase we are trying to handle here is when the SA token is changed. In this case, Prometheus will fail to read metrics from the webhook for maximum 5 minutes. I think we can live with it. WDYT?

nunnatsa:
new change - rebase

PR #3713, opened 2025-08-17
Try to fix issue in nightly build

nunnatsa:
The nightly build fails from time to time on unit tests, when trying to create dashboard handlers based on a non-exists directory of dashboard yamls.

The directory name is computed with:
```
dir := path.Join(os.TempDir(), fmt.Sprint(time.Now().UTC().Unix()))
```

This pattern is used in two additional unit tests. The guess is that the directory is aleady exist, when the tests assumes it isn't.

This commit gives a unique directory name for each one of the tests that uses the temp directory name pattern, to ensure each test is using its own directory. Also, try to make these test more robust, by taking logic out of the `By` functions, and by adding some error checks.

**Release note**:
```release-note
None
```


nunnatsa:
> * Instead of manually generating timestamp‐based temp directories, consider using os.MkdirTemp (or Ginkgo’s TempDir) for guaranteed uniqueness and automatic cleanup.

This is not working for these tests. We need to first create the name, have a test with non-exist directory, and only then create the directory. The suggestions above will create the directory itself.

controllers/handlers/golden-images/golden_images_test.go
```diff
@@ -29,7 +29,7 @@ func TestOperators(t *testing.T) {
 }
 
 var _ = Describe("Test data import cron template", func() {
-	dir := path.Join(os.TempDir(), fmt.Sprint(time.Now().UTC().Unix()))
+	dir := path.Join(os.TempDir(), fmt.Sprintf("custom-dict-dir-%d", time.Now().UTC().Unix()))
```
enp0s3, line 32: If the directory syntax doesn't matter perhaps you can use `GinkgoT().TempDir()`, you have the cleanup for free.

cmd/hyperconverged-cluster-operator/main.go
```diff
@@ -330,11 +330,9 @@ func getCacheOption(operatorNamespace string, ci hcoutil.ClusterInfo) cache.Opti
 			},
 			&rbacv1.ClusterRole{}: {
 				Label: labelSelector,
-				Field: namespaceSelector,
```
enp0s3, line 333: +1 nice catch!

controllers/handlers/golden-images/golden_images_test.go
nunnatsa: Will do in a followup PR

PR #3722, opened 2025-08-21
New alert: `HCOMultiArchGoldenImagesDisabled`

nunnatsa:

Added the new `kubevirt_hco_multi_arch_boot_images_enabled` metric. It set if running on a multi-arch cluster, and there are valid DataImportCronTemplates. If the `enableMultiArchBootImageImport` FG is enabled, the metric value will be 1, if not, the value will be 0.

Added the `HCOMultiArchGoldenImagesDisabled` alert. The alert is triggered if the the `enableMultiArchBootImageImport` FG is disabled, while running on a multi-arch cluster (i.e., when the `kubevirt_hco_multi_arch_boot_images_enabled` is with value of 0).

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-63043
https://issues.redhat.com/browse/CNV-64419
```

**Release note**:

```release-note
Added the new `kubevirt_hco_multi_arch_boot_images_enabled` metric.
Added the `HCOMultiArchGoldenImagesDisabled` alert.
```

### DIFF
--- a/pkg/monitoring/hyperconverged/collectors/controllers.go
+++ b/pkg/monitoring/hyperconverged/collectors/controllers.go
@@ -1,0 +1,18 @@
+package collectors
+
+import (
+	"github.com/rhobs/operator-observability-toolkit/pkg/operatormetrics"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func SetupCollectors(cli client.Client, namespace string) error {
+	err := operatormetrics.RegisterCollector(
+		getMultiArchBootImagesStatusCollector(cli, namespace),
+	)
+
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/pkg/monitoring/hyperconverged/collectors/operator_collectors.go
+++ b/pkg/monitoring/hyperconverged/collectors/operator_collectors.go
@@ -1,0 +1,81 @@
+package collectors
+
+import (
+	"context"
+
+	"github.com/rhobs/operator-observability-toolkit/pkg/operatormetrics"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+
+	hcov1beta1 "github.com/kubevirt/hyperconverged-cluster-operator/api/v1beta1"
+	"github.com/kubevirt/hyperconverged-cluster-operator/pkg/nodeinfo"
+)
+
+const (
+	multiArchBootImagesFeatureEnabled  = float64(1)
+	multiArchBootImagesFeatureDisabled = float64(0)
+)
+
+var (
+	multiArchBootImagesStatus = operatormetrics.NewGauge(
+		operatormetrics.MetricOpts{
+			Name: "kubevirt_hco_multi_arch_boot_images_enabled",
+			Help: "indicates if the Multi-Arch Boot Images feature is enabled (1) or not (0)",
+		},
+	)
+
+	logger = logf.Log.WithName("operator-metrics-collector")
+)
+
+func getMultiArchBootImagesStatusCollector(cli client.Client, operatorNamespace string) operatormetrics.Collector {
+	return operatormetrics.Collector{
+		Metrics: []operatormetrics.Metric{
+			multiArchBootImagesStatus,
+		},
+		CollectCallback: getMultiArchBootImagesStatusCallback(cli, operatorNamespace),
+	}
+}
+
+func getMultiArchBootImagesStatusCallback(cli client.Client, operatorNamespace string) func() []operatormetrics.CollectorResult {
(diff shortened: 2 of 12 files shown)


pkg/monitoring/hyperconverged/collectors/operator_collectors.go
```diff
@@ -0,0 +1,83 @@
+package collectors
```
avlitman, line 1: please move this logic under pkg/monitoring/hyperconverged/metrics

pkg/monitoring/hyperconverged/collectors/controllers.go
```diff
@@ -0,0 +1,18 @@
+package collectors
+
+import (
+	"github.com/rhobs/operator-observability-toolkit/pkg/operatormetrics"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func SetupCollectors(cli client.Client, namespace string) error {
```
avlitman, line 8: can you move this login under pkg/monitoring/hyperconverged/metrics/metrics.go? then instead of calling  SetupCollectors in cmd/hyperconverged-cluster-operator/main.go you can call only SetupMetrics

pkg/monitoring/hyperconverged/collectors/operator_collectors_test.go
```diff
@@ -0,0 +1,464 @@
+package collectors
```
avlitman, line 1: same move this logic under pkg/monitoring/hyperconverged/metrics

pkg/monitoring/hyperconverged/collectors/operator_collectors.go
avlitman: I think it can also be in this file pkg/monitoring/hyperconverged/metrics/operator_metrics.go

pkg/monitoring/hyperconverged/collectors/operator_collectors.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+
+		if !ptr.Deref(hc.Spec.EnableCommonBootImageImport, true) &&
+			!enabledCustomDICTsFound(hc.Spec.DataImportCronTemplates) {
+			// Don't set the metric if no common nor custom DataImportCronTemplates found
+			return []operatormetrics.CollectorResult{}
+		}
+
+		// Set the metric based on the FeatureGate value
+		value := multiArchBootImagesFeatureDisabled
```
avlitman, line 66: for more clarity I would use similar code as in SetHCOMetricHyperConvergedExists() and SetHCOMetricHyperConvergedNotExists(), see https://github.com/kubevirt/hyperconverged-cluster-operator/blob/main/pkg/monitoring/hyperconverged/metrics/operator_metrics.go#L151

pkg/monitoring/hyperconverged/collectors/operator_collectors.go
```diff
@@ -0,0 +1,83 @@
...
+		if ptr.Deref(hc.Spec.FeatureGates.EnableMultiArchBootImageImport, false) {
```
avlitman, line 67: maybe need here some error handling

nunnatsa: why? Deref does not return error.

pkg/monitoring/hyperconverged/collectors/operator_collectors.go
nunnatsa: This implementation is not working in this case, because we must distinguish between non-set and zero.

The collector is the only option to achieve that.

pkg/monitoring/hyperconverged/collectors/operator_collectors.go
nunnatsa: It's not a metric. It's a collector.

avlitman: so can we move only the metric definition?
```
var (
	multiArchBootImagesStatus = operatormetrics.NewGauge(
		operatormetrics.MetricOpts{
			Name: "kubevirt_hco_multi_arch_boot_images_enabled",
			Help: "indicates if the Multi-Arch Boot Images feature is enabled (1) or not (0)",
		},
	)
)
```

pkg/monitoring/hyperconverged/collectors/operator_collectors.go
```diff
@@ -0,0 +1,83 @@
...
+			logger.Info("Multi-Arch boot images feature is enabled")
+			value = multiArchBootImagesFeatureEnabled
+		} else {
+			logger.Info("Multi-Arch boot images feature is disabled, but running on a multi-arch cluster with boot images enabled")
+		}
+
+		multiArchBootImagesStatus.Set(value)
```
machadovilaca: this can be removed

nunnatsa:
> let's add a E2E test on a follow-up PR if possible

Can't really be done, as it requires a real heterogeneous cluster, that we don't have in our CI.

PR #3730, opened 2025-08-27
Improve the performance of the HyperConverged CR deletion + some chores

nunnatsa:

This PR is handling several issues in the code:
### Improve the performance of the HyperConverged CR deletion
When the HyperConverged CR is deleted, the current behavior is to remove the KubeVirt, CDI, CNAO, SSP, and the AAQ CRs (with some additional resources and CRs).

The deletion itself takes a lot of time. By looking at the SSP operator code, it appears that it is trying to access several CDI and KubeVirt CRs, like VirtualMachines or DataImportCron, and that these operations are failed, because the relevant CRDs were already removed by the virt operator and the cdi operator, when the KubeVirt and the CDI CRs were removed. 

That delays the removal of the SSP CR for a few minutes. Eventually, the SSP CR is deleted, and then the HyperConverged
is deleted too, but it takes long time.

This PR changes the order of the resource deletions. It removes the KubeVirt and the CDI CRs, only after all the other resources and CRs are successfully removed.

The result is much faster deletion of the HyperConverged CR.

### Get rid of the `JustBeforeComplete` method
The `HCOOperandHooks` interface defines the `JustBeforeComplete` method to be called after a specific operand was successfully ensured.

After a few years, it is very clear that we never used this hook method, except for the SSP handler. Any other operand handler must implement this method to satisfy the interface, but has no implementation. This is just a redundant code that we don't need, and make the code a bit more complex for no good reason, make it harder to read and understand, to add new handlers and so on. Also, in general, we prefer interfaces to be as small as possible.

This PR removes the `JustBeforeComplete` method from the `HCOOperandHooks` interface, and all the method implementations from all the handlers.

The SSP handler was modified to be a `GenericOperand` wrapper, instead of just a `GenericOperand` object, then the same logic the SSP handler used to have in its `JustBeforeComplete` method, is added to this new wrapper.

### Jira Ticket
```jira-ticket
None
```

### Release note
```release-note
Improve the performance of the HyperConverged CR
```


nunnatsa:
/hold

#3725 should get in first, and it will conflict this one.

nunnatsa:
close and reopen to try fix the merging issue

nunnatsa:
/override pull-hyperconverged-cluster-operator-unit-test-s390x

The s390x test server is broken.

PR #3744, opened 2025-09-05
Add declarativeHotplugVolumes feature gate

mhenriks:
With this [PR](https://github.com/kubevirt/kubevirt/pull/13847), KubeVirt added a new feature gate called `DeclarativeHotplugVolumes` that enables the following features for `VirtualMachines`:
1. Declarative hotplug volumes
2. Inject/eject CD-ROM

Unfortunately, `DeclarativeHotplugVolumes` is not compatible with the existing `HotplugVolumes` feature gate which is currently defaulted in HCO. To make it easy for users that want to take advantage of the new behavior, a new flag was added to the `HyperConvergedFeatureGates` struct. When the `DeclarativeHotplugVolumes` flag is set, it will enable the new feature gate and remove the old one. Accomplishing this without the new flag would be pretty awkward using the existing jsonpatch functionality given that one feature gate has to be removed while another is added. Would have to do something like this:

```bash
kubectl annotate --overwrite -n kubevirt-hyperconverged hyperconverged kubevirt-hyperconverged \
  'kubevirt.kubevirt.io/jsonpatch=[{"op": "remove", "path": "/spec/configuration/developerConfiguration/featureGates/2"}, {"op": "add", "path": "/spec/configuration/developerConfiguration/featureGates/-", "value": "DeclarativeHotplugVolumes"}]'
```

 kubevirt-dev post announcing new behavior and risks [here](https://groups.google.com/g/kubevirt-dev/c/-tRpvU6CvVY)

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-68802
```

**Release note**:

```release-note
Add declarativeHotplugVolumes feature gate
```


controllers/handlers/kubevirt_test.go
```diff
@@ -2348,8 +2383,11 @@ Version: 1.2.3`)
 					).ToNot(HaveOccurred())
 
 					Expect(foundResource.Spec.Configuration.DeveloperConfiguration).ToNot(BeNil())
-					Expect(foundResource.Spec.Configuration.DeveloperConfiguration.FeatureGates).To(HaveLen(len(hardCodeKvFgs)))
+					// Should have base KVM emulation featuregate count (hardcoded + conditional volume hotplug)
+					Expect(foundResource.Spec.Configuration.DeveloperConfiguration.FeatureGates).To(HaveLen(defaultFeatureGateCount))
 					Expect(foundResource.Spec.Configuration.DeveloperConfiguration.FeatureGates).To(ContainElements(hardCodeKvFgs))
+					// Should contain HotplugVolumes by default when DeclarativeHotplugVolumes is not set
+					Expect(foundResource.Spec.Configuration.DeveloperConfiguration.FeatureGates).To(ContainElement("HotplugVolumes"))
```
nunnatsa: nit: Please use the `kvHotplugVolumesGate` constant instead

mhenriks:
@nunnatsa Thanks for the review, updated the description!

orenc1:
according to @batyana, this should be also in 4.20
/cherry-pick release-1.16

nunnatsa:
> according to @batyana, this should be also in 4.20 /cherry-pick release-1.16

We can't accept new api changes for 1.16 at this stage. This is way way way behind schedule.

Use json patch if needed.

nunnatsa:
We've got a waiver to do it now.

@dankenigsberg: FYI

/cherry-pick release-1.16

PR #3747, opened 2025-09-07
Fix failures in e2e tests, when calling the alert manager

nunnatsa:

The current test logic relays on the existence of a bearer token.

However, In some test environments, the kubeconfig file is client certificate file, rather than a bearer token one.

In this case the test will fail when trying to access send HTTP requests to the alert manager.

This PR allow also using the certificate identification, if the bearer token does not exist.

**Release note**:
```release-note
None
```


orenc1:
We tested that approach and it doesn't work.
Still getting 401 Unauthorized error when sending a request towards the alert manager.
The fix will be on the test env side.
/close

PR #3755, opened 2025-09-10
[release-1.15] Unblock the webhook metrics endpoint

nunnatsa:

> **NOTE**: This is a semi-manual cherry-pick of #3708 and #3739 

PR #3303 introduced an authorization protection for the metrics endpoints in the operator and the webhook pods.

The HCO logic creates the Bearer token for the operator, but not for the webhook, as the operator does not have access to the webhook secrets, and it can't produce the bearer token for it.

The meaning of this is that the metrics endpoint of the webhook became not accessible, because there is no way to get the bearer token. And the meaning of that, is that Prometheus can't get any metric information from the webhook.

This PR fixes the issue by adding a new controller to the webhook pod, to create the required resources (a service, a bearer token secret, and a ServiceMonitor). Now Prometheus can fetch the bearer token and use it, and webhook metrics are visible.

**Release note**:

```release-note
Fix a bug where the webhook's metrics endpoint was not accessible.
```

### DIFF
--- a/cmd/cmdcommon/cmdcommon.go
+++ b/cmd/cmdcommon/cmdcommon.go
@@ -2,6 +2,7 @@ package cmdcommon
 
 import (
 	"context"
+	"crypto/tls"
 	"flag"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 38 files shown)


nunnatsa:
Wrong target branch. closing.

PR #3766, opened 2025-09-17
Fix the flaky MediatedDevices e2e test

nunnatsa:

Give some more time to update the HyperConverged CR, to avoid conflict error.

Also, the current test is actually two tests, with big "if". This Pr splits the test into two tests, to get better readability and maintenance.

**Additional small changes**:
* On successful tests, the install namespace is deleted, with all the pods and all the other resources.

  When extra-gather is called, the HCO pods are not running, and we don't have their logs. Many time these logs are needed, even if the tests passed. For example, to compare with failed test lane. This PR also reads the operator and the webhook logs before killing them.

* Add some missing log lines in the alerts controller.

**Release note**:
```release-note
None
```

### DIFF
--- a/controllers/alerts/secret.go
+++ b/controllers/alerts/secret.go
@@ -96,6 +96,7 @@ func (r *SecretReconciler) UpdateExistingResource(ctx context.Context, cl client
 	}
 
 	// If the token is incorrect, delete the old secret and create a new one
+	logger.Info("the Secret token is outdated, deleting the old secret and creating a new one", "namespace", found.Namespace, "name", found.Name)
 	if err = cl.Delete(ctx, found); err != nil {
 		if !errors.IsNotFound(err) {
 			logger.Error(err, "failed to delete old secret")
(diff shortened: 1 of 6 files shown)


controllers/alerts/secret.go
```diff
@@ -96,6 +96,7 @@ func (r *SecretReconciler) UpdateExistingResource(ctx context.Context, cl client
 	}
 
 	// If the token is incorrect, delete the old secret and create a new one
+	logger.Info("the Secret token is outdated, deleting the old secret and creating a new one", "namespace", found.Namespace, "name", found.Name)
```
orenc1, line 99: we can be here also if the secret does not exist, and then we're not deleting anything, just creating a new secret token.
Also, not sure why one occurrence of `secret` is capitalized and the other one it not.

machadovilaca (approved):

controllers/alerts/secret.go
nunnatsa: > we can be here also if the secret does not exist

No. New resources are handled elsewhere, in https://github.com/kubevirt/hyperconverged-cluster-operator/blob/63ec654802b27dc399616d77c4c0ab69d77d79fd/controllers/alerts/reconciler.go#L138.

PR #3778, opened 2025-09-22
Handle empty Network string

yiraeChristineKim:

This PR fixes an issue where liveMigrationConfig.network: "" (empty string) causes an error. When the network field is null or omitted, the default network is used. With this fix, an empty string will also default to the network instead of throwing an error.

This fix is needed because, in ACM, we deliver HyperConverged with an [AddonTemplate](https://open-cluster-management.io/docs/concepts/add-on-extensibility/addon/)
. We use a customVariable for the network field.
Users can provide a network name for their managed clusters. However, when this variable is left empty, it becomes an empty string (""), which causes an error and prevents HyperConverged from installing CNV.

**Reviewer Checklist**

- [x] PR Message
- [ ] Commit Messages
- [x] How to test
- [ v] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket

```

**Release note**:

```release-note

```

### DIFF
--- a/controllers/handlers/kubevirt.go
+++ b/controllers/handlers/kubevirt.go
@@ -681,6 +681,10 @@ func hcLiveMigrationToKv(lm hcov1beta1.LiveMigrationConfigurations) (*kubevirtco
 		bandwidthPerMigration = &bandwidthPerMigrationObject
 	}
 
+	if lm.Network != nil && *lm.Network == "" {
+		lm.Network = nil
+	}
+
 	return &kubevirtcorev1.MigrationConfiguration{
 		BandwidthPerMigration:             bandwidthPerMigration,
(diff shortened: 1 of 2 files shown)


nunnatsa:
After reading the description, it looks like an ACM issue, not HCO's. 

The field is omitted if empty. in yaml, empty string is not "empty". This is true in k8s in general. If the user (== ACM) chose to set a value (and yes, an empty string is a value), then this is the user decision to do so. 

Fixing this specific issue is:
not following the regular behavior in K8s and K8s operators
Will cause inconsistent behavior: if we will create a special behavior for this specific field, when we have many similar string fields - why this one should behave differently?

/hold

PR #3789, opened 2025-09-29
kubevirt: Deprecate the highBurst HyperConvergedTuningPolicy

lyarwood:

This policy is no longer required as KubeVirt now defaults to the burst and QPS client rate limits provided by this policy. As such it is now deprecated ahead of removal in a future release.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-69442
```

**Release note**:

```release-note
The `highBurst` tuning policy is now deprecated ahead of removal in a future release as KubeVirt now provides the client rate burst and qps limits by default previously provided by this policy.
```

### DIFF
--- a/pkg/webhooks/validator/validator.go
+++ b/pkg/webhooks/validator/validator.go
@@ -161,6 +161,10 @@ func (wh *WebhookHandler) ValidateCreate(_ context.Context, dryrun bool, hc *v1b
 		return err
 	}
 
+	if err := wh.validateTuningPolicy(hc); err != nil {
+		return err
+	}
+
 	if err := wh.validateAffinity(hc); err != nil {
 		return err
 	}
@@ -229,6 +233,10 @@ func (wh *WebhookHandler) ValidateUpdate(ctx context.Context, dryrun bool, reque
 	}
 
 	if err := wh.validateFeatureGatesOnUpdate(requested, exists); err != nil {
+		return err
+	}
+
+	if err := wh.validateTuningPolicy(requested); err != nil {
 		return err
 	}
 
@@ -449,6 +457,13 @@ func (wh *WebhookHandler) validateMediatedDeviceTypes(hc *v1beta1.HyperConverged
 				return fmt.Errorf("mediatedDevicesTypes is deprecated, please use mediatedDeviceTypes instead")
 			}
 		}
+	}
+	return nil
+}
+
+func (wh *WebhookHandler) validateTuningPolicy(hc *v1beta1.HyperConverged) error {
+	if hc.Spec.TuningPolicy == v1beta1.HyperConvergedHighBurstProfile { //nolint SA1019
+		return newValidationWarning([]string{"spec.tuningPolicy: the highBurst profile is deprecated as of v1.16.0 and will be removed in a future release"})
 	}
 	return nil
 }
--- a/pkg/webhooks/validator/validator_test.go
+++ b/pkg/webhooks/validator/validator_test.go
@@ -784,6 +784,24 @@ var _ = Describe("webhooks validator", func() {
 				))
 			})
 		})
+
+		Context("validate tuning policy", func() {
+			It("should return warning for deprecated highBurst tuning policy", func() {
+				cr.Spec.TuningPolicy = v1beta1.HyperConvergedHighBurstProfile //nolint SA1019
+				err := wh.ValidateCreate(ctx, dryRun, cr)
+				Expect(err).To(HaveOccurred())
+				expected := &ValidationWarning{}
+				Expect(errors.As(err, &expected)).To(BeTrue())
+				Expect(expected.warnings).To(HaveLen(1))
+				Expect(expected.warnings[0]).To(ContainSubstring("highBurst profile is deprecated"))
+				Expect(expected.warnings[0]).To(ContainSubstring("v1.16.0"))
+			})
+
+			It("should not return warning when tuning policy is not set", func() {
+				cr.Spec.TuningPolicy = ""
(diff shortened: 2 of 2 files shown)


controllers/handlers/kubevirt.go
```diff
@@ -374,7 +374,7 @@ func hcoTuning2Kv(hc *hcov1beta1.HyperConverged) (*kubevirtcorev1.ReloadableComp
 	switch hc.Spec.TuningPolicy {
 	case hcov1beta1.HyperConvergedAnnotationTuningPolicy:
 		return getHcoAnnotationTuning(hc)
-	case hcov1beta1.HyperConvergedHighBurstProfile:
+	case hcov1beta1.DeprecatedHyperConvergedHighBurstProfile:
```
nunnatsa:
```suggestion
	case hcov1beta1.DeprecatedHyperConvergedHighBurstProfile: //nolint SA1019
```

controllers/handlers/kubevirt_test.go
```diff
@@ -3749,7 +3749,7 @@ Version: 1.2.3`)
 				})
 
 				It("Should create the fields and populate them using the highBurst profile values", func() {
-					hco.Spec.TuningPolicy = hcov1beta1.HyperConvergedHighBurstProfile
+					hco.Spec.TuningPolicy = hcov1beta1.DeprecatedHyperConvergedHighBurstProfile
```
nunnatsa: same

controllers/handlers/kubevirt_test.go
```diff
@@ -3738,7 +3738,7 @@ Version: 1.2.3`)
 			Context("with highBurst profile", func() {
 
 				It("Should return error if the json annotation tuningPolicy is present", func() {
-					hco.Spec.TuningPolicy = hcov1beta1.HyperConvergedHighBurstProfile
+					hco.Spec.TuningPolicy = hcov1beta1.DeprecatedHyperConvergedHighBurstProfile
```
nunnatsa:
```suggestion
					//nolint SA1019
					hco.Spec.TuningPolicy = hcov1beta1.DeprecatedHyperConvergedHighBurstProfile
```

tests/func-tests/tuningpolicy_test.go
```diff
@@ -59,7 +59,7 @@ var _ = Describe("Check that the TuningPolicy annotation is configuring the KV o
 		hc := tests.GetHCO(ctx, cli)
 
 		delete(hc.Annotations, common.TuningPolicyAnnotationName)
-		hc.Spec.TuningPolicy = v1beta1.HyperConvergedHighBurstProfile
+		hc.Spec.TuningPolicy = v1beta1.DeprecatedHyperConvergedHighBurstProfile
```
nunnatsa: same

lyarwood:
> Added a few inline comments.
> 
> Maybe you also want to add a deprecation warning from the validation webhook?

@nunnatsa thanks for the review, I wanted to ask about this, is an admission webhook warning enough or do we want an actionable condition raised to let users know that they are using a deprecated configurable or maybe even both?

nunnatsa:
I would do the validation warning anyway. Not sure about the condition. What are the actual implication of the deprecated setting in 1.7? How bad is it?

Also, I noticed the PR does not change the logic of the deprecated setting, and still update KV. Is there a point to do that? Don't we want to remove this logic?

lyarwood:
> I would do the validation warning anyway. Not sure about the condition. What are the actual implication of the deprecated setting in 1.7? How bad is it?

No functional changes, I'm just looking for ways to make users aware who already have it set and might not trigger the admission warning.

> Also, I noticed the PR does not change the logic of the deprecated setting, and still update KV. Is there a point to do that? Don't we want to remove this logic?

Technically we can drop the logic once https://github.com/kubevirt/kubevirt/pull/15712 lands in KubeVirt but you don't typically change behaviour during a deprecation and would only remove this after the deprecation period ends. I don't mind either way tbh.

nunnatsa:
> Technically we can drop the logic once kubevirt/kubevirt#15712 lands in KubeVirt 

This is not enough, BTW. We'll need a KubeVirt release with this change.

lyarwood:
> > Technically we can drop the logic once [kubevirt/kubevirt#15712](https://github.com/kubevirt/kubevirt/pull/15712) lands in KubeVirt
> 
> This is not enough, BTW. We'll need a KubeVirt release with this change.

Yes sorry I did mean to say that once the PR lands, ships in a release and is consumed by HCO.

lyarwood:
@nunnatsa would you mind taking a look at this if you have time?

api/v1beta1/hyperconverged_types.go
```diff
@@ -31,7 +31,8 @@ type HyperConvergedTuningPolicy string
 // through annotation values.
 const (
 	HyperConvergedAnnotationTuningPolicy HyperConvergedTuningPolicy = "annotation"
-	HyperConvergedHighBurstProfile       HyperConvergedTuningPolicy = "highBurst"
+	// Deprecated: The highBurst profile is deprecated as of v1.16.0 ahead of removal in a future release
+	DeprecatedHyperConvergedHighBurstProfile HyperConvergedTuningPolicy = "highBurst"
```
nunnatsa: I don't think we need to rename the constant. The `// Deprecated` comment is well known to linters and IDEs.

PR #3801, opened 2025-10-13
Ensure secret always appears at related objects

machadovilaca:

Ensure secret always appears at related objects

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-70251
```

**Release note**:

```release-note
none
```

### DIFF
--- a/controllers/alerts/secret.go
+++ b/controllers/alerts/secret.go
@@ -92,7 +92,7 @@ func (r *SecretReconciler) UpdateExistingResource(ctx context.Context, cl client
 
 	// Check if the secret has the correct token
 	if string(found.Data["token"]) == token {
-		return nil, false, nil
+		return found, false, nil
 	}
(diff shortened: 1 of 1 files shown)


nunnatsa:
> thinking about all of this again, I'm not sure it's valid to place the metric objects in the related images. We create them even if the HyperConverged CR is not exist, so are they really relates to it?

OK, I'll approve it for now. @machadovilaca - let's think about it again when you'll back.

/unhold

PR #3803, opened 2025-10-15
Fix the failing nightly test

nunnatsa:

The test, as preparations, installs the latest version of OLM, then it installs the cert-manager.

But the latest version of OLM, rleased 2 days ago, requires cert-manager to be already installed.

This PR changes the order so the test will first install cert-manager, and only then install OLM.

Also, stop using the deploy/cert-manager.yaml file. Instead, use the file from the cert-manager source, directly.

We'll remove this file in a future PR.

**Jira Ticket**:

```jira-ticket
None
```

**Release note**:

```release-note
None
```

### DIFF
--- a/deploy/deploy.sh
+++ b/deploy/deploy.sh
@@ -4,6 +4,8 @@ set -ex
 
 hco_namespace=kubevirt-hyperconverged
 
+CERT_MANAGER_VERSION=${CERT_MANAGER_VERSION:-"v1.18.2"}
+
 IS_OPENSHIFT=${IS_OPENSHIFT:-false}
 if kubectl api-resources |grep clusterversions |grep config.openshift.io; then
   IS_OPENSHIFT="true"
@@ -36,7 +38,7 @@ kubectl apply ${LABEL_SELECTOR_ARG} -f https://raw.githubusercontent.com/kubevir
 kubectl apply ${LABEL_SELECTOR_ARG} -f https://raw.githubusercontent.com/kubevirt/hyperconverged-cluster-operator/main/deploy/crds/application-aware-quota00.crd.yaml
 
 # Deploy cert-manager for webhook certificates
-kubectl apply ${LABEL_SELECTOR_ARG} -f https://raw.githubusercontent.com/kubevirt/hyperconverged-cluster-operator/main/deploy/cert-manager.yaml
+kubectl apply -f "https://github.com/cert-manager/cert-manager/releases/download/${CERT_MANAGER_VERSION}/cert-manager.yaml"
 kubectl -n cert-manager wait deployment/cert-manager --for=condition=Available --timeout="300s"
 kubectl -n cert-manager wait deployment/cert-manager-webhook --for=condition=Available --timeout="300s"
 
--- a/hack/deploy-cert-manager.sh
+++ b/hack/deploy-cert-manager.sh
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+
+source hack/common.sh
+
+set -ex
+
+CERT_MANAGER_TIMEOUT=${CERT_MANAGER_TIMEOUT:-"120s"}
+
+echo "Installing cert-manager ${CERT_MANAGER_VERSION}"
+${CMD} apply -f https://github.com/cert-manager/cert-manager/releases/download/${CERT_MANAGER_VERSION}/cert-manager.yaml
+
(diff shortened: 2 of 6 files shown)


hack/deploy-cert-manager.sh
```diff
@@ -0,0 +1,24 @@
...
+${CMD} wait --for=condition=Available --namespace=cert-manager deployment/cert-manager --timeout=${CERT_MANAGER_TIMEOUT}
+${CMD} wait --for=condition=Available --namespace=cert-manager deployment/cert-manager-cainjector --timeout=${CERT_MANAGER_TIMEOUT}
+${CMD} wait --for=condition=Available --namespace=cert-manager deployment/cert-manager-webhook --timeout=${CERT_MANAGER_TIMEOUT}
+
+echo "Waiting for cert-manager webhook to be ready..."
+${CMD} wait --for=condition=Ready --namespace=cert-manager pod -l app=webhook --timeout=${CERT_MANAGER_TIMEOUT}
+
+echo "Waiting for cert-manager CRDs to be available..."
+${CMD} wait --for condition=established --timeout=${CERT_MANAGER_TIMEOUT} crd/certificates.cert-manager.io
+${CMD} wait --for condition=established --timeout=${CERT_MANAGER_TIMEOUT} crd/issuers.cert-manager.io
```
orenc1: Don't we need the CRDs to be available before we check for the pods/deployments?

hack/deploy-cert-manager.sh
```diff
@@ -0,0 +1,24 @@
...
+echo "Installing cert-manager ${CERT_MANAGER_VERSION}"
+${CMD} apply -f https://github.com/cert-manager/cert-manager/releases/download/${CERT_MANAGER_VERSION}/cert-manager.yaml
+
+echo "Waiting for cert-manager to be ready..."
+${CMD} wait --for=condition=Available --namespace=cert-manager deployment/cert-manager --timeout=${CERT_MANAGER_TIMEOUT}
+${CMD} wait --for=condition=Available --namespace=cert-manager deployment/cert-manager-cainjector --timeout=${CERT_MANAGER_TIMEOUT}
+${CMD} wait --for=condition=Available --namespace=cert-manager deployment/cert-manager-webhook --timeout=${CERT_MANAGER_TIMEOUT}
+
+echo "Waiting for cert-manager webhook to be ready..."
+${CMD} wait --for=condition=Ready --namespace=cert-manager pod -l app=webhook --timeout=${CERT_MANAGER_TIMEOUT}
```
orenc1: isn't that redundant if we already checked the webhook deployment above?
condition=available of deployment means that all pods of the deployment are ready.

deploy/deploy.sh
```diff
@@ -4,6 +4,8 @@ set -ex
 
 hco_namespace=kubevirt-hyperconverged
 
+CERT_MGR_VERSION=${CERT_MGR_VERSION:-"v1.18.2"}
```
orenc1: are we dynamically setting the cert manager version somewhere?
or it is always hard-coded to v1.18.2 until updated manually?

hack/deploy-cert-manager.sh
nunnatsa: right. copied from OLM code... will remove

deploy/deploy.sh
nunnatsa: I also put it in hack/common.sh

But this script does not source it, and I don't want to change too much

nunnatsa: More details:

The deploy/deploy.sh script is only used in our README file to deploy HCO. We guide the users to run the script directly from github:

>
> \## Using the HCO without OLM or Marketplace
>
> Run the following script to apply the HCO operator:
>
> \```bash
> $ curl https://raw.githubusercontent.com/kubevirt/hyperconverged-cluster-operator/main/deploy/deploy.sh | > bash
> \```

So we can't refer any other script there.

orenc1: but in `hack/common.sh` it's also:
```
CERT_MANAGER_VERSION=${CERT_MANAGER_VERSION:-"v1.18.2"}
```
is the `CERT_MANAGER_VERSION` env var being set from the outside somewhere?
or we'll need to bump it from time to time?

nunnatsa: Currently it does not set from the outside, and yes - we'll need to bump from time to time.

orenc1 (approved):
/approve

PR #3813, opened 2025-10-22
CI: Fix the upgrade tests

nunnatsa:

The upgrade tests are failing because the hack/check-uptime.sh is using the virtctl `--local-ssh=true` flag, that is no longer supported by virtctl in v1.7

The prev-upgrade tests are passing because the test uses the previous KubeVirt version to get virtctl version, and it's still v1.6.2. This is also why the bump PR for KV 1.7-beta.0 was passed.

**Release note**:

```release-note
None
```


orenc1 (approved):
/approve

PR #3833, opened 2025-11-03
Add e2e test for network policies

nunnatsa:
**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-70737
```

**Release note**:

```release-note
None
```

### DIFF
--- a/tests/func-tests/network_policy_test.go
+++ b/tests/func-tests/network_policy_test.go
@@ -1,0 +1,435 @@
+package tests_test
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	consolev1 "github.com/openshift/api/console/v1"
+	corev1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	hcov1beta1 "github.com/kubevirt/hyperconverged-cluster-operator/api/v1beta1"
+	"github.com/kubevirt/hyperconverged-cluster-operator/controllers/operands"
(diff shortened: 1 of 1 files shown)


nunnatsa:
> * The error assertions using MatchError with k8serrors.IsNotFound are incorrect—use Expect(errors.IsNotFound(err)).To(BeTrue()) or a similar pattern to properly detect NotFound errors.

This is not a good advice. First the `MatchError` matcher is the way to check error types in gomega. Also, in general, using the BeTrue matcher give almost no information about the failure. Always choose another matcher instead of BeTrue, if possible.

rlobillo:
Hello @nunnatsa 

IIUC, this PR is only checking the blocked ingress traffic from the console plugin. It's missing the egress traffic from the apiserver proxy. That will require the creation of a server pod + service.

tests/func-tests/network_policy_test.go
```diff
@@ -0,0 +1,435 @@
...
+	hcov1beta1 "github.com/kubevirt/hyperconverged-cluster-operator/api/v1beta1"
+	"github.com/kubevirt/hyperconverged-cluster-operator/controllers/operands"
+	hcoutil "github.com/kubevirt/hyperconverged-cluster-operator/pkg/util"
+	tests "github.com/kubevirt/hyperconverged-cluster-operator/tests/func-tests"
+)
+
+const (
+	clientPodName        = "client"
+	testServerPodName    = "my-server"
+	testServerPort       = int32(8080)
```
rlobillo: I think it's better to use a high, non-standard port for the test server (e.g., 9876) for clarity in test debugging, since port 8080 is already used by kubevirt-apiserver-proxy-service. What do you think?

tests/func-tests/network_policy_test.go
```diff
@@ -0,0 +1,435 @@
...
+
+	By(fmt.Sprintf("The %s/%s pod was created; waiting for it to start running...", pod.Namespace, pod.Name))
+	key := client.ObjectKey{Namespace: pod.Namespace, Name: pod.Name}
+	Eventually(func(g Gomega, ctx context.Context) {
+		resPod := &corev1.Pod{}
+		g.Expect(cli.Get(ctx, key, resPod)).To(Succeed())
+
+		g.Expect(resPod.Status.Phase).To(Equal(corev1.PodRunning))
+	}).WithContext(ctx).
+		WithTimeout(5 * time.Minute).
```
avlitman: reduce to 3 min

PR #3835, opened 2025-11-04
Add Cluster Memory Load Perses dashboard

sradco:

Add first-class Perses dashboard support to HCO:
- Ship a namespaced PersesDatasource and a Node Memory Overview PersesDashboard as HCO assets.
- Auto-provision a bearer token for the datasource from the Perses ServiceAccount.
- Apply assets only when Perses CRDs are present.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-71909
```

**Release note**:

```release-note
Add a Node Memory Overview Perses Dashboard
```


sradco:
@jgbernalp @saswatamcode, This is the first Perses dashboard we will be shipping.
Can you please review the datasource and dashboards I created in this PR and see if its fits to how we should be deploying the Perses dashboards?

controllers/observability/perses.go
```diff
@@ -0,0 +1,95 @@
...
+		kind, _ := objMap["kind"].(string)
+		if apiVersion == "" || kind == "" {
+			return fmt.Errorf("missing apiVersion/kind in %s", path)
+		}
+
+		u := &unstructured.Unstructured{Object: objMap}
+		u.SetGroupVersionKind(schema.FromAPIVersionAndKind(apiVersion, kind))
+
+		// Apply with SSA so we don't need to read existing objects
+		if err := r.Patch(ctx, u, client.Apply, client.FieldOwner("hyperconverged-operator"), client.ForceOwnership); err != nil {
```
jgbernalp: we are missing to check if the Perses CRD are present in the cluster, otherwise this step will always fail.

saswatamcode: Yup, maybe this could also be behind a flag?

sradco:
@nunnatsa @avlitman @machadovilaca please review this PR.

controllers/observability/perses.go
sradco: Updated.  Please review.

sradco:
@fabiand @tiraboschi  FYI, the PR for deploying the Perses dashboard is ready for review.

controllers/observability/perses.go
```diff
@@ -0,0 +1,213 @@
...
+			persesNs = item.GetNamespace()
+			if persesNs != "" {
+				break
+			}
+		}
+	}
+	if persesNs == "" {
+		// Nothing to do if Perses is not installed
+		return nil
+	}
```
nunnatsa: Can you use the actual `Perses` go type, instead of this tricky unstructured type?

controllers/observability/perses.go
```diff
@@ -0,0 +1,213 @@
...
+
+	// Request a short-lived token for the "perses" SA in the Perses namespace
+	token, err := r.requestServiceAccountToken(ctx, persesNs, "perses")
+	if err != nil || token == "" {
+		// Do not fail reconcile; just skip creating the secret
+		return nil
+	}
+
+	// Create/Update secret in operator namespace
+	secretName := "perses-thanos-datasource-secret"
```
nunnatsa: Please change it to const, instead of a variable.

controllers/observability/perses.go
```diff
@@ -0,0 +1,213 @@
...
+	return r.Update(ctx, secret)
+}
+
+func (r *Reconciler) getSecret(ctx context.Context, s *corev1.Secret) (*corev1.Secret, error) {
+	current := &corev1.Secret{}
+	if err := r.Get(ctx, types.NamespacedName{Namespace: s.Namespace, Name: s.Name}, current); err != nil {
+		return nil, err
+	}
+	return current, nil
+}
```
nunnatsa: not sure why do we need this function. Also, the returned value is eventually ignored.

controllers/observability/perses.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	// Create/Update secret in operator namespace
+	secretName := "perses-thanos-datasource-secret"
+	secret := &corev1.Secret{}
+	secret.Namespace = r.namespace
+	secret.Name = secretName
+	_, getErr := r.getSecret(ctx, secret)
+	data := map[string][]byte{"token": []byte(token)}
+	if getErr != nil {
```
nunnatsa: Please check the error. this is only right in the error is "not found". for every other error, you'll need to return error.

See example here:
https://github.com/kubevirt/hyperconverged-cluster-operator/blob/fbf6052c296667f34d994c93bee0073185930fcf/controllers/observability/alerts.go#L27-L28

controllers/observability/perses.go
nunnatsa:
```suggestion
func (r *Reconciler) getSecret(ctx context.Context, s *corev1.Secret) (*corev1.Secret, error) {
	current := &corev1.Secret{}
	return r.Get(ctx, client.ObjectKeyFromObject(s), current)
}
```

controllers/observability/perses.go
```diff
@@ -0,0 +1,213 @@
...
+		Spec: authv1.TokenRequestSpec{
+			Audiences:         []string{"https://thanos-querier.openshift-monitoring.svc.cluster.local"},
+			ExpirationSeconds: ptrInt64(3600), // 1h
+		},
+	}
+	resp, err := clientset.CoreV1().ServiceAccounts(namespace).CreateToken(ctx, saName, tr, metav1.CreateOptions{})
+	if err != nil {
+		return "", err
+	}
+	return resp.Status.Token, nil
```
nunnatsa: Can't we use our own service account for that?

controllers/observability/perses.go
```diff
@@ -0,0 +1,227 @@
...
+			return false
+		}
+	}
+	return true
+}
+
+// readPersesAssets walks the assets dir and returns parsed objects as generic maps.
+func readPersesAssets(d fs.FS) ([]map[string]any, error) {
+	var out []map[string]any
+	err := fs.WalkDir(d, persesAssetsDir, func(path string, entry fs.DirEntry, err error) error {
```
nunnatsa: Reading from disk is expensive, and doing it again and again for files that will never changed, is not so efficient. This whole function must be called only once, on boot time: Collect the files into objects in the memory, and then use these object for reconcile.

 See the implementation of quickstart, or imagestream for example.

controllers/observability/perses.go
```diff
@@ -0,0 +1,227 @@
...
+		"persesdashboards.perses.dev",
+		"persesdatasources.perses.dev",
+	} {
+		u := &unstructured.Unstructured{}
+		u.SetGroupVersionKind(crdGVK)
+		if err := r.Get(ctx, types.NamespacedName{Name: crdName}, u); err != nil {
+			return false
+		}
+	}
+	return true
```
nunnatsa: just read into the CRD go type. No need to use unstructured here. Also, please check the existing CRD controller - maybe you can reuse it somehow?

controllers/observability/perses.go
```diff
@@ -0,0 +1,227 @@
...
+			return fmt.Errorf("failed to read %s: %w", path, err)
+		}
+		jsonBytes, err := yaml.YAMLToJSON(content)
+		if err != nil {
+			return fmt.Errorf("failed to convert yaml to json for %s: %w", path, err)
+		}
+		var objMap map[string]any
+		if err := json.Unmarshal(jsonBytes, &objMap); err != nil {
+			return fmt.Errorf("failed to unmarshal json for %s: %w", path, err)
+		}
```
nunnatsa: can we just marshal into the actual Preses type? it will be much shorter and simpler code.

controllers/observability/perses.go
```diff
@@ -0,0 +1,215 @@
...
+// It uses server-side apply to create/update resources idempotently without relying on cache reads.
+func (r *Reconciler) ReconcilePersesResources(ctx context.Context) error {
+	// Use embedded assets instead of runtime file system
+	d := assets.GetPersesDashboardsFS()
+
+	// Ensure Perses CRDs are present before attempting to apply resources.
+	// Without CRDs, server-side apply will fail on every reconcile loop.
+	if !r.persesCRDsReady(ctx) {
+		return nil
+	}
```
nunnatsa: let's move it up, to the beginning of the method.

controllers/observability/perses.go
```diff
@@ -0,0 +1,215 @@
...
+
+// persesAssetsDir is the directory (inside the repo image) containing Perses CR manifests (yaml).
+// Each YAML should define a single resource: PersesDashboard or PersesDatasource.
+const persesAssetsDir = "."
+
+// ReconcilePersesResources applies all Perses resources shipped with the operator under assets/perses.
+// It uses server-side apply to create/update resources idempotently without relying on cache reads.
+func (r *Reconciler) ReconcilePersesResources(ctx context.Context) error {
+	// Use embedded assets instead of runtime file system
+	d := assets.GetPersesDashboardsFS()
```
nunnatsa: Can be moved, to be inside `r.assetsOnce.Do()`.

controllers/observability/perses.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	// Load and cache assets once
+	var loadErr error
+	r.assetsOnce.Do(func() {
+		r.cachedAssets, loadErr = readPersesAssets(d)
+	})
+	if loadErr != nil {
+		return loadErr
+	}
```
nunnatsa: This error is not recoverable. On the other hand, if the asset files are wrong, then we should not pass unit tests.

I think it will be safe to panic here, because if our unit tests are good, then we shall never get error here, in production.

WDYT about this?

```suggestion
	r.assetsOnce.Do(func() {
		var err error
		r.cachedAssets, err = readPersesAssets(d)
		if err != nil {
			panic("failed to parse perses files; " + err.Error())
		}
	})
```

controllers/observability/perses.go
```diff
@@ -0,0 +1,215 @@
...
+
+	// Apply cached assets each reconcile
```
nunnatsa: This seems to be very pricey.

We should read the CRs, compare them to the required assets, and only update if something was actually changed. if we'll [add the resources to the client cache](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/e4ae5da5fd878d811749dca08bfe19709e3da982/cmd/hyperconverged-cluster-operator/main.go#L307), and then we'll make the controller watch them, we'll only handle them if something was changed.

controllers/observability/perses.go
nunnatsa: We have a utility here https://github.com/kubevirt/hyperconverged-cluster-operator/blob/cae67e758968fda8c1c71efa1509f0f917897465/pkg/util/cluster.go#L228 

Please reuse it: either expose it (rename it to `IsCRDExists`), or add a new utility; e.g.
```go
func IsPersesAvailable(ctx context.Context, cl client.Client) bool {
	return isCRDExists(ctx, cl, "persesdashboards.perses.dev") &&
			isCRDExists(ctx, cl, "persesdatasources.perses.dev") 
}

File: controllers/observability/perses.go [hunk 14]
```diff
@@ -0,0 +1,215 @@
...
+			if p.Namespace != "" {
+				persesNs = p.Namespace
+				break
+			}
+		}
+	}
+	if persesNs == "" {
+		// Nothing to do if Perses is not installed
+		return nil
+	}
```
nunnatsa: let's move this into a new function, for better readability; e.g.
```go
func (r *Reconciler) findPersesNamespace(ctx context.Context) string {
	persesList := &persesv1alpha1.PersesList{}
	err := r.List(ctx, &persesList)
	id err != nil || len(persesList.Items) == 0 {
		return ""
	}

	return persesList.Items[0].Namespace
}
```

Then just:

```suggestion
	persesNS := r.findPersesNamespace(ctx)
	if persesNS == "" {
		return nil
	}
```

controllers/observability/perses.go
```diff
@@ -0,0 +1,215 @@
...
+
+	// Request a short-lived token for the "perses" SA in the Perses namespace
+	token, err := r.requestServiceAccountToken(ctx, persesNs, "perses")
```
nunnatsa: Should we do it in each and every reconcile loop? this is super pricey.

controllers/observability/perses.go
```diff
@@ -0,0 +1,215 @@
...
+		// Do not fail reconcile; just skip creating the secret
+		return nil
+	}
+
+	// Create/Update secret in operator namespace
+	secret := &corev1.Secret{}
+	secret.Namespace = r.namespace
+	secret.Name = persesDatasourceSecretName
+	getErr := r.Get(ctx, types.NamespacedName{Namespace: secret.Namespace, Name: secret.Name}, secret)
+	data := map[string][]byte{"token": []byte(token)}
```
nunnatsa: we don't need this variable. instead, please consider to modify line 185 below:
```go
		secret.Data = map[string][]byte{"token": []byte(token)}
```

controllers/observability/perses.go
```diff
@@ -0,0 +1,215 @@
...
+		secret.Type = corev1.SecretTypeOpaque
+		secret.Data = data
+		return r.Create(ctx, secret)
+	}
+	// Update existing secret
+	if secret.Data == nil {
+		secret.Data = map[string][]byte{}
+	}
+	secret.Data["token"] = []byte(token)
+	return r.Update(ctx, secret)
```
nunnatsa: Check the data before updating; e.g.

```suggestion
	if foundToken, ok := secret.Data["token"]; ok && slices.Equal(data, foundData) {
		return nil
	}

	// Update existing secret
	if secret.Data == nil {
		secret.Data = map[string][]byte{}
	}
	secret.Data["token"] = []byte(token)
	return r.Update(ctx, secret)
```

controllers/observability/perses.go
```diff
@@ -0,0 +1,215 @@
...
+}
+
+// requestServiceAccountToken requests a token for the provided service account using TokenRequest API.
+func (r *Reconciler) requestServiceAccountToken(ctx context.Context, namespace, saName string) (string, error) {
+	clientset, err := kubernetes.NewForConfig(r.config)
```
nunnatsa: This should be done only once. Please add `clientset` as a new field to the Reconciler, and move this logic into the `NewReconciler` function.

sradco:
@nunnatsa thank you for this very detailed review.
I believe I addressed all you comments.

sradco:
@nunnatsa I updated the PR based on your review comments.
I also found that t the Perses operator creates the secret and and how to configure the datasource so that it can use,
so removed all this logic.
It is ready for review.

sradco:
@nunnatsa please review.
I dont think the failed tests are related to my PR.

cmd/hyperconverged-cluster-operator/main.go
```diff
@@ -231,6 +234,11 @@ func main() {
 			logger.Error(err, "unable to create controller", "controller", "Observability")
 			os.Exit(1)
 		}
+		// Register Perses controller separately for clear separation of concerns
+		if err = perses.SetupPersesWithManager(mgr, ownresources.GetDeploymentRef()); err != nil {
+			logger.Error(err, "unable to create controller", "controller", "ObservabilityPerses")
+			os.Exit(1)
+		}
```
nunnatsa: The operator pod is still getting into crash loop, because it is trying to watch unknown types, when perses is not installed.

Please check that before starting the controller; e.g.

```suggestion
		hcoutil.IsPersesAvailable(ctx, apiClient) {
			if err = perses.SetupPersesWithManager(mgr, ownresources.GetDeploymentRef()); err != nil {
				logger.Error(err, "unable to create controller", "controller", "ObservabilityPerses")
				os.Exit(1)
			}
		}
```

controllers/crd/crd_controller.go
```diff
@@ -57,6 +57,10 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 				switch object.GetName() {
 				case hcoutil.DeschedulerCRDName:
 					return true
+				case "persesdashboards.perses.dev":
+					return true
```
nunnatsa: let's add these CRD names as constants in the util package

controllers/perses/perses_resources.go
```diff
@@ -0,0 +1,92 @@
...
+			return err
+		}
+		defer file.Close()
+
+		dec := yaml.NewYAMLToJSONDecoder(file)
+		dashboard := persesv1alpha1.PersesDashboard{}
+		if err := dec.Decode(&dashboard); err != nil {
+			return err
+		}
+		dashboard.Namespace = namespace
```
nunnatsa, line 72: add common label:

```suggestion
		dashboard.Namespace = namespace
		dashboard.Labels = hcoutil.GetLabels(hcoutil.HyperConvergedName, hcoutil.AppComponentMonitoring)
```

controllers/perses/perses_resources.go
```diff
@@ -0,0 +1,92 @@
...
+	managedDashboards = slices.Collect(maps.Keys(dashboards))
+	return dashboards, nil
+}
+
+func initDatasource(namespace string) (*persesv1alpha1.PersesDatasource, error) {
+	ds := &persesv1alpha1.PersesDatasource{}
+	if err := yaml.Unmarshal(persesDatasourcesBytes, ds); err != nil {
+		return nil, err
+	}
+	ds.Namespace = namespace
```
nunnatsa, line 43: Add common labels:

```suggestion
	ds.Namespace = namespace
	ds.Labels = hcoutil.GetLabels(hcoutil.HyperConvergedName, hcoutil.AppComponentMonitoring)
```

controllers/perses/perses_controller.go
```diff
@@ -0,0 +1,278 @@
...
+		if k8serrors.IsNotFound(err) {
+			obj := db.DeepCopy()
+			r.ensureOwnerReference(obj)
+			return r.Create(ctx, obj)
+		}
+		return err
+	}
+	if !reflect.DeepEqual(found.Spec, db.Spec) {
+		db.Spec.DeepCopyInto(&found.Spec)
+		r.ensureOwnerReference(found)
```
nunnatsa: please take it out of the if block, because currently we'll only reconcile the owner reference if the spec was change.

Please also check the labels, as done for all the resources in HCO.

controllers/perses/perses_controller.go
```diff
@@ -0,0 +1,278 @@
...
+	err := r.Get(ctx, types.NamespacedName{Namespace: namespace, Name: name}, found)
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			obj := db.DeepCopy()
+			r.ensureOwnerReference(obj)
+			return r.Create(ctx, obj)
+		}
+		return err
+	}
+	if !reflect.DeepEqual(found.Spec, db.Spec) {
```
nunnatsa: add a boolean, e.g. `modified`, and use it to check if something was changed: the spec, the owner reference or the labels, and call r.Update only if something was changed.

Wrap this code into a new method to improve readability.

controllers/perses/perses_controller.go
```diff
@@ -0,0 +1,278 @@
...
+	err := r.Get(ctx, types.NamespacedName{Namespace: namespace, Name: name}, found)
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			obj := r.cachedDatasource.DeepCopy()
+			r.ensureOwnerReference(obj)
+			return r.Create(ctx, obj)
+		}
+		return err
+	}
+	if !reflect.DeepEqual(r.cachedDatasource.Spec, found.Spec) {
```
nunnatsa: same comments as in reconcileDashboard

controllers/perses/perses_test.go
```diff
@@ -0,0 +1,139 @@
+package perses
```
nunnatsa, line 1: many use cases are not covered. we'll need to add more tests.

controllers/crd/crd_controller.go
```diff
@@ -57,6 +57,10 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 				switch object.GetName() {
 				case hcoutil.DeschedulerCRDName:
```
nunnatsa: nit:

```go
case hcoutil.DeschedulerCRDName, "persesdashboards.perses.dev", "persesdashboards.perses.dev":
    return true
```

controllers/crd/crd_controller.go
```diff
@@ -99,5 +103,12 @@ func (r *ReconcileCRD) Reconcile(ctx context.Context, req reconcile.Request) (re
 		}
 	}
 
+	// If Perses CRDs are now available, restart to register Perses controller and cache the new GVKs.
+	if hcoutil.IsPersesAvailable(ctx, r.client) {
```
nunnatsa: If something is changed in the CRD, this code will cause restart even if the CRD were already installed.

WDYT about adding a new boolean field in the reconciler struct, to hold the result of `hcoutil.IsPersesAvailable` on boot time (when the reconciler is created), then check it here; for example:


```suggestion
	if !r.presesAvalableOnBoot && hcoutil.IsPersesAvailable(ctx, r.client) {
```

This will make sure we are doing the restart only once, only when perses is installed.

controllers/crd/crd_controller.go
nunnatsa: continue from previous comment:

```suggestion
		eventEmitter: hcoutil.GetEventEmitter(),
		presesAvalableOnBoot: hcoutil.IsPersesAvailable(ctx, mgr.GetClient())
		
```

nunnatsa: Add context, to be used in the code from the previous comment.


```suggestion
func newReconciler(ctx context.Context, mgr manager.Manager, restartCh chan<- struct{}) reconcile.Reconciler {
```

nunnatsa: same:

```suggestion
func RegisterReconciler(ctx context.Context, mgr manager.Manager, restartCh chan<- struct{}) error {
```

nunnatsa (commented):
Added several comments inline.

cmd/hyperconverged-cluster-operator/main.go
```diff
@@ -219,8 +225,8 @@ func main() {
 	restartCh := make(chan struct{})
 	defer close(restartCh)
 
-	// Create a new CRD reconciler
-	if err = crd.RegisterReconciler(mgr, restartCh); err != nil {
+	// Create a new CRD reconciler (compute Perses availability inside using uncached client)
+	if err = crd.RegisterReconciler(ctx, apiClient, mgr, restartCh); err != nil {
```
nunnatsa: Now, that we're reading the initial state of the crds (`persesAvailable`), we don't need to read it again from the crd controller. WDTY about just pass the boolean, and drop the ctx and the apiClient?

cmd/hyperconverged-cluster-operator/main.go
```diff
@@ -231,6 +237,13 @@ func main() {
 			logger.Error(err, "unable to create controller", "controller", "Observability")
 			os.Exit(1)
 		}
+		// Register Perses controller only if CRDs are available; otherwise avoid watching unknown types.
+		if persesAvailable {
```
nunnatsa, line 241: nice!

cmd/hyperconverged-cluster-operator/main.go
```diff
@@ -389,6 +402,16 @@ func getCacheOption(operatorNamespace string, ci hcoutil.ClusterInfo) cache.Opti
...
 
+	// Perses resources (namespaced): restrict cache to operator namespace, only when CRDs are available
+	if persesAvailable {
+		cacheOptionsByObjectForMonitoring[&persesv1alpha1.PersesDashboard{}] = cache.ByObject{
+			Field: namespaceSelector,
+		}
+		cacheOptionsByObjectForMonitoring[&persesv1alpha1.PersesDatasource{}] = cache.ByObject{
+			Field: namespaceSelector,
+		}
+	}
```
nunnatsa, line 413: nice!

controllers/crd/crd_controller.go
```diff
@@ -21,18 +21,21 @@ var (
...
+// RegisterReconciler creates a new CRD Reconciler and registers it into manager.
+// nonCachedClient should be a direct client (pre-manager) to query CRDs at boot.
+func RegisterReconciler(ctx context.Context, nonCachedClient client.Client, mgr manager.Manager, restartCh chan<- struct{}) error {
+	persesAvailableOnBoot := hcoutil.IsPersesAvailable(ctx, nonCachedClient)
+	return add(mgr, newReconciler(ctx, mgr, restartCh, persesAvailableOnBoot))
 }
 
 // newReconciler returns a new reconcile.Reconciler
-func newReconciler(mgr manager.Manager, restartCh chan<- struct{}) reconcile.Reconciler {
+func newReconciler(_ context.Context, mgr manager.Manager, restartCh chan<- struct{}, persesAvailableOnBoot bool) reconcile.Reconciler {
```
nunnatsa: we don't need the context anymore

controllers/crd/crd_controller.go
```diff
@@ -21,18 +21,21 @@ var (
 	log = logf.Log.WithName("controller_crd")
 )
 
-// RegisterReconciler creates a new Descheduler Reconciler and registers it into manager.
-func RegisterReconciler(mgr manager.Manager, restartCh chan<- struct{}) error {
-	return add(mgr, newReconciler(mgr, restartCh))
+// RegisterReconciler creates a new CRD Reconciler and registers it into manager.
+// nonCachedClient should be a direct client (pre-manager) to query CRDs at boot.
+func RegisterReconciler(ctx context.Context, nonCachedClient client.Client, mgr manager.Manager, restartCh chan<- struct{}) error {
```
nunnatsa: let's pass the boolean from main, instead of reading it again here


```suggestion
func RegisterReconciler(mgr manager.Manager, restartCh chan<- struct{}, persesAvailableOnBoot bool) error {
```

controllers/crd/crd_controller.go
```diff
@@ -21,18 +21,21 @@ var (
...
+	persesAvailableOnBoot := hcoutil.IsPersesAvailable(ctx, nonCachedClient)
```
nunnatsa: now we can drop this line

deploy/olm-catalog/community-kubevirt-hyperconverged/1.13.0/manifests/kubevirt-hyperconverged-operator.v1.13.0.clusterserviceversion.yaml
```diff
@@ -523,6 +523,19 @@ spec:
...
+          - persesdashboards
+          - persesdatasources
+          verbs:
+          - get
+          - list
+          - watch
+          - create
+          - update
+          - patch
+          - delete
```
nunnatsa: this is (a) old csv, not the one that will be used with this version; and (b) auto generated.

Please add these types in https://github.com/kubevirt/hyperconverged-cluster-operator/blob/94f9a99b6ae21650e3cbcb39c05801f4cf60b9bf/pkg/components/components.go#L689, to add:

```go
		{
			APIGroups: stringListToSlice(persesv1alpha1.GroupVersion.Group),
			Resources: stringListToSlice("persesdashboards", "persesdatasources"),
		},
```

Then run `make build manifests` and add commit the changes under `deploy/`.

config/rbac/role.yaml
```diff
@@ -13,11 +13,47 @@ rules:
   verbs:
   - create
   - delete
+- apiGroups:
```
nunnatsa: Please revert all the changes in this file. we're not using it, but the auto generated files under deploy/

PR #3837, opened 2025-11-05
Rename the NetworkPolicy common labels

nunnatsa:

Rename the `hco.kubevirt.io/allow-access-cluster-services` label to `np.kubevirt.io/allow-access-cluster-services`, and the `hco.kubevirt.io/allow-prometheus-access` label, to `np.kubevirt.io/allow-prometheus-access`m to match the KubeVirt and CDI implementations.

**Jira Ticket**:
```jira-ticket
https://issues.redhat.com/browse/CNV-72058
```

**Release note**:
```release-note
None
```

### DIFF
--- a/pkg/util/consts.go
+++ b/pkg/util/consts.go
@@ -92,11 +92,12 @@ const (
 	DataImportCronEnabledAnnotation = "dataimportcrontemplate.kubevirt.io/enable"
 
 	HCOAnnotationPrefix = "hco.kubevirt.io/"
+	NPLabelPrefix       = "np.kubevirt.io/"
 
 	// AllowEgressToDNSAndAPIServerLabel if this label is set, the network policy will allow egress to DNS and API server
-	AllowEgressToDNSAndAPIServerLabel = HCOAnnotationPrefix + "allow-access-cluster-services"
+	AllowEgressToDNSAndAPIServerLabel = NPLabelPrefix + "allow-access-cluster-services"
 	// AllowIngressToMetricsEndpointLabel if this label is set, the network policy will allow ingress to the metrics endpoint
-	AllowIngressToMetricsEndpointLabel = HCOAnnotationPrefix + "allow-prometheus-access"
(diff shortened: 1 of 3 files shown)


pkg/util/consts.go
```diff
@@ -92,11 +92,12 @@ const (
 	DataImportCronEnabledAnnotation = "dataimportcrontemplate.kubevirt.io/enable"
 
 	HCOAnnotationPrefix = "hco.kubevirt.io/"
+	NPLabelPrefix       = "np.kubevirt.io"
```
orenc1: nit:
it's inconsistent that `HCOAnnotationPrefix` is suffixed with `/`, and `NPLabelPrefix` is not.
anyway every time when using `NPLabelPrefix` it is followed by a `/` in all usages.

tools/csv-merger/network-policies-template.tmpl
```diff
@@ -72,18 +72,18 @@ spec:
...
+    - namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: openshift-monitoring
+      podSelector:
+        matchLabels:
+          alertmanager: main
+          app.kubernetes.io/component: alert-router
+          app.kubernetes.io/instance: main
+  policyTypes:
+  - Egress
```
orenc1, line 86: i think this network policy will always be rendered, even if `deployK8sDNSNetworkPolicy = true`
and in plain kubernetes there is no such namespace

nunnatsa: This is right, but there is nothing to do about it. anyway, it won't harm.

The `deployK8sDNSNetworkPolicy` flag is planed to be used to build the community version, assuming it is installed mostly on OKD/OCP clusters. But to allow also install on plain K8s with OLM, e.g. as we're doing in our nightly test, we need to add support for plain K8s as well. 

As this is done on bundle image build time, and we have no way to add conditions after the build, we must support both options.

PR #3844, opened 2025-11-10
csv-merger: allow passing csv files

nunnatsa:
The hack/build-manifests.sh pass the component CSVs to the csv-merger and the manifest-templator tools as very long strings. This is very hard to debug. Also, some tools that run the csv-merger fail to pass these long string properly.

This PR adds the `--cna-csv-file`, `--virt-csv-file`, `--ssp-csv-file`, `--cdi-csv-file`, `--hpp-csv-file`, and the `--aaq-csv-file` flags, to the csv-merger and the manifest-templator tools, in order to pass the CSV file names instead the very long csv strings.

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-72184
```

**Release note**:

```release-note
None
```

### DIFF
--- a/hack/build-manifests.sh
+++ b/hack/build-manifests.sh
@@ -278,12 +278,12 @@ done
 # Build and write deploy dir
 ${TOOLS}/manifest-templator \
   --api-sources=${PROJECT_ROOT}/api/... \
-  --cna-csv="$(<${cnaCsv})" \
-  --virt-csv="$(<${virtCsv})" \
-  --ssp-csv="$(<${sspCsv})" \
-  --cdi-csv="$(<${cdiCsv})" \
-  --hpp-csv="$(<${hppCsv})" \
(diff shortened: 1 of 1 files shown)


tools/util/csv_with_component.go
```diff
@@ -1,9 +1,110 @@
...
+		{str: aaqCsv, fileName: *aaqCsvFile, flagName: "aaq-csv"},
+	} {
+		if err := fileOrString(f.str, f.fileName, f.flagName); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func fileOrString(str *string, fileName, scvName string) error {
```
orenc1: typo?
```suggestion
func fileOrString(str *string, fileName, csvName string) error {
```

nunnatsa: yap. fixing

PR #3845, opened 2025-11-11
Free disk-space on image build

nunnatsa:
### Fix image build in github action

The github actions that builds the images, sometimes fail for out of storage in the device.

To fix it, we are now removing the local images after pushing them to the remote registry.

### Allow modifying arch for image manifests 
When building a multi-arch images, we're using a hard coded list of architectures. This is sometime take a lot of time, and during development, we sometimes want to build only for our local architecture.

Now it is possible to override the arch list by setting the `ARCHITECTURES` environment variable.

### Jira Ticket
```jira-ticket
None
```

### Release note
```release-note
None
```

### DIFF
--- a/hack/build-push-multi-arch-images.sh
+++ b/hack/build-push-multi-arch-images.sh
@@ -12,6 +12,8 @@ if [[ -z ${DOCKER_FILE} ]]; then
   exit 1
 fi
 
+CLEANUP=${CLEANUP:-"true"}
+
 SHA=$(git describe --no-match  --always --abbrev=40 --dirty)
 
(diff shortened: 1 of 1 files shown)


hack/build-push-multi-arch-images.sh
```diff
@@ -12,6 +12,8 @@ if [[ -z ${DOCKER_FILE} ]]; then
   exit 1
 fi
 
+CLEANUP=${CLEANUP:-"true"}
```
avlitman, line 15: maybe it's not needed and always clean?

nunnatsa: I can think about cases where the developer wants to keep the images, when building them locally, for testing and so on.

PR #3867, opened 2025-11-17
adding rsync install to builder

Ronilerr:
Adding install of rsync inside the builder 

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-70894

```

**Release note**:

```release-note
None

```

### DIFF
--- a/hack/builder/Dockerfile
+++ b/hack/builder/Dockerfile
@@ -1,9 +1,11 @@
 FROM quay.io/fedora/fedora:42
-RUN dnf install -y make jq git curl && \
+RUN dnf install -y make jq git curl rsync rsync-daemon && \
+    dnf clean all && \
     curl -L -O https://go.dev/dl/go1.24.7.linux-amd64.tar.gz && \
     tar -C /usr/local -xzf go1.24.7.linux-amd64.tar.gz && \
     rm -f go1.24.7.linux-amd64.tar.gz && \
     mkdir -p /go
+COPY rsyncd.conf /etc/rsyncd.conf
 ADD entrypoint.sh /usr/bin/entrypoint.sh
 ENTRYPOINT ["/usr/bin/entrypoint.sh"]
 ENV PATH=$PATH:/usr/local/go/bin:/go/bin \
@@ -11,7 +13,7 @@ ENV PATH=$PATH:/usr/local/go/bin:/go/bin \
     PACKAGE=github.com/kubevirt/hyperconverged-cluster-operator \
     API_FOLDER=api \
     API_VERSION=v1beta1 \
-    PROJECT_ROOT=/home/hco-builder/hco
+    PROJECT_ROOT=/root/hco
 # Build args for code-generator versions
 ARG K8S_VER
 ARG KUBEOPENAPI_VER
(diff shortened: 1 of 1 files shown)


hack/builder/Dockerfile
```diff
@@ -18,21 +15,22 @@ ENV PATH=$PATH:/usr/local/go/bin:/go/bin \
...
-	go install \
-	k8s.io/kube-openapi/cmd/openapi-gen@${KUBEOPENAPI_VER}
-
+    k8s.io/code-generator/cmd/deepcopy-gen@${K8S_VER} \
+    k8s.io/code-generator/cmd/defaulter-gen@${K8S_VER} && \
+    go install \
+    k8s.io/kube-openapi/cmd/openapi-gen@${KUBEOPENAPI_VER}
+# Install rsync and rsync-daemon (for compatibility across different distributions)
+# In Fedora, rsync includes daemon functionality, but rsync-daemon may exist in other distros
+RUN dnf install -y rsync && \
```
nunnatsa: Please combine with the RUN command above. I don’t think we need to split the two dnf’s

hack/builder/Dockerfile
```diff
@@ -18,21 +15,22 @@ ENV PATH=$PATH:/usr/local/go/bin:/go/bin \
...
+    (dnf install -y rsync-daemon || true) && \
```
nunnatsa: Not sure about this. Either we have the package installed in the image or it fails. Not familiar with optional step in a Dockerfile. Do I miss something?

hack/builder/rsyncd.conf
```diff
@@ -0,0 +1,19 @@
+gid = 0
+uid = 0
+log file = /dev/stdout
+reverse lookup = no
+[build]
+    hosts allow = *
+    read only = false
+    path = /root/go/src/kubevirt/hyperconverged-cluster-operator
```
nunnatsa: What do we have in this path? Nothing is there from the Dockerfile. Also,the conversion of ${HOME}/go/src/project-name is very old and we don’t have to use it.

Ronilerr: You are right, the path will be used in the Containerize script. But I had to add the rsync.conf to the PR because of: 
`COPY rsyncd.conf /etc/rsyncd.conf`
in the Docker file.

I will change the path to be /root/hco

hack/builder/Dockerfile
Ronilerr: You don't miss- I will change it.

hack/builder/Dockerfile
```diff
@@ -1,12 +1,9 @@
 FROM quay.io/fedora/fedora:42
-
-
-RUN dnf install -y make jq rsync git curl && \
+RUN dnf install -y make jq git curl && \
```
nunnatsa: Please combine with the 2nd `RUN` below; e.g.


```suggestion
RUN dnf install -y make jq git curl rsync rsync-daemon && \
    dnf clean all
```

And remove dnf from there.

hack/builder/rsyncd.conf
```diff
@@ -0,0 +1,19 @@
+gid = 0
+uid = 0
+log file = /dev/stdout
+reverse lookup = no
+[build]
+    hosts allow = *
+    read only = false
+    path = /root/hco
```
nunnatsa, line 8: This is still confusing me.

The Dockerfile's WORKDIR is `"/home/hco-builder/hco"`, but here it's `"/root/hco"`.

Is it on purpose? How are you going to run the image, eventually?

PR #3871, opened 2025-11-18
Allow virt-operator deployment on Hosted Control Planes Cluster

orenc1:
virt-operator deploymnet now requires to be scheduled only on control-plane nodes. In Hosted Control Plane cluster (HCP), there are no control plane nodes, but only worker nodes. This PR aims to address this issue, by labeling the worker nodes with additional 'node-role.kubernetes.io/control-plane' label on them, so the virt-operator pods will be scheduled and the deployment of kubevirt will succeed on such topology.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-61721
```

**Release note**:

```release-note
Allow virt-operator deployment on Hosted Control Planes Cluster
```

### DIFF
--- a/controllers/nodes/nodes_controller.go
+++ b/controllers/nodes/nodes_controller.go
@@ -3,6 +3,7 @@ package nodes
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/go-logr/logr"
 	"github.com/google/uuid"
@@ -25,6 +26,11 @@ import (
 	hcoutil "github.com/kubevirt/hyperconverged-cluster-operator/pkg/util"
 )
 
+const (
+	// HyperShift label value for worker nodes
+	hypershiftLabelValue = "set-to-allow-kubevirt-deployment"
+)
+
 var (
 	log               = logf.Log.WithName("controller_nodes")
 	randomConstSuffix = uuid.New().String()
@@ -37,19 +43,102 @@ var (
 	}
 )
 
+// startupNodeLabeler is a runnable that labels all nodes after the manager cache starts
+type startupNodeLabeler struct {
+	reconciler *ReconcileNodeCounter
+}
+
+// Start implements manager.Runnable
+func (s *startupNodeLabeler) Start(ctx context.Context) error {
+	log.Info("Starting node labeling after cache is ready")
+
+	const (
+		maxRetries    = 5
+		retryInterval = 10 * time.Second
+	)
+
+	// Label all nodes now that the cache is ready
+	err := s.reconciler.labelAllNodesAtStartup(ctx)
+	if err != nil {
+		log.Error(err, "Failed to label nodes at startup, will retry", "maxRetries", maxRetries, "retryInterval", retryInterval)
+
+		// Retry up to maxRetries times
+		for attempt := 1; attempt <= maxRetries && err != nil; attempt++ {
+			select {
+			case <-ctx.Done():
+				log.Info("Context cancelled during retry, stopping node labeling")
+				return nil
+			case <-time.After(retryInterval):
+				log.Info("Retrying node labeling at startup", "attempt", attempt, "maxRetries", maxRetries)
+				err = s.reconciler.labelAllNodesAtStartup(ctx)
+				if err == nil {
+					log.Info("Successfully labeled nodes at startup after retry", "attempt", attempt)
+				} else {
+					log.Error(err, "Failed to label nodes at startup", "attempt", attempt, "maxRetries", maxRetries)
+				}
+			}
+		}
+
+		if err != nil {
+			log.Error(err, "Failed to label nodes at startup after all retries, giving up", "maxRetries", maxRetries)
+		}
+	}
+
+	// Keep running until context is cancelled
+	<-ctx.Done()
+	return nil
+}
+
+// NeedLeaderElection implements manager.LeaderElectionRunnable
+func (s *startupNodeLabeler) NeedLeaderElection() bool {
+	return true
+}
+
 // RegisterReconciler creates a new Nodes Reconciler and registers it into manager.
 func RegisterReconciler(mgr manager.Manager, nodeEvents chan<- event.GenericEvent) error {
-	return add(mgr, newReconciler(mgr, nodeEvents))
+	reconciler := newReconciler(mgr, nodeEvents)
+
+	// Add a runnable to label all nodes after the cache starts if we're in a HyperShift cluster
+	clusterInfo := hcoutil.GetClusterInfo()
+	if clusterInfo.IsOpenshift() && clusterInfo.IsHyperShiftManaged() {
+		if err := mgr.Add(&startupNodeLabeler{reconciler: reconciler}); err != nil {
+			return fmt.Errorf("failed to add startup node labeler: %w", err)
+		}
+	}
+
+	return add(mgr, reconciler)
 }
 
 // newReconciler returns a new reconcile.Reconciler
-func newReconciler(mgr manager.Manager, nodeEvents chan<- event.GenericEvent) reconcile.Reconciler {
+func newReconciler(mgr manager.Manager, nodeEvents chan<- event.GenericEvent) *ReconcileNodeCounter {
+	clusterInfo := hcoutil.GetClusterInfo()
+
+	// Evaluate once at initialization whether we should label nodes for HyperShift
+	shouldLabelNodes := clusterInfo.IsOpenshift() && clusterInfo.IsHyperShiftManaged()
+
+	log.Info("Initializing nodes controller",
+		"isOpenshift", clusterInfo.IsOpenshift(),
+		"isHyperShiftManaged", clusterInfo.IsHyperShiftManaged(),
+		"shouldLabelNodes", shouldLabelNodes,
+	)
+
 	r := &ReconcileNodeCounter{
 		Client:     mgr.GetClient(),
 		nodeEvents: nodeEvents,
 	}
 
+	if shouldLabelNodes {
+		r.HandleHyperShiftNodeLabeling = HandleHyperShiftNodeLabeling
+	} else {
+		r.HandleHyperShiftNodeLabeling = staleHyperShiftNodeLabeling
+	}
+
 	return r
+}
+
+// staleHyperShiftNodeLabeling is a no-op function used when HyperShift node labeling is not needed
+func staleHyperShiftNodeLabeling(_ context.Context, _ client.Client, _ string, _ logr.Logger) error {
+	return nil
 }
 
 // add adds a new Controller to mgr with r as the reconcile.Reconciler
@@ -84,8 +173,9 @@ type ReconcileNodeCounter struct {
 	// This client, initialized using mgr.Client() above, is a split client
 	// that reads objects from the cache and writes to the apiserver
 	client.Client
-	HyperConvergedQueue workqueue.TypedRateLimitingInterface[reconcile.Request]
-	nodeEvents          chan<- event.GenericEvent
+	HyperConvergedQueue          workqueue.TypedRateLimitingInterface[reconcile.Request]
+	nodeEvents                   chan<- event.GenericEvent
+	HandleHyperShiftNodeLabeling func(ctx context.Context, cli client.Client, nodeName string, logger logr.Logger) error
 }
 
 // Reconcile updates the nodes count on ClusterInfo singleton
@@ -110,6 +200,15 @@ func (r *ReconcileNodeCounter) Reconcile(ctx context.Context, req reconcile.Requ
 	nodeInfoChanged, err := nodeinfo.HandleNodeChanges(ctx, r, hc, logger)
 	if err != nil {
 		return reconcile.Result{}, err
+	}
+
+	// Handle HyperShift node labeling for hosted control plane clusters
+	// Only process if this is a node event (not HCO event)
+	if req != hcoReq {
+		if err := r.HandleHyperShiftNodeLabeling(ctx, r.Client, req.Name, logger); err != nil {
+			logger.Error(err, "Failed to handle HyperShift node labeling")
+			return reconcile.Result{}, err
+		}
 	}
 
 	if hc == nil || !hc.DeletionTimestamp.IsZero() {
@@ -139,4 +238,78 @@ func (r *ReconcileNodeCounter) readHyperConverged(ctx context.Context) (*hcov1be
 	}
 
 	return hc, nil
+}
+
+// labelAllNodesAtStartup labels all worker nodes at controller startup for HyperShift clusters
+func (r *ReconcileNodeCounter) labelAllNodesAtStartup(ctx context.Context) error {
+	log.Info("Labeling all worker nodes at startup for HyperShift")
+
+	// Get all nodes
+	nodesList := &corev1.NodeList{}
+	if err := r.List(ctx, nodesList); err != nil {
+		return fmt.Errorf("failed to list nodes for HyperShift labeling at startup: %w", err)
+	}
+
+	var errs []error
+	for i := range nodesList.Items {
+		node := &nodesList.Items[i]
+		if err := labelNode(ctx, r.Client, node, log); err != nil {
+			log.Error(err, "Failed to label node at startup", "node", node.Name)
+			errs = append(errs, fmt.Errorf("node %s: %w", node.Name, err))
+		}
+	}
+
+	if len(errs) > 0 {
+		return fmt.Errorf("failed to label %d node(s): %v", len(errs), errs)
+	}
+
+	log.Info("Completed labeling nodes at startup", "totalNodes", len(nodesList.Items))
+	return nil
+}
+
+// HandleHyperShiftNodeLabeling manages the control-plane label on a specific worker node for HyperShift managed clusters
+func HandleHyperShiftNodeLabeling(ctx context.Context, cli client.Client, nodeName string, logger logr.Logger) error {
+	// Get the specific node
+	node := &corev1.Node{}
+	if err := cli.Get(ctx, client.ObjectKey{Name: nodeName}, node); err != nil {
+		if errors.IsNotFound(err) {
+			// Node was deleted, nothing to do
+			logger.V(1).Info("Node not found, skipping HyperShift labeling", "node", nodeName)
+			return nil
+		}
+		return fmt.Errorf("failed to get node %s for HyperShift labeling: %w", nodeName, err)
+	}
+
+	return labelNode(ctx, cli, node, logger)
+}
+
+// labelNode applies the HyperShift label on a single node
+func labelNode(ctx context.Context, cli client.Client, node *corev1.Node, logger logr.Logger) error {
+	if !isWorkerNode(node) {
+		return nil
+	}
+
+	logger.Info("Adding control-plane label to worker node",
+		"node", node.Name,
+		"labelValue", hypershiftLabelValue,
+	)
+
+	patch := client.MergeFrom(node.DeepCopy())
+	node.Labels[nodeinfo.LabelNodeRoleControlPlane] = hypershiftLabelValue
+
+	if err := cli.Patch(ctx, node, patch); err != nil {
+		return fmt.Errorf("failed to patch node %s: %w", node.Name, err)
+	}
+
+	logger.Info("Successfully patched node labels", "node", node.Name)
+	return nil
+}
+
+// isWorkerNode checks if a node has the worker role label
+func isWorkerNode(node *corev1.Node) bool {
+	// A node is a worker if it has the worker label and doesn't have control-plane label
+	_, hasWorkerLabel := node.Labels[nodeinfo.LabelNodeRoleWorker]
+	_, hasControlPlaneLabel := node.Labels[nodeinfo.LabelNodeRoleControlPlane]
+
+	return hasWorkerLabel && !hasControlPlaneLabel
 }
--- a/controllers/nodes/nodes_controller_test.go
+++ b/controllers/nodes/nodes_controller_test.go
@@ -8,6 +8,8 @@ import (
 	"github.com/go-logr/logr"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/event"
@@ -58,8 +60,9 @@ var _ = Describe("NodesController", func() {
 				cl := commontestutils.InitClient(resources)
 
 				r := &ReconcileNodeCounter{
-					Client:     cl,
-					nodeEvents: nodeEvents,
+					Client:                       cl,
+					nodeEvents:                   nodeEvents,
+					HandleHyperShiftNodeLabeling: staleHyperShiftNodeLabeling,
 				}
 
 				// Reconcile
@@ -77,8 +80,9 @@ var _ = Describe("NodesController", func() {
 				cl := commontestutils.InitClient(nil)
 
 				r := &ReconcileNodeCounter{
-					Client:     cl,
-					nodeEvents: nodeEvents,
+					Client:                       cl,
+					nodeEvents:                   nodeEvents,
+					HandleHyperShiftNodeLabeling: staleHyperShiftNodeLabeling,
 				}
 
 				// Reconcile
@@ -98,8 +102,9 @@ var _ = Describe("NodesController", func() {
 				cl := commontestutils.InitClient(resources)
 
 				r := &ReconcileNodeCounter{
-					Client:     cl,
-					nodeEvents: nodeEvents,
+					Client:                       cl,
+					nodeEvents:                   nodeEvents,
+					HandleHyperShiftNodeLabeling: staleHyperShiftNodeLabeling,
 				}
 
 				// Reconcile
@@ -119,8 +124,9 @@ var _ = Describe("NodesController", func() {
 				cl := commontestutils.InitClient(resources)
 
 				r := &ReconcileNodeCounter{
-					Client:     cl,
-					nodeEvents: nodeEvents,
+					Client:                       cl,
+					nodeEvents:                   nodeEvents,
+					HandleHyperShiftNodeLabeling: staleHyperShiftNodeLabeling,
 				}
 
 				// Reconcile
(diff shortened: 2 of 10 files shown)


controllers/nodes/nodes_controller.go
```diff
@@ -140,3 +183,95 @@ func (r *ReconcileNodeCounter) readHyperConverged(ctx context.Context) (*hcov1be
...
 	return hc, nil
 }
+
+// labelAllNodesAtStartup labels all worker nodes at controller startup for HyperShift clusters
+func (r *ReconcileNodeCounter) labelAllNodesAtStartup(ctx context.Context) error {
+	log.Info("Labeling all worker nodes at startup for HyperShift")
+
+	// Get all nodes
+	nodesList := &corev1.NodeList{}
+	if err := r.List(ctx, nodesList); err != nil {
```
nunnatsa, line 249: Should we filter only worker node, or we assume that in hypershift this is the status anyway?

controllers/nodes/nodes_controller.go
```diff
@@ -140,3 +183,95 @@ func (r *ReconcileNodeCounter) readHyperConverged(ctx context.Context) (*hcov1be
...
+				updatedNode.Labels[labelNodeRoleControlPlane] = hypershiftLabelValue
+				needsUpdate = true
+				logger.Info("Adding control-plane label to worker node",
+					"node", node.Name,
+					"labelValue", hypershiftLabelValue,
+				)
+			}
+		}
+	} else {
+		// Remove the label if it exists with our specific value
```
nunnatsa: What is the use case here? is it a real issue? The reason I'm asking is that if not, we can simplify the whole process by not calling the `labelAllNodesAtStartup()` and `handleHyperShiftNodeLabeling()` function at all, if `shouldLabelNodes` is false.

controllers/nodes/nodes_controller.go
```diff
@@ -140,3 +183,95 @@ func (r *ReconcileNodeCounter) readHyperConverged(ctx context.Context) (*hcov1be
...
+
+	return nil
+}
+
+// isWorkerNode checks if a node has the worker role label
+func isWorkerNode(node *corev1.Node) bool {
+	// A node is a worker if it has the worker label and doesn't have control-plane or master labels
+	_, hasWorkerLabel := node.Labels[labelNodeRoleWorker]
+	_, hasControlPlaneLabel := node.Labels[labelNodeRoleControlPlane]
+	_, hasMasterLabel := node.Labels[labelNodeRoleMaster]
```
nunnatsa: Not sure we need it. This is no longer set.

controllers/nodes/nodes_controller.go
```diff
@@ -140,3 +183,95 @@ func (r *ReconcileNodeCounter) readHyperConverged(ctx context.Context) (*hcov1be
...
+// labelNode applies or removes the HyperShift label on a single node
+func (r *ReconcileNodeCounter) labelNode(ctx context.Context, node *corev1.Node, logger logr.Logger) error {
+	needsUpdate := false
+	updatedNode := node.DeepCopy()
+
+	if r.shouldLabelNodes {
+		// Add label to worker nodes that don't have it
+		if isWorkerNode(node) {
+			currentValue, hasLabel := updatedNode.Labels[labelNodeRoleControlPlane]
+			if !hasLabel || currentValue != hypershiftLabelValue {
```
nunnatsa: If we won't check for the old master label in `isWorkerNode()`, we could know for sure that the label does not exist.

Also, with the current implementation of `isWorkerNode()`, we won't get here if the control-plane label does exist, and so the whole logic of replacing the label value can't really happen.

controllers/nodes/nodes_controller.go
```diff
@@ -140,3 +183,95 @@ func (r *ReconcileNodeCounter) readHyperConverged(ctx context.Context) (*hcov1be
...
+
+// isWorkerNode checks if a node has the worker role label
+func isWorkerNode(node *corev1.Node) bool {
+	// A node is a worker if it has the worker label and doesn't have control-plane or master labels
+	_, hasWorkerLabel := node.Labels[labelNodeRoleWorker]
+	_, hasControlPlaneLabel := node.Labels[labelNodeRoleControlPlane]
+	_, hasMasterLabel := node.Labels[labelNodeRoleMaster]
+
+	return hasWorkerLabel && !hasControlPlaneLabel && !hasMasterLabel
+}
```
nunnatsa, line 315: WDYT about this?

```suggestion
func isWorkerNode(node *corev1.Node) bool {
	// A node is a worker if it has the worker label and doesn't have control-plane or master labels
	_, hasWorkerLabel := node.Labels[labelNodeRoleWorker]
	lblVal, hasControlPlaneLabel := node.Labels[labelNodeRoleControlPlane]

	return hasWorkerLabel &&
		(!hasControlPlaneLabel || lblVal != hypershiftLabelValue)
}
```

controllers/nodes/nodes_controller.go
```diff
@@ -39,14 +49,37 @@ var (
 
 // RegisterReconciler creates a new Nodes Reconciler and registers it into manager.
 func RegisterReconciler(mgr manager.Manager, nodeEvents chan<- event.GenericEvent) error {
-	return add(mgr, newReconciler(mgr, nodeEvents))
+	reconciler := newReconciler(mgr, nodeEvents)
+
+	// Label all nodes at startup if needed
+	if r, ok := reconciler.(*ReconcileNodeCounter); ok && r.shouldLabelNodes {
```
nunnatsa: We can maybe simplify this code by returning `*ReconcileNodeCounter` from the `newReconciler()` function, then do this:


```suggestion
	if reconciler.shouldLabelNodes {
```

I think it should work.

controllers/nodes/nodes_controller.go
```diff
@@ -37,16 +47,62 @@ var (
...
+
+// Start implements manager.Runnable
+func (s *startupNodeLabeler) Start(ctx context.Context) error {
+	log.Info("Starting node labeling after cache is ready")
+
+	// Label all nodes now that the cache is ready
+	if err := s.reconciler.labelAllNodesAtStartup(ctx); err != nil {
+		log.Error(err, "Failed to label nodes at startup")
+		// Don't return error to avoid stopping the manager
+	}
```
nunnatsa: Maybe something like this?

```suggestion
	err := s.reconciler.labelAllNodesAtStartup(ctx)
	if err != nil {
		log.Error(err, "Failed to label nodes at startup")
		// Don't return error to avoid stopping the manager
		for range time.Tick(30 * time.Second) {
			if err = s.reconciler.labelAllNodesAtStartup(ctx); err == nil {
				break
			}
			
			log.Error(err, "Failed to label nodes at startup")
		}
	}
```

controllers/nodes/nodes_controller.go
```diff
@@ -37,16 +47,62 @@ var (
...
+		"shouldLabelNodes", shouldLabelNodes,
+	)
+
 	r := &ReconcileNodeCounter{
-		Client:     mgr.GetClient(),
-		nodeEvents: nodeEvents,
+		Client:           mgr.GetClient(),
+		nodeEvents:       nodeEvents,
+		shouldLabelNodes: shouldLabelNodes,
 	}
```
nunnatsa, line 129: What if we'll add a new `handleHyperShiftNodeLabeling` filed to the reconciler, with the type of `func (ctx context.Context, cli client.Client, nodeName string, logger logr.Logger) error`?

Then we'll change the `handleHyperShiftNodeLabeling` and the `labelNode` methods, to be functions (not part of the reconciler), and to get new parameter of client,

then we could add here:

```suggestion
	if shouldLabelNodes {
		r.handleHyperShiftNodeLabeling = handleHyperShiftNodeLabeling
	} else {
		r.handleHyperShiftNodeLabeling = func(_ context.Context, _ client.Client, _ string, _ logr.Logger) error {
			return nil
		}
	}
```

And then, we we can call it (almost) as before (just need to pass the client).

Thank wat we wan't need to ask every time is we're in an hypershift cluster, while we already know we do.

nunnatsa (commented):
Impressive work @orenc1 .

Added a few questions and comments.

controllers/nodes/nodes_controller.go
orenc1: yes, makes sense.

controllers/nodes/nodes_controller.go
```diff
@@ -25,6 +25,16 @@ import (
 	hcoutil "github.com/kubevirt/hyperconverged-cluster-operator/pkg/util"
 )
 
+const (
+	// Node role labels
+	labelNodeRoleControlPlane = "node-role.kubernetes.io/control-plane"
+	labelNodeRoleMaster       = "node-role.kubernetes.io/master"
+	labelNodeRoleWorker       = "node-role.kubernetes.io/worker"
```
nunnatsa: I guess you had to re-define these constants, because you can't access the consts already defined in pkg/internal/nodeinfo/high_availability.go

If you wish to solve it, you can do one of the following (3 - easiest; 1 - best solution):
1. move the `pkg/internal` dir to the project root - it will allow any package in the project to use it, and maybe it's a good idea anyway.
2. move them into a new package out of the `internal` dir, but make sure this package import nothing, to avoid unintended dependencies.
3. redefine the consts in pkg/nodeinfo/nodeinfo.go; e.g.
    ```go
    const (
    	// LabelNodeRoleControlPlane is the label used to identify control plane nodes
    	LabelNodeRoleControlPlane = internal.LabelNodeRoleControlPlane
    	// LabelNodeRoleMaster is the old label used to identify control plane nodes
    	LabelNodeRoleMaster = internal.LabelNodeRoleMaster
    	// LabelNodeRoleWorker is the label used to identify worker nodes
    	LabelNodeRoleWorker = internal.LabelNodeRoleWorker
    )
    ````

controllers/nodes/nodes_controller.go
orenc1: done, thanks.
it required a slight refactoring, and to convert some methods to standalone functions.
but now we don't need to check the condition at every reconciliation run, since the decision is made in the controller initialization.

controllers/nodes/nodes_controller.go
orenc1: if we are here we're in an hypershift cluster, and that means all nodes are workers (and we should label them with control-plane as well).

controllers/nodes/nodes_controller.go
orenc1: we're not replacing the label, but adding a control plane label for nodes that have only the worker label.
the CP label can exist if they set previously by a previous run of hco-operator.

controllers/nodes/nodes_controller.go
orenc1: the `node-role.kubernetes.io/master` label was removed in kuberentes since 1.24, but in openshift it still exists.
both `node-role.kubernetes.io/control-plane` and `node-role.kubernetes.io/master` are being set on master nodes.
but for our use case, i think we can remove it since it won't be seen anyway on hypershift cluster nodes.

controllers/nodes/nodes_controller.go
orenc1: removed master.

controllers/nodes/nodes_controller.go
```diff
@@ -37,16 +48,78 @@ var (
...
+	if err != nil {
+		log.Error(err, "Failed to label nodes at startup")
+		// Don't return error to avoid stopping the manager
+		for range time.Tick(30 * time.Second) {
+			if err = s.reconciler.labelAllNodesAtStartup(ctx); err == nil {
+				break
+			}
+
+			log.Error(err, "Failed to label nodes at startup")
+		}
```
nunnatsa, line 84: Reading it again, it's a bit risky. If there is a constant error, the controller is stuck, as there is no exit from the loop. We can limit the number of retries, e.g.

```go
const ( // just examples
	numRetries = 5
	waitTime = 10 * time.Second)
)
```

```suggestion
	if err != nil {
		retries := 0
		// Don't return error to avoid stopping the manager
		for range time.Tick(10 * time.Second) {
			if err = s.reconciler.labelAllNodesAtStartup(ctx); err == nil {
				break
			}
			retries++
			
			if retries >= numRetries {
				log.Error(err, "Failed to label nodes at startup")
				break
			}
		}
```

controllers/nodes/nodes_controller.go
```diff
@@ -37,16 +48,78 @@ var (
...
-		nodeEvents: nodeEvents,
+		Client:           mgr.GetClient(),
+		nodeEvents:       nodeEvents,
+		shouldLabelNodes: shouldLabelNodes,
+	}
+
+	if shouldLabelNodes {
+		r.HandleHyperShiftNodeLabeling = HandleHyperShiftNodeLabeling
+	} else {
+		r.HandleHyperShiftNodeLabeling = func(_ context.Context, _ client.Client, _ string, _ logr.Logger) error {
```
nunnatsa: Let's also have something like:
```go
func staleHyperShiftNodeLabeling(_ context.Context, _ client.Client, _ string, _ logr.Logger) error {
	return nil
}
```

And use it here, and also in the unit tests (it re-defined there multiple times).

controllers/nodes/nodes_controller.go
```diff
@@ -140,3 +224,83 @@ func (r *ReconcileNodeCounter) readHyperConverged(ctx context.Context) (*hcov1be
...
+	nodesList := &corev1.NodeList{}
+	if err := r.List(ctx, nodesList); err != nil {
+		return fmt.Errorf("failed to list nodes for HyperShift labeling at startup: %w", err)
+	}
+
+	for i := range nodesList.Items {
+		node := &nodesList.Items[i]
+		if err := labelNode(ctx, r.Client, node, log); err != nil {
+			log.Error(err, "Failed to label node at startup", "node", node.Name)
+			// Continue with other nodes even if one fails
```
nunnatsa: this is the right thing to do, not returning error here. However, we want to make sure we had completed the labeling within `startupNodeLabeler.Start()`, and according the current implementation, we only check for error.

WDYT about adding a bool return value? e.g.

```go
	completed := true
	for i := range nodesList.Items {
		if err := labelNode(ctx, r.Client, node, log); err != nil {
			log.Error(err, "Failed to label node at startup", "node", node.Name)
			completed = false
		}
	}

	if completed {
		g.Info("Completed labeling nodes at startup", "totalNodes", len(nodesList.Items))
	}

	return completed, nil
```

But maybe we don't even need the error return value.

controllers/nodes/nodes_controller.go
```diff
@@ -140,3 +224,83 @@ func (r *ReconcileNodeCounter) readHyperConverged(ctx context.Context) (*hcov1be
...
+	updatedNode := node.DeepCopy()
+
+	// Add label to worker nodes that don't have it
+	if isWorkerNode(node) {
+		currentValue, hasLabel := updatedNode.Labels[labelNodeRoleControlPlane]
+		if !hasLabel || currentValue != hypershiftLabelValue {
+			if updatedNode.Labels == nil {
+				updatedNode.Labels = make(map[string]string)
+			}
+			updatedNode.Labels[labelNodeRoleControlPlane] = hypershiftLabelValue
```
nunnatsa: According to the current implementation of `isWorkerNode()`, we get here only if `hasWorkerLabel && !hasControlPlaneLabel`.

That means that `hasLabel` is ***always false***, and the `currentValue` can never be populated; i.e. `!hasLabel` is always true, and `currentValue != hypershiftLabelValue` is always false, the condition here is redundant.

Next: since `hasWorkerLabel` is always true, then `updatedNode.Labels` can't be nil, making the second condition redundant as well.

If we want to handle the case of CP label with different value, we can use the suggested implementation of `isWorkerNode` in my previous review.

But either ways, we can just
```go
if isWorkerNode(node) {
	updatedNode.Labels[labelNodeRoleControlPlane] = hypershiftLabelValue
	...
}
```

controllers/nodes/nodes_controller.go
```diff
@@ -140,3 +224,83 @@ func (r *ReconcileNodeCounter) readHyperConverged(ctx context.Context) (*hcov1be
...
+			needsUpdate = true
```
nunnatsa: since we no longer remove the label, and actually have only one use-case, then we don't need the `needsUpdate` variable at all - just add the update logic here instead.

controllers/nodes/nodes_controller.go
```diff
@@ -140,3 +224,83 @@ func (r *ReconcileNodeCounter) readHyperConverged(ctx context.Context) (*hcov1be
...
+			logger.Info("Adding control-plane label to worker node",
+				"node", node.Name,
+				"labelValue", hypershiftLabelValue,
+			)
+		}
+	}
+
+	if needsUpdate {
+		if err := cli.Update(ctx, updatedNode); err != nil {
```
nunnatsa: Please consider using Patch instead of Update.

deploy/index-image/community-kubevirt-hyperconverged/1.17.0/manifests/kubevirt-hyperconverged-operator.v1.17.0.clusterserviceversion.yaml
```diff
@@ -322,6 +322,7 @@ spec:
           - get
           - list
           - watch
+          - update
```
nunnatsa: patch?

pkg/components/components.go
```diff
@@ -544,7 +544,7 @@ func GetClusterPermissions() []rbacv1.PolicyRule {
 		{
 			APIGroups: emptyAPIGroup,
 			Resources: stringListToSlice("pods", "nodes"),
-			Verbs:     stringListToSlice("get", "list", "watch"),
+			Verbs:     stringListToSlice("get", "list", "watch", "update"),
```
nunnatsa: patch?

controllers/nodes/nodes_controller_test.go
```diff
@@ -60,6 +62,9 @@ var _ = Describe("NodesController", func() {
 				r := &ReconcileNodeCounter{
 					Client:     cl,
 					nodeEvents: nodeEvents,
+					HandleHyperShiftNodeLabeling: func(_ context.Context, _ client.Client, _ string, _ logr.Logger) error {
```
nunnatsa: If you'll have a stale function - use it here.

controllers/nodes/nodes_controller_test.go
```diff
@@ -79,6 +84,9 @@ var _ = Describe("NodesController", func() {
...
```
nunnatsa: same

controllers/nodes/nodes_controller_test.go
```diff
@@ -100,6 +108,9 @@ var _ = Describe("NodesController", func() {
...
```
nunnatsa: same

controllers/nodes/nodes_controller_test.go
```diff
@@ -121,6 +132,9 @@ var _ = Describe("NodesController", func() {
...
```
nunnatsa: same

controllers/nodes/nodes_controller_test.go
```diff
@@ -131,5 +145,431 @@ var _ = Describe("NodesController", func() {
...
+				Expect(res.IsZero()).To(BeTrue())
+
+				// Verify the node was not labeled
+				updatedNode := &corev1.Node{}
+				err = cl.Get(context.TODO(), client.ObjectKey{Name: "worker-1"}, updatedNode)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(updatedNode.Labels).NotTo(HaveKey(labelNodeRoleControlPlane))
+			})
+
+			It("Should remove HyperShift label when shouldLabelNodes is false", func() {
```
nunnatsa: I don't understand this test. It does not check the actual behavior, but adds a fake one, and check it.

Please drop this test.

controllers/nodes/nodes_controller_test.go
```diff
@@ -131,5 +145,431 @@ var _ = Describe("NodesController", func() {
...
+				Expect(res.IsZero()).To(BeTrue())
+
+				// Verify the HyperShift label was removed
+				updatedNode := &corev1.Node{}
+				err = cl.Get(context.TODO(), client.ObjectKey{Name: "worker-1"}, updatedNode)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(updatedNode.Labels).NotTo(HaveKey(labelNodeRoleControlPlane))
+			})
+
+			It("Should not remove control-plane label with different value", func() {
```
nunnatsa: Same

controllers/nodes/nodes_controller.go
```diff
@@ -84,8 +157,10 @@ type ReconcileNodeCounter struct {
 	// This client, initialized using mgr.Client() above, is a split client
 	// that reads objects from the cache and writes to the apiserver
 	client.Client
-	HyperConvergedQueue workqueue.TypedRateLimitingInterface[reconcile.Request]
-	nodeEvents          chan<- event.GenericEvent
+	HyperConvergedQueue          workqueue.TypedRateLimitingInterface[reconcile.Request]
+	nodeEvents                   chan<- event.GenericEvent
+	shouldLabelNodes             bool // Cached value for HyperShift node labeling decision
```
nunnatsa: I tend to think this filed can be dropped. It can be moved to be a local variable in `RegisterReconciler()`, and then passed to `newReconciler` as a new boolean parameter. But these are the only places we really need it: decide if we use the `startupNodeLabeler`, and if to use the actual label function, or the stale one. both are happening on startup, and so the field is not needed later.

controllers/nodes/nodes_controller.go
orenc1: i think that in this case it's equivalent. if there is no error - the labeling has been completed successfully, and if an error is returned, the labeling did not complete.
i'm not sure there's a need to return both err + boolean

controllers/nodes/nodes_controller_test.go
orenc1: this handles a theoretical use case in which the control plane topology is no longer external, and then the CP labels should be removed.
but this will not happen in reality, so i can drop this test.

controllers/nodes/nodes_controller.go
orenc1: `shouldLabelNodes` field has been removed from the struct and is now a local variable in `newReconciler()`

controllers/nodes/nodes_controller.go
```diff
@@ -37,21 +43,103 @@ var (
...
+		for attempt := 1; attempt <= maxRetries; attempt++ {
+			select {
+			case <-ctx.Done():
+				log.Info("Context cancelled during retry, stopping node labeling")
+				return nil
+			case <-time.After(retryInterval):
+				log.Info("Retrying node labeling at startup", "attempt", attempt, "maxRetries", maxRetries)
+				if err = s.reconciler.labelAllNodesAtStartup(ctx); err == nil {
+					log.Info("Successfully labeled nodes at startup after retry", "attempt", attempt)
+					break
```
nunnatsa: I think will break from the case but not from the loop. See here: https://go.dev/ref/spec#Break_statements

controllers/nodes/nodes_controller.go
```diff
@@ -37,21 +43,103 @@ var (
...
+
+	// Label all nodes now that the cache is ready
+	err := s.reconciler.labelAllNodesAtStartup(ctx)
+	if err != nil {
+		log.Error(err, "Failed to label nodes at startup, will retry", "maxRetries", maxRetries, "retryInterval", retryInterval)
+
+		// Retry up to maxRetries times
+		for attempt := 1; attempt <= maxRetries; attempt++ {
+			select {
+			case <-ctx.Done():
```
nunnatsa, line 68: Not sure I understand the required logic of the Start function. I read that we must wait for `ctx.Done()`. But here, if we exit for completing the labeling, then we miss it. We'll also miss it it if the very first attempt, before the loop, will succeeded (and there is no real reason it won't).

So I think we'll need another one after the loop.  But again - not sure about the requirements.

controllers/nodes/nodes_controller.go
nunnatsa: Since we don't returning error for a single node failure, but only for failing to read the node list, then the cases are not equivalent.

so if you want to use the error return value, but not adding the completed, you can maybe do something like:
```go
	var labelingErr error
	for i := range nodesList.Items {
		if err := labelNode(ctx, r.Client, node, log); err != nil {
			log.Error(err, "Failed to label node at startup", "node", node.Name)
			labelingErr = err
		}
	}

	if labelingErr == nil {
		g.Info("Completed labeling nodes at startup", "totalNodes", len(nodesList.Items))
	}

	return labelingErr
```

controllers/nodes/nodes_controller.go
orenc1: the idea was to break the loop if the context is cancelled (if the operator is stopped, manager is shutting down or the pod terminated). that way the loop will exit cleanly.
we don't need to wait for the context to cancel, if we successfully labeled all nodes or if the max retries has reached.

controllers/nodes/nodes_controller.go
orenc1: good point.
i modified the function to append errors for every node labeling failure, and return an error if the slice length is non-zero.

controllers/nodes/nodes_controller.go
nunnatsa: The documentation is not clear to me.

```go
// Runnable allows a component to be started.
// It's very important that Start blocks until
// it's done running.
type Runnable interface {
	// Start starts running the component.  The component will stop running
	// when the context is closed. Start blocks until the context is closed or
	// an error occurs.
	Start(context.Context) error
}
```

Not sure I understand how to use it.

controllers/nodes/nodes_controller.go
```diff
@@ -140,3 +239,77 @@ func (r *ReconcileNodeCounter) readHyperConverged(ctx context.Context) (*hcov1be
...
+		node := &nodesList.Items[i]
+		if err := labelNode(ctx, r.Client, node, log); err != nil {
+			log.Error(err, "Failed to label node at startup", "node", node.Name)
+			errs = append(errs, fmt.Errorf("node %s: %w", node.Name, err))
+		}
+	}
+
+	if len(errs) > 0 {
+		return fmt.Errorf("failed to label %d node(s): %v", len(errs), errs)
+	}
```
nunnatsa, line 264: Nit: you can use the standard lib `errors.Join` function; e.g.

```suggestion
	if err := errors.Join(errs...); err != nil {
		return fmt.Errorf("failed to label %d node(s): %w", len(errs), err)
	}
```

PR #3944, opened 2025-12-30
Report recommended CPU models for heterogeneous clusters

dankenigsberg:
In heterogeneous clusters with multiple CPU vendors and generations, choosing the best CPU model for VMs is challenging:

- Using a newer CPU model limits VM placement to a subset of nodes
- Using an older model impacts guest performance unnecessarily

This change adds automatic CPU model recommendations to the HyperConverged status. The recommendations are sorted by a weighted score that balances:

- PassMark performance score (50%) - for guest performance
- Available CPU cores (20%) - for workload capacity
- Available memory (15%) - for workload capacity
- Node count (15%) - for availability and redundancy

The top recommendations appear in status.nodeInfo.recommendedCpuModels, helping cluster administrators make informed decisions when setting spec.defaultCPUModel.

The CPU model data is gathered from node labels (cpu-model.node.kubevirt.io/*) set by KubeVirt's virt-handler, and is automatically refreshed when nodes change or hourly.

Assisted-by: claude-4-sonnet

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-42906
```

**Release note**:

```release-note
HCO CR status recommends cpu models for VMs
```


api/v1beta1/hyperconverged_types.go
```diff
@@ -856,6 +856,22 @@ type NodeInfoStatus struct {
 	WorkloadsArchitectures []string `json:"workloadsArchitectures,omitempty"`
 	// ControlPlaneArchitectures is a distinct list of the CPU architecture of the control-plane nodes.
 	ControlPlaneArchitectures []string `json:"controlPlaneArchitectures,omitempty"`
+	// RecommendedCpuModels is a list of recommended CPU models for the cluster based on available nodes
+	RecommendedCpuModels []CpuModelInfo `json:"recommendedCpuModels,omitempty"`
```
nunnatsa, line 863: Please add list type and key. This will ensure that we only have one object for each CPU name:

```suggestion
	// +listType=map
	// +listMapKey=name
	RecommendedCpuModels []CpuModelInfo `json:"recommendedCpuModels,omitempty"`
```

api/v1beta1/hyperconverged_types.go
```diff
@@ -856,6 +856,22 @@ type NodeInfoStatus struct {
...
+// CpuModelInfo contains information about a CPU model and its availability in the cluster
+type CpuModelInfo struct {
+	// Name is the CPU model name
+	Name string `json:"name"`
+	// Benchmark is the CPU performance score for this model
+	Benchmark int `json:"benchmark"`
+	// Nodes is the number of nodes that support this CPU model
+	Nodes int `json:"nodes"`
+	// Cpu is the total CPU cores available across all nodes supporting this model (in Kubernetes resource format, e.g., "4" or "4000m")
+	Cpu string `json:"cpu"`
```
nunnatsa: `Cpu` => `CPU` ?

api/v1beta1/hyperconverged_types.go
```diff
@@ -856,6 +856,22 @@ type NodeInfoStatus struct {
 	WorkloadsArchitectures []string `json:"workloadsArchitectures,omitempty"`
 	// ControlPlaneArchitectures is a distinct list of the CPU architecture of the control-plane nodes.
 	ControlPlaneArchitectures []string `json:"controlPlaneArchitectures,omitempty"`
+	// RecommendedCpuModels is a list of recommended CPU models for the cluster based on available nodes
+	RecommendedCpuModels []CpuModelInfo `json:"recommendedCpuModels,omitempty"`
+}
+
+// CpuModelInfo contains information about a CPU model and its availability in the cluster
+type CpuModelInfo struct {
```
nunnatsa, line 867: All the fields in this struct are mandatory, because there in no "emitempty" directive for them.

Is that intentional?

api/v1beta1/hyperconverged_types.go
nunnatsa: Can we use `*resource.Quantity` instead of string?

api/v1beta1/hyperconverged_types.go
```diff
@@ -856,6 +856,22 @@ type NodeInfoStatus struct {
...
+	// Name is the CPU model name
+	Name string `json:"name"`
+	// Benchmark is the CPU performance score for this model
+	Benchmark int `json:"benchmark"`
+	// Nodes is the number of nodes that support this CPU model
+	Nodes int `json:"nodes"`
+	// Cpu is the total CPU cores available across all nodes supporting this model (in Kubernetes resource format, e.g., "4" or "4000m")
+	Cpu string `json:"cpu"`
+	// Memory is the total memory available across all nodes supporting this model (in Kubernetes resource format, e.g., "24Gi")
+	Memory string `json:"memory"`
```
nunnatsa: Can we use `*resource.Quantity` instead of string?

pkg/internal/nodeinfo/cpu_models.go
```diff
@@ -0,0 +1,333 @@
...
+// calculateWeightedScore computes a weighted recommendation score considering:
+// - PassMark performance (50% weight) - CPU performance indicator
+// - Available CPU cores (20% weight) - fraction of cluster CPU
+// - Memory (15% weight) - fraction of cluster memory
+// - Number of nodes (15% weight) - fraction of cluster nodes
+func calculateWeightedScore(model v1beta1.CpuModelInfo, totalNodes int, totalCpu, totalMemory float64) float64 {
+	// PassMark score (50% weight) - normalized to 0-100 scale
+	// Assuming max PassMark of ~65000 (Zen 4 family)
+	passMark := float64(model.Benchmark)
+	passMarkScore := (passMark / 65000.0) * 100.0 * 0.50
```
nunnatsa: Please avoid magic number, and use a const instead of `65000.0`, and `100.0 * 0.50`

pkg/internal/nodeinfo/cpu_models.go
```diff
@@ -0,0 +1,333 @@
...
+
+	// CPU (20% weight) - fraction of total cluster CPU
+	cpu := parseCpuString(model.Cpu)
+	cpuScore := 0.0
+	if totalCpu > 0 {
+		cpuScore = (cpu / totalCpu) * 100.0 * 0.20
```
nunnatsa: Please avoid magic numbers

pkg/internal/nodeinfo/cpu_models.go
```diff
@@ -0,0 +1,333 @@
...
+	}
+
+	// Memory (15% weight) - fraction of total cluster memory
+	memoryGB := parseMemoryString(model.Memory)
+	memoryScore := 0.0
+	if totalMemory > 0 {
+		memoryScore = (memoryGB / totalMemory) * 100.0 * 0.15
```
nunnatsa: same

pkg/internal/nodeinfo/cpu_models.go
```diff
@@ -0,0 +1,333 @@
...
+
+func getCpuModelPassMarkScore(cpuModel string) int {
+	// Try exact match first
+	if score, exists := cpuModelPassMarkScores[cpuModel]; exists {
+		return score
+	}
+
+	// Try to find a partial match for models with suffixes
+	for modelName, score := range cpuModelPassMarkScores {
+		if strings.Contains(cpuModel, modelName) || strings.HasPrefix(cpuModel, modelName) {
```
nunnatsa: if cpuModek has a prefix of modelName, it also contains it, right?


```suggestion
		if strings.Contains(cpuModel, modelName) {
```

pkg/internal/nodeinfo/cpu_models.go
```diff
@@ -0,0 +1,333 @@
...
+
+		memStr := "0"
+		if memory[model] != nil {
+			memStr = memory[model].String()
+		}
+
+		part := model + ":" + strconv.Itoa(counts[model]) + ":" +
+			cpuStr + ":" + memStr
+		parts = append(parts, part)
+	}
```
nunnatsa: what is the purpose of the `models` slice? also, we do need the value, not only the key. can we do someting like this? Or do I miss something?

```suggestion
	for model, count := range counts {
		cpuStr := "0"
		if cores[model] != nil {
			cpuStr = cores[model].String()
		}

		memStr := "0"
		if memory[model] != nil {
			memStr = memory[model].String()
		}

		part := model + ":" + strconv.Itoa(count) + ":" +
			cpuStr + ":" + memStr
		parts = append(parts, part)
	}
```

pkg/internal/nodeinfo/cpu_models.go
```diff
@@ -0,0 +1,333 @@
...
+			cpuStr = cores[model].String()
+		}
+
+		memStr := "0"
+		if memory[model] != nil {
+			memStr = memory[model].String()
+		}
+
+		part := model + ":" + strconv.Itoa(counts[model]) + ":" +
+			cpuStr + ":" + memStr
```
nunnatsa: IIRC, string concatenation is not so efficient. `fmt.Sprinf` is not great either, but more readable. 

so we can do:

```suggestion
		part := fmt.Sprintf("%s:%d:%s:%s", model, strconv.Itoa(counts[model]), cpuStr, memStr)
```

`strings.Builder` considered to be more efficient, but less readable; e.g.

```suggestion
	b := strings.Builder{}
	b.WriteString(model)
	b.WriteByte(':')
	...
	parts = append(parts, b.String())
```

pkg/internal/nodeinfo/cpu_models_test.go
```diff
@@ -0,0 +1,164 @@
...
+						},
+						Allocatable: corev1.ResourceList{
+							corev1.ResourceCPU:    resource.MustParse("1.5"),
+							corev1.ResourceMemory: resource.MustParse("3Gi"),
+						},
+					},
+				},
+			}
+
+			ctx := context.Background()
```
nunnatsa: Please use the `It()` context instead. 

e.g.
```go
It("should exclude models that appear only once on a weak node", func(ctx context.Context) {
```

pkg/internal/nodeinfo/cpu_models_test.go
```diff
@@ -0,0 +1,164 @@
...
+		It("should correctly aggregate CPU across nodes", func() {
+			if skylake, exists := modelMap["Skylake"]; exists {
+				Expect(skylake.Cpu).To(Equal("16"))
+			}
+			if haswell, exists := modelMap["Haswell"]; exists {
+				Expect(haswell.Cpu).To(Equal("24")) // 16 + 8
+			}
+			if broadwell, exists := modelMap["Broadwell"]; exists {
+				Expect(broadwell.Cpu).To(Equal("8"))
+			}
```
nunnatsa: Since we control the input, I think we should expect the map keys, because if we did something wrong, and we won't have these key in the map, the test won't find them.

WDYT about this (two options)?

```suggestion
			Expext(odelMap).To(HaveKey("Skylake")
			Expect(modelMap["Skylake"].Cpu).To(Equal("16"))
			// OR
			haswell, exists := modelMap["Haswell"]
			Expect(exists).To(BeTrue())			
			Expect(haswell.Cpu).To(Equal("24")) // 16 + 8
			...
```

pkg/internal/nodeinfo/cpu_models.go
```diff
@@ -0,0 +1,333 @@
...
+	"github.com/kubevirt/hyperconverged-cluster-operator/api/v1beta1"
+)
+
+const (
+	// CPU model labels prefix used by KubeVirt
+	CpuModelLabelPrefix = "cpu-model.node.kubevirt.io/"
+)
+
+// CPU model PassMark scores mapping
+var cpuModelPassMarkScores = map[string]int{
```
nunnatsa, line 29: what is the source for that? can we somehow auto-generate it?

deploy/olm-catalog/community-kubevirt-hyperconverged/1.16.0/manifests/hco00.crd.yaml
```diff
@@ -5130,6 +5130,54 @@ spec:
                     items:
```
nunnatsa, line 5130: Please revert the changes in this file.

pkg/internal/nodeinfo/cpu_models.go
```diff
@@ -0,0 +1,333 @@
...
+
+var (
+	cpuModelInfo = &CpuModelInfo{
+		models:         make(map[string]*v1beta1.CpuModelInfo),
+		lock:           &sync.RWMutex{},
+		lastUpdateHash: "",
+	}
+)
+
+type CpuModelInfo struct {
```
nunnatsa: I find it confusing that this struct type, shares the same name with the API type.

Can we rename it?

pkg/internal/nodeinfo/cpu_models.go
```diff
@@ -0,0 +1,333 @@
...
+	models         map[string]*v1beta1.CpuModelInfo
+	lock           *sync.RWMutex
+	lastUpdateHash string
+	// Cluster totals for normalization
+	totalNodes  int
+	totalCpu    float64
+	totalMemory float64
+}
+
+func GetRecommendedCpuModels() []v1beta1.CpuModelInfo {
```
nunnatsa, line 202: This is a relatively pricey function, but it called for each reconcile loop, while the input for it only modified upon a node event. Can we run its logic as part of `processCpuModels`, and this function will only return the cached result?

pkg/internal/nodeinfo/cpu_models.go
```diff
@@ -0,0 +1,333 @@
...
+			Memory:    memStr,
+		}
+	}
+
+	cpuModelInfo.lastUpdateHash = newHash
+	return true
+}
+
+func extractCpuModels(labels map[string]string) []string {
+	cpuModels := sets.New[string]()
```
nunnatsa, line 348: Not sure we need a set here, since `labels` is a map, we can have more than one `CpuModelLabelPrefix + aSpecificModelName` key. Consider use a simple slice here.

api/v1beta1/hyperconverged_types.go
dankenigsberg: Is map fine for a list like this, where order is of high significance?

api/v1beta1/hyperconverged_types.go
dankenigsberg: yes, it would be a bug to recommend a cpu model if we lack information about it.

api/v1beta1/hyperconverged_types.go
dankenigsberg: ok

pkg/internal/nodeinfo/cpu_models.go
dankenigsberg: oh, I don't like this sloppy search. I'll try to be exact.

pkg/internal/nodeinfo/cpu_models.go
dankenigsberg: oh, it's far worse. Generating the hash and comparing it is as complex as doing a deep comparison.

pkg/internal/nodeinfo/cpu_models_test.go
dankenigsberg: I'll try

api/v1beta1/hyperconverged_types.go
nunnatsa: It's not a map, it's a validation rule, makes sure to only have one object for a specific name. The go type is still a slice.

nunnatsa (commented):
Thanks for fixing the comment, Added some more inline. 

Please also handle the linter issues (redefinition of the builtin function `max`).

pkg/internal/nodeinfo/cpu_models.go
dankenigsberg: answered in comment.

pkg/internal/nodeinfo/cpu_models.go
dankenigsberg: I don't follow. We can **not** have more than one CpuModelLabelPrefix + aSpecificModelName key.
List would work just as well, but to me "set" says "order is irrelevant, no repetitions", which is the case here. I is not the idiomatic-go way to express that? Am I still poisoned by Python?

nunnatsa: Sorry. Misspelled. Should be "can't", as you said.

I just think using a slice here will drive a bit simpler code, and we have no concern to have duplication. But this is not a critical comment.

pkg/internal/nodeinfo/cpu_models.go
```diff
@@ -0,0 +1,368 @@
...
+	recommendedModels []v1beta1.CpuModelInfo
+}
+
+func GetRecommendedCpuModels() []v1beta1.CpuModelInfo {
+	cpuModelInfo.lock.RLock()
+	defer cpuModelInfo.lock.RUnlock()
+
+	// Return a copy of the pre-computed result
+	result := make([]v1beta1.CpuModelInfo, len(cpuModelInfo.recommendedModels))
+	copy(result, cpuModelInfo.recommendedModels)
```
nunnatsa: not sure about using `copy`, because the `CPU` and the `Memory` fields are pointers. We can try to replace their type to not be a pointer, or use loop of DeepCopy/DeepCopyInto

pkg/internal/nodeinfo/cpu_models.go
```diff
@@ -0,0 +1,368 @@
...
+
+	return cpuModels.UnsortedList()
+}
+
+func getCpuModelPassMarkScore(cpuModel string) int {
+	if score, exists := cpuModelPassMarkScores[cpuModel]; exists {
+		return score
+	}
+	return 0
+}
```
nunnatsa: If a `key` is not exist in a map, or if the map is nil, the value of `m[key]` is the zero value of the value type, meaning `cpuModelPassMarkScores[cpuModel]` will return 0 if the key does not exist, or if the map is not initialized.

In another words, this function can be replaced with just:
```go
cpuModelPassMarkScores[cpuModel]
```

pkg/internal/nodeinfo/cpu_models.go
```diff
@@ -0,0 +1,368 @@
...
+
+	maxPassMark = func() int {
+		m := 0
+		for _, score := range cpuModelPassMarkScores {
+			if score > m {
+				m = score
+			}
+		}
+		return m
+	}()
```
nunnatsa: Can we have this as a regular function, or is there a reason to have it as variable?

We can use the standard library iterator to get the same, but I'm not sure it's any better (it's shouter, though):
```go
return slices.Max(slices.Collect(maps.Values(cpuModelPassMarkScores)))
```

controllers/hyperconverged/hyperconverged_controller.go
```diff
@@ -514,6 +514,15 @@ func updateStatus(req *common.HcoRequest) {
...
 	}
+
+	if cpuModels := nodeinfo.GetRecommendedCpuModels(); !slices.EqualFunc(req.Instance.Status.NodeInfo.RecommendedCpuModels, cpuModels, func(a, b hcov1beta1.CpuModelInfo) bool {
+		return a.Name == b.Name && a.Benchmark == b.Benchmark && a.Nodes == b.Nodes &&
+			((a.CPU == nil && b.CPU == nil) || (a.CPU != nil && b.CPU != nil && a.CPU.Equal(*b.CPU))) &&
+			((a.Memory == nil && b.Memory == nil) || (a.Memory != nil && b.Memory != nil && a.Memory.Equal(*b.Memory)))
+	}) {
+		req.Instance.Status.NodeInfo.RecommendedCpuModels = cpuModels
+		req.StatusDirty = true
+	}
```
nunnatsa, line 521: 1. we already have a comparison function in the nodeinfo package, we can't expose it for reuse, because it's in an internal package, but maybe we can add it as the CpuModel type's method, instead?
2. this code is not tested. Can we have a few unit tests for it? We can use a simple mock, as done for the rest of the functions in the nodeinfo package, to return any desired value.

nunnatsa (commented):
Thanks @dankenigsberg 

Added some more comments.

pkg/internal/nodeinfo/cpu_models.go
dankenigsberg: ok, you're the ruler regarding golang taste.

Barakmor1:
Thank you for this important effort. Have you considered clusters with multiple vendors or architectures? Would this recommendation still be valid in those cases?

dankenigsberg:
> Thank you for this important effort. Have you considered clusters with multiple vendors or architectures? Would this recommendation still be valid in those cases?

Yes: if most of the CPU power of the cluster come from an "exotic" architecture, the code would recommend models from that architecture.

pkg/internal/nodeinfo/cpu_models.go
```diff
@@ -0,0 +1,361 @@
...
+
+	// Weights for CPU model recommendation scoring
+	benchmarkWeight = 0.50
+	cpuWeight       = 0.20
+	memoryWeight    = 0.15
+	nodeWeight      = 0.15
+)
+
+// cpuModelPassMarkScores maps libvirt CPU model names to approximate PassMark scores.
+// Keys must match exactly the cpu-model.node.kubevirt.io/* label values.
```
Barakmor1, line 27: Just a heads-up that the keys in this list tend to change with different libvirt versions, so this would require continued maintenance.

PR #4129, opened 2026-03-26
Fix the optional operator: ensure removal of related objects

nunnatsa:
* In the optional operator logic, if the operand should be removed, we also remove the object from the relatedObjects list in the HyperConverged status. If there was a problem during this process, the related image stays there forever. 
  
  This PR fixes the optional operator logic to make sure the deleted object is removed from the relatedObjects list, even if it was already deleted. It also should fix the E2 flaky sub-resource label test. 

* The conversion test sometimes fails when comparing the results of getting the HyperConverged in v1, to the result when reading in v1beta1 format. 
  
  This is happening for unrelated changes in the status or the metadata fields. The fix is to only compare the spec fields.

**Jira Ticket**:

```jira-ticket
None
```

**Release note**:

```release-note
None
```

### DIFF
--- a/controllers/operands/conditionalHandler.go
+++ b/controllers/operands/conditionalHandler.go
@@ -62,16 +62,24 @@ func (ch *ConditionalHandler) ensureDeleted(req *common.HcoRequest) *EnsureResul
 
 		if deleted {
 			res.SetDeleted()
-			objectRef, err := reference.GetReference(ch.operand.Scheme, cr)
-			if err != nil {
-				return res.Error(err)
-			}
-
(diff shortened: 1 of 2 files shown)


machadovilaca (commented):
can we rename the 1st commit and PR title to something like
"Remove stale RelatedObjects reference for already deleted operands"?

nunnatsa:
> can we rename the 1st commit and PR title to something like "Remove stale RelatedObjects reference for already deleted operands"?

The logic was there. Just move it around so it will happen even if it was failed somehow on the reconcile loop that removed the CR, so it will happen in the next loop.

machadovilaca (approved):

PR #4164, opened 2026-04-12
[Refactor - internal] Simplify handlers

nunnatsa:

The current implementation of initializing the operand handlers, sets up many operand handlers only on the first time it can get the HyperConverged CR. This is originally done for handlers that should created dynamically, like quickstart or imagestream handlers. But over time, it got abused and many operand handlers are now created dynamically. 

The main reason for that is the implementation of the `operands.GetLabels()` function, that requires the HyperConverged CR, only for its name; however, the name is hard coded, and no other name is allowed for this CR, so we don't actually need the CR for its name. In other cases, the CR namespace is also required, but this can be fetched from an environment variable, and reading the HyperConverged CR is not actually required in order to get its namespace.

This PR refactors the `NewOperandHandler()` and the `FirstUseInitiation()` functions, so all the handlers are pre-initialized on boot, and only very few once are created on the first read of the HyperConverged CR, when it's really a must.

**Jira Ticket**:
```jira-ticket
None
```

**Release note**:
```release-note
None
```

### DIFF
--- a/controllers/handlers/kubevirtConsolePlugin.go
+++ b/controllers/handlers/kubevirtConsolePlugin.go
@@ -116,13 +116,13 @@ func NewKvUINginxCMHandler(_ log.Logger, Client client.Client, Scheme *runtime.S
 }
 
 // **** UI user settings config map Handler ****
-func NewKvUIUserSettingsCMHandler(_ log.Logger, Client client.Client, Scheme *runtime.Scheme, hc *hcov1beta1.HyperConverged) (operands.Operand, error) {
-	return operands.NewEditableCmHandler(Client, Scheme, NewKvUIUserSettingsCM(hc)), nil
+func NewKvUIUserSettingsCMHandler(cli client.Client, Scheme *runtime.Scheme) operands.Operand {
+	return operands.NewEditableCmHandler(cli, Scheme, NewKvUIUserSettingsCM())
 }
 
 // **** UI features config map Handler ****
-func NewKvUIFeaturesCMHandler(_ log.Logger, Client client.Client, Scheme *runtime.Scheme, hc *hcov1beta1.HyperConverged) (operands.Operand, error) {
-	return operands.NewEditableCmHandler(Client, Scheme, NewKvUIFeaturesCM(hc), "ipStackType"), nil
+func NewKvUIFeaturesCMHandler(cli client.Client, Scheme *runtime.Scheme) operands.Operand {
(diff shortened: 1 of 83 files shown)


controllers/operandhandler/operandHandler.go
```diff
@@ -59,27 +59,59 @@ func NewOperandHandler(client client.Client, scheme *runtime.Scheme, ci hcoutil.
...
 	}
 
 	if ci.IsOpenshift() {
 		operandList = append(operandList, []operands.Operand{
 			handlers.NewSspHandler(client, scheme),
 			handlers.NewCliDownloadHandler(client, scheme),
 			handlers.NewCliDownloadsRouteHandler(client, scheme),
-			operands.NewServiceHandler(client, scheme, handlers.NewCliDownloadsService),
+			operands.NewServiceHandler(client, scheme, handlers.NewCliDownloadsService()),
 			passt.NewPasstServiceAccountHandler(client, scheme),
```
orenc1, line 77: why on openshift-specific? service account is available in general k8s

nunnatsa: not in this PR, as it's not a change.

controllers/handlers/kubevirtConsolePlugin.go
```diff
@@ -410,7 +409,7 @@ func NewKVUINginxCM(hc *hcov1beta1.HyperConverged) (*corev1.ConfigMap, error) {
 	return &corev1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      nginxConfigMapName,
-			Labels:    operands.GetLabels(hc, hcoutil.AppComponentUIPlugin),
+			Labels:    operands.GetLabels(hcoutil.AppComponentUIPlugin),
 			Namespace: hc.Namespace,
```
orenc1: i think the namespace of the nginx conf CM should be also taken from `hcoutil.GetOperatorNamespaceFromEnv()` instead of `hc.Namespace`

controllers/operandhandler/operandHandler.go
orenc1: and you don't want to fix this on the way?
we already have 2 passt handlers in the general list.
it means that on plain k8s there is no service account for passt? is that even working?

orenc1: can be fixed in a followup PR.
/approve
/lgtm

nunnatsa: Thanks @orenc1 . After looking on it a bit more, this is done on purpose, and the ServiceAccount is not in use in plain K8s.

Not need to fix it.

PR #4165, opened 2026-04-12
Add ipStackType to kubevirt-ui-features ConfigMap

orenc1:
Detect the cluster IP stack type (`IPv4SingleStack`, `IPv6SingleStack`, or `DualStack`) at operator startup by examining the clusterNetwork CIDRs from the `network.config.openshift.io/cluster` status, and include it as the `ipStackType` key in the `kubevirt-ui-features` ConfigMap.

Signed-off-by: Oren Cohen <ocohen@redhat.com>

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://redhat.atlassian.net/browse/CNV-79245
```

**Release note**:

```release-note
Add ipStackType to kubevirt-ui-features ConfigMap
```

### DIFF
--- a/controllers/operands/cmHandler.go
+++ b/controllers/operands/cmHandler.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"maps"
 	"reflect"
+	"slices"
 	"sync"
 
 	corev1 "k8s.io/api/core/v1"
@@ -19,6 +20,10 @@ func NewCmHandler(Client client.Client, Scheme *runtime.Scheme, required *corev1
 	return NewGenericOperand(Client, Scheme, "ConfigMap", &cmHooks{required: required}, false)
 }
 
+func NewEditableCmHandler(Client client.Client, Scheme *runtime.Scheme, required *corev1.ConfigMap, managedKeys ...string) *GenericOperand {
+	return NewGenericOperand(Client, Scheme, "ConfigMap", &cmHooks{required: required, editable: true, managedKeys: managedKeys}, false)
+}
+
 type newDynamicConfigMapFunc func(*hcov1beta1.HyperConverged) (*corev1.ConfigMap, error)
 
 func NewDynamicCmHandler(Client client.Client, Scheme *runtime.Scheme, makeCM newDynamicConfigMapFunc) *GenericOperand {
@@ -26,7 +31,9 @@ func NewDynamicCmHandler(Client client.Client, Scheme *runtime.Scheme, makeCM ne
 }
 
 type cmHooks struct {
-	required *corev1.ConfigMap
+	required    *corev1.ConfigMap
+	editable    bool
+	managedKeys []string
 }
 
 func (h cmHooks) GetFullCr(_ *hcov1beta1.HyperConverged) (client.Object, error) {
@@ -49,16 +56,8 @@ func (h cmHooks) UpdateCR(req *common.HcoRequest, Client client.Client, exists r
 		util.MergeLabels(&h.required.ObjectMeta, &found.ObjectMeta)
 	}
 
-	// Don't reconcile contents of UI settings or AIE webhook config maps
-	if label, exist := found.Labels[util.AppLabelComponent]; exist &&
-		(label == string(util.AppComponentUIConfig) || label == string(util.AppComponentAIEWebhook)) {
-		if labelChanged {
-			err := Client.Update(req.Ctx, found)
-			if err != nil {
-				return false, false, err
-			}
-		}
-		return labelChanged, false, nil
+	if h.editable {
+		return h.reconcileUserEditableCM(req, Client, found, labelChanged)
 	}
 
 	if !reflect.DeepEqual(found.Data, h.required.Data) || labelChanged {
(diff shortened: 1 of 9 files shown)


orenc1:
FYI @adamviktora

nunnatsa (commented):
Please do not extend the `ClusterInfo` interface. We're in a process to split it into a smaller interfaces.

Let's move it to a new package, and then we could easily maintain and mock it. See the nodeinfo package for example.

orenc1:
> Please do not extend the `ClusterInfo` interface. We're in a process to split it into a smaller interfaces.
> 
> Let's move it to a new package, and then we could easily maintain and mock it. See the nodeinfo package for example.

Done.
Moved it to a new package named `ipstacktype`

controllers/operands/cmHandler.go
```diff
@@ -49,16 +49,27 @@ func (h cmHooks) UpdateCR(req *common.HcoRequest, Client client.Client, exists r
...
+		dataChanged := false
+		for k, v := range h.required.Data {
+			if _, exists := found.Data[k]; !exists {
+				if found.Data == nil {
+					found.Data = make(map[string]string)
+				}
+				found.Data[k] = v
+				dataChanged = true
+			}
+		}
```
nunnatsa: let's move the map initialization out of the loop; e.g.

```suggestion
		if len(h.required.Data) > 0 && found.Data == nil {
			found.Data = make(map[string]string)
		}
		
		for k, v := range h.required.Data {
			if _, exists := found.Data[k]; !exists {
				found.Data[k] = v
				dataChanged = true
			}
		}
```

pkg/util/cluster.go
```diff
@@ -141,6 +142,11 @@ func (c *ClusterInfoImp) initOpenshift(ctx context.Context, cl client.Client) er
 		)
 	}
 	c.singlestackipv6 = len(cn) == 1 && net.IsIPv6CIDRString(cn[0].CIDR)
+
+	stackType := ipstacktype.Compute(cn)
+	ipstacktype.Set(stackType)
+	c.logger.Info("Detected IP stack type", "ipStackType", stackType)
+
```
nunnatsa, line 149: Eventually I would move everything related to the `clusterNetwork` to the new package. but that can be done in a followup PR.

controllers/operands/cmHandler.go
```diff
@@ -49,16 +49,27 @@ func (h cmHooks) UpdateCR(req *common.HcoRequest, Client client.Client, exists r
...
 
-	// Don't reconcile contents of UI settings or AIE webhook config maps
+	// Don't reconcile contents of UI settings or AIE webhook config maps,
+	// but seed any new operator-managed keys that are missing.
 	if label, exist := found.Labels[util.AppLabelComponent]; exist &&
 		(label == string(util.AppComponentUIConfig) || label == string(util.AppComponentAIEWebhook)) {
-		if labelChanged {
+		dataChanged := false
+		for k, v := range h.required.Data {
+			if _, exists := found.Data[k]; !exists {
```
nunnatsa: We're adding the new `"ipStackType"` key, that is based on the the actual status of the cluster - what if this status was changed? We'll miss it because we don't override existing keys.

nunnatsa (commented):
Thanks @orenc1 . Looks good.

Added a few comments inline. Also, please add unit test to the changes in controllers/handlers/kubevirtConsolePlugin.go, including the new logic in controllers/operands/cmHandler.go - e.g. check that HCO does not override existing keys in the map.

controllers/operands/cmHandler.go
orenc1: thanks for the comment, the method has been refactored to adhere such a flow

controllers/operands/cmHandler.go
orenc1: done.

controllers/operands/cmHandler.go
```diff
@@ -78,6 +71,40 @@ func (h cmHooks) UpdateCR(req *common.HcoRequest, Client client.Client, exists r
 	return labelChanged, false, nil
 }
 
+func (h cmHooks) isUserEditableCM(found *corev1.ConfigMap) bool {
+	label, exist := found.Labels[util.AppLabelComponent]
+	return exist && (label == string(util.AppComponentUIConfig) || label == string(util.AppComponentAIEWebhook))
+}
```
nunnatsa: Not sure about this. The general case known it specific implementations. This is usually a bad practice.

Let's just add a new `editable` boolean field to the hook, and set it to true on the hook creation, in the specific implementations. then just do
```go
if h.editable {}
```

controllers/operands/cmHandler.go
```diff
@@ -78,6 +71,40 @@ func (h cmHooks) UpdateCR(req *common.HcoRequest, Client client.Client, exists r
...
+			found.Data[k] = v
+			dataChanged = true
+		}
+	}
+	for _, k := range h.managedKeys {
+		if v, exists := h.required.Data[k]; exists && found.Data[k] != v {
+			found.Data[k] = v
+			dataChanged = true
+		}
+	}
```
nunnatsa, line 93: nit: maybe something like this? WDYT?

```suggestion
	for k, v := range h.required.Data {
		if _, exists := found.Data[k]; !exists || slices.Contains(h.managedKeys, k) {
			found.Data[k] = v
			dataChanged = true
		}
	}
```

orenc1: ok, simplified and doing it in a single loop :+1:

controllers/operands/cmHandler.go
orenc1: added a new `editable` field to the cmHook

nunnatsa (approved):

controllers/operands/cmHandler.go
```diff
@@ -78,6 +77,29 @@ func (h cmHooks) UpdateCR(req *common.HcoRequest, Client client.Client, exists r
...
+// reconcileUserEditableCM seeds any missing keys from the required data and
+// always reconciles keys explicitly listed as operator-managed, without
+// overwriting other user-modified values.
+func (h cmHooks) reconcileUserEditableCM(req *common.HcoRequest, Client client.Client, found *corev1.ConfigMap, labelChanged bool) (bool, bool, error) {
+	dataChanged := false
+	if found.Data == nil {
+		found.Data = make(map[string]string)
+	}
+	for k, v := range h.required.Data {
+		if _, exists := found.Data[k]; !exists || (slices.Contains(h.managedKeys, k) && found.Data[k] != v) {
```
nunnatsa: nit:

```suggestion
		if foundVal, exists := found.Data[k]; !exists || (slices.Contains(h.managedKeys, k) && foundVal != v) {
```

orenc1: updated.

PR #4171, opened 2026-04-15
Fix bugs in the perses controller

nunnatsa:
The perses controller is not working properly, and fails to recognise that the perses CRDs are installed.

This PR fixes the issue and adds an e2e test to verify.

Based on #3943 by @sradco 

**Jira Ticket**:

```jira-ticket
None
```

**Release note**:

```release-note
Fix the perses controller
```


controllers/crd/crd_controller.go
```diff
@@ -92,21 +93,36 @@ func (r *ReconcileCRD) operatorRestart() {
...
+	logger = logger.WithName("crd-reconciler").WithValues("Request.Name", req.Name)
+
+	logger.Info("Triggered by a CRD")
+
+	switch req.Name {
+	case hcoutil.DeschedulerCRDName:
+		if !hcoutil.GetClusterInfo().IsDeschedulerAvailable() {
+			if hcoutil.GetClusterInfo().IsDeschedulerCRDDeployed(ctx, r.client) {
+				logger.Info("KubeDescheduler CRD got deployed, restarting the operator to reconfigure the operator for the new kind")
+				r.eventEmitter.EmitEvent(nil, corev1.EventTypeNormal, "KubeDescheduler CRD got deployed, restarting the operator to reconfigure the operator for the new kind", "Restarting the operator to be able to read KubeDescheduler CRs ")
```
sradco: If text for the log and event is the same wouldnt it make sense to create a const? Also for the pereses crd log and event. Perhaps even use the same string and just replace the name of the CRD

nunnatsa: Thanks. Done

PR #4186, opened 2026-04-20
Use v1 status

nunnatsa:

Since we didn't change the status field in API v1, we actually have the exact duplicate of v1beta1 status structure.

To ease the transition to v1, this PR removes the status type from v1beta1, and uses the v1 status type instead.

**Sanity**: the CRD was not change: that proves the theory that v1 status types can be used instead of v1beta1, as they are the same.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
None
```

**Release note**:

```release-note
None
```


docs/api.md
```diff
@@ -66,7 +38,7 @@ HyperConverged is the Schema for the hyperconvergeds API
...
-| status |  | [HyperConvergedStatus](#hyperconvergedstatus) |  | false |
+| status |  | hcov1.HyperConvergedStatus |  | false |
```
avlitman, line 41: nit: now no link

nunnatsa: This is auto generated. There is no link because previously it was an internal link within the same page. The tool does not know how to link to a type that is defined out of the package.

PR #4199, opened 2026-04-27
Deprecate disableMDevConfiguration in favor of mediatedDevicesConfiguration.enabled (v1 + v1beta1)

Ronilerr:

- Deprecates the legacy disableMDevConfiguration surface while keeping it readable for backward compatibility during migration.
- Introduces / uses spec.mediatedDevicesConfiguration.enabled (v1beta1) and spec.virtualization.mediatedDevicesConfiguration.enabled (v1) as the preferred way to disable mediated device handling in KubeVirt.
- Regenerates API boilerplate (deepcopy, openapi, defaults, conversion) and syncs CRDs, deploy manifests, tooling CRD copies, sample CR, feature-gate metadata, and upgrade patch fixtures to match the API changes.
- Updates the KubeVirt handler so an explicit enabled value wins over the deprecated gate; if enabled is unset, the legacy gate can still drive enabled: false for existing CRs.
- Implements mutating webhooks for v1 and v1beta1 to keep the deprecated gate and enabled in sync (including materializing mediatedDevicesConfiguration when missing, and preferring remove JSON patches where appropriate).
- Adjusts func-test default expectations for feature gates after default/metadata changes.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://issues.redhat.com/browse/CNV-79304
```

**Release note**:

```release-note
Deprecate disableMDevConfiguration in favor of mediatedDevicesConfiguration.enabled (v1 + v1beta1)
```


api/v1/featuregates/feature_gates.go
```diff
@@ -119,3 +123,17 @@ func (fgs *HyperConvergedFeatureGates) IsEnabled(name string) bool {
...
+	for _, fg := range fgs {
+		if fg.Name != DeprecatedDisableMDevConfigurationFeatureGateName {
+			continue
+		}
+
+		return ptr.Deref(fg.State, Enabled) == Enabled
+	}
+
+	return false
+}
```
nunnatsa: I don't think this is needed. First, we still use v1beta1 in our code. Also, HyperConvergedFeatureGates already have the IsEnabled method.

api/v1beta1/zz_generated.conversion.go
```diff
@@ -415,6 +415,7 @@ func Convert_v1_HyperConvergedStatus_To_v1beta1_HyperConvergedStatus(in *v1.Hype
 }
 
 func autoConvert_v1beta1_MediatedDevicesConfiguration_To_v1_MediatedDevicesConfiguration(in *MediatedDevicesConfiguration, out *v1.MediatedDevicesConfiguration, s conversion.Scope) error {
+	out.Enabled = (*bool)(unsafe.Pointer(in.Enabled))
```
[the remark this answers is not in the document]
nunnatsa: this is auto generated.

api/v1/hyperconverged_types.go
```diff
@@ -658,7 +660,14 @@ type MediatedHostDevice struct {
 
 // MediatedDevicesConfiguration holds information about MDEV types to be defined, if available
 // +k8s:openapi-gen=true
+// +kubebuilder:validation:XValidation:rule="(has(self.enabled) && self.enabled == false) || (has(self.mediatedDeviceTypes) && size(self.mediatedDeviceTypes)>0)",message="for mediatedDevicesConfiguration a non-empty mediatedDeviceTypes is required, unless enabled is false"
```
orenc1: there might be an issue in which the mutator can produce a CEL-invalid state.
When the create-time mutator sets `mediatedDevicesConfiguration: {enabled: false}` (no mediatedDeviceTypes), and the user later removes the FG, the update mutator strips the enabled field -- leaving `mediatedDevicesConfiguration: {}`, and then this CEL rule will be rejected.

i think that before removing `enabled`, check if `mediatedDeviceTypes` is populated; if not, remove the entire `mediatedDevicesConfiguration` object instead.

config/crd/bases/hco.kubevirt.io_hyperconvergeds.yaml
```diff
@@ -3111,6 +3113,12 @@ spec:
                     description: MediatedDevicesConfiguration holds information about
                       MDEV types to be defined on nodes, if available
                     properties:
+                      enabled:
+                        description: |-
+                          Enable the creation and removal of mediated devices by virt-handler
```
orenc1:
```suggestion
                          Enables the creation and removal of mediated devices by virt-handler
```

controllers/handlers/kubevirt.go
```diff
@@ -617,7 +618,9 @@ func toKvMediatedDevicesConfiguration(hc *hcov1.HyperConverged) *kubevirtcorev1.
 		kvMdev.NodeMediatedDeviceTypes = toKvNodeMediatedDevicesConfiguration(mdevsConfig.NodeMediatedDeviceTypes)
 	}
 
-	if disabled {
+	if mdevsConfig != nil && mdevsConfig.Enabled != nil {
+		kvMdev.Enabled = mdevsConfig.Enabled
+	} else if disabled {
```
orenc1: why do you check `disabled` here if it's based on the deprecated FG?
it's not enough to set `kvMdev.Enabled` to false if `mdevsConfig == nil` or `mdevsConfig.Enabled == nil` (the opposite of the last `if` statement) ?

pkg/webhooks/validator/validator.go
```diff
@@ -408,13 +421,82 @@ func (wh *WebhookHandler) validateDeprecatedFeatureGates(fgMap, oldFgMap map[str
...
+func effectiveV1MDevEnabled(hc *hcov1.HyperConverged) bool {
+	mdc := hc.Spec.Virtualization.MediatedDevicesConfiguration
+	if mdc == nil || mdc.Enabled == nil {
+		return true
+	}
+
+	return *mdc.Enabled
+}
+
+func v1DisableMDevFGValue(fgs hcov1fg.HyperConvergedFeatureGates) (present bool, disabled bool) {
```
orenc1: the second return var is confusing, because to the reader it's unclear at first glance whether the FG is disabled, or we're disabling the mediated devices.
consider changing to `(present bool, fgActive bool)`

pkg/webhooks/mutator/hyperConvergedMutator.go
orenc1, line 1: When a user changes the FG from `state: Enabled` to `state: Disabled` (not removed), the mutator removes enabled but leaves the inert FG entry. This triggers deprecation warnings on every subsequent update. The mutator should also clean up the FG in this case.

api/v1/hyperconverged_types.go
```diff
@@ -658,7 +660,14 @@ type MediatedHostDevice struct {
...
 type MediatedDevicesConfiguration struct {
+	// Enable the creation and removal of mediated devices by virt-handler
+	// Replaces the deprecated DisableMDEVConfiguration feature gate
+	// Defaults to true
```
orenc1: you say "Defaults to true", but there is no annotation that sets it to true on the CRD level (no `+kubebuilder:default=true` annotation).

pkg/webhooks/mutator/hyperConvergedMutator.go
```diff
@@ -167,3 +183,153 @@ func mutateTuningPolicy(hc *hcov1.HyperConverged, patches []jsonpatch.JsonPatchO
...
+func hcV1MDevEnabledValue(hc *hcov1.HyperConverged) bool {
+	mdc := hc.Spec.Virtualization.MediatedDevicesConfiguration
+	if mdc == nil || mdc.Enabled == nil {
+		return true
+	}
+
+	return *mdc.Enabled
+}
+
+func hcPatchV1MDevEnabledFalseWhenUnset(hasConfig bool, patches []jsonpatch.JsonPatchOperation) []jsonpatch.JsonPatchOperation {
```
orenc1, line 197: `hcPatchV1MDevEnabledFalseWhenUnset` function appears to be identical to `hcPatchV1MDevEnabledFalseOnExistingConfig` for all inputs. consider consolidate.

orenc1 (dismissed):
thanks for the PR!
good structure overall, please see my comments.
also there are some code rabbit comments that were not addressed.

controllers/handlers/kubevirt.go
Ronilerr: @orenc1 The else if disabled is only for back-compat when the deprecated FG is on but mediatedDevicesConfiguration.enabled is unset. Setting false whenever enabled is nil would also affect clusters that have device types configured with no enabled and no FG ,  those should leave KV Enabled nil so KubeVirt defaults to true . Once we drop FG support entirely and rely on the mutator always setting enabled, we can remove this.

Ronilerr:
@orenc1 Please review again

Ronilerr:
@orenc1 Please review again

api/v1/hyperconverged_types.go
```diff
@@ -658,7 +660,15 @@ type MediatedHostDevice struct {
 
 // MediatedDevicesConfiguration holds information about MDEV types to be defined, if available
 // +k8s:openapi-gen=true
+// +kubebuilder:validation:XValidation:rule="(has(self.enabled) && self.enabled == false) || (has(self.mediatedDeviceTypes) && size(self.mediatedDeviceTypes)>0)",message="for mediatedDevicesConfiguration a non-empty mediatedDeviceTypes is required, unless enabled is false"
 type MediatedDevicesConfiguration struct {
+	// Enables the creation and removal of mediated devices by virt-handler
+	// Replaces the deprecated DisableMDEVConfiguration feature gate
+	// Defaults to true
+	// +optional
+	// +kubebuilder:default=true
```
orenc1: i think there might be an issue with upgrade when setting a default=true on the CRD level.

The problem: Kubernetes applies CRD defaults on every read from the API server -- including for existing objects in etcd that don't have enabled stored. Consider an existing cluster that has:
```
spec:
  featureGates:
    - name: disableMDevConfiguration   # user wants mdev DISABLED
  virtualization:
    mediatedDevicesConfiguration:
      mediatedDeviceTypes: ["nvidia-222"]
      # no explicit 'enabled' field in etcd
```
After upgrading to this version:
* API server reads the CR from etcd and applies `default: true` to `enabled`
* Controller sees `mdevsConfig.Enabled = ptr.To(true)`:

```
	if mdevsConfig != nil && mdevsConfig.Enabled != nil {
		kvMdev.Enabled = mdevsConfig.Enabled
	} else if disabled {
		kvMdev.Enabled = ptr.To(false)
	}
```
* Controller uses Enabled=true, ignoring the FG entirely
* KubeVirt CR gets Enabled: true -- mdev is silently re-enabled
The FG fallback (`else if disabled`) becomes unreachable whenever `mediatedDevicesConfiguration` exists, because the CRD default always populates Enabled.

The mutator's normalization path also breaks:
```
func hcMutateV1HandleNormalization(hc *hcov1.HyperConverged, patches []jsonpatch.JsonPatchOperation) []jsonpatch.JsonPatchOperation {
	mdc := hc.Spec.Virtualization.MediatedDevicesConfiguration
	if mdc == nil || mdc.Enabled == nil {
		return hcPatchV1MDevEnabledFalseWhenUnset(mdc != nil, patches)
	}
	return patches
}
```
With the default, `mdc.Enabled` is always `ptr.To(true)` (never nil) when `mdc` exists, so normalization is a no-op.

i recommend to remove `+kubebuilder:default=true` from the `Enabled` field in both v1 and v1beta1. Keep the implicit "nil means true" semantics in Go code (which is already correctly implemented in `hcV1MDevEnabledValue` and `effectiveV1MDevEnabled`). The field description already says "Defaults to true" which is sufficient documentation. A CRD-level default on a nested optional field inside an optional struct creates exactly this kind of upgrade hazard.

api/v1/hyperconverged_types.go
orenc1: disregard. `+kubebuilder:default=true` shouldn't be defined, due to the potential issue described in https://github.com/kubevirt/hyperconverged-cluster-operator/pull/4199/changes#r3378982944

api/v1beta1/hyperconverged_types.go
```diff
@@ -519,8 +519,14 @@ type HyperConvergedFeatureGates struct {
...
 // MediatedDevicesConfiguration holds information about MDEV types to be defined, if available
 // +k8s:openapi-gen=true
-// +kubebuilder:validation:XValidation:rule="(has(self.mediatedDeviceTypes) && size(self.mediatedDeviceTypes)>0) || (has(self.mediatedDevicesTypes) && size(self.mediatedDevicesTypes)>0)",message="for mediatedDevicesConfiguration a non-empty mediatedDeviceTypes or mediatedDevicesTypes(deprecated) is required"
+// +kubebuilder:validation:XValidation:rule="(has(self.enabled) && self.enabled == false) || (has(self.mediatedDeviceTypes) && size(self.mediatedDeviceTypes)>0) || (has(self.mediatedDevicesTypes) && size(self.mediatedDevicesTypes)>0)",message="for mediatedDevicesConfiguration a non-empty mediatedDeviceTypes or mediatedDevicesTypes(deprecated) is required, unless enabled is false"
 type MediatedDevicesConfiguration struct {
+	// Enables the creation and removal of mediated devices by virt-handler
+	// Replaces the deprecated DisableMDEVConfiguration feature gate
+	// Defaults to true
+	// +optional
+	Enabled *bool `json:"enabled,omitempty"`
```
nunnatsa: This should only be in v1. We don't change v1beta1 anymore.

api/v1beta1/hyperconverged_types.go
```diff
@@ -414,10 +414,10 @@ type HyperConvergedFeatureGates struct {
 	// +hco:fgphase:discontinued
 	NonRoot *bool `json:"nonRoot,omitempty"`
 
-	// Disable mediated devices handling on KubeVirt
+	// Deprecated: use spec.mediatedDevicesConfiguration.enabled instead. This
+	// field is still read for backward compatibility and will be removed in a future release.
 	// +optional
-	// +kubebuilder:default=false
-	// +default=false
+	// +hco:fgphase:deprecated
```
nunnatsa: This is not longer the source of truth. Please change it in the FG json file (pkg/featuregatedetails/feature-gates.json).

api/v1beta1/hyperconverged_types.go
```diff
@@ -519,8 +519,14 @@ type HyperConvergedFeatureGates struct {
 
 // MediatedDevicesConfiguration holds information about MDEV types to be defined, if available
 // +k8s:openapi-gen=true
-// +kubebuilder:validation:XValidation:rule="(has(self.mediatedDeviceTypes) && size(self.mediatedDeviceTypes)>0) || (has(self.mediatedDevicesTypes) && size(self.mediatedDevicesTypes)>0)",message="for mediatedDevicesConfiguration a non-empty mediatedDeviceTypes or mediatedDevicesTypes(deprecated) is required"
+// +kubebuilder:validation:XValidation:rule="(has(self.enabled) && self.enabled == false) || (has(self.mediatedDeviceTypes) && size(self.mediatedDeviceTypes)>0) || (has(self.mediatedDevicesTypes) && size(self.mediatedDevicesTypes)>0)",message="for mediatedDevicesConfiguration a non-empty mediatedDeviceTypes or mediatedDevicesTypes(deprecated) is required, unless enabled is false"
```
nunnatsa: Please drop these changes, because we're not adding here.

api/v1/hyperconverged_types.go
```diff
@@ -658,7 +660,14 @@ type MediatedHostDevice struct {
 
 // MediatedDevicesConfiguration holds information about MDEV types to be defined, if available
 // +k8s:openapi-gen=true
+// +kubebuilder:validation:XValidation:rule="(has(self.enabled) && self.enabled == false) || (has(self.mediatedDeviceTypes) && size(self.mediatedDeviceTypes)>0)",message="for mediatedDevicesConfiguration a non-empty mediatedDeviceTypes is required, unless enabled is false"
 type MediatedDevicesConfiguration struct {
+	// Enables the creation and removal of mediated devices by virt-handler
+	// Replaces the deprecated DisableMDEVConfiguration feature gate
```
nunnatsa: Not sure we need the API history here. Please add it to the FG description, to point here, instead (in the json file).

api/v1beta1/conversion_test.go
```diff
@@ -585,6 +585,7 @@ var _ = Describe("api/v1beta1", func() {
 			It("should convert MediatedDevicesConfiguration", func() {
 				v1VirtConfig := hcov1.VirtualizationConfig{
 					MediatedDevicesConfiguration: &hcov1.MediatedDevicesConfiguration{
+						Enabled:             ptr.To(false),
```
nunnatsa: after moving to go v1.26, please use the `new()` builtin function; e.g.

```suggestion
						Enabled:             new(false),
```

controllers/handlers/kubevirt.go
```diff
@@ -599,6 +599,7 @@ func getObsoleteCPUConfig(hcObsoleteCPUModels []string) map[string]bool {
 
 func toKvMediatedDevicesConfiguration(hc *hcov1.HyperConverged) *kubevirtcorev1.MediatedDevicesConfiguration {
 	mdevsConfig := hc.Spec.Virtualization.MediatedDevicesConfiguration
+	//nolint:staticcheck // Read deprecated feature gate for backwards compatibility.
```
nunnatsa: not needed. The deprecation is not in a golang code.

controllers/handlers/kubevirt.go
```diff
@@ -599,6 +599,7 @@ func getObsoleteCPUConfig(hcObsoleteCPUModels []string) map[string]bool {
...
 	disabled := hc.Spec.FeatureGates.IsEnabled("disableMDevConfiguration")
```
nunnatsa: let's rename to something like `disabledByFG`, because now we have two ways to enable.

pkg/webhooks/mutator/hyperConvergedMutator.go
```diff
@@ -68,12 +74,24 @@ func (hcm *HyperConvergedMutator) mutateHyperConverged(req admission.Request, lo
...
 	}
 
+	var oldHC *hcov1.HyperConverged
+	if req.Operation == admissionv1.Update && req.OldObject.Raw != nil {
+		oldHC = &hcov1.HyperConverged{}
+		if err := hcm.decoder.DecodeRaw(req.OldObject, oldHC); err != nil {
+			logger.Error(err, "failed to read the old HyperConverged custom resource")
+			return admission.Errored(http.StatusBadRequest, fmt.Errorf("failed to parse the old HyperConverged"))
+		}
+	}
```
nunnatsa: Let's move this logic under the `else` in line 93

pkg/webhooks/mutator/hyperConvergedMutator.go
```diff
@@ -167,3 +185,186 @@ func mutateTuningPolicy(hc *hcov1.HyperConverged, patches []jsonpatch.JsonPatchO
...
+			Operation: "remove",
+			Path:      fmt.Sprintf("/spec/featureGates/%d", idx),
+		})
+	}
+
+	return patches
+}
+
+func hcMutateV1MDevFGAndEnabledOnCreate(hc *hcov1.HyperConverged, patches []jsonpatch.JsonPatchOperation) []jsonpatch.JsonPatchOperation {
+	//nolint:staticcheck // Read deprecated feature gate for backwards compatibility.
```
nunnatsa: not needed, as above.

nunnatsa (commented):
Thanks @Ronilerr. 

Added several comments inline.

api/v1beta1/conversion_test.go
```diff
@@ -995,6 +997,7 @@ var _ = Describe("api/v1beta1", func() {
 			It("should convert MediatedDevicesConfiguration", func() {
 				v1beta1Spec := HyperConvergedSpec{
 					MediatedDevicesConfiguration: &MediatedDevicesConfiguration{
+						Enabled:             ptr.To(false),
```
nunnatsa: `ptr.To` ==> `new`

api/v1beta1/hyperconverged_types.go
```diff
@@ -415,9 +415,11 @@ type HyperConvergedFeatureGates struct {
 	NonRoot *bool `json:"nonRoot,omitempty"`
 
 	// Disable mediated devices handling on KubeVirt
+	// Deprecated: use v1.MediatedDevicesConfiguration.Enabled instead. This feature gate will be removed in a future release.
```
nunnatsa: WDYT about something like this?

> Deprecated: use v1's spec.virtualization.mediatedDevicesConfiguration.Enabled instead. This feature gate will be removed in a future release.

api/v1beta1/hyperconverged_types.go
```diff
@@ -415,9 +415,11 @@ type HyperConvergedFeatureGates struct {
...
 	// +optional
 	// +kubebuilder:default=false
 	// +default=false
+	// +hco:fgphase:discontinued
```
nunnatsa: I think we can drop this

api/v1/hyperconverged_types.go
```diff
@@ -135,6 +131,11 @@ type HyperConvergedSpec struct {
 	//   gate causes the redeployment of the virt-handler pod.
 	//   Phase: alpha
 	//
+	// * disableMDevConfiguration:
+	//   Deprecated: use mediatedDevicesConfiguration.enabled instead. This feature
```
nunnatsa: please use the full path: `spec.virtualization.mediatedDevicesConfiguration.enabled`

api/v1beta1/conversion.go
```diff
@@ -531,3 +535,67 @@ func setPtr[T comparable](orig *T) *T {
 	*dst = *orig
 	return dst
 }
+
+// Convert_v1_MediatedDevicesConfiguration_To_v1beta1_MediatedDevicesConfiguration converts
+// v1 MediatedDevicesConfiguration to v1beta1. Enabled is v1-only and is mapped at the
+// HyperConverged spec level via convertMDevEnabledV1ToV1beta1.
+func Convert_v1_MediatedDevicesConfiguration_To_v1beta1_MediatedDevicesConfiguration(in *hcov1.MediatedDevicesConfiguration, out *MediatedDevicesConfiguration, s conversion2.Scope) error {
```
nunnatsa: For consistency and readability, please use `src` and `dst` instead of `in` and `out`.

nunnatsa: Is this function in use? if not, please drop it.

api/v1beta1/conversion.go
```diff
@@ -531,3 +535,67 @@ func setPtr[T comparable](orig *T) *T {
...
+	} else {
+		out.NodeMediatedDeviceTypes = nil
+	}
+	return nil
+}
+
+// Convert_v1beta1_MediatedDevicesConfiguration_To_v1_MediatedDevicesConfiguration converts
+// v1beta1 MediatedDevicesConfiguration to v1. Enabled is v1-only and is mapped at the
+// HyperConverged spec level via convertMDevEnabledV1beta1ToV1.
+func Convert_v1beta1_MediatedDevicesConfiguration_To_v1_MediatedDevicesConfiguration(in *MediatedDevicesConfiguration, out *hcov1.MediatedDevicesConfiguration, s conversion2.Scope) error {
```
nunnatsa: Is this function in use? if not, please drop it. if yes, please change `in` and `out` to `src` and `dst`.

api/v1beta1/conversion_test.go
nunnatsa, line 1: Please also add some code in the `conversion_fuzz_test.go` file as well, for random testing (both for v1 => v1beta1 and for v1beta1 => v1)

pkg/webhooks/mutator/hyperConvergedMutator.go
```diff
@@ -167,3 +184,184 @@ func mutateTuningPolicy(hc *hcov1.HyperConverged, patches []jsonpatch.JsonPatchO
...
+	hc *hcov1.HyperConverged,
+	newFGEnabled bool,
+	hasConfig bool,
+	patches []jsonpatch.JsonPatchOperation,
+) []jsonpatch.JsonPatchOperation {
+	if newFGEnabled {
+		return hcPatchV1MDevEnabledFalseWhenUnset(hasConfig, patches)
+	}
+
+	patches = hcPatchV1RemoveMDevEnabled(hc, patches)
```
nunnatsa, line 335: Not sure about it. I this we should be explicit.

nunnatsa (commented):
Thanks @Ronilerr. we're almost there. Added a few comments inline.

Some general comments:
1. I can't see any validation logic in this PR (in the webhook). Did I missed it?
2. Please update cluster-configuration.md with the changes
3. Please update cluster-configuration-v1beta1-api.md with the v1beta1 deprecation (refer to the v1 doc)

api/v1beta1/conversion_fuzz_test.go
```diff
@@ -499,6 +537,47 @@ func randomV1HC(r *rand.Rand) *hcov1.HyperConverged {
...
+	g.Expect(fg).ToNot(BeNil())
+	g.Expect(v1Virt.MediatedDevicesConfiguration).ToNot(BeNil())
+	g.Expect(*v1Virt.MediatedDevicesConfiguration.Enabled).To(Equal(!*fg))
+}
+
+func assertMDevV1ToV1beta1Mapping(g Gomega, v1Virt hcov1.VirtualizationConfig, v1beta1Spec HyperConvergedSpec) {
+	mdc := v1Virt.MediatedDevicesConfiguration
+	g.Expect(mdc).ToNot(BeNil())
+	g.Expect(mdc.Enabled).ToNot(BeNil())
+	g.Expect(*v1beta1Spec.FeatureGates.DisableMDevConfiguration).To(Equal(!*mdc.Enabled))
```
nunnatsa: Please use the `HaveValue` matcher to compare pointer values. It also checks for not nil.

e.g.

```suggestion
	g.Expect(v1beta1Spec.FeatureGates.DisableMDevConfiguration).To(HaveValue(Equal(!*mdc.Enabled)))
```

api/v1beta1/hyperconverged_types.go
nunnatsa:
```suggestion
	// +kubebuilder:default={"downwardMetrics": false, "deployKubeSecondaryDNS": false,  "persistentReservation": false, "enableMultiArchBootImageImport": false, "decentralizedLiveMigration": true, "declarativeHotplugVolumes": false, "videoConfig": true, "objectGraph": false, "incrementalBackup": false, "containerPathVolumes": false}
```

api/v1beta1/hyperconverged_types.go
```diff
@@ -415,6 +415,8 @@ type HyperConvergedFeatureGates struct {
 	NonRoot *bool `json:"nonRoot,omitempty"`
 
 	// Disable mediated devices handling on KubeVirt
+	// Deprecated: use v1's spec.virtualization.mediatedDevicesConfiguration.enabled instead. This feature gate will be removed in a future release.
+	// +hco:fgphase:deprecated
 	// +optional
 	// +kubebuilder:default=false
 	// +default=false
```
nunnatsa: Please drop these two lines

api/v1beta1/hyperconverged_types.go
nunnatsa:
```suggestion
	// +kubebuilder:default={"certConfig": {"ca": {"duration": "48h0m0s", "renewBefore": "24h0m0s"}, "server": {"duration": "24h0m0s", "renewBefore": "12h0m0s"}},"featureGates": {"downwardMetrics": false, "deployKubeSecondaryDNS": false, "persistentReservation": false, "enableMultiArchBootImageImport": false, "decentralizedLiveMigration": true, "declarativeHotplugVolumes": false, "videoConfig": true, "objectGraph": false, "incrementalBackup": false, "containerPathVolumes": false}, "liveMigrationConfig": {"completionTimeoutPerGiB": 150, "parallelMigrationsPerCluster": 5, "parallelOutboundMigrationsPerNode": 2, "progressTimeout": 150, "allowAutoConverge": false, "allowPostCopy": false}, "resourceRequirements": {"vmiCPUAllocationRatio": 10}, "uninstallStrategy": "BlockUninstallIfWorkloadsExist", "virtualMachineOptions": {"disableFreePageReporting": false, "disableSerialConsoleLog": false}, "enableApplicationAwareQuota": false, "enableCommonBootImageImport": true, "deployVmConsoleProxy": false}
```

nunnatsa:
/lgtm
/approve

Thanks @Ronilerr !!

In a followup PR we'll need to:
1. Fix the conversion to restore the new `enable` field as done in #4346
2. Fix the fg-json-generator tool to properly generate a conversion logic for deprecated feature gates (and to use the json file as source of truth instead of the v1beta1 file).

PR #4211, opened 2026-04-29
tests: fix observability controller test causing cascading e2e failures

nunnatsa:

The observability test was causing subsequent tests (e.g. labels) to fail because it deleted HCO operator pods without waiting for them to be ready again. Additionally, there was a race condition where the test could find the original silence before Alertmanager finished expiring it, making the test pass without actually validating the controller's reconciliation.

Fix by:
- Verifying the silence is actually removed before proceeding
- Waiting for the HCO operator rollout after pod deletion
- Adding DeferCleanup to ensure operator readiness even on test failure
- Using g Gomega in Eventually blocks to allow retries on transient errors

**Jira Ticket**:

```jira-ticket
None
```

**Release note**:

```release-note
None
```

### DIFF
--- a/tests/func-tests/observability_controller_test.go
+++ b/tests/func-tests/observability_controller_test.go
@@ -5,6 +5,7 @@ import (
 	"crypto/tls"
 	"fmt"
 	"net/http"
+	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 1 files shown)


nunnatsa:
> * You now call `tests.WaitForHCOOperatorRollout(ctx)` both immediately and via `DeferCleanup`, which may add unnecessary delay on success; consider keeping only the DeferCleanup for failure cleanup and relying on the existing Eventually block for readiness, or vice versa, to avoid redundant waiting.

The two calls serves different purposes. The one in the DeferCleanup is for case of a failure, to make sure the HCO pod is running again, before continuing to the next test. If the test passes, the waiting time is minimal, because the other call already validated the pod is running, so this does not worry.

The second call makes sure the pod is running again, before validating that the silence exists.

machadovilaca (approved):

PR #4218, opened 2026-04-30
Add discontinued feature gate phase

Ronilerr:
Adds discontinued as a feature-gate phase on v1beta1. Discontinued gates are excluded from feature-gates.json at generation time.
v1 API comments and related doc generation stay consistent with that.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://redhat.atlassian.net/browse/CNV-85871
```

**Release note**:

```release-note
Add discontinued feature gate phase
```


api/v1/hyperconverged_types.go
```diff
@@ -57,8 +57,8 @@ type HyperConvergedSpec struct {
 	//   can be disabled.
 	// * GA: the feature is graduated and is always enabled. There is no way to
 	//   disable it.
-	// * deprecated: the feature is no longer supported. There is no way to enable
-	//   it
+	// * deprecated: the feature is deprecated. Unless noted otherwise, it is
+	//   disabled by default, but can still be enabled until it is removed.
```
nunnatsa: WDYT about something like this?


```suggestion
	// * deprecated: the feature is deprecated. The feature gate will be removed in a future release.
```

tools/fg-json-generator/main.go
```diff
@@ -195,26 +201,55 @@ func fieldJSONName(field *ast.Field) string {
...
-					return featuregates.PhaseBeta
-				}
-				return featuregates.PhaseAlpha
+	for _, comment := range field.Doc.List {
+		text := strings.TrimPrefix(comment.Text, "//")
+		text = strings.TrimSpace(text)
+		if strings.HasPrefix(strings.ToLower(text), "deprecated") {
+			return featuregates.PhaseDeprecated
+		}
+	}
```
nunnatsa: I think we can drop this code - we are handling it in `phaseFromHcoAnnotation`. But if you think it's safer, then keep it.

tools/fg-json-generator/main.go
```diff
@@ -195,26 +201,55 @@ func fieldJSONName(field *ast.Field) string {
...
-			if strings.HasPrefix(comment.Text, kubebuilderDefaultPrefix) {
-				val := comment.Text[len(kubebuilderDefaultPrefix):]
-				if val == "true" {
-					return featuregates.PhaseBeta
-				}
-				return featuregates.PhaseAlpha
+	for _, comment := range field.Doc.List {
+		text := strings.TrimPrefix(comment.Text, "//")
+		text = strings.TrimSpace(text)
+		if strings.HasPrefix(strings.ToLower(text), "deprecated") {
```
nunnatsa: it may be too general. please consider to do

```suggestion
		if strings.HasPrefix("Deprecated:") {
```

tools/fg-v1-comments/featuregates_list.gotmplt
```diff
@@ -10,8 +10,8 @@
 //   can be disabled.
 // * GA: the feature is graduated and is always enabled. There is no way to
 //   disable it.
-// * deprecated: the feature is no longer supported. There is no way to enable
-//   it
+// * deprecated: the feature is deprecated. Unless noted otherwise, it is
+//   disabled by default, but can still be enabled until it is removed.
```
nunnatsa: Same as the above comment (adding again because here is the source):

```suggestion
// * deprecated: the feature is deprecated. The feature gate will be removed in a future release.
```

nunnatsa (commented):
Thanks @Ronilerr .

Added some more comments

PR #4220, opened 2026-05-03
Improvements in wasp-agent testing

enp0s3:

There were few missing parts in the wasp-agent e2e testing. The suite should run only on
supported clusters.
Since there was a bug with OCI files cleanup, the wasp-agent version was manually bumped.

**Jira Ticket**:

```jira-ticket
NONE
```

**Release note**:

```release-note
NONE
```

## Summary by CodeRabbit

* **Chores**
  * Updated WASP Agent component to v1.2.2
  * Updated WASP Agent container image digest

* **Bug Fixes**
  * Added validation to prevent runtime errors when configuration is missing

* **Tests**
  * Refactored test utilities for improved code reuse
  * Enhanced WASP Agent tests to validate high-availability cluster requirements

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubevirt/hyperconverged-cluster-operator/pull/4220?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


controllers/handlers/wasp-agent/daemonset.go
```diff
@@ -79,6 +79,17 @@ func newWaspAgentDaemonSet(hc *hcov1beta1.HyperConverged) *appsv1.DaemonSet {
 		SecurityContext: &corev1.SecurityContext{
 			Privileged: ptr.To(true),
 		},
+		Lifecycle: &corev1.Lifecycle{
+			PreStop: &corev1.LifecycleHandler{
+				Exec: &corev1.ExecAction{
+					Command: []string{
```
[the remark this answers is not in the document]
enp0s3: Good catch! Will update the PR

tests/func-tests/wasp_agent_test.go
```diff
@@ -32,6 +34,10 @@ const (
 	setMemoryOvercommitTemplate = `[{"op": "replace", "path": "/spec/higherWorkloadDensity/memoryOvercommitPercentage", "value": %d}]`
 	overcommitPercent           = 150
 
+	ociHookScriptPath = "/host/opt/oci-hook-swap.sh"
+	ociHookConfigPath = "/host/run/containers/oci/hooks.d/swap-for-burstable.json"
```
avlitman: duplicate constants, why not use wasp_agent.OciHookSwapScriptPath and wasp_agent.OciHookSwapConfigPath?

controllers/handlers/wasp-agent/daemonset_test.go
```diff
@@ -205,6 +205,34 @@ var _ = Describe("Wasp Agent DaemonSet", func() {
 		})
 
 	})
+	Context("Wasp agent DaemonSet spec definition", func() {
```
avlitman: maybe add also a test for nil: It("should not create DaemonSet when HigherWorkloadDensity is nil", func() {

controllers/handlers/wasp-agent/daemonset.go
```diff
@@ -79,6 +82,17 @@ func newWaspAgentDaemonSet(hc *hcov1beta1.HyperConverged) *appsv1.DaemonSet {
 		SecurityContext: &corev1.SecurityContext{
 			Privileged: ptr.To(true),
 		},
+		Lifecycle: &corev1.Lifecycle{
+			PreStop: &corev1.LifecycleHandler{
```
nunnatsa: My only concern is what will happen in upgrade to a newer version; two use-case:
1. upgrade from previous version to this one: what if the old pod is created but then, because we changed the condition to create the daemonset, the current one won't be created? then we still have the files in place on the node.
2. upgrade from this version to a new one: will we have a race condition of deleting/creating the files at the same time?

tests/func-tests/wasp_agent_test.go
```diff
@@ -196,6 +202,35 @@ var _ = Describe("Test wasp-agent", Label(tests.OpenshiftLabel, "wasp-agent"), S
...
+
+			Eventually(func(g Gomega, ctx context.Context) bool {
+				ds, err := getWaspDS(ctx, cli)
+				g.Expect(err).ToNot(HaveOccurred())
+				g.Expect(ds.Status.DesiredNumberScheduled).ToNot(BeZero())
+				return ds.Status.DesiredNumberScheduled == ds.Status.CurrentNumberScheduled
+			}).WithTimeout(5 * time.Minute).WithPolling(time.Second).WithContext(ctx).Should(BeTrue())
+
+			By("verify OCI hook files are present on all worker nodes after deployment")
+			expectOCIHookFilesPresentOnAllNodes(ctx, cli)
```
nunnatsa: @enp0s3 : e2e test is failing here

controllers/handlers/wasp-agent/daemonset.go
enp0s3: Hi @nunnatsa
> upgrade from previous version to this one: what if the old pod is created but then, because we changed the condition to create the daemonset, the current one won't be created? then we still have the files in place on the node.

You are correct, we don't handle cleanup upon upgrade, the cleanup scope is only for the current release. However we can backport the bugfix.

> upgrade from this version to a new one: will we have a race condition of deleting/creating the files at the same time?

I don't think its possible, such race condition assumes that two pods of the same DS will run on the same node, but we deploy wasp-agent with `maxSurge: 0` value, which mean zero additional wasp-agent pods allowed to be run on the same worker node during rollout. Therefore the preStop will be synced with creation of the new pod.

enp0s3: @nunnatsa Fwiw if the nodes are being re-provisioned or rebooted during the upgrade then the OCI files will be removed.

orenc1: hi @enp0s3 ,
you are making changes to the node's filesystem without syncing this with MachineConfig.
how do you know the machine-config-controller won't become degraded due to this with `unexpected on-disk state` ?

enp0s3: @orenc1 Thanks for looking at this! These files aren't part of the MCO ignition spec, wasp-agent application is rendering these files in runtime based on CRIO configuration and then copy them.

tests/func-tests/wasp_agent_test.go
enp0s3: @avlitman ~~Sorry but I can't determine the similarity you are talking about. The variables have different names and values.~~
I see what you mean. I want to achieve maximum test isolation from the business logic. Therefore I would like to fail if these consts changed in the product code.

controllers/handlers/wasp-agent/daemonset_test.go
enp0s3: @avlitman I am testing the `newWaspAgentDaemonSet` unit which creates the DS struct unconditionally. The `HigherWorkloadDensity` API field controls the memory overcommit percentage and also serves as a trigger for the wasp-agent operand deployment.

enp0s3:
Converting to draft in order to debug the e2e failures. Going to add timeout and programmatic focus in order to manually debug the cluster.

enp0s3:
@nunnatsa I think I found the root cause:
```
The issue is that oc debug node/<name> creates a debug pod in the current namespace context. Inside the functest pod, oc is resolving the current namespace from the kubeconfig or environment, which in CI points to the temporary Prow job namespace ci-op-3y97c20m. That namespace either doesn't exist anymore or isn't accessible from within the functest pod's context.

The fix is to explicitly tell oc debug which namespace to create the debug pod in. We should use the install namespace since the functest service account has cluster-admin there.
```

controllers/handlers/wasp-agent/daemonset.go
nunnatsa: Do you also want to check for high availability here??

You can check the HCO's `status.infrastructureHighlyAvailable` filed, or use the functions in the `nodeinfo` package.

tests/func-tests/wasp_agent_test.go
```diff
@@ -370,6 +409,96 @@ func unpauseWorkerMachineConfigPool(ctx context.Context, cli client.Client) {
...
+
+		for i := range nodes {
+			nodeName := nodes[i].Name
+			g.Expect(checkFileExistsOnNode(nodeName, ociHookScriptPath)).To(Succeed())
+			g.Expect(checkFileExistsOnNode(nodeName, ociHookConfigPath)).To(Succeed())
+		}
+	}).WithTimeout(1 * time.Minute).
+		WithPolling(10 * time.Second).
+		WithContext(ctx).
+		Should(Succeed())
```
nunnatsa: WDYT about embeding the eventually inside the loop? something like


```suggestion
	nodes, err := listWorkerNodes(ctx, cli)
	Expect(err).ToNot(HaveOccurred())
	Expect(nodes).ToNot(BeEmpty(), "no worker nodes found")
	
	for i := range nodes {
		Eventually(func(g Gomega, ctx context.Context) {
			nodeName := nodes[i].Name
			g.Expect(checkFileExistsOnNode(nodeName, ociHookScriptPath)).To(Succeed())
			g.Expect(checkFileExistsOnNode(nodeName, ociHookConfigPath)).To(Succeed())
		}
	}).WithTimeout(1 * time.Minute).
		WithPolling(10 * time.Second).
		WithContext(ctx).
		Should(Succeed())
	}
```

enp0s3: @nunnatsa Sounds great! updated

nunnatsa (approved):

PR #4223, opened 2026-05-04
Disable KubeVirt beta feature gates

nunnatsa:
Once kubevirt/kubevirt#17405 is merged, KubeVirt enables all it's beta FGs by default. This behavior is not desired by HCO. 

HCO uses a json file produced by KubeVirt to get the list of the beta feature gates, and disable them, unless they are enabled by HCO logic.

The json file will be auto generated in a future PR.

**Jira Ticket**:

```jira-ticket
https://redhat.atlassian.net/browse/CNV-86398
```

**Release note**:

```release-note
Disable KubeVirt beta feature gates
```


pkg/internal/kvfeaturegates/kv-beta-feature-gates.json
```diff
@@ -0,0 +1,12 @@
...
+  "ImageVolume",
+  "KubevirtSeccompProfile",
+  "MigrationPriorityQueue",
+  "NodeRestriction",
+  "PasstBinding",
+  "PodSecondaryInterfaceNamingUpgrade",
+  "SecureExecution",
+  "Snapshot",
+  "VideoConfig"
+]
```
iholder101, line 182: [feature-gate-report.txt](https://github.com/user-attachments/files/27368501/feature-gate-report.txt)

This is temporary I suppose?
Why not use the real output? It is attached

controllers/handlers/kubevirt.go
```diff
@@ -1105,9 +1109,24 @@ func getKvFeatureGateList(spec hcov1beta1.HyperConvergedSpec, annotations map[st
...
+
+	disabled := make([]string, 0, len(betaFGs))
+	for _, fg := range betaFGs {
+		if !slices.Contains(enabledFGs, fg) {
+			disabled = append(disabled, fg)
+		}
+	}
+
+	return disabled
+}
```
iholder101, line 1095: Let's say we decide to not disable a beta FG moving forward. Would this where the code should reside?

iholder101 (commented):
Thank you @nunnatsa!

pkg/internal/kvfeaturegates/kv-beta-feature-gates.json
nunnatsa: yap. will fix.

controllers/handlers/kubevirt.go
nunnatsa: It will be in the `enabledFGs` slice, which is the result of `getKvFeatureGateList()`. All the FG logic is in this function. If we choose to enable the FG, we can add the relevant logic (e.g. HCO new FG, logic that base on the status of the cluster and so on) , to add this FG to the enabled list.

We can maybe remove this beta FG from the allow list as well, in order to use its default. Not sure its critical.

controllers/handlers/kubevirt.go
```diff
@@ -815,6 +798,9 @@ func getKVDevConfig(hc *hcov1.HyperConverged) *kubevirtcorev1.DeveloperConfigura
 	if len(fgs) > 0 {
 		devConf.FeatureGates = fgs
 	}
+	if disabledFGs := getKvDisabledFeatureGateList(fgs); len(disabledFGs) > 0 {
+		devConf.DisabledFeatureGates = disabledFGs
+	}
```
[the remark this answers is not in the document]
nunnatsa: Good point. But this is an edge case that is the user responsibility, as the json patch annotation is not formally supported. Also, the same way the use can modify the feature gate list, they can modify the disabled list, and this is an endless loop. And finally, KubeVirt behavior when the same FG is in both lists, is to enable the feature gate, so this is covered.

controllers/handlers/kubevirt_test.go
```diff
@@ -2379,6 +2388,85 @@ Version: 1.2.3`)
...
+
+				It("should not have any FG in both enabled and disabled lists", func() {
+					mandatoryKvFeatureGates = getMandatoryKvFeatureGates(false)
+					fgs := getKvFeatureGateList(hco)
+					disabled := getKvDisabledFeatureGateList(fgs)
+					for _, fg := range disabled {
+						Expect(fgs).NotTo(ContainElement(fg), "feature gate %q must not appear in both enabled and disabled lists", fg)
+					}
+				})
+			})
```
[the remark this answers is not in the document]
nunnatsa: @coderabbitai - Fixed

controllers/handlers/kubevirt_test.go
```diff
@@ -2626,6 +2647,85 @@ Version: 1.2.3`)
...
+					disabled := getKvDisabledFeatureGateList(kvfeaturegates.GetBetaFeatureGates())
+					Expect(disabled).To(BeEmpty())
+				})
+
+				It("should not include Snapshot or KubevirtSeccompProfile when they are in mandatory FGs", func() {
+					mandatoryKvFeatureGates = getMandatoryKvFeatureGates(false)
+					fgs := getKvFeatureGateList(hco.Spec, nil)
+					disabled := getKvDisabledFeatureGateList(fgs)
+					Expect(disabled).NotTo(ContainElement(kvSnapshotGate))
+					Expect(disabled).NotTo(ContainElement(kvKubevirtSeccompProfile))
```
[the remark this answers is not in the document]
nunnatsa: fixed

nunnatsa:
/unhold

We actually don't need to wait for the KV PR.

controllers/handlers/kubevirt.go
```diff
@@ -1077,7 +1063,7 @@ func translateKubeVirtConds(orig []kubevirtcorev1.KubeVirtCondition) []metav1.Co
 func getMandatoryKvFeatureGates(isKVMEmulation bool) []string {
 	mandatoryFeatureGates := hardCodeKvFgs
 	if !isKVMEmulation {
-		mandatoryFeatureGates = append(mandatoryFeatureGates, sspConditionKvFgs...)
+		mandatoryFeatureGates = append(mandatoryFeatureGates, kvHypervStrictCheck)
 	}
 
 	return mandatoryFeatureGates
```
orenc1, line 1069: when `isKVMEmulation` is true, we return the very same slice of `hardCodeKvFgs`, while when it's false, we return a copy, because the slice length exceeds the capacity allocated (due to the addition of `kvHypervStrictCheck`).
if in the future something will mutate the slice in place by index, it would create inconsistency.
consider something like:
```
func getMandatoryKvFeatureGates(isKVMEmulation bool) []string {
	mandatoryFeatureGates := slices.Clone(hardCodeKvFgs)
	if !isKVMEmulation {
		mandatoryFeatureGates = append(mandatoryFeatureGates, kvHypervStrictCheck)
	}
	return mandatoryFeatureGates
}
```

pkg/internal/kvfeaturegates/kv_feature_gates_test.go
```diff
@@ -0,0 +1,21 @@
...
+func TestKvFeatureGates(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "KV Feature Gates Suite")
+}
+
+var _ = Describe("KV Beta Feature Gates", func() {
+	It("just make sure init is working, and does not panic", func() {
+		_ = kvfeaturegates.GetBetaFeatureGates()
+	})
+})
```
orenc1, line 21: this will pass even if the embedded json will contain 0 beta feature gates, or if the beta feature gates have empty names.
the test can be stronger by adding:

- `Expect(gates).ToNot(BeEmpty()`
- `Expect(slices.IsSorted(gates)).To(BeTrue())`
// Verify at least one known stable gate
- `Expect(gates).To(ContainElement("Snapshot"))`

orenc1 (commented):
looks good in general, just a couple of nits for your consideration

controllers/handlers/kubevirt.go
nunnatsa: ok. This is a good protection for the `hardCodeKvFgs` slice.

pkg/internal/kvfeaturegates/kv_feature_gates_test.go
nunnatsa: I think no beta FGs is a legit use-case. This test is only to make sure that we don't panic with the auto generated file (it will eventually be auto generated)

orenc1 (approved):
/approve

PR #4229, opened 2026-05-06
Remove Passt operand now that Passt is core Network plugin in KubeVirt

oshoval:

Since KubeVirt 1.8 / HCO 1.18, Passt is provided as a core KubeVirt network plugin.
HCO no longer needs to reconcile passt resources directly.

This PR removes passt-specific handler/test code, 
drops the related dependency from `go.mod`/`go.sum`/`vendor`, 
and regenerates manifests/docs/upgrade patches to match.

Annotation is kept because it still enables the Passt Binding FG, which is now core network plugin.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://redhat.atlassian.net/browse/CNV-81997
```

**Release note**:

```release-note
HCO stops managing passt resources because passt is core in KubeVirt 1.8 / HCO 1.18.
The passt CNI binary can be removed manually from nodes.
```

## Summary by CodeRabbit

* **Features Removed**
  * Passt network binding support removed (Passt images, flags, handlers, tests, and vendored CNI types dropped).

* **Chores**
  * Upgrade cleanup extended to remove Passt-related objects for affected versions.
  * Build/packaging updated to stop including Passt images and flags; image lists adjusted.

* **Behavioral Change**
  * NetworkAttachmentDefinition changes no longer trigger operator reconciliations.

* **Documentation**
  * Annotation docs updated to describe enabling the PasstBinding feature gate.


controllers/handlers/kubevirt.go
```diff
@@ -166,7 +165,6 @@ const (
 	kvObjectGraph                = "ObjectGraph"
 	kvUtilityVolumes             = "UtilityVolumes"
 	kvIncrementalBackup          = "IncrementalBackup"
-	kvPasstBinding               = "PasstBinding"
```
oshoval, line 169: Hi @nirdothan @nunnatsa
on 1.9 since we are removing the annotation, this FG will be need to be set manually right ?
since it now affects the passt core network binding as far as i understand

pkg/components/components.go
```diff
@@ -590,11 +580,6 @@ func GetClusterPermissions() []rbacv1.PolicyRule {
 			Resources: stringListToSlice("serviceaccounts"),
 			Verbs:     stringListToSlice("get", "list", "watch", "create", "update", "delete"),
 		},
-		{
-			APIGroups: stringListToSlice("k8s.cni.cncf.io"),
-			Resources: stringListToSlice("network-attachment-definitions"),
-			Verbs:     stringListToSlice("get", "list", "watch", "create", "update", "delete"),
-		},
```
oshoval, line 597: Hi @nunnatsa 
we are removing NAD on upgrade,
does the removal happens with old RBACs hence safe to remove those, or not safe?

same question more or less on the scheme CRD registration that was removed, but it depends on the implementation of upgrade, will try to learn about those deeper as well meanwhile

controllers/handlers/kubevirt.go
nirdothan: As far as I understand, beta FGs will be enabled by fefault in kv 1.9

oshoval: Thanks

by HCO or on Kubevirt itself ?
anyhow it can be done on follow-up if it will be on HCO
meanwhile dropping it here, as it was part of the annotation effects anyhow

assets/upgradePatches.json
```diff
@@ -342,6 +342,54 @@
         "name": "kubevirt-user-settings-reader-binding",
         "namespace": "default"
       }
+    },
+    {
+      "semverRange": ">=1.18.0 <1.19.0",
```
oshoval: upgrade is always only one step at a time ?

asking because if from 1.16.0 / 1.17.0 we can directly  jump to 1.19.0
those 2 also had this feature so we might need to start with >=1.16.0

but since you said start with 1.18.0, i believe it is one step at a time

Another question:
we don't need to backport to 4.22 (HCO v1.18) right ?

controllers/handlers/kubevirt.go
nunnatsa: @nirdothan - this is right, but not simple. In HCO we are required to disable all KV beta FG, unless there is a specific reason to set them. See the #4223 WIP PR.

The way we are implementing it is to keep the current logic to assemble the KV FG list, and then add all the beta FG to KV's disabled FG list, unless they are in the enabled list.

Bottom line - if you want this FG enabled in KV, it should be enabled here.

assets/upgradePatches.json
nunnatsa: I was maybe wrong about 1.18. If it got in before that, then we need to set the semverRange accordingly

nunnatsa: > we don't need to backport to 4.22 (HCO v1.18) right ?

No, we still need this feature in 1.18

controllers/handlers/kubevirt.go
oshoval: Since this PR removes the annotation as a whole, 
i think that enabling the FG merits a follow-up, instead making them related please
allowing to advance in parallel
in any case, even if we wouldnt remove the annotation, next release would enable the FG instead telling to use the annotation, it shows it is ortogonal

nunnatsa: In addition, there is a long time until KV 1.9 get here. we're currently using v1.8.2

assets/upgradePatches.json
oshoval: >I was maybe wrong about 1.18. If it got in before that, then we need to set the semverRange accordingly

fixed, and rebased by mistake in the same push

we can see here it is on 1.16
https://github.com/kubevirt/hyperconverged-cluster-operator/pull/3601/files#diff-2269e12c59476c013a2c8aabd7f10650154c5a662b254e19465423fd09bada92

controllers/handlers/kubevirt.go
```diff
@@ -952,10 +945,6 @@ func getFeatureGateChecks(spec hcov1beta1.HyperConvergedSpec, annotations map[st
 		fgs = append(fgs, kvHotplugVolumesGate)
 	}
 
-	if annotations[passt.DeployPasstNetworkBindingAnnotation] == "true" {
-		fgs = append(fgs, kvPasstBinding)
-	}
```
nunnatsa, line 957: See my response to @nirdothan question above. Be sure you want to drop this FG, even with a different logic.

oshoval:
change:

please look only on
	modified:   assets/upgradePatches.json
	modified:   controllers/hyperconverged/test-files/upgradePatches/upgradePatches.json
	modified:   pkg/upgradepatch/test-files/upgradePatches.json

changed to `>=1.16.0 <1.19.0`

controllers/handlers/kubevirt.go
orelmisan: @oshoval please note that`passtBinding` is a tech preview feature which requires explicit enabling.
It is not planned for graduation in KubeVirt v1.9.

assets/upgradePatches.json
```diff
@@ -342,6 +342,54 @@
...
+    {
+      "semverRange": ">=1.16.0 <1.19.0",
+      "groupVersionKind": {
+        "group": "apps",
+        "version": "v1",
+        "kind": "DaemonSet"
+      },
+      "objectKey": {
+        "name": "passt-binding-cni",
+        "namespace": "kubevirt-hyperconverged"
```
oshoval, line 355: OKD same NS as far as i remember
so we are good on U/S code

deploy/index-image/community-kubevirt-hyperconverged/1.19.0/manifests/kubevirt-hyperconverged-operator.v1.19.0.clusterserviceversion.yaml
```diff
@@ -644,17 +644,6 @@ spec:
           - create
           - update
           - delete
-        - apiGroups:
```
nunnatsa, line 647: We must keep it here, for the upgradePatch logic.

controllers/handlers/kubevirt.go
oshoval: Thanks
it is fine as far as i understand, this PR just removes the annotation and the logic below it,
this PR doesn't plan to enable the FG
my task is focused on removing the annotation and it's logic only

Kubevirt FG enabling is unrelated and other task
meanwhile 1.9 should say on RN to enable the FG (and stop saying to enable the annotation)

pkg/components/components.go
nunnatsa: We'll need to keep it for the removal logic. We don't need the CRD registration, I think, because we use Unstructured for that. But we do need the permissions.

nunnatsa (commented):
Thanks for the PR @oshoval ! it looks good in general. It's fun to see old code goes away...

Added a few comments.

controllers/handlers/kubevirt.go
orelmisan: > Thanks it is fine as far as i understand, this PR just removes the annotation and the logic below it, this PR doesn't plan to enable the FG my task is focused on removing the annotation and it's logic only
> 
> Kubevirt FG enabling is unrelated and other task meanwhile 1.9 should say on RN to enable the FG (and stop saying to enable the annotation)

@oshoval the annotation is the knob that enables the KubeVirt feature gate.

oshoval: HCO allows to update the HCO CR with Kubevirt FGs that it populate, isnt it ?
better if there is a formal way instead keeping the proprietary annotation just for that
(of course a managed way)

`deployPasstNetworkBinding` doesnt make sense to use for core network plugin if you ask me

anyhow whatever you decide

nunnatsa: The question is simple:

Do you need the FG enabled in KubeVirt? yes or no? Do you want to leave it to the user decision? I'm not following. What is the requirement?

orelmisan: @nunnatsa passtBinding is keeping its tech previous status.
I don't follow why we need to drop the annotation, as it is our API to enable the KubeVirt feature gate.

orelmisan: The requirement is to let cluster admins explicitly enable it.

nunnatsa: The reason we asked to do it in an annotation was that it was supposed to be relevant for a single release and we didn't want to add a new field to the v1beta1 api that will be deprecated almost immediately. 

So, you can leave the annotation if it's the know API. You can also add a new FG to v1, when we'll freeze the v1beta1 changes (very soon!). But we must have a way to add it to KV. We just can't drop this logic - it breaks the whole feature if we do that.

nunnatsa: And this is true for both KV 1.8 and KV 1.9 - even when it is enabled by default - if we won't set it here, we'll add it to the disabled list once it become beta in KV

orelmisan: Thank you @nunnatsa.
I support keeping the annotation for now.

oshoval: will keep

pkg/components/components.go
oshoval: thanks, 
note that on k8s upstream, we don't have SCC type,
but from your answer i believe that unknown CRDS are supported,
and that it `--ignore-not-found`

nunnatsa: Note sure what you mean. Didn't talk about SCC. also, from your answer I'm not sure what you meant by "CRD registration"

assets/upgradePatches.json
nunnatsa: yap.

pkg/components/components.go
oshoval: https://github.com/kubevirt/hyperconverged-cluster-operator/pull/4229/files#diff-a73b4d9c18bfef19861752d9071a045588f0a3406d8d4be1c136f67b1cff03fcL121

I mean we are removing the Add to scheme, as far as i understand you said CRD registration is not needed.

I derived from that, this is the general case, for every CRD,
so i mentioned SCC, which is another CRD we remove in the json, but isn't registered on vanilla k8s.

so for each CRD the upgrade is believed to support those two cases
1. even if the CRD isn't deployed (i.e NAD or SCC on k8s vanilla) the cleaner use unstructed so will be fine
2. if the resources are deployed, it is fine, it clean them only if they exist (sounds trivial, since they are opt-in)

nunnatsa: Oh. So the AddToScheme is a golang mechanism to convert K8s resource to a golang struct and vice versa. It's not a k8s issue, and it not related to a specific K8s cluster, and if the CRD is installed there or not.

nunnatsa: And since we don't use NAD anymore in the code, and sice we use the `Unstructured` type for the removal logic, then we don't need to run the NAD's AddToScheme.

oshoval:
restored the RBACs and rebased

still to TODO - restore the FG, will update once ready

oshoval:
/hold

didnt return yet the annotation for FG
thanks

nunnatsa:
> /hold
> 
> didnt return yet the annotation for FG thanks

OK.

/approve cancel

/lgtm cancel

oshoval:
Restored FG, as standalone commit, so it will be easier to drop it once desired
mentioned it on PR desc

basically it is the original code, just with hc instead hco, same as the other tests

https://github.com/kubevirt/hyperconverged-cluster-operator/pull/4229/files#diff-d7f85f01df8e1ce50ba2d3f7081518872c43eca662f49c1cb7dbb2eb5fdd6453R2133

nunnatsa (commented):
Let's don't preserve the passt package just to have the constant. Please move the constant (as a package private; e.g. `deployPasstNetworkBindingAnnotation`), to the controllers/handlers/kubevirt.go file, and remove the passt package.

oshoval:
done
also changed the comment of the unit test to reflect the real FG name PasstBinding

nunnatsa (approved):
/approve

deploy/olm-catalog/community-kubevirt-hyperconverged/1.19.0/manifests/kubevirt-hyperconverged-operator.v1.19.0.clusterserviceversion.yaml
```diff
@@ -9,7 +9,7 @@ metadata:
     certified: "false"
     console.openshift.io/disable-operand-delete: "true"
     containerImage: quay.io/kubevirt/hyperconverged-cluster-operator:1.19.0-unstable
-    createdAt: "2026-05-07 06:04:08"
+    createdAt: "2026-05-08 09:18:33"
```
oshoval, line 12: generic topic, not related to this PR
i think i saw that if one change this file manually, and omits the createdAt change, CI doesn't catch it, and it should
(unless i am wrong)

PR #4264, opened 2026-05-19
CNV-86222: Move the `upgradepatch` package to v1

nunnatsa:
This PR moves the internal implementation of the `upgradepatch` package to use the v1 API version.

In addition, a massive cleanup done for the upgradePatches.json file, to remove almost all the entries, as they related to a very old versions. We don't support upgrading from these versions to the current one anyway. That saves the need to translate the v1beta1 paths to v1 paths, in the json patch objects.

Note: Most of the changes in this PR were done by claude code, using the Claude Opus 4.6 model

**Jira Ticket**:

```jira-ticket
https://redhat.atlassian.net/browse/CNV-86222
```

**Release note**:

```release-note
None
```


tools/crwriter/crwriter.go
```diff
@@ -19,12 +18,14 @@ var (
...
 	outputFile string
 	header     string
+	apiVersion string
 )
 
 func init() {
 	flag.StringVar(&format, "format", "yaml", `output format. May be "json", "yaml" or "go"`)
 	flag.StringVar(&outputFile, "out", "", "output file name")
 	flag.StringVar(&header, "header", "", "path to an optional header text file, for go format")
+	flag.StringVar(&apiVersion, "api-version", "v1beta1", `API version to generate CR for. May be "v1beta1" or "v1"`)
```
orenc1, line 28: why the default is `v1beta1`, if all code paths are now using the api version of `v1`?

nunnatsa: We'll change it as part of CNV-86236.

pkg/upgradepatch/test-files/upgradePatches.json
orenc1, line 1: this file contains 16 old entries in `objectsToBeRemoved` list.
better to clean them up as well, to align with production jsonpatches file

pkg/upgradepatch/upgradePatches_test.go
```diff
@@ -195,78 +195,87 @@ var _ = Describe("upgradePatches", func() {
...
+
+			hc := components.GetOperatorV1CR()
+
+			ver, err := semver.Parse("1.18.0")
+			Expect(err).NotTo(HaveOccurred())
+
+			newHc, err := ApplyUpgradePatch(GinkgoLogr, hc, ver)
+			Expect(err).NotTo(HaveOccurred())
+
+			Expect(newHc).NotTo(BeNil())
```
orenc1, line 298: this only verifies that the operation didn't crash, but not that the remove actually took effect. consider something like
```
Expect(newHc.Spec.Virtualization.VirtualMachineOptions.DisableSerialConsoleLog).To(BeNil())
```

pkg/upgradepatch/test-files/upgradePatches.json
nunnatsa: done

pkg/upgradepatch/upgradePatches_test.go
nunnatsa: This is a good point, but not 100% accurate. We want to check that if a path does not include, it does not return error, so we can't check this. however, we don't have remove patch in our test - adding.

nunnatsa: Added a dedicated json file for this test.

orenc1 (approved):
/approve

PR #4296, opened 2026-06-09
Fix 409 conflict on webhook dry-run validation of operand CRs

orenc1:
The validating webhook's dry-run update of operand CRs (KubeVirt, CDI, etc.) could fail with a 409 Conflict when the HCO reconciler was concurrently updating the same CR. This race is more likely after an upgrade, where the reconciler performs additional reconciliation rounds.

Add retry-on-conflict logic to the webhook's `updateOperatorCr` so it re-fetches the operand and retries the dry-run update up to 3 times. Also wrap the `PatchHCO` call in the `RoleAggregationStrategy` func-test with `Eventually` to tolerate transient admission errors.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
NONE
```

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/webhooks/validator/validator.go
+++ b/pkg/webhooks/validator/validator.go
@@ -544,7 +544,39 @@ func getOperands(ctx context.Context, cli client.Client, isOpenshift bool) ([]cl
 	return resources, nil
 }
 
+const dryRunMaxRetries = 3
+
 func updateOperatorCr(ctx context.Context, cli client.Client, logger logr.Logger, hc *hcov1.HyperConverged, exists client.Object, opts *client.UpdateOptions) error {
+	for attempt := range dryRunMaxRetries {
+		if attempt > 0 {
+			if err := cli.Get(ctx, client.ObjectKeyFromObject(exists), exists); err != nil {
+				logger.Error(err, "failed to re-fetch object for dry-run retry", "kind", exists.GetObjectKind())
+				return err
+			}
+		}
+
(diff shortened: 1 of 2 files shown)


pkg/webhooks/validator/validator.go
```diff
@@ -544,7 +544,39 @@ func getOperands(ctx context.Context, cli client.Client, isOpenshift bool) ([]cl
 	return resources, nil
 }
 
+const dryRunMaxRetries = 3
+
 func updateOperatorCr(ctx context.Context, cli client.Client, logger logr.Logger, hc *hcov1.HyperConverged, exists client.Object, opts *client.UpdateOptions) error {
+	for attempt := 0; attempt < dryRunMaxRetries; attempt++ {
```
nunnatsa: [Modernize] Please use range; e.g.

```suggestion
	for attempt := range dryRunMaxRetries {
```

nunnatsa (commented):
Thanks @orenc1. Added one nit comment.

pkg/webhooks/validator/validator.go
orenc1: thanks, done.

PR #4297, opened 2026-06-09
Operands: Introduce network-resources-injector

nirdothan:
 ## Add network-resources-injector operand to HCO

  Add the network-resources-injector mutating admission webhook as an HCO-managed operand. This webhook automatically injects network resource requests/limits into pods based on their network-attachment-definitions, enabling proper resource accounting for
  SRIOV and other device-plugin-based networks.

Deployment mimics https://github.com/openshift/sriov-network-operator/tree/release-4.23/bindata/manifests/webhook

  **Why CNV needs its own instance:**

  The sriov-network-operator also deploys network-resources-injector, but CNV requires an independent instance to:
  1. Avoid dependency on whether SRIOV operator is installed
  2. Maintain control over the webhook configuration and lifecycle
  3. Prevent naming conflicts (both would create `network-resources-injector` ClusterRole/Deployment/etc.)

  The `cnv-` prefix ensures HCO's instance is clearly distinguished from upstream or SRIOV-operator deployments.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket

```

**Release note**:

```release-note
Deploy network-resources-injector 
```


controllers/handlers/netresinjector/common.go
```diff
@@ -0,0 +1,17 @@
+package netresinjector
+
+import (
+	hcoutil "github.com/kubevirt/hyperconverged-cluster-operator/pkg/util"
+)
+
+const (
+	appComponent       = hcoutil.AppComponentNetResInjector
```
nunnatsa: should we redefine this const?

controllers/handlers/netresinjector/webhook_config.go
```diff
@@ -0,0 +1,154 @@
...
+		}
+	} else {
+		mwc.Annotations = map[string]string{
+			"cert-manager.io/inject-ca-from": util.GetOperatorNamespaceFromEnv() + "/" + tlsCertificateName,
+		}
+	}
+
+	failPolicy := admissionregistrationv1.Ignore
+	sideEffects := admissionregistrationv1.SideEffectClassNone
+	scope := admissionregistrationv1.AllScopes
```
nunnatsa: Please use `ptr.To` instead of these new variables.

nunnatsa (commented):
Thanks @nirdothan 

Added a few nit comments.

pkg/components/components.go
```diff
@@ -473,9 +473,20 @@ func GetClusterPermissions() []rbacv1.PolicyRule {
 		roleWithAllPermissions("", stringListToSlice("services")),
 		{
 			APIGroups: emptyAPIGroup,
-			Resources: stringListToSlice("pods", "nodes"),
+			Resources: stringListToSlice("nodes"),
```
nunnatsa: Why?

nirdothan: A mistake off course. Fixed.

controllers/handlers/netresinjector/webhook_config.go
nirdothan: Done.

nirdothan:
[Rebased](https://github.com/kubevirt/hyperconverged-cluster-operator/compare/e309a9780d1859526932b55628d40dae3c40e85b..8f3d42008252e9ad58814d4ede671906902e9ee8) and [addressed](https://github.com/kubevirt/hyperconverged-cluster-operator/compare/8f3d42008252e9ad58814d4ede671906902e9ee8..41c21bb711644eb5b34d963aaeb47ee4aa594a60) comments by @nunnatsa

nirdothan:
[Fixed](https://github.com/kubevirt/hyperconverged-cluster-operator/compare/41c21bb711644eb5b34d963aaeb47ee4aa594a60..43132d0d83fc0c4e666bad146fdfd8cb8401af4f) another RBAC mistake, and linter failure.

nirdothan:
[Rebased](https://github.com/kubevirt/hyperconverged-cluster-operator/compare/43132d0d83fc0c4e666bad146fdfd8cb8401af4f..4fad0af86b5c1c31a01b1037dd7d58164ae57588) to avoid merge conflicts.

oshoval:
can you please make sure CNV isn't always shipped with it ?
if it does, would we be able to unrequire this PR ?
as they don't live together anyhow at the same time, so rest of the reasons won't catch, right ?

nirdothan:
> can you please make sure CNV isn't always shipped with it ? if it does, would we be able to unrequire this PR ? as they don't live together anyhow at the same time, so rest of the reasons won't catch, right ?

I'm not sure I follow. We do want CNV to always ship with this. However once we accomplish it, we plan to have `autopilot` disable it if `sriov-networks-operator` is installed and has it enabled.

oshoval:
> > can you please make sure CNV isn't always shipped with it ? if it does, would we be able to unrequire this PR ? as they don't live together anyhow at the same time, so rest of the reasons won't catch, right ?
> 
> I'm not sure I follow. We do want CNV to always ship with this. However once we accomplish it, we plan to have `autopilot` disable it if `sriov-networks-operator` is installed and has it enabled.

This what i mean, first we must always have it disabled in this PR if the other is already installed, we can't do it in phases, unless you have some knob already (i didn't review).
But if CNV already comes always with the SRIOV operator, then i think we don't need to be defensive, unless a customer raise a case.

nirdothan:
> > > can you please make sure CNV isn't always shipped with it ? if it does, would we be able to unrequire this PR ? as they don't live together anyhow at the same time, so rest of the reasons won't catch, right ?
> > 
> > 
> > I'm not sure I follow. We do want CNV to always ship with this. However once we accomplish it, we plan to have `autopilot` disable it if `sriov-networks-operator` is installed and has it enabled.
> 
> This what i mean, first we must always have it disabled in this PR if the other is already installed, we can't do it in phases, unless you have some knob already (i didn't review). But if CNV already comes always with the SRIOV operator, then i think we don't need to be defensive, unless a customer raise a case.

We're actually taking the opposite approach: Since 2 instances of the webhook **can safely live together** being idempotent, they will (potentially) be deployed duplicate at first, and only later we will optimize the deployment with autopilot (and potentially an HCO knob).

oshoval:
> > > > can you please make sure CNV isn't always shipped with it ? if it does, would we be able to unrequire this PR ? as they don't live together anyhow at the same time, so rest of the reasons won't catch, right ?
> > > 
> > > 
> > > I'm not sure I follow. We do want CNV to always ship with this. However once we accomplish it, we plan to have `autopilot` disable it if `sriov-networks-operator` is installed and has it enabled.
> > 
> > 
> > This what i mean, first we must always have it disabled in this PR if the other is already installed, we can't do it in phases, unless you have some knob already (i didn't review). But if CNV already comes always with the SRIOV operator, then i think we don't need to be defensive, unless a customer raise a case.
> 
> We're actually taking the opposite approach: Since 2 instances of the webhook **can safely live together** being idempotent, they will (potentially) be deployed duplicate at first, and only later we will optimize the deployment with autopilot (and potentially an HCO knob).

i don't like it tbh, better to not postpone the autopilot, or at least hide it behind annotation that can disable it,
moreover, the first question is the original - does CNV has a case where it isn't deployed ?
or the default includes it.
Moreover what about just having RN, if you need this NetRI, simply deploy SR-IOV operator, it is true it is heavier, but having this grace will allow us to maintain less and see it is a real deal or not.

EDIT  -in the long run, if this is a real case, maybe the NetRI can be extracted as standalone that both HCO and SR-IOV Operator can deploy, reusable.

nunnatsa:
@nirdothan - please rebase. there was an issue with the upgrade lanes that was solved last week.

nirdothan:
> @nirdothan - please rebase. there was an issue with the upgrade lanes that was solved last week.

[Done ](https://github.com/kubevirt/hyperconverged-cluster-operator/compare/4fad0af86b5c1c31a01b1037dd7d58164ae57588..e897731a0cad1fec0d524c916b1cf19911c1a6e4).

nirdothan:
> does CNV has a case where it isn't deployed ? 

No we always need it. Bridge Marker, AFAIC it's an extension of `multus` that is always required if multus is used. 

>  Moreover what about just having RN, if you need this NetRI, simply deploy SR-IOV operator, it is true it is heavier, but having this grace will allow us to maintain less and see it is a real deal or not.

Most CNV customers do not use SR-IOV and would needlessly "suffer" from deploying sriov-network-operator. The rest of them can manually toggle off the second copy in the sriov-network-operator CR. 

> 
> EDIT -in the long run, if this is a real case, maybe the NetRI can be extracted as standalone that both HCO and SR-IOV Operator can deploy, reusable.

AFAIC in the long run they would both be gone due to DRA, and perhaps if we discontinue Brige Marker at some point we might also endup reverting this PR.

oshoval:
> > does CNV has a case where it isn't deployed ?
> 
> No we always need it. Bridge Marker, AFAIC it's an extension of `multus` that is always required if multus is used.

If we always need it, it means it is always deployed already, and then we don't need it via HCO (this PR).
This what i mean, i strive to say it is wrong to have it on HCO imho.
It worked like that for years,
dual ways to deploy / dual deployment - just confuse and prone for mistakes.

> > Moreover what about just having RN, if you need this NetRI, simply deploy SR-IOV operator, it is true it is heavier, but having this grace will allow us to maintain less and see it is a real deal or not.
> 
> Most CNV customers do not use SR-IOV and would needlessly "suffer" from deploying sriov-network-operator. The rest of them can manually toggle off the second copy in the sriov-network-operator CR.

Anyhow, we always have it already, so this is a secondary question, but even if not, i would say lets first see someone miss it and then think.

> > EDIT -in the long run, if this is a real case, maybe the NetRI can be extracted as standalone that both HCO and SR-IOV Operator can deploy, reusable.
> 
> AFAIC in the long run they would both be gone due to DRA, and perhaps if we discontinue Brige Marker at some point we might also endup reverting this PR.

it would take time, still i think the right thing is the make it standalone product as described and not overload HCO.

EDIT:
1. About co existence, beside the fact it is heavy, not required, confusing, the order i believe can be random / depend on deployment order - hence the behavior might depend on the first one that fails,
it makes the configurable params confusing.
2. Lets say one user report a bug that happens with a specific setup (i.e which one / both is deployed and what the params), it is confusing to debug and not a good SLA, to have such cases imo.

controllers/operands/pdbHandler.go
```diff
@@ -0,0 +1,72 @@
...
+		}
+		util.MergeLabels(&pdb.ObjectMeta, &found.ObjectMeta)
+		pdb.Spec.DeepCopyInto(&found.Spec)
+		err := Client.Update(req.Ctx, found)
+		if err != nil {
+			return false, false, err
+		}
+		return true, !req.HCOTriggered, nil
+	}
+	return false, false, nil
```
nunnatsa: Let's reduce the cognitive load here; WDYT about something like that?


```suggestion
	if hasPDBRightFields(found, pdb) {
		return false, false, nil
	}
	
	if req.HCOTriggered {
		req.Logger.Info("Updating existing PodDisruptionBudget Spec to new opinionated values")
	} else {
		req.Logger.Info("Reconciling an externally updated PodDisruptionBudget's Spec to its opinionated values")
	}
	util.MergeLabels(&pdb.ObjectMeta, &found.ObjectMeta)
	pdb.Spec.DeepCopyInto(&found.Spec)
	err := Client.Update(req.Ctx, found)
	if err != nil {
		return false, false, err
	}

	return true, !req.HCOTriggered, nil
```

pkg/components/components.go
```diff
@@ -473,9 +473,30 @@ func GetClusterPermissions() []rbacv1.PolicyRule {
...
+		{
+			APIGroups: stringListToSlice("extensions"),
+			Resources: stringListToSlice("pods", "daemonsets", "replicasets", "statefulsets", "replicationcontrollers"),
+			Verbs:     stringListToSlice("*"),
+		},
+		{
+			APIGroups: stringListToSlice("k8s.cni.cncf.io"),
+			Resources: stringListToSlice("network-attachment-definitions"),
+			Verbs:     stringListToSlice("*"),
+		},
```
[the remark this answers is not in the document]
nunnatsa: @nirdothan - please fix this too

nirdothan:
- [Reduced](https://github.com/kubevirt/hyperconverged-cluster-operator/compare/e897731a0cad1fec0d524c916b1cf19911c1a6e4..8f4af3bffce541a080d2825922dc8f48111d089f) complexity per SonarCloud's analysis.
- [Rebased](https://github.com/kubevirt/hyperconverged-cluster-operator/compare/8f4af3bffce541a080d2825922dc8f48111d089f..11099ad1a588388697f90601e6af4e00b2e09ab8) over main.

controllers/handlers/netresinjector/pdb.go
```diff
@@ -0,0 +1,44 @@
...
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/kubevirt/hyperconverged-cluster-operator/controllers/operands"
+	hcoutil "github.com/kubevirt/hyperconverged-cluster-operator/pkg/util"
+)
+
+func NewPDBHandler(cli client.Client, scheme *runtime.Scheme) operands.Operand {
+	return operands.NewPDBHandler(cli, scheme, newPDB())
+}
```
nunnatsa, line 67: Please simplify the pdb handling. We're not expecting an additional pdb soon, and it's kind of pre-mature optimization. We don't need the generalization of it. could you please move all the logic here and remove the operands/pdbHandler.go file? It will (probably) allow you to simplify the code a bit.

controllers/handlers/netresinjector/webhook_config.go
```diff
@@ -0,0 +1,201 @@
...
+// preserveExistingCABundles copies CA bundle values from existing webhooks into required webhooks.
+// This prevents wiping out CA bundles that were injected by the cluster (e.g., service-ca operator
+// on OpenShift or cert-manager on non-OpenShift).
+func preserveExistingCABundles(required, existing *admissionregistrationv1.MutatingWebhookConfiguration) {
+	for i := range required.Webhooks {
+		if i < len(existing.Webhooks) && len(required.Webhooks[i].ClientConfig.CABundle) == 0 {
+			required.Webhooks[i].ClientConfig.CABundle = existing.Webhooks[i].ClientConfig.CABundle
+		}
+	}
+}
```
nunnatsa, line 73: Can we assume the order of the slices is the same?

controllers/handlers/netresinjector/webhook_config.go
```diff
@@ -0,0 +1,201 @@
...
+
+// needsUpdate determines if the existing webhook configuration needs to be updated.
+func needsUpdate(required, existing *admissionregistrationv1.MutatingWebhookConfiguration) bool {
+	return !hcoutil.CompareLabels(required, existing) ||
+		!reflect.DeepEqual(required.Webhooks, existing.Webhooks) ||
+		!hasRequiredAnnotations(existing.Annotations, required.Annotations)
+}
```
nunnatsa, line 80: Shouldn't we also check the `webhooks` filed? In the Current code, changes in the `webhooks` field are ignored and are not reconciled.

controllers/handlers/netresinjector/webhook_config.go
```diff
@@ -0,0 +1,201 @@
...
+		return
+	}
+
+	if existing.Annotations == nil {
+		existing.Annotations = make(map[string]string)
+	}
+
+	for k, v := range required.Annotations {
+		existing.Annotations[k] = v
+	}
```
nunnatsa: Can be replaced by:

```suggestion
	maps.Copy(existing.Annotations, required.Annotations)
```

controllers/handlers/netresinjector/webhook_config.go
```diff
@@ -0,0 +1,201 @@
...
+			AdmissionReviewVersions: []string{"v1", "v1beta1"},
+			SideEffects:             ptr.To(admissionregistrationv1.SideEffectClassNone),
+			FailurePolicy:           ptr.To(admissionregistrationv1.Ignore),
+			TimeoutSeconds:          ptr.To[int32](10),
+			ClientConfig: admissionregistrationv1.WebhookClientConfig{
+				Service: &admissionregistrationv1.ServiceReference{
+					Name:      serviceName,
+					Namespace: hcoutil.GetOperatorNamespaceFromEnv(),
+					Path:      ptr.To("/mutate"),
+					Port:      ptr.To[int32](443),
```
nunnatsa: Please change all `ptr.To()` ==> `new()`

nunnatsa: (in the whole PR, please)

controllers/operands/pdbHandler.go
```diff
@@ -0,0 +1,72 @@
+package operands
```
nunnatsa: Please drop this general implementation, and move it to `handlers/netresinjector/pdb.go` file.

nunnatsa (commented):
Thanks for the PR @nirdothan 

code rabbit added a few good comments - please address them.

Please also add unit tests for the code in the new `netresinjector` package.

controllers/operands/pdbHandler.go
nirdothan: Done.

controllers/handlers/netresinjector/pdb.go
nirdothan: Done.

pkg/components/components.go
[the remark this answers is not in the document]
nirdothan: Rebased over https://github.com/kubevirt/hyperconverged-cluster-operator/pull/4314 which removes duplicate RBAC orphaned from `passt`.

controllers/handlers/netresinjector/webhook_config.go
nirdothan: Done.

controllers/handlers/netresinjector/webhook_config.go
nirdothan: No we can't, it was copied from [controllers/handlers/aie/webhook_config.go ](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/release-1.18/controllers/handlers/aie/webhook_config.go#L51).
Made it safer to match by name. Note that it's a single webhook anyway.

controllers/handlers/netresinjector/webhook_config.go
nirdothan: Isn't that what the row above checks? 
`!reflect.DeepEqual(required.Webhooks, existing.Webhooks)`

controllers/handlers/netresinjector/webhook_config.go
nirdothan: Replaced where possible. Not all instances are.

nirdothan (commented):
Thanks for your thorough review. Please note that I've also published https://github.com/kubevirt/hyperconverged-cluster-operator/pull/4314 and rebased on top.

nirdothan:
[Changes](https://github.com/kubevirt/hyperconverged-cluster-operator/compare/11099ad1a588388697f90601e6af4e00b2e09ab8..1f82d8bd393aa8ae8b93a70088cea72c1105f621) addressing @nunnatsa's last https://github.com/kubevirt/hyperconverged-cluster-operator/pull/4297#pullrequestreview-4496467296

nunnatsa:
> Replaced where possible. Not all instances are.

@nirdothan it is possible :smiley: If you refer to this form:
```go
Replicas: ptr.To[int32](2)
```
it can be changed to:
```go
Replicas: new(int32(2))
```

controllers/handlers/netresinjector/deployment.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	podLabels := operands.GetLabels(hcoutil.AppComponentNetResInjector)
+
+	infrastructureHighlyAvailable := nodeinfo.IsInfrastructureHighlyAvailable()
+	affinity := operands.GetPodAntiAffinity(podLabels[hcoutil.AppLabelComponent], infrastructureHighlyAvailable)
+
+	dep := NewDeploymentWithNameOnly()
+	dep.Spec = appsv1.DeploymentSpec{
+		Replicas: ptr.To[int32](2),
```
nunnatsa:
```suggestion
		Replicas: new(int32(2)),
```

controllers/handlers/netresinjector/deployment_test.go
```diff
@@ -0,0 +1,145 @@
...
+	Context("newDeployment", func() {
+		It("should have all default values", func() {
+			dep := newDeployment(hco)
+
+			Expect(dep.Name).To(Equal(deploymentName))
+			Expect(dep.Namespace).To(Equal(hco.Namespace))
+			Expect(dep.Labels).To(HaveKeyWithValue(hcoutil.AppLabel, hcoutil.HyperConvergedName))
+			Expect(dep.Labels).To(HaveKeyWithValue(hcoutil.AppLabelComponent, string(hcoutil.AppComponentNetResInjector)))
+
+			Expect(dep.Spec.Replicas).To(Equal(ptr.To[int32](2)))
```
nunnatsa:
```suggestion
			Expect(dep.Spec.Replicas).To(HaveValue(Equal(int32(2))))
```

controllers/handlers/netresinjector/deployment_test.go
```diff
@@ -0,0 +1,145 @@
...
+			container := dep.Spec.Template.Spec.Containers[0]
+			Expect(container.Name).To(Equal("webhook-server"))
+			Expect(container.Image).To(Equal(testImage))
+			Expect(container.ImagePullPolicy).To(Equal(corev1.PullIfNotPresent))
+			Expect(container.Command).To(ConsistOf("webhook"))
+
+			Expect(container.SecurityContext).ToNot(BeNil())
+			Expect(container.SecurityContext.AllowPrivilegeEscalation).To(Equal(ptr.To(false)))
+			Expect(container.SecurityContext.ReadOnlyRootFilesystem).To(Equal(ptr.To(true)))
+			Expect(container.SecurityContext.RunAsNonRoot).To(Equal(ptr.To(true)))
```
nunnatsa:
```suggestion
			Expect(container.SecurityContext.AllowPrivilegeEscalation).To(HaveValue(BeFalse()))
			Expect(container.SecurityContext.ReadOnlyRootFilesystem).To(HaveValue(BeTrue()))
			Expect(container.SecurityContext.RunAsNonRoot).To(HaveValue(BeTrue()))
```

controllers/handlers/netresinjector/deployment_test.go
```diff
@@ -0,0 +1,145 @@
...
+			Expect(foundDeps.Items[0].Name).To(Equal(deploymentName))
+		})
+	})
+
+	Context("Deployment update", func() {
+		It("should update Deployment fields if not matched to the requirements", func() {
+			originalDep := newDeployment(hco)
+			modifiedDep := originalDep.DeepCopy()
+			modifiedDep.Spec.Template.Spec.Containers[0].Image = "malicious:tag"
+			modifiedDep.Spec.Template.Spec.Containers[0].SecurityContext.RunAsNonRoot = ptr.To(false)
```
nunnatsa:
```suggestion
			modifiedDep.Spec.Template.Spec.Containers[0].SecurityContext.RunAsNonRoot = new(false)
```

controllers/handlers/netresinjector/deployment_test.go
```diff
@@ -0,0 +1,145 @@
...
+			handler := NewDeploymentHandler(cl, commontestutils.GetScheme())
+			res := handler.Ensure(req)
+
+			Expect(res.Err).ToNot(HaveOccurred())
+			Expect(res.Updated).To(BeTrue())
+
+			reconciledDep := &appsv1.Deployment{}
+			Expect(cl.Get(context.Background(), client.ObjectKey{Name: res.Name, Namespace: hco.Namespace}, reconciledDep)).To(Succeed())
+			Expect(reconciledDep.Spec.Template.Spec.Containers[0].Image).To(Equal(testImage))
+			Expect(reconciledDep.Spec.Template.Spec.Containers[0].SecurityContext.RunAsNonRoot).To(Equal(ptr.To(true)))
```
nunnatsa:
```suggestion
			Expect(reconciledDep.Spec.Template.Spec.Containers[0].SecurityContext.RunAsNonRoot).To(HaveValue(BeTrue()))
```

controllers/handlers/netresinjector/service.go
```diff
@@ -0,0 +1,47 @@
...
+				hcoutil.AppLabelComponent: string(hcoutil.AppComponentNetResInjector),
+			},
+		},
+	}
+
+	if hcoutil.GetClusterInfo().IsOpenshift() {
+		svc.Annotations = map[string]string{
+			"service.beta.openshift.io/serving-cert-secret-name": tlsSecretName,
+		}
+	}
```
nunnatsa, line 44: Please add unit tests for the two cases (plain K8s vs openshift).

You can use the ClusterInfoMock like this:
```go
It(... func(){
		origGetClusterInfo := hcoutil.GetClusterInfo
		hcoutil.GetClusterInfo = clusterinfo.NewGetClusterInfo(clusterinfo.WithIsOpenshift(true))

		DeferCleanup(func() {
			hcoutil.GetClusterInfo = origGetClusterInfo
		})

		...
})
```
See example here: https://github.com/kubevirt/hyperconverged-cluster-operator/blob/0a0cdd3bbecc351cd3ba6209d450fdd551327b41/controllers/webhooks/conversion-webhook-controller/controller_test.go#L35

controllers/handlers/netresinjector/webhook_config.go
```diff
@@ -0,0 +1,203 @@
...
+		}
+	}
+
+	mwc.Webhooks = []admissionregistrationv1.MutatingWebhook{
+		{
+			Name:                    "cnv-network-resources-injector-config.k8s.io",
+			AdmissionReviewVersions: []string{"v1", "v1beta1"},
+			SideEffects:             new(admissionregistrationv1.SideEffectClassNone),
+			FailurePolicy:           new(admissionregistrationv1.Ignore),
+			TimeoutSeconds:          ptr.To[int32](10),
```
nunnatsa:
```suggestion
			TimeoutSeconds:          new(int32(10)),
```

controllers/handlers/netresinjector/webhook_config.go
```diff
@@ -0,0 +1,203 @@
...
+			ClientConfig: admissionregistrationv1.WebhookClientConfig{
+				Service: &admissionregistrationv1.ServiceReference{
+					Name:      serviceName,
+					Namespace: hcoutil.GetOperatorNamespaceFromEnv(),
+					Path:      new("/mutate"),
+					Port:      ptr.To[int32](443),
```
nunnatsa:
```suggestion
					Port:      new(int32(443)),
```

controllers/handlers/netresinjector/webhook_config.go
nunnatsa: you are right. Thanks

controllers/handlers/netresinjector/webhook_config.go
```diff
@@ -0,0 +1,203 @@
...
+
+	if hcoutil.GetClusterInfo().IsOpenshift() {
+		mwc.Annotations = map[string]string{
+			"service.beta.openshift.io/inject-cabundle": "true",
+		}
+	} else {
+		mwc.Annotations = map[string]string{
+			"cert-manager.io/inject-ca-from": hcoutil.GetOperatorNamespaceFromEnv() + "/" + tlsCertificateName,
+		}
+	}
```
nunnatsa, line 139: Please unit test this as explained above.

nunnatsa (commented):
Thanks for fixing the comments. Added some more ): 

We'll also need to watch the PDB resource. Add it here: https://github.com/kubevirt/hyperconverged-cluster-operator/blob/f5499c2045ed40cdd2c4f22e75f0f7ed30b87e65/controllers/hyperconverged/hyperconverged_controller.go#L173-L230

pkg/components/components.go
```diff
@@ -585,11 +596,6 @@ func GetClusterPermissions() []rbacv1.PolicyRule {
 			Resources: stringListToSlice("serviceaccounts"),
 			Verbs:     stringListToSlice("get", "list", "watch", "create", "update", "delete"),
 		},
-		{
-			APIGroups: stringListToSlice("k8s.cni.cncf.io"),
-			Resources: stringListToSlice("network-attachment-definitions"),
-			Verbs:     stringListToSlice("get", "list", "watch", "create", "update", "delete"),
-		},
```
nunnatsa, line 592: Please keep this. We need it in order to remove passt on upgrade.

pkg/components/components.go
```diff
@@ -476,6 +476,17 @@ func GetClusterPermissions() []rbacv1.PolicyRule {
 			Resources: stringListToSlice("pods", "nodes"),
 			Verbs:     stringListToSlice("get", "list", "watch", "patch"),
 		},
+		// Permissions needed to create cnv-network-resources-injector ClusterRole
+		{
+			APIGroups: stringListToSlice("k8s.cni.cncf.io"),
+			Resources: stringListToSlice("network-attachment-definitions"),
+			Verbs:     stringListToSlice("watch", "list", "get"),
+		},
```
nunnatsa: Please drop this, and restore the old one.

pkg/components/components.go
```diff
@@ -476,6 +476,17 @@ func GetClusterPermissions() []rbacv1.PolicyRule {
...
+		{
+			APIGroups: emptyAPIGroup,
+			Resources: stringListToSlice("configmaps"),
+			Verbs:     stringListToSlice("watch", "list", "get"),
+		},
```
nunnatsa: Please drop this. It duplicates line 467.

controllers/handlers/netresinjector/deployment.go
nirdothan: You're a go genius. Done.

controllers/handlers/netresinjector/deployment_test.go
nirdothan: Done.

controllers/handlers/netresinjector/deployment_test.go
nirdothan: Done.

controllers/handlers/netresinjector/deployment_test.go
nirdothan: Done.

controllers/handlers/netresinjector/deployment_test.go
nirdothan: Done.

controllers/handlers/netresinjector/service.go
nirdothan: Done.

controllers/handlers/netresinjector/webhook_config.go
nirdothan: Done.

controllers/handlers/netresinjector/webhook_config.go
nirdothan: Done.

pkg/components/components.go
nirdothan: Done.

pkg/components/components.go
nirdothan: Done.

pkg/components/components.go
nirdothan: Done.

nirdothan:
[Addressed](https://github.com/kubevirt/hyperconverged-cluster-operator/compare/1f82d8bd393aa8ae8b93a70088cea72c1105f621..16cada1947d39f99598e3a259fed3148b6582db4) last [review](https://github.com/kubevirt/hyperconverged-cluster-operator/pull/4297#pullrequestreview-4503826456) and [rebased](https://github.com/kubevirt/hyperconverged-cluster-operator/compare/16cada1947d39f99598e3a259fed3148b6582db4..d560719c0eeddc49ca0dc4161f9639d2ca8e0013).

dsanatar:
hpp-operator was patched to fix this mounter issue however, I believe we need to make a new release and hco needs to bump to that new version before we can see the CI pick up these changes

nunnatsa:
There is a good chance that the same issue also exists in non-upgrade test, but it then causes nothing. But then it is visible in the HCO log.

Running several simple non-upgrade tests

/test pull-hyperconverged-cluster-operator-e2e-k8s-1.36 hco-e2e-operator-sdk-gcp hco-e2e-operator-sdk-azure

nirdothan:
Attempted [fix](https://github.com/kubevirt/hyperconverged-cluster-operator/compare/6210a1035c15815086e11aba1a56734467ec8c71..dcf4d354a541c2a8c1b2e09355ae99aea23686f0) of webhook reconciliation and [rebased](https://github.com/kubevirt/hyperconverged-cluster-operator/compare/dcf4d354a541c2a8c1b2e09355ae99aea23686f0..6fe2d73578e478843be8ef82291e77fe5ffaf428).

nirdothan:
[Squashed](https://github.com/kubevirt/hyperconverged-cluster-operator/compare/6fe2d73578e478843be8ef82291e77fe5ffaf428..4d50ec80d3a58d3dd670a294363b6f3c8c6a027f) commits.

nunnatsa (approved):

nunnatsa (approved):
/approve

PR #4303, opened 2026-06-11
CNV-83939: replace `node-role.kubernetes.io/control-plane` label with `node-role.kubevirt.io/control-plane` for HCP Clusters

orenc1:
Use a kubevirt-specific label for HCP worker nodes instead of the standard Kubernetes control-plane label, in order to prevent misidentification of workers as control-plane nodes. Handle upgrade by removing the old HCO-set label when the new one is applied.
the new label of `node-role.kubevirt.io/control-plane` for `virt-operator` placement has been introduced at https://github.com/kubevirt/kubevirt/pull/17763

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://redhat.atlassian.net/browse/CNV-83939
```

**Release note**:

```release-note
Label HCP worker nodes with node-role.kubevirt.io/control-plane rather than node-role.kubernetes.io/control-plane
```

### DIFF
--- a/controllers/nodes/nodes_controller.go
+++ b/controllers/nodes/nodes_controller.go
@@ -284,19 +284,41 @@ func HandleHyperShiftNodeLabeling(ctx context.Context, cli client.Client, nodeNa
 	return labelNode(ctx, cli, node, logger)
 }
 
-// labelNode applies the HyperShift label on a single node
+// labelNode applies the HyperShift label on a single node.
+// It adds node-role.kubevirt.io/control-plane and removes the old
+// node-role.kubernetes.io/control-plane label if it was previously set by HCO (upgrade path).
 func labelNode(ctx context.Context, cli client.Client, node *corev1.Node, logger logr.Logger) error {
 	if !isWorkerNode(node) {
 		return nil
 	}
 
-	logger.Info("Adding control-plane label to worker node",
-		"node", node.Name,
-		"labelValue", hypershiftLabelValue,
-	)
-
 	patch := client.MergeFrom(node.DeepCopy())
-	node.Labels[nodeinfo.LabelNodeRoleControlPlane] = hypershiftLabelValue
+	needsPatch := false
+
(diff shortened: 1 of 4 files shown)


controllers/nodes/nodes_controller.go
```diff
@@ -284,19 +284,41 @@ func HandleHyperShiftNodeLabeling(ctx context.Context, cli client.Client, nodeNa
...
+			"node", node.Name,
+			"label", nodeinfo.LabelNodeRoleKubevirtControlPlane,
+			"labelValue", hypershiftLabelValue,
+		)
+		node.Labels[nodeinfo.LabelNodeRoleKubevirtControlPlane] = hypershiftLabelValue
+		needsPatch = true
+	}
+
+	// Upgrade path: remove the old kubernetes control-plane label if it was set by HCO
+	if val, exists := node.Labels[nodeinfo.LabelNodeRoleControlPlane]; exists && val == hypershiftLabelValue {
```
nunnatsa: It's enough to check the value. if the key does not exist, it will return an empty string, which is also != `hypershiftLabelValue`.; e.g.

```suggestion
	if node.Labels[nodeinfo.LabelNodeRoleControlPlane] == hypershiftLabelValue {
```

nunnatsa (commented):
Added a nit comment.

controllers/nodes/nodes_controller.go
orenc1: right, the `exists` check is redundant in this case. fixed. thanks.

nunnatsa (approved):
/approve

PR #4306, opened 2026-06-11
Enable DeclarativeHotplugVolume by default

dsanatar:

`DeclarativeHotplugVolumes` supersedes the now deprecated `HotplugVolumes`. This PR intends to enable Declarative by default removing the logic that previously defaulted to the using legacy `HotplugVolumes` when Declarative wasn't explicitly enabled

Addresses: https://redhat.atlassian.net/browse/CNV-81481

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://redhat.atlassian.net/browse/CNV-69390
```

**Release note**:

```release-note
Enable DeclarativeHotplugVolume by default
```


nunnatsa:
Hi @dsanatar .

`DeclarativeHotplugVolumes` is not GA in KubeVirt, but it's beta.

/hold

dsanatar:
> `DeclarativeHotplugVolumes` is not GA in KubeVirt, but it's beta.
>

Hi @nunnatsa, yes I understand this is unconventional however,  we are planning to announce this feature as GA in the upcoming 5.0 release and we don't want to require users to still opt in to use the feature.

controllers/handlers/kubevirt.go
```diff
@@ -915,12 +915,8 @@ func getFeatureGateChecks(hc *hcov1.HyperConverged) []string {
 		fgs = append(fgs, kvDecentralizedLiveMigration)
 	}
 
-	// Add the appropriate volume hotplug featuregate based on DeclarativeHotplugVolumes setting
 	if featureGates.IsEnabled("declarativeHotplugVolumes") {
 		fgs = append(fgs, kvDeclarativeHotplugVolumesGate)
-	} else {
-		// Default behavior: use the original HotplugVolumes featuregate
-		fgs = append(fgs, kvHotplugVolumesGate)
```
nunnatsa, line 945: I think we still need this, if the FG is disabled.

nunnatsa: This FG is in alpha phase in the current version `v1.9.0-beta.0`. When the next version will be released, where the FG will be in beta and `HotplugVolumes` FG will be deprecated, we could continue.

nunnatsa:
To make the test work, please update the default functional test in https://github.com/kubevirt/hyperconverged-cluster-operator/blob/50cd65175855ec98db72c8d3632df8a3272a3e78/tests/func-tests/defaults_test.go#L78 

Change `BeFalesBecause()` to `BeTrueBecause()`

nunnatsa:
Please update the phase and the documentation here:
https://github.com/kubevirt/hyperconverged-cluster-operator/blob/485f9352820854b63776264aca0d5829bfa0ab54/docs/cluster-configuration.md?plain=1#L160-L164

and here:
https://github.com/kubevirt/hyperconverged-cluster-operator/blob/485f9352820854b63776264aca0d5829bfa0ab54/docs/cluster-configuration-v1beta1-api.md?plain=1#L230-L234

controllers/handlers/kubevirt.go
dsanatar: if the `DeclarativeHotplugVolumes` gate is explicitly disabled, we do not want to fall back to the legacy `HotplugVolumes` anymore since it will be deprecated. But yes these changes reflect the feature gate graduations that will be included in the 1.9 release.

dsanatar:
Hi @nunnatsa could you take another look at this when you get the chance, thanks!

nunnatsa:
Done in #4381

@dsanatar - you can close this PR.

dsanatar:
@nunnatsa awesome, thank you!
/close

PR #4315, opened 2026-06-16
fix: make DeprecatedMachineType alert test multi-arch aware

carterpewpew:

The DeprecatedMachineType alert functional test hardcodes pc-q35-rhel7.6.0 as the deprecated machine type to test with. This is an x86-only machine type and fails on s390x clusters with:
```
spec.template.spec.domain.machine.type is not supported: pc-q35-rhel7.6.0
```
This PR makes the test multi-arch aware by using the existing getArchs() helper to detect the cluster architecture and select the appropriate deprecated machine type:
1. amd64: pc-q35-rhel7.6.0 (existing behavior, no change)
2. s390x: s390-ccw-virtio-rhel7.6.0
3. Other: gracefully skips the test

Both machine types are the oldest RHEL-versioned types for their respective architecture, guaranteed to be marked Deprecated: "yes" by QEMU/libvirt on any modern (RHEL 9-based) CNV cluster.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
NONE
```

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/func-tests/monitoring_test.go
+++ b/tests/func-tests/monitoring_test.go
@@ -209,8 +209,19 @@ var _ = Describe("[crit:high][vendor:cnv-qe@redhat.com][level:system]Monitoring"
 			})
 	})
 
-	It("should fire the DeprecatedMachineType alert when a VM is using a deprecated machine type", func(ctx context.Context) {
-		const deprecatedMachineType = "pc-q35-rhel7.6.0"
+	It("should fire the DeprecatedMachineType alert when a VM is using a deprecated machine type", Label(tests.DeprecatedArchLabel), func(ctx context.Context) {
+		archs, err := getArchs(ctx)
+		Expect(err).ToNot(HaveOccurred())
+
+		var deprecatedMachineType string
+		switch {
+		case slices.Contains(archs, "amd64"):
+			deprecatedMachineType = "pc-q35-rhel7.6.0"
+		case slices.Contains(archs, "s390x"):
+			deprecatedMachineType = "s390-ccw-virtio-rhel7.6.0"
+		default:
+			Fail(fmt.Sprintf("no known deprecated machine type for architectures %v; Use the \"!%s\" label filter in order to skip this test", archs, tests.DeprecatedArchLabel))
+		}
 
 		By("Creating a VM with a deprecated machine type")
 		vm := &kubevirtcorev1.VirtualMachine{
--- a/tests/func-tests/utils.go
+++ b/tests/func-tests/utils.go
@@ -37,6 +37,7 @@ const (
 	HighlyAvailableClusterLabel = "HIGHLY_AVAILABLE_CLUSTER"
 	DestructiveLabel            = "DESTRUCTIVE"
 	OpenshiftLabel              = "OpenShift"
+	DeprecatedArchLabel         = "DEPRECATED_ARCH"
 
 	TestNamespace = "hco-test-default"
 )


carterpewpew:
cc @nekkunti

tests/func-tests/monitoring_test.go
```diff
@@ -196,21 +196,25 @@ var _ = Describe("[crit:high][vendor:cnv-qe@redhat.com][level:system]Monitoring"
...
+		Expect(err).ToNot(HaveOccurred())
+
+		var deprecatedMachineType string
+		switch {
+		case slices.Contains(archs, "amd64"):
+			deprecatedMachineType = "pc-q35-rhel7.6.0"
+		case slices.Contains(archs, "s390x"):
+			deprecatedMachineType = "s390-ccw-virtio-rhel7.6.0"
+		default:
+			Skip(fmt.Sprintf("no known deprecated machine type for architectures %v", archs))
```
nunnatsa: We prefer not using `Skip()`, as it tend to hide issues over time.

Instead, please add a label to this test (to the `It()` function), and fail the test if no known architecture is found. The label name can be something like `DEPRECATED_ARCH` or something similar.

The caller should know their system, and they should skip the test using the label (if running on an arm only cluster, for example).


```suggestion
			Fail(fmt.Sprintf("no known deprecated machine type for architectures %v", archs))
```

See example in this file. In line 53, we are adding the `"OpenShift"` label (`tests.OpenshiftLabel`), and then in line 72, we call the `tests.FailIfNotOpenShift()` function, to fail the test if running in plain K8s cluster. The test user should filter tests by the `"OpenShift"` label when running on plain K8s. (no need to define a special function for that. The `OpenShift` label is very common. That why there is a common function for it).

For example, here are the functional tests configurations for the kubevirt-ci lanes: https://github.com/kubevirt/project-infra/blob/bbf8dbabdbc6102078d8786979d3f0acce1bbc73/github/ci/prow-deploy/files/jobs/kubevirt/hyperconverged-cluster-operator/hyperconverged-cluster-operator-presubmits.yaml#L34-L35

nunnatsa (commented):
Thanks @carterpewpew 

This makes a lot of sense to me. My only concern is that we have no s390x cluster to check if this fix works.

Added one comment inline.

tests/func-tests/monitoring_test.go
carterpewpew: Makes sense, updated ! Swapped Skip for Fail and added a DEPRECATED_ARCH label to the test so lanes on unsupported architectures can just filter it out with GINKGO_LABELS. Thanks for taking a look

tests/func-tests/monitoring_test.go
```diff
@@ -209,8 +209,19 @@ var _ = Describe("[crit:high][vendor:cnv-qe@redhat.com][level:system]Monitoring"
...
+		Expect(err).ToNot(HaveOccurred())
+
+		var deprecatedMachineType string
+		switch {
+		case slices.Contains(archs, "amd64"):
+			deprecatedMachineType = "pc-q35-rhel7.6.0"
+		case slices.Contains(archs, "s390x"):
+			deprecatedMachineType = "s390-ccw-virtio-rhel7.6.0"
+		default:
+			Fail(fmt.Sprintf("no known deprecated machine type for architectures %v", archs))
```
nunnatsa: Sorry for missing this in my original comment. 

Let's make it a bit more user friendly:

```suggestion
			Fail(fmt.Sprintf("no known deprecated machine type for architectures %v; . Use the "!%s" label filter in order to skip this test", archs, tests.DeprecatedArchLabel))
```

carterpewpew: Done, Thanks for the catch

nunnatsa (approved):

PR #4329, opened 2026-06-21
Set the default architecture in the KubeVirt CR

nunnatsa:

This PR sets the KubeVirt's `spec.configuration.architectureConfiguration.defaultArchitecture` field.
 
Assuming a single control plan architecture, HCO will choose this architecture as default, if the workload architecture contains it. If not, HCO will choose the architecture used by most of the workload nodes.

**Jira Ticket**:

```jira-ticket
https://redhat.atlassian.net/browse/CNV-90106
```

**Release note**:

```release-note
Set the default architecture in the KubeVirt CR
```

### DIFF
--- a/pkg/internal/nodeinfo/architectures.go
+++ b/pkg/internal/nodeinfo/architectures.go
@@ -1,6 +1,7 @@
 package nodeinfo
 
 import (
+	"maps"
 	"slices"
 	"sync"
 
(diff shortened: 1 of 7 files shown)


pkg/internal/nodeinfo/architectures.go
```diff
@@ -26,6 +26,27 @@ func GetWorkloadsArchitectures() []string {
...
+
+	if cpArches := GetControlPlaneArchitectures(); len(cpArches) > 0 {
+		cpArch = cpArches[0]
+		if slices.Contains(wrkArches, cpArch) {
+			return cpArch
+		}
+	}
+
+	if len(wrkArches) > 0 {
+		return wrkArches[0]
```
orenc1: why are you taking specifically the first architecture on the list to be the default?
wouldn't it be more reasonable to put the default arch as the one that is in the highest number of workloads nodes in the cluster?

nunnatsa: That could be a future improvement. Currently we don't hold this info, but just a set of workload node architectures. But I thought about something else: adding a new field to let the user choose the default, and use this algorithm if this field is empty.

pkg/internal/nodeinfo/architectures.go
```diff
@@ -26,6 +26,27 @@ func GetWorkloadsArchitectures() []string {
 	return workloadArchitectures.get()
 }
 
+// GetDefaultArchitecture returns the default architecture for virtual machines
+// Assuming a single control plane architecture, HCO will choose this architecture as default, if the workload
+// architecture contains it. If not, HCO will choose the first architecture from the workload architecture list
+func GetDefaultArchitecture() string {
+	var cpArch string
```
orenc1: this var is declared in the outer scope, but only used in the if block in line 37.
you can replace with `:=` there

pkg/internal/nodeinfo/architectures.go
```diff
@@ -37,25 +51,61 @@ func newArchitectures() *Architectures {
...
+		modified = true
 	}
 
-	return false
+	if modified {
+		if len(cpArchesList) > 0 {
+			if cpArch := cpArchesList[0]; slices.Contains(a.workloadArches, cpArch) {
+				a.defaultArch = cpArch
+				return true
+			}
```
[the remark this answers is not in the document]
nunnatsa: Right. We assume CP only have a single architecture, common to all CP nodes.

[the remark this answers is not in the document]
nunnatsa: not really. I think for this is better

nunnatsa:
This will probably fail, but let's try:
/cherry-pick release-1.18

PR #4331, opened 2026-06-22
[release 1.18] many resources are not deleted when deleting the HCO CR

nunnatsa:

This is a manual cherry pick of #4322.

After moving the operands creation to be static in #4164, we introduced a bug where most of the resources created by HCO are not deleted when HCO is removed.

The reason for that is that the old dynamic addition of operand also added the resource to a list in the operandHandler, to be used for the deletion. But when we stopped using the dynamic code, the resources are not listed anymore, and are not deleted.

This commit fixes this issue by re-creating the list.

**Jira Ticket**:

```jira-ticket
https://redhat.atlassian.net/browse/CNV-90394
```

**Release note**:
```release-note
fix CNV-90394: many resources are not deleted when deleting
```

### DIFF
--- a/controllers/commontestutils/eventEmitterMock.go
+++ b/controllers/commontestutils/eventEmitterMock.go
@@ -3,6 +3,7 @@ package commontestutils
 import (
 	"context"
 	"reflect"
+	"slices"
 	"sync"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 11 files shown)


orenc1 (approved):
/approve

PR #4337, opened 2026-06-23
fix: make functest build and TLS check multi-arch aware

carterpewpew:

Prep work for adding an s390x e2e test lane. The functest Dockerfiles and TLS check script had hardcoded x86 binary URLs that prevent building/running on other architectures. This makes them architecture-aware using Docker's built-in TARGETARCH and distro packages instead of x86-only downloads.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
NONE
```

**Release note**:

```release-note
NONE
```

### DIFF
--- a/build/Dockerfile.functest
+++ b/build/Dockerfile.functest
@@ -9,6 +9,7 @@ RUN ARCH=${TARGETARCH} make build-functest
 
 FROM --platform=${TARGETPLATFORM} registry.access.redhat.com/ubi9/ubi-minimal
 ARG TARGETPLATFORM
+ARG TARGETARCH
 
 ENV USER_UID=1001 \
     TEST_OUT_PATH=/test
@@ -16,12 +17,11 @@ ENV USER_UID=1001 \
 WORKDIR ${TEST_OUT_PATH}
 ENTRYPOINT ["./hack/run-tests.sh"]
 
-RUN microdnf install -y tar gzip which openssl diffutils && \
-    curl -Lo /usr/local/bin/kubectl "https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl" && \
+RUN microdnf install -y tar gzip which openssl diffutils jq && \
+    microdnf clean all && \
+    curl -fLo /usr/local/bin/kubectl "https://storage.googleapis.com/kubernetes-release/release/$(curl -fs https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/${TARGETARCH}/kubectl" && \
     chmod a+x /usr/local/bin/kubectl && \
-    curl -Lsv --fail https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/stable/openshift-client-linux.tar.gz | tar xvzf - -C /usr/local/bin/ oc && \
-    curl -Lo /usr/local/bin/jq https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64 && \
-    chmod a+x /usr/local/bin/jq
+    curl -fsSL "https://mirror.openshift.com/pub/openshift-v4/${TARGETARCH}/clients/ocp/stable/openshift-client-linux.tar.gz" | tar xzf - -C /usr/local/bin/ oc
 
 COPY --from=builder --chown=${USER_UID}:0 /go/src/github.com/kubevirt/hyperconverged-cluster-operator/tests/func-tests/_out/func-tests.test  ${TEST_OUT_PATH}/
 COPY --from=builder --chown=${USER_UID}:0 /go/src/github.com/kubevirt/hyperconverged-cluster-operator/hack  ${TEST_OUT_PATH}/hack
--- a/build/Dockerfile.functest.ci
+++ b/build/Dockerfile.functest.ci
@@ -1,11 +1,14 @@
 # registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.26-openshift-5.0
-FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.26-openshift-5.0 AS builder
+FROM --platform=${BUILDPLATFORM} registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.26-openshift-5.0 AS builder
+
+ARG TARGETARCH
 
 WORKDIR /go/src/github.com/kubevirt/hyperconverged-cluster-operator/
 COPY . .
-RUN make build-functest
+RUN ARCH=${TARGETARCH} make build-functest
 
 FROM registry.access.redhat.com/ubi9/ubi-minimal
+ARG TARGETARCH
 
 ENV USER_UID=1001 \
     TEST_OUT_PATH=/test
@@ -13,12 +16,11 @@ ENV USER_UID=1001 \
 WORKDIR ${TEST_OUT_PATH}
 ENTRYPOINT ["./hack/run-tests.sh"]
 
-RUN microdnf install -y tar gzip which openssl diffutils && \
-    curl -Lo /usr/local/bin/kubectl "https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl" && \
+RUN microdnf install -y tar gzip which openssl diffutils jq && \
+    microdnf clean all && \
+    curl -fLo /usr/local/bin/kubectl "https://storage.googleapis.com/kubernetes-release/release/$(curl -fs https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/${TARGETARCH}/kubectl" && \
     chmod a+x /usr/local/bin/kubectl && \
-    curl -Lsv --fail https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/stable/openshift-client-linux.tar.gz | tar xvzf - -C /usr/local/bin/ oc && \
-    curl -Lo /usr/local/bin/jq https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64 && \
-    chmod a+x /usr/local/bin/jq
+    curl -fsSL "https://mirror.openshift.com/pub/openshift-v4/${TARGETARCH}/clients/ocp/stable/openshift-client-linux.tar.gz" | tar xzf - -C /usr/local/bin/ oc
 
 COPY --from=builder --chown=${USER_UID}:0 /go/src/github.com/kubevirt/hyperconverged-cluster-operator/tests/func-tests/_out/func-tests.test  ${TEST_OUT_PATH}/
 COPY --from=builder --chown=${USER_UID}:0 /go/src/github.com/kubevirt/hyperconverged-cluster-operator/hack  ${TEST_OUT_PATH}/hack
--- a/hack/check_tlsprofile.sh
+++ b/hack/check_tlsprofile.sh
@@ -20,7 +20,7 @@
 
 set -ex
 
-KUBECTL_BINARY=oc
+KUBECTL_BINARY=${KUBECTL_BINARY:-oc}
 INSTALLED_NAMESPACE=kubevirt-hyperconverged
 HCO_TYPE="hyperconvergeds"
 TLS_PATH="/spec/security/tlsSecurityProfile"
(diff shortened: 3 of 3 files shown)


build/Dockerfile.functest
```diff
@@ -9,19 +9,19 @@ RUN ARCH=${TARGETARCH} make build-functest
...
     TEST_OUT_PATH=/test
 
 WORKDIR ${TEST_OUT_PATH}
 ENTRYPOINT ["./hack/run-tests.sh"]
 
-RUN microdnf install -y tar gzip which openssl diffutils && \
-    curl -Lo /usr/local/bin/kubectl "https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl" && \
+RUN microdnf install -y tar gzip which openssl diffutils jq && \
+    curl -Lo /usr/local/bin/kubectl "https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/${TARGETARCH}/kubectl" && \
     chmod a+x /usr/local/bin/kubectl && \
```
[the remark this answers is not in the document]
carterpewpew: Good catch, added -f flag to both the kubectl download and the inner curl that fetches the stable version string.

build/Dockerfile.functest.ci
```diff
@@ -13,12 +13,13 @@ ENV USER_UID=1001 \
 WORKDIR ${TEST_OUT_PATH}
 ENTRYPOINT ["./hack/run-tests.sh"]
 
-RUN microdnf install -y tar gzip which openssl diffutils && \
-    curl -Lo /usr/local/bin/kubectl "https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl" && \
+RUN GOARCH=$(uname -m | sed 's/x86_64/amd64/' | sed 's/aarch64/arm64/') && \
+    OC_ARCH=$(uname -m | sed 's/aarch64/arm64/') && \
```
nunnatsa: TARGRETARCH is auto feed. It's much easier to use it.

nunnatsa (commented):
Thanks for the PR @carterpewpew 

I think the Dockerfile architecture logic can be simplified in some of the places. Also, the coderabbit comment seems to be useful.

build/Dockerfile.functest.ci
carterpewpew: Thanks for the catch. Both Dockerfiles now use TARGETARCH consistently. Dropped the uname -m + sed approach from Dockerfile.functest.ci.

hack/check_tlsprofile.sh
```diff
@@ -62,9 +62,17 @@ fi
...
+    if command -v microdnf &>/dev/null; then
+        microdnf install -y nmap
+    elif command -v dnf &>/dev/null; then
+        dnf install -y nmap
+    elif command -v yum &>/dev/null; then
+        yum install -y nmap
+    else
+        rpm -vhU --nodeps https://nmap.org/dist/nmap-7.94-1.x86_64.rpm
+        rpm -vhU https://nmap.org/dist/ncat-7.94-1.x86_64.rpm
+        rpm -vhU https://nmap.org/dist/nping-0.7.94-1.x86_64.rpm
```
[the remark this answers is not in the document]
carterpewpew: This fallback is intentionally kept as the original behavior. The script runs inside the functest container which is UBI9 minimal, so microdnf is always available and the RPM fallback is never reached. nmap.org only publishes x86_64 RPMs anyway, so there's nothing to substitute with uname -m. I'd rather keep the working original as a last-resort path than remove it or add an error that would never trigger in practice.

carterpewpew:
cc @nekkunti

build/Dockerfile.functest
```diff
@@ -9,19 +9,20 @@ RUN ARCH=${TARGETARCH} make build-functest
...
-    curl -Lo /usr/local/bin/kubectl "https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl" && \
+RUN microdnf install -y tar gzip which openssl diffutils jq && \
+    microdnf clean all && \
+    curl -fLo /usr/local/bin/kubectl "https://storage.googleapis.com/kubernetes-release/release/$(curl -fs https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/${TARGETARCH}/kubectl" && \
     chmod a+x /usr/local/bin/kubectl && \
-    curl -Lsv --fail https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/stable/openshift-client-linux.tar.gz | tar xvzf - -C /usr/local/bin/ oc && \
-    curl -Lo /usr/local/bin/jq https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64 && \
-    chmod a+x /usr/local/bin/jq
+    OC_ARCH=$([ "${TARGETARCH}" = "amd64" ] && echo "x86_64" || echo "${TARGETARCH}") && \
+    curl -Lsv --fail "https://mirror.openshift.com/pub/openshift-v4/${OC_ARCH}/clients/ocp/stable/openshift-client-linux.tar.gz" | tar xvzf - -C /usr/local/bin/ oc
```
nunnatsa: It seems like the architecture in the URL is ignored. 

We can use `https://mirror.openshift.com/pub/openshift-v4/clients/ocp/stable/` instead, and get to the same place. But when getting there, there are per-architecture version. 

I can only guess that the openshift-client-linux.tar.gz file is amd64.

WDYT about something like this? It will also cover arm64:
```suggestion
    curl -Lsv --fail https://mirror.openshift.com/pub/openshift-v4/clients/ocp/stable/openshift-client-linux-${TARGETARCH}-rhel9.tar.gz" | tar xvzf - -C /usr/local/bin/ oc
```

build/Dockerfile.functest.ci
```diff
@@ -3,22 +3,23 @@ FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.26-openshift-5.0 AS b
...
-    curl -Lo /usr/local/bin/kubectl "https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl" && \
+RUN microdnf install -y tar gzip which openssl diffutils jq && \
+    microdnf clean all && \
+    curl -fLo /usr/local/bin/kubectl "https://storage.googleapis.com/kubernetes-release/release/$(curl -fs https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/${TARGETARCH}/kubectl" && \
     chmod a+x /usr/local/bin/kubectl && \
-    curl -Lsv --fail https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/stable/openshift-client-linux.tar.gz | tar xvzf - -C /usr/local/bin/ oc && \
-    curl -Lo /usr/local/bin/jq https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64 && \
-    chmod a+x /usr/local/bin/jq
+    OC_ARCH=$([ "${TARGETARCH}" = "amd64" ] && echo "x86_64" || echo "${TARGETARCH}") && \
+    curl -Lsv --fail "https://mirror.openshift.com/pub/openshift-v4/${OC_ARCH}/clients/ocp/stable/openshift-client-linux.tar.gz" | tar xvzf - -C /usr/local/bin/ oc
```
nunnatsa: same

build/Dockerfile.functest
nunnatsa: After plying with it, it seems that this will give the best outcome:

```
"https://mirror.openshift.com/pub/openshift-v4/${TARGETARCH}/clients/ocp/stable/openshift-client-linux.tar.gz"
```

carterpewpew: Done. Both Dockerfiles now use https://mirror.openshift.com/pub/openshift-v4/${TARGETARCH}/clients/ocp/stable/openshift-client-linux.tar.gz directly, no OC_ARCH mapping needed.

build/Dockerfile.functest.ci
carterpewpew: Applied the same fix here.

build/Dockerfile.functest.ci
```diff
@@ -3,22 +3,22 @@ FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.26-openshift-5.0 AS b
 
 WORKDIR /go/src/github.com/kubevirt/hyperconverged-cluster-operator/
 COPY . .
-RUN make build-functest
+RUN ARCH=$(go env GOARCH) make build-functest
```
nunnatsa: Nit: Can we use the same method we used in `build/Dockerfile.functest`, for consistency?

e.g.

```suggestion
FROM --platform=${BUILDPLATFORM} registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.26-openshift-5.0 AS builder

ARG TARGETARCH

WORKDIR /go/src/github.com/kubevirt/hyperconverged-cluster-operator/
COPY . .
RUN ARCH=${TARGETARCH} make build-functest
```

This is not critical. But this way we can build it both in architecture dedicated build machine, or on any other machine.

hack/check_tlsprofile.sh
```diff
@@ -62,9 +62,17 @@ fi
...
     echo "Try to install nmap"
-    rpm -vhU --nodeps https://nmap.org/dist/nmap-7.94-1.x86_64.rpm
-    rpm -vhU https://nmap.org/dist/ncat-7.94-1.x86_64.rpm
-    rpm -vhU https://nmap.org/dist/nping-0.7.94-1.x86_64.rpm
+    if command -v microdnf &>/dev/null; then
+        microdnf install -y nmap
+    elif command -v dnf &>/dev/null; then
+        dnf install -y nmap
+    elif command -v yum &>/dev/null; then
+        yum install -y nmap
```
nunnatsa, line 70: not sure it will work. we need the ncat and nping packages. are they installed as part of nmap?

nunnatsa: BTW, we want to move the TLS test to go. then we won't need nmap at all

nunnatsa (commented):
added two questions. thanks!

build/Dockerfile.functest.ci
carterpewpew: Done. Updated the builder stage to use `--platform=${BUILDPLATFORM}` on the FROM line, declared `ARG TARGETARCH`, and switched to `RUN ARCH=${TARGETARCH} make build-functest`. Both Dockerfiles are now consistent.

hack/check_tlsprofile.sh
carterpewpew: Yes, nmap-ncat is a required dependency of the nmap RPM on UBI9 / CentOS Stream 9 (and 10). Installing nmap automatically pulls in nmap-ncat. nping is also bundled in the nmap package itself (not a separate RPM). So `microdnf install -y nmap` covers all three binaries: nmap, ncat, and nping.

Verified from the CentOS Stream 9 nmap spec: nmap-ncat is listed as `Requires: nmap-ncat` in the main nmap package, and nping is built as a subpackage from the same source RPM.

carterpewpew: Agreed, moving the TLS check to Go would be the cleaner long-term solution and eliminates the nmap dependency entirely. Happy to help with that in a follow-up PR if needed.

build/Dockerfile.functest.ci
```diff
@@ -1,24 +1,26 @@
...
-RUN microdnf install -y tar gzip which openssl diffutils && \
-    curl -Lo /usr/local/bin/kubectl "https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl" && \
+RUN microdnf install -y tar gzip which openssl diffutils jq && \
+    microdnf clean all && \
+    curl -fLo /usr/local/bin/kubectl "https://storage.googleapis.com/kubernetes-release/release/$(curl -fs https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/${TARGETARCH}/kubectl" && \
     chmod a+x /usr/local/bin/kubectl && \
-    curl -Lsv --fail https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/stable/openshift-client-linux.tar.gz | tar xvzf - -C /usr/local/bin/ oc && \
-    curl -Lo /usr/local/bin/jq https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64 && \
-    chmod a+x /usr/local/bin/jq
+    curl -fsSL "https://mirror.openshift.com/pub/openshift-v4/${TARGETARCH}/clients/ocp/stable/openshift-client-linux.tar.gz" | tar xzf - -C /usr/local/bin/ oc
```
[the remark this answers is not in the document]
carterpewpew: This was already verified by the reviewer (@nunnatsa) who tested the URL and specifically recommended using `${TARGETARCH}` directly in the path. The OpenShift mirror supports Docker-style arch names:

- `https://mirror.openshift.com/pub/openshift-v4/amd64/...` -> HTTP 200
- `https://mirror.openshift.com/pub/openshift-v4/arm64/...` -> HTTP 200
- `https://mirror.openshift.com/pub/openshift-v4/s390x/...` -> HTTP 200

No mapping needed. The curl command also uses `-L` to follow redirects as an extra safety net.

nunnatsa (approved):
/approve

Thanks @carterpewpew !!

PR #4341, opened 2026-06-23
Fix flaky functional tests

nunnatsa:

Several functional tests are randomly failing when trying to patch HCO. This is usually happening when the webhook is down, usually when the the leader election is lost.

This PR wraps the HCO Patch helper functions with Eventually of two minutes. This should be enough time for the webhook pod to be back online.

In addition, this PR introduces better versions of the enable/disable feature gate helpers, to handle all cases.

**Release note**:
```release-note
None
```

### DIFF
--- a/tests/func-tests/golden_image_test.go
+++ b/tests/func-tests/golden_image_test.go
@@ -24,7 +24,6 @@ import (
 	sspv1beta3 "kubevirt.io/ssp-operator/api/v1beta3"
 
 	hcov1 "github.com/kubevirt/hyperconverged-cluster-operator/api/v1"
-	"github.com/kubevirt/hyperconverged-cluster-operator/api/v1/featuregates"
 	goldenimages "github.com/kubevirt/hyperconverged-cluster-operator/controllers/handlers/golden-images"
 	tests "github.com/kubevirt/hyperconverged-cluster-operator/tests/func-tests"
 )
@@ -110,7 +109,7 @@ var _ = Describe("golden image test", Label("data-import-cron"), Serial, Ordered
 			Eventually(func(g Gomega, ctx context.Context) string {
 				is = getImageStream(ctx, cli, expectedIS.Name, imageNamespace)
 				return is.GetLabels()["app.kubernetes.io/part-of"]
-			}).WithTimeout(time.Second * 15).WithPolling(time.Millisecond * 100).WithContext(ctx).Should(Equal(expectedValue))
+			}).WithTimeout(time.Minute).WithPolling(time.Second).WithContext(ctx).Should(Equal(expectedValue))
 		},
 			isEntries,
 		)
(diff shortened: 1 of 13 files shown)


tests/func-tests/hyperconverged.go
```diff
@@ -130,25 +148,83 @@ func EnableFG(ctx context.Context, cli client.Client, fgName string) error {
...
+		patch = fmt.Appendf(nil, `[{ "op": "remove", "path": "/spec/featureGates/%d"}]`, idx)
+
+	default:
+		// alpha/deprecated FG are false by default. We need to explicitly enable them.
 		if idx == -1 {
 			patch = fmt.Appendf(nil, `[{"op": "add", "path": "/spec/featureGates/-", "value": {"name": %q}}]`, fgName)
+
+		} else {
+			patch = fmt.Appendf(nil, `[{"op": "add", "path": "/spec/featureGates/%d", "value": {"name": %q, "state": "%s"}}]`, idx, fgName, featuregates.Enabled)
+		}
```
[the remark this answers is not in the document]
nunnatsa: Assuming most of the usage of the featureGates will be to enable FGs, it is clearer to just have a list of FG names.

nunnatsa:
Change: rebase + fix a @coderabbitai comment

tests/func-tests/golden_image_test.go
```diff
@@ -915,14 +889,17 @@ var _ = Describe("golden image test", Label("data-import-cron"), Serial, Ordered
 
 func removeCustomDICTFromHC(ctx context.Context, cli client.Client) {
 	GinkgoHelper()
+	hc, err := tests.GetHCO(ctx, cli)
+	Expect(err).NotTo(HaveOccurred())
+	if hc.Spec.WorkloadSources.DataImportCronTemplates == nil {
+		return
+	}
```
orenc1, line 896: you can check if the field is not nil, and remove it if so, using a single jsonpatch call, using `test`, e.g.
```
[
  {
    "op": "test",
    "path": "/spec/workloadSources/dataImportCronTemplates",
    "value": null
  },
  {
    "op": "remove",
    "path": "/spec/workloadSources/dataImportCronTemplates"
  }
]
```

nunnatsa: This is a nice idea. Unfortunately, we need the opposite condition (`/spec/workloadSources/dataImportCronTemplates != nil`) and this is not supported by jsonpatch (no logical `NOT` operator, nor "not exists" value).

tests/func-tests/hyperconverged.go
```diff
@@ -130,25 +169,90 @@ func EnableFG(ctx context.Context, cli client.Client, fgName string) error {
...
+	}
+
+	idx := slices.IndexFunc(hc.Spec.FeatureGates, func(fg featuregates.FeatureGate) bool {
+		return fg.Name == fgName
+	})
+
+	var patch []byte
+	switch phase, _ := featuregatedetails.GetFeatureGatePhase(fgName); phase {
+	case fgs.PhaseUnknown:
+		return fmt.Errorf("unknow feature gate %q", fgName)
```
orenc1:
```suggestion
		return fmt.Errorf("unknown feature gate %q", fgName)
```

tests/func-tests/hyperconverged.go
```diff
@@ -97,31 +105,62 @@ func UpdateHCO(ctx context.Context, cli client.Client, input *hcov1.HyperConverg
...
+			}
+
+			b.WriteByte('\n')
+
+			return b.String()
+		})
+
+}
+
 // PatchHCO updates the HCO CR using a DynamicClient, it can return errors on failures
```
orenc1: after refactoring, this comment is no longer accurate, it doesn't return anything now, and assert internally via Gomega.
same comment for `PatchMergeHCO` in line 151

orenc1 (commented):
looks good, a couple of small comments below

orenc1 (approved):
/approve

PR #4346, opened 2026-06-23
 netresinjector: Make deployment conditional on spec field

nirdothan:

Currently there are 2 operators in the OLM catalog that deploy the `netwrok-resources-injector` webhook: `HCO` and `SR-IOV Network Operator`. While they are safe to use 
concurrently, users can benefit from disabling one instance.
In such cases the instance that should be disabled is the HCO one, because its scope is specific to CNV, while the other one has a more general scope. 

Adds spec.deployment.deployNetworkResourcesInjector boolean field to
control deployment of the network-resources-injector component.

When enabled (default: true), the network-resources-injector mutating
webhook is deployed to automatically inject resource requests for
custom resources annotated in NetworkAttachmentDefinition.

This is a deployment configuration field (not a feature gate),
allowing users to opt-out of the network-resources-injector if needed.

In addition
- Modify affinity to run injectors in control plain
- Change webhook failure policy to `Fail`

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://redhat.atlassian.net/browse/CNV-89798
```

**Release note**:

```release-note
DeploymentConfig in in HCO API now includes the deployNetworkResourcesInjector switch to support opting out of network-resources-injector deployment.
```

### DIFF
--- a/api/v1/hyperconverged_types.go
+++ b/api/v1/hyperconverged_types.go
@@ -164,7 +164,7 @@ type HyperConvergedSpec struct {
 	Security SecurityConfig `json:"security,omitempty"`
 
 	// Deployment contains all the configurations related to deployment of KubeVirt components
-	// +kubebuilder:default={"uninstallStrategy": "BlockUninstallIfWorkloadsExist", "deployVmConsoleProxy": false, "applicationAwareConfig": {"enable": false}}
+	// +kubebuilder:default={"uninstallStrategy": "BlockUninstallIfWorkloadsExist", "deployVmConsoleProxy": false, "deployNetworkResourcesInjector": true, "applicationAwareConfig": {"enable": false}}
 	// +optional
 	// +k8s:conversion-gen=false
 	Deployment DeploymentConfig `json:"deployment,omitempty"`
@@ -418,6 +418,14 @@ type DeploymentConfig struct {
 	// +kubebuilder:default=false
 	// +default=false
 	DeployVMConsoleProxy *bool `json:"deployVmConsoleProxy,omitempty"`
+
+	// DeployNetworkResourcesInjector enables deployment of the network-resources-injector component.
+	// When enabled, the network-resources-injector mutating webhook will be deployed to automatically
+	// inject resource requests for custom resources annotated in NetworkAttachmentDefinition.
+	// +optional
+	// +kubebuilder:default=true
+	// +default=true
+	DeployNetworkResourcesInjector *bool `json:"deployNetworkResourcesInjector,omitempty"`
 }
 
 // CertRotateConfigCA contains the tunables for TLS certificates.
@@ -925,7 +933,7 @@ type HyperConverged struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	// +kubebuilder:default={"security": {"certConfig": {"ca": {"duration": "48h0m0s", "renewBefore": "24h0m0s"}, "server": {"duration": "24h0m0s", "renewBefore": "12h0m0s"}}}, "virtualization": {"liveMigrationConfig": {"completionTimeoutPerGiB": 150, "parallelMigrationsPerCluster": 5, "parallelOutboundMigrationsPerNode": 2, "progressTimeout": 150, "allowAutoConverge": false, "allowPostCopy": false}, "virtualMachineOptions": {"disableFreePageReporting": false, "disableSerialConsoleLog": false}, "vmiCPUAllocationRatio": 10},"workloadSources":{"enableCommonBootImageImport":true}, "deployment": {"uninstallStrategy": "BlockUninstallIfWorkloadsExist", "deployVmConsoleProxy": false, "applicationAwareConfig": {"enable": false}}}
+	// +kubebuilder:default={"security": {"certConfig": {"ca": {"duration": "48h0m0s", "renewBefore": "24h0m0s"}, "server": {"duration": "24h0m0s", "renewBefore": "12h0m0s"}}}, "virtualization": {"liveMigrationConfig": {"completionTimeoutPerGiB": 150, "parallelMigrationsPerCluster": 5, "parallelOutboundMigrationsPerNode": 2, "progressTimeout": 150, "allowAutoConverge": false, "allowPostCopy": false}, "virtualMachineOptions": {"disableFreePageReporting": false, "disableSerialConsoleLog": false}, "vmiCPUAllocationRatio": 10},"workloadSources":{"enableCommonBootImageImport":true}, "deployment": {"uninstallStrategy": "BlockUninstallIfWorkloadsExist", "deployVmConsoleProxy": false, "deployNetworkResourcesInjector": true, "applicationAwareConfig": {"enable": false}}}
 	// +optional
 	Spec   HyperConvergedSpec   `json:"spec,omitempty"`
 	Status HyperConvergedStatus `json:"status,omitempty"`
--- a/controllers/handlers/netresinjector/conditional_test.go
+++ b/controllers/handlers/netresinjector/conditional_test.go
@@ -1,0 +1,249 @@
+package netresinjector
+
+import (
+	"context"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	policyv1 "k8s.io/api/policy/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	hcov1 "github.com/kubevirt/hyperconverged-cluster-operator/api/v1"
+	"github.com/kubevirt/hyperconverged-cluster-operator/controllers/common"
+	"github.com/kubevirt/hyperconverged-cluster-operator/controllers/commontestutils"
+)
+
+var _ = Describe("Network Resources Injector Conditional Handlers", func() {
+	var (
+		hco *hcov1.HyperConverged
+		req *common.HcoRequest
+		cl  client.Client
+	)
+
+	BeforeEach(func() {
+		hco = commontestutils.NewHco()
+		req = commontestutils.NewReq(hco)
+	})
+
+	Context("Conditional ClusterRole Handler", func() {
+		It("should create ClusterRole when deployNetworkResourcesInjector is true", func() {
+			hco.Spec.Deployment.DeployNetworkResourcesInjector = ptr.To(true)
+			cl = commontestutils.InitClient([]client.Object{hco})
+
+			handler := NewClusterRoleHandler(cl, commontestutils.GetScheme())
+			res := handler.Ensure(req)
+
+			Expect(res.Err).ToNot(HaveOccurred())
+			Expect(res.Created).To(BeTrue())
+
+			foundCR := &rbacv1.ClusterRole{}
+			Expect(cl.Get(context.Background(), client.ObjectKey{Name: clusterRoleName}, foundCR)).To(Succeed())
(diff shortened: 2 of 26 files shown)


api/v1/hyperconverged_types.go
```diff
@@ -329,6 +329,14 @@ type NetworkingConfig struct {
 	// Those bindings can be used when defining virtual machine interfaces.
 	// +optional
 	NetworkBinding map[string]v1.InterfaceBindingPlugin `json:"networkBinding,omitempty"`
+
+	// DeployNetworkResourcesInjector enables deployment of the network-resources-injector component.
+	// When enabled, the network-resources-injector mutating webhook will be deployed to automatically
+	// inject resource requests for custom resources annotated in NetworkAttachmentDefinition.
+	// +optional
+	// +kubebuilder:default=true
+	// +default=true
```
nunnatsa: Few comments:
* Not sure about the right place for this new field. The `deployment` object may be a good choice as well. This field is not optional.
* The `networking` object is a pointer and may be empty. Setting the default to `true` will only take affect if this field is populated. If this is the desired behavior, that fine. If not, you'll need to add this default to the default of the `networking` field (see the `virtualization` field for reference) **and** to the `spec` field. This is relevant also if you choose to move the new field to the `deployment` field.

controllers/handlers/netresinjector/deployment.go
```diff
@@ -45,8 +45,55 @@ func newDeployment(_ *hcov1.HyperConverged) *appsv1.Deployment {
 
 	podLabels := operands.GetLabels(hcoutil.AppComponentNetResInjector)
 
-	infrastructureHighlyAvailable := nodeinfo.IsInfrastructureHighlyAvailable()
-	affinity := operands.GetPodAntiAffinity(podLabels[hcoutil.AppLabelComponent], infrastructureHighlyAvailable)
+	// Configure affinity to run on control plane nodes only
+	affinity := &corev1.Affinity{
+		NodeAffinity: &corev1.NodeAffinity{
```
nunnatsa, line 57: Will it work in a cluster with worker node only (like in HCP)? Will it work for single node clusters (where the single node is both a worker and a control plane)?

controllers/handlers/netresinjector/conditional.go
```diff
@@ -0,0 +1,91 @@
+package netresinjector
```
nunnatsa: Personally, the concept used in this file feels like a bit over engineering and confusing. I would just modify the existing handlers to be conditional. For example, in common.go, you can add something like:
```go
func shouldDeploy(hc *hcov1.HyperConverged) bool {
		return hc.Spec.Networking == nil || ptr.Deref(hc.Spec.Networking.DeployNetworkResourcesInjector, true)
}
```

and then in deployment.go, modify NewDeploymentHandler() to something like
```go
func NewDeploymentHandler(cli client.Client, scheme *runtime.Scheme) operands.Operand {
	return operands.NewConditionalHandler(
		operands.NewDeploymentHandler(cli, scheme, newDeployment),
		shouldDeploy,
		func(hc *hcov1.HyperConverged) client.Object {
			return NewDeploymentWithNameOnly()
		},
	)
}
```

nunnatsa (commented):
Thanks for the PR @nirdothan 

I know that its work in progress, but I already have several comments that I hope you can use.

General question: I understand the networking resources injector is now optional (but is enabled by default). Is that the long term plan? will it always be optional? if not, let's use a feature gate instead an enabler. If yes, that fine, but please notice the comments about the default value.

controllers/handlers/netresinjector/deployment.go
nunnatsa: If needed, it can be set dynamically with the help of the `nodeinfo` package.

nirdothan: It's a critical component that must always exist. It has no relation to worker nodes and its entire work is done against the API Server. I tried to follow virt-operator affinity. Please recommend an optimal approach that assures it will always be deployed.

nirdothan:
> Thanks for the PR @nirdothan
> 
> I know that its work in progress, but I already have several comments that I hope you can use.
> 
> General question: I understand the networking resources injector is now optional (but is enabled by default). Is that the long term plan? will it always be optional? if not, let's use a feature gate instead an enabler. If yes, that fine, but please notice the comments about the default value.

Thanks for reviewing. It's not exactly optional but rather mandatory and also critical, but it can be toggled off here if another instance is running under SR-IOV operator. 
A FG would not be a good fit because it is permanent and not version dependent.

nunnatsa:
> > Thanks for the PR @nirdothan
> > I know that its work in progress, but I already have several comments that I hope you can use.
> > General question: I understand the networking resources injector is now optional (but is enabled by default). Is that the long term plan? will it always be optional? if not, let's use a feature gate instead an enabler. If yes, that fine, but please notice the comments about the default value.
> 
> Thanks for reviewing. It's not exactly optional but rather mandatory and also critical, but it can be toggled off here if another instance is running under SR-IOV operator. A FG would not be a good fit because it is permanent and not version dependent.

OK, great. Just an idea (and only if it's not too complex): do you want HCO to detect the situation of another instance(s) and only deploy netresinjector if not found? in general, if we can do something for the user and spare it from them, it's preferred over adding a new API field.

controllers/handlers/netresinjector/deployment.go
nunnatsa: I think it should work on a cluster without CP. We solved it for the virt pods.

I think it should also work on an SNO node, but we can check it by adding a new functional test to check deployment, and make sure the pod is running. If it will fail in the SNO test lanes, then we'll need to fix it.

api/v1/hyperconverged_types.go
nirdothan: Moved DeployNetworkResourcesInjector from NetworkingConfig to DeploymentConfig.

controllers/handlers/netresinjector/conditional.go
nirdothan: Done.

orelmisan (commented):
Thanks @nirdothan.
Can you please share the motivation behind this change in the PR description?

controllers/handlers/netresinjector/deployment.go
nirdothan: Done (attempted).

controllers/handlers/netresinjector/deployment.go
```diff
@@ -45,8 +45,55 @@ func newDeployment(_ *hcov1.HyperConverged) *appsv1.Deployment {
...
+					{
+						MatchExpressions: []corev1.NodeSelectorRequirement{
+							{
+								Key:      nodeinfo.LabelNodeRoleControlPlane,
+								Operator: corev1.NodeSelectorOpExists,
+							},
+						},
+					},
+				},
+			},
```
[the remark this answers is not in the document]
nirdothan: > However, in HyperShift clusters, the codebase relabels worker nodes with node-role.kubevirt.io/control-plane 

Is this a typo? does it relabel worker nodes or control-plane nodes?

nirdothan:
[Addressed](https://github.com/kubevirt/hyperconverged-cluster-operator/compare/4597b7d01b8b2b4ff6662bd62b7dbc6ebb84ebd8..ba0dc3fec1467b53e0eb273cdd8cb0478591f357) @nunnatsa's https://github.com/kubevirt/hyperconverged-cluster-operator/pull/4346#pullrequestreview-4559402112.

nirdothan:
[Rebased](https://github.com/kubevirt/hyperconverged-cluster-operator/compare/ba0dc3fec1467b53e0eb273cdd8cb0478591f357..c0e4aeb20adc6e7b255efed6e2ecc7dffe7c3b73) and updated commit message (1st commit).

nirdothan:
> Thanks @nirdothan. Can you please share the motivation behind this change in the PR description?

Done.

nirdothan:
> > > Thanks for the PR @nirdothan
> > > I know that its work in progress, but I already have several comments that I hope you can use.
> > > General question: I understand the networking resources injector is now optional (but is enabled by default). Is that the long term plan? will it always be optional? if not, let's use a feature gate instead an enabler. If yes, that fine, but please notice the comments about the default value.
> > 
> > 
> > Thanks for reviewing. It's not exactly optional but rather mandatory and also critical, but it can be toggled off here if another instance is running under SR-IOV operator. A FG would not be a good fit because it is permanent and not version dependent.
> 
> OK, great. Just an idea (and only if it's not too complex): do you want HCO to detect the situation of another instance(s) and only deploy netresinjector if not found? in general, if we can do something for the user and spare it from them, it's preferred over adding a new API field.

We are planning to do that with `virt-platform-autopilot` mainly because we wouldn't want HCO to have any knowledge about non CNV components and introduce "hidden complexity" that would not be intuitive to users. In addition, the option to manually disable the webhook can also be beneficial in various use cases.

controllers/handlers/netresinjector/conditional.go
nunnatsa: Thanks for fixing it @nirdothan 

I would remove this file and change the original handler constructors, in their original location. Also, I think we can drop "Conditional" from their names, as been conditional is just an internal implementation. 

I can look at it like the user code at controllers/operandhandler/operandHandler.go should not know about this change.

Anyway, currently we have two constructors for the handlers and the original ones are not in use.

nunnatsa:
Thanks for the fixes @nirdothan 

We have an issue with the conversion webhook. (this is why the tests are failing) This is the first field we only adding to v1, as planned, so we face a new general issue, not related to this PR.

If someone modified the HyperConverged CR (update/patch/edit) using v1beta1, it will be converted to v1 and stored in ETCD. But since the v1-only field is not known to the conversion webhook, it is set to the zero value (nil in this case), and then get the default value (true in this case). So if the user set the field to false, and then the CR was modified again using v1beta1 API, we are loosing the change, and it will become true again.

To solve it we'll use the unknown field annotation pattern:
* in v1 => v1beta1 conversion, we'll add an annotation with the v1 only fields
* in v1beta1 => v1 conversion, we'll restore the v1 only fields from the annotation

Please consider the attached diff file: 
[suggested-fix.txt](https://github.com/user-attachments/files/29497247/suggested-fix.txt)

controllers/handlers/netresinjector/conditional_test.go
```diff
@@ -0,0 +1,248 @@
...
+			handler := NewClusterRoleHandler(cl, commontestutils.GetScheme())
+			res := handler.Ensure(req)
+
+			Expect(res.Err).ToNot(HaveOccurred())
+			Expect(res.Deleted).To(BeTrue())
+
+			foundCR := &rbacv1.ClusterRole{}
+			err := cl.Get(context.Background(), client.ObjectKey{Name: clusterRoleName}, foundCR)
+			Expect(err).To(HaveOccurred())
+			Expect(client.IgnoreNotFound(err)).ToNot(HaveOccurred())
```
nunnatsa: nit: can be simplified like this:



```suggestion
			Expect(err).To(MatchError(apierrors.IsNotFound, "not found error"))
```

controllers/handlers/netresinjector/deployment.go
nunnatsa: @nirdothan - I don't see any change in the functional tests in this PR. Do I miss something?

nunnatsa (commented):
Thanks @nirdothan 

The PR looks good to me.
I still have two general comments (and one nit inline)

I think we do need a minimal functional test to make sure everything is deployed. Mostly I concern about the node affinity on SNO clusters.

Also, please update the docs/cluster-configuration.md user-guide with the new field.

controllers/handlers/netresinjector/conditional_test.go
nirdothan: Done.

controllers/handlers/netresinjector/deployment.go
nirdothan: There were no functional tests. Added.

nunnatsa (approved):

PR #4376, opened 2026-07-05
Fine tuning the MDev's enabled field and FG

nunnatsa:
Fixing several bug and behaviors, like
* never remove the enabled field if already set.
* when dropping the FG, if it's the only FG in the list, remove the list (to restore the default value).
* differentiate between disabled FG and not set FG

Another changes:
* Use the v1-only-field mechanism to store and restore the `enabled` filed and the `disableMDevConfiguration` FG.
* Move he v1 MDev validation logic to the mutating webhook, to achieve more accurate code, and to simplify the implementation.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
None
```

**Release note**:

```release-note
None
```


nunnatsa:
Changes: fixed @coderabbitai comments

pkg/webhooks/mutator/hyperConvergedMutator.go
```diff
@@ -78,26 +82,57 @@ func (hcm *HyperConvergedMutator) mutateHyperConverged(req admission.Request, lo
...
 			oldHC = &hcov1.HyperConverged{}
-			if err := hcm.decoder.DecodeRaw(req.OldObject, oldHC); err != nil {
+			if err = hcm.decoder.DecodeRaw(req.OldObject, oldHC); err != nil {
 				logger.Error(err, "failed to read the old HyperConverged custom resource")
 				return admission.Errored(http.StatusBadRequest, fmt.Errorf("failed to parse the old HyperConverged"))
 			}
 		}
-		patches = hcMutateV1MDevFGAndEnabledOnUpdate(hc, oldHC, patches)
+
+		allowed, triggerWarning, newPatches := hcMutateV1MDevFGAndEnabledOnUpdate(hc, oldHC, patches)
```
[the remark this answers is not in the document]
nunnatsa: @coderabbitai, this is done in line 105 - 107

nunnatsa:
Change: reduced complexity in hcMutateV1MDevFGAndEnabledOnUpdate()

pkg/webhooks/mutator/hyperConvergedMutator.go
```diff
@@ -185,194 +217,125 @@ func mutateTuningPolicy(hc *hcov1.HyperConverged, patches []jsonpatch.JsonPatchO
...
 	}
 
-	return hcPatchV1DisableMDevFG(fgs, false, patches)
+	return true, []string{mdevWarning}, mutateMdevEnabled(mdc, !fgEnabled, patches)
 }
 
-func hcMutateV1HandleFGOnlyChanged(
-	hc *hcov1.HyperConverged,
-	newFGEnabled bool,
+func mutateMdevEnabled(
```
Ronilerr, line 272: `mutateMdevEnabled` always uses JSON Patch `add` on `.../enabled`.
On update, if the user already sent `enabled` in the request and the FG-change sync tries to flip it (e.g. `enabled: true` + newly added `disableMDevConfiguration: Enabled`), `add` fails because the path already exists.
Consider `replace` when `enabled` is already set (`newEnabledFound`) 
Also consider to add test for it

Ronilerr:
Hey @nunnatsa, Thanks for the PR! 
Added one comment.

pkg/webhooks/mutator/hyperConvergedMutator.go
nunnatsa: Thanks for the comment @Ronilerr 

`add` will not fail if the field is already populated. See the [jsonpatch spec](https://datatracker.ietf.org/doc/html/rfc6902/#section-4.1). 

If the field already exists, it behaves like `replace`.

Ronilerr: Thanks, sorry for the mistake.

nunnatsa:
/approve

the s390x cluster is down, while the unit tests passed in two github actions.
/override pull-hyperconverged-cluster-operator-unit-test-s390x

PR #4377, opened 2026-07-05
netresinjector: Temporarily disable by default due to bug

nirdothan:
Changes the network-resources-injector deployment default from enabled to disabled to work around [1].

This is a temporary measure to prevent breaking environments until the underlying bug is fixed. Users can still explicitly enable the feature by setting spec.deployment.deployNetworkResourcesInjector to true.

[1] https://redhat.atlassian.net/browse/OCPBUGS-96963

Assisted by: Claude Sonnet 4.5 <noreply@anthropic.com>
**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket

```

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tests/func-tests/netresinjector_test.go
+++ b/tests/func-tests/netresinjector_test.go
@@ -2,6 +2,7 @@ package tests_test
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
@@ -15,7 +16,8 @@ import (
 )
 
 const (
-	netResInjectorDeploymentName = "virt-network-resources-injector"
+	netResInjectorDeploymentName          = "virt-network-resources-injector"
+	netResInjectorDeploymentPatchTemplate = `{"spec":{"deployment":{"deployNetworkResourcesInjector": %t}}}`
 )
 
 var _ = Describe("Test Network Resources Injector", Label("NetResInjector"), Serial, Ordered, func() {
@@ -31,7 +33,21 @@ var _ = Describe("Test Network Resources Injector", Label("NetResInjector"), Ser
 		restoreNetResInjectorToDefault(ctx, cli)
 	})
 
-	Context("when deployNetworkResourcesInjector is true (default)", func() {
+	Context("when deployNetworkResourcesInjector is not set (default false)", func() {
+		BeforeAll(func(ctx context.Context) {
+			restoreNetResInjectorToDefault(ctx, cli)
+		})
+
+		It("should not deploy the network resources injector", func(ctx context.Context) {
+			validateNetResInjectorDeleted(ctx, cli)
+		})
+	})
+
+	Context("when deployNetworkResourcesInjector is true", func() {
+		BeforeAll(func(ctx context.Context) {
+			enableNetResInjector(ctx, cli)
+		})
+
 		It("should deploy the network resources injector", func(ctx context.Context) {
 			By("verifying the deployment exists and is ready")
 			Eventually(func(g Gomega, ctx context.Context) {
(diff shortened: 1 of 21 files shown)


nunnatsa (commented):
I would remove the default value from the crd for now, relaying on the default in shouldDeploy(). I think it would be easier to upgrade, specially if we'll not get the fixed image on time. Because once the field is set in the CR, it won't bd updated from the crd default.

Once we'll have the fix, we could decide iv we want to explicitly set the default value in the CR.

nirdothan:
> I would remove the default value from the crd for now, relaying on the default in shouldDeploy(). I think it would be easier to upgrade, specially if we'll not get the fixed image on time. Because once the field is set in the CR, it won't bd updated from the crd default.
> 
> Once we'll have the fix, we could decide iv we want to explicitly set the default value in the CR.

[Done]( https://github.com/kubevirt/hyperconverged-cluster-operator/compare/85cc62706a6011f998d1cdde9e712e38cdf48308..e009fb1328ed67f888396e09d047b0f6001c90d1)

nunnatsa (approved):
/approve

tests/func-tests/netresinjector_test.go
nunnatsa: Now, that the `deployNetworkResourcesInjector` is optional, this function is failing if the field does not exist. 

Please use merge patch instead of json. 

For example:

```suggestion
patch := []byte(`{"spec":{"deployment":{"deployNetworkResourcesInjector": true}}}`)
tests.PatchMergeHCO(ctx, cli, patch)
```

you can define a constant like `{"spec":{"deployment":{"deployNetworkResourcesInjector": %t}}}` and use for both enable or disable.

nunnatsa: same

tests/func-tests/netresinjector_test.go
```diff
@@ -139,16 +138,4 @@ func restoreNetResInjectorToDefault(ctx context.Context, cli client.Client) {
 	By("restoring deployNetworkResourcesInjector to default")
 	patch := []byte(`[{"op": "remove", "path": "/spec/deployment/deployNetworkResourcesInjector"}]`)
```
nunnatsa: This will fail if the filed is missing. Please read the HyperConverged CR first, to make sure the field is there, before pathing.

nunnatsa (commented):
The functional test is failing. Please fix. See specific comments inline.

tests/func-tests/netresinjector_test.go
nunnatsa, line 61: This will fail because of the the image issue. Please remove or comment (with TODO, to uncomment after the image fix).

nirdothan: The image issue is not supposed to affect its readiness, if we're speaking about[1].
Leaving it as is for now, unless we'll see that it is failing.

[1] https://redhat.atlassian.net/browse/OCPBUGS-96963

tests/func-tests/netresinjector_test.go
nirdothan: Done.

tests/func-tests/netresinjector_test.go
nirdothan: Done.

nirdothan: Done.

nirdothan (commented):
[Changes](https://github.com/kubevirt/hyperconverged-cluster-operator/compare/e009fb1328ed67f888396e09d047b0f6001c90d1..0c7bc7dc573797062a3ad70dc6dd571fc3a712c9) addressing last review.

nunnatsa (approved):

nirdothan:
[Changes](https://github.com/kubevirt/hyperconverged-cluster-operator/compare/3a93897da997aadd060ad4677a896d669455dfc7..571a3ffc8dde66f20d06b687049a1631c9613b29) to support inter HCO version upgrade test.

PR #4386, opened 2026-07-08
Fix a bug in NetResInjector: deployment's replicas

nunnatsa:

The number of the virt-network-resources-injector deployment is constantly 2. This is a problem in a non-highly available or single-node clusters.

This PR set the replica to 1 on a non-highly available clusters.

**Release note**:

```release-note
None
```

### DIFF
--- a/controllers/handlers/netresinjector/deployment.go
+++ b/controllers/handlers/netresinjector/deployment.go
@@ -44,6 +44,13 @@ func NewDeploymentWithNameOnly() *appsv1.Deployment {
 func newDeployment(_ *hcov1.HyperConverged) *appsv1.Deployment {
 	image := os.Getenv(hcoutil.NetworkResourcesInjectorImageEnvV)
 
+	var replicas int32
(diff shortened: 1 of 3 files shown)


controllers/handlers/netresinjector/deployment.go
```diff
@@ -43,6 +43,14 @@ func NewDeploymentWithNameOnly() *appsv1.Deployment {
 
 func newDeployment(_ *hcov1.HyperConverged) *appsv1.Deployment {
 	image := os.Getenv(hcoutil.NetworkResourcesInjectorImageEnvV)
+	infrastructureHighlyAvailable := nodeinfo.IsInfrastructureHighlyAvailable()
```
nirdothan: nit: Consider inlining the function, if this variable is only used once.

nirdothan (approved):
Thanks. Looks right.

nunnatsa:
Thanks @nirdothan - fixed your comment

nirdothan (approved):

PR #4387, opened 2026-07-08
netresinjector: Add Certificate handler for non-OpenShift deployments

nirdothan:

In non-OpenShift environments, the TLS secret for network-resources-injector needs to be created by cert-manager. Previously, this relied on a static Certificate resource in deploy/webhooks.yaml, which is not applied in OLM deployments (like nightly-bundle).

Add a minimal Certificate handler that:
- Only deploys when network-resources-injector is enabled
- Only in non-OpenShift clusters (OpenShift uses service-ca)
- Creates a cert-manager Certificate resource dynamically
- Tells cert-manager to generate virt-network-resources-injector-secret

This fixes the "secret not found" error in non-OpenShift OLM deployments.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket

```

**Release note**:

```release-note
NONE
```

### DIFF
--- a/controllers/handlers/certmanager_issuer.go
+++ b/controllers/handlers/certmanager_issuer.go
@@ -1,0 +1,78 @@
+package handlers
+
+import (
+	"reflect"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	hcov1 "github.com/kubevirt/hyperconverged-cluster-operator/api/v1"
+	"github.com/kubevirt/hyperconverged-cluster-operator/controllers/common"
+	"github.com/kubevirt/hyperconverged-cluster-operator/controllers/operands"
+	hcoutil "github.com/kubevirt/hyperconverged-cluster-operator/pkg/util"
+)
+
+const (
+	certManagerAPIVersion = "cert-manager.io/v1"
+	issuerName            = "selfsigned"
+	issuerKind            = "Issuer"
+)
+
+func NewCertManagerIssuerHandler(cli client.Client, scheme *runtime.Scheme) operands.Operand {
+	return operands.NewGenericOperand(cli, scheme, issuerKind, &issuerHooks{}, false)
+}
+
+type issuerHooks struct{}
+
+func (h *issuerHooks) GetFullCr(_ *hcov1.HyperConverged) (client.Object, error) {
+	obj := &unstructured.Unstructured{
+		Object: map[string]any{
+			"apiVersion": certManagerAPIVersion,
+			"kind":       issuerKind,
+			"metadata": map[string]any{
+				"name":      issuerName,
+				"namespace": hcoutil.GetOperatorNamespaceFromEnv(),
+			},
+			"spec": map[string]any{
+				"selfSigned": map[string]any{},
+			},
+		},
+	}
+	_ = unstructured.SetNestedStringMap(obj.Object, operands.GetLabels(hcoutil.AppComponentCompute), "metadata", "labels")
+	return obj, nil
+}
+
+func (h *issuerHooks) GetEmptyCr() client.Object {
+	return &unstructured.Unstructured{
+		Object: map[string]any{
+			"apiVersion": certManagerAPIVersion,
+			"kind":       issuerKind,
+		},
+	}
+}
+
+func (h *issuerHooks) UpdateCR(req *common.HcoRequest, Client client.Client, exists, desired runtime.Object) (bool, bool, error) {
+	existingIssuer, ok := exists.(*unstructured.Unstructured)
+	if !ok {
+		return false, false, nil
+	}
+
+	desiredIssuer, ok := desired.(*unstructured.Unstructured)
+	if !ok {
+		return false, false, nil
+	}
+
+	existingLabels, _, _ := unstructured.NestedStringMap(existingIssuer.Object, "metadata", "labels")
+	desiredLabels, _, _ := unstructured.NestedStringMap(desiredIssuer.Object, "metadata", "labels")
+
+	if !reflect.DeepEqual(existingLabels, desiredLabels) {
+		_ = unstructured.SetNestedStringMap(existingIssuer.Object, desiredLabels, "metadata", "labels")
+		if err := Client.Update(req.Ctx, existingIssuer); err != nil {
+			return false, false, err
+		}
+		return true, !req.HCOTriggered, nil
+	}
+
+	return false, false, nil
+}
--- a/controllers/handlers/netresinjector/certificate.go
+++ b/controllers/handlers/netresinjector/certificate.go
@@ -1,0 +1,115 @@
+package netresinjector
+
+import (
+	"reflect"
+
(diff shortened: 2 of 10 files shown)


controllers/handlers/netresinjector/certificate.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/kubevirt/hyperconverged-cluster-operator/controllers/common"
+	"github.com/kubevirt/hyperconverged-cluster-operator/controllers/operands"
+	hcoutil "github.com/kubevirt/hyperconverged-cluster-operator/pkg/util"
+)
+
+func NewCertificateHandler(cli client.Client, scheme *runtime.Scheme) operands.Operand {
+	return operands.NewConditionalHandler(
+		operands.NewGenericOperand(cli, scheme, "Certificate", &certHooks{}, false),
+		func(hc *hcov1.HyperConverged) bool {
+			return shouldDeploy(hc) && !hcoutil.GetClusterInfo().IsOpenshift()
```
nunnatsa: There is no point to recheck for openshift in each reconcile loop, while the answer will not changed during runtime. Instead, just don't create the handler in openshift.

nirdothan: Done.

nirdothan:
[Changes](https://github.com/kubevirt/hyperconverged-cluster-operator/compare/eb95efb8c35fffd37428f62bf4844d20d0510471..ef2c070518db391cf058fef82f56c24a1b5eacfb) addressed https://github.com/kubevirt/hyperconverged-cluster-operator/pull/4387#pullrequestreview-4657645542, fixed cert-manager RBAC, added cert unit test.

nirdothan:
[Last push](https://github.com/kubevirt/hyperconverged-cluster-operator/compare/ef2c070518db391cf058fef82f56c24a1b5eacfb..f4a780504205da35cecfdecd83771ffdd5cca305) adds a cert-manager Issuer to the the controller, to overcome failing functional tests. 
It is possible that the issuer in `deploy/webhooks.yaml` is now redundant.

controllers/handlers/certmanager_issuer.go
```diff
@@ -0,0 +1,93 @@
...
+const (
+	certManagerAPIVersion = "cert-manager.io/v1"
+	issuerName            = "selfsigned"
+	issuerKind            = "Issuer"
+)
+
+func NewCertManagerIssuerHandler(cli client.Client, scheme *runtime.Scheme) operands.Operand {
+	return operands.NewConditionalHandler(
+		operands.NewGenericOperand(cli, scheme, issuerKind, &issuerHooks{}, false),
+		func(_ *hcov1.HyperConverged) bool { return true },
```
nunnatsa: If the condition is always true, then we don't need it to be a conditional handler. Please simplify by making it a regular GenericHandler.

controllers/handlers/netresinjector/certificate.go
```diff
@@ -0,0 +1,130 @@
...
+		return false, false, nil
+	}
+	desiredCert, ok := desired.(*unstructured.Unstructured)
+	if !ok {
+		return false, false, nil
+	}
+
+	needsUpdate := false
+
+	existingLabels, _, _ := unstructured.NestedStringMap(existingCert.Object, "metadata", "labels")
```
nunnatsa, line 75: Not sure about the unstructured solution. Is it because you want to avoid adding a new meaningful dependency? That fair. it's just not the way we usually do things here.

Let's keep it like this for now and see if it's working for us.

controllers/handlers/netresinjector/certificate_test.go
```diff
@@ -0,0 +1,288 @@
...
+		)
+
+		BeforeEach(func() {
+			ns := "test-namespace"
+			hco := commontestutils.NewHco()
+			req = commontestutils.NewReq(hco)
+			cl = &updateSpy{}
+
+			existing = &unstructured.Unstructured{
+				Object: map[string]interface{}{
```
nunnatsa: Please - for the new code, change `interface{}` to `any`.

controllers/handlers/netresinjector/certificate.go
```diff
@@ -0,0 +1,130 @@
...
+
+const (
+	certManagerAPIVersion = "cert-manager.io/v1"
+	certificateKind       = "Certificate"
+)
+
+func NewCertManagerCertHandler(cli client.Client, scheme *runtime.Scheme) operands.Operand {
+	return operands.NewConditionalHandler(
+		operands.NewGenericOperand(cli, scheme, certificateKind, &certHooks{}, false),
+		func(_ *hcov1.HyperConverged) bool { return true },
```
nunnatsa: If this is always true, then please simplify and make it a regular GenericHandler.

controllers/operandhandler/operandHandler.go
```diff
@@ -122,6 +122,13 @@ func NewOperandHandler(client client.Client, scheme *runtime.Scheme, ci hcoutil.
 		operandList = append(operandList, handlers.NewCsvHandler(client))
 	}
 
+	if !ci.IsOpenshift() {
+		operandList = append(operandList,
+			handlers.NewCertManagerIssuerHandler(client, scheme),
+			netresinjector.NewCertManagerCertHandler(client, scheme),
+		)
+	}
```
nunnatsa: tiny nit: can be moved to line 119, as `else` for the condition in line 77.

Feel free to ignore this comment, if you think, for example, that it improves the readability (as the line 77 condition is pretty long).

nunnatsa (commented):
Thanks @nirdothan 

Added a few comments inline. Mostly about making the new two handlers non-conditional. Also, please avoid `interface{}` and prefer `any` instead.

controllers/handlers/netresinjector/certificate.go
nirdothan: Yes because I think it's not worth adding the dependency for this "rare" scenario.

controllers/handlers/netresinjector/certificate_test.go
nirdothan: Cool. Done.

controllers/handlers/certmanager_issuer.go
nirdothan: Done.

controllers/operandhandler/operandHandler.go
nirdothan: Done.

controllers/handlers/netresinjector/certificate.go
nirdothan: Done.

nirdothan:
[Changes](https://github.com/kubevirt/hyperconverged-cluster-operator/compare/387ba29d114615bcd5b93e3a1f4e49f295a766c0..84cc1bc91545d95f3abd265d087ed3e091d8cb11) addressing https://github.com/kubevirt/hyperconverged-cluster-operator/pull/4387#pullrequestreview-4682157609.

nunnatsa (approved):
Thanks for this fix @nirdothan!

PR #4389, opened 2026-07-09
Deprecate persistent reservation feature gate in follow up to upstream GA/toggle

akalenyu:
Follow up upstream kubevirt's GAing plan in https://github.com/kubevirt/kubevirt/pull/16674 which made it to 1.9,
Which graduates the gate alongside a toggle to enable this niche feature for interested users.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://redhat.atlassian.net/browse/CNV-88999
```

**Release note**:

```release-note
Deprecate reservation feature gate in follow up to upstream GA/toggle
```


nunnatsa:
Thanks @akalenyu .

If you can wait for #4384, it would be great (or if you want to base your PR on it, but it wasn't reviewed yet).

nunnatsa:
@akalenyu 

#4384 was merged and should change your logic a bit: 
* you don't need the `v1OnlyFields.PersistentReservationFG` filed anymore.
* Please update the new `v1OnlyFields.isEmpty()` method, to check the `PersistentReservationEnabled` field.

Please also notice #4390 (EDITED: merged)

api/v1/hyperconverged_types.go
```diff
@@ -310,6 +307,22 @@ type StorageConfig struct {
...
+}
+
+// PersistentReservationConfiguration holds the configuration for SCSI persistent reservation support
+// +k8s:openapi-gen=true
+type PersistentReservationConfiguration struct {
+	// Enabled controls the deployment of additional resources like the pr-helper container
+	// for enabling the use of the SCSI persistent reservation in VMs, defaults to false.
+	// +optional
+	// +k8s:conversion-gen=false
+	Enabled *bool `json:"enabled,omitempty"`
```
nunnatsa, line 333: Do we expect additional configurations here? Do you think it should be a nested field?

api/v1/hyperconverged_types.go
```diff
@@ -310,6 +307,22 @@ type StorageConfig struct {
 	// resource
 	// +optional
 	WorkloadResourceRequirements *corev1.ResourceRequirements `json:"workloadResourceRequirements,omitempty"`
+
+	// PersistentReservationConfiguration controls the deployment of additional resources
+	// required for using SCSI persistent reservation in VMs
+	// +optional
+	// +k8s:conversion-gen=false
+	PersistentReservationConfiguration *PersistentReservationConfiguration `json:"persistentReservationConfiguration,omitempty"`
```
nunnatsa, line 323: can we somehow make the name of the field shorter? Even drop the "Configuration" word from it?

api/v1beta1/conversion.go
```diff
@@ -55,7 +58,9 @@ func (src *HyperConverged) ConvertTo(dstRaw conversion.Hub) error { //revive:dis
 
 	convertMDevEnabledV1beta1ToV1(src.Spec, &dst.Spec)
 
-	dst.Spec.Storage = convertStorageV1beta1ToV1(src.Spec)
+	convertStorageV1beta1ToV1(src.Spec, &dst.Spec)
+
+	convertPersistentReservationV1beta1ToV1(src.Spec, &dst.Spec)
```
nunnatsa, line 63: maybe it belong inside the `convertStorageV1beta1ToV1()` function?

api/v1beta1/conversion.go
```diff
@@ -89,6 +94,8 @@ func (dst *HyperConverged) ConvertFrom(srcRaw conversion.Hub) error { //revive:d
 
 	convertStorageV1ToV1beta1(src.Spec.Storage, &dst.Spec)
 
+	convertPersistentReservationV1ToV1beta1(src.Spec.Storage, &dst.Spec)
```
nunnatsa, line 97: Maybe it can be inside the `convertStorageV1ToV1beta1()` function?

docs/cluster-configuration-v1beta1-api.md
```diff
@@ -154,29 +154,14 @@ For additional information, see here: [KubeSecondaryDNS](https://github.com/kube
...
-      disks:
-      - name: mypvcdisk
-        lun:
-          reservations: true
-```
-**Note**: An important aspect of this feature is that the SCSI persistent reservation doesn't support migration. Even if you apply the reservation to an RWX PVC provisioning SCSI devices, the restriction is due to the reservation done by the initiator on the node. The VM could be migrated but not the reservation.
-
-**Note**: this feature is in Developer Preview.
+The `persistentReservation` feature gate is deprecated. Use
+`spec.storage.persistentReservationConfiguration.enabled` in the v1 API instead.
```
nunnatsa: The FG is still active, so I think it should be still documented, with a clear deprecation notice (as already done).

api/v1beta1/hyperconverged_types.go
```diff
@@ -420,13 +420,10 @@ type HyperConvergedFeatureGates struct {
 	// +hco:fgphase:deprecated
 	DisableMDevConfiguration *bool `json:"disableMDevConfiguration,omitempty"`
 
-	// Enable persistent reservation of a LUN through the SCSI Persistent Reserve commands on Kubevirt.
-	// In order to issue privileged SCSI ioctls, the VM requires activation of the persistent reservation flag.
-	// Once this feature gate is enabled, then the additional container with the qemu-pr-helper is deployed inside the virt-handler pod.
-	// Enabling (or removing) the feature gate causes the redeployment of the virt-handler pod.
+	// This feature gate has graduated to a dedicated configuration field.
```
nunnatsa, line 423: since it's still active, I would add back the documentation, in addition to the existing deprecation notice.

pkg/webhooks/mutator/hyperConvergedMutator.go
```diff
@@ -336,3 +364,119 @@ func hcMutateV1MDevFGAndEnabledOnUpdate(hc, oldHC *hcov1.HyperConverged, patches
...
+	if hc.Spec.Storage == nil ||
+		hc.Spec.Storage.PersistentReservationConfiguration == nil ||
+		hc.Spec.Storage.PersistentReservationConfiguration.Enabled == nil {
+		return false, false
+	}
+
+	return *hc.Spec.Storage.PersistentReservationConfiguration.Enabled, true
+}
+
+func dropPRFG(fgs hcov1fg.HyperConvergedFeatureGates, patches []jsonpatch.JsonPatchOperation) []jsonpatch.JsonPatchOperation {
```
nunnatsa: This is exactly the same function as `dropMdevFG`. Maybe you can create a common function that will get the FG name as parameter?

pkg/webhooks/mutator/hyperConvergedMutator.go
```diff
@@ -336,3 +364,119 @@ func hcMutateV1MDevFGAndEnabledOnUpdate(hc, oldHC *hcov1.HyperConverged, patches
...
+		if fgEnabled != enabled {
+			return false, nil, nil
+		}
+		return true, []string{prWarning}, patches
+	}
+
+	return true, []string{prWarning}, mutatePREnabled(hc.Spec.Storage, fgEnabled, patches)
+}
+
+func hcMutateV1PRFGAndEnabledOnUpdate(hc, oldHC *hcov1.HyperConverged, patches []jsonpatch.JsonPatchOperation) (allow bool, warningList []string, newPatches []jsonpatch.JsonPatchOperation) {
```
nunnatsa, line 419: I wonder if it can be combined with hcMutateV1MDevFGAndEnabledOnUpdate. I guess not ):

nunnatsa (commented):
few more comments

api/v1/hyperconverged_types.go
akalenyu: Can't rule out, yes, this is why it's nested in both kubevirt/hco

nunnatsa:
@akalenyu, you'll need to rebase and to run `make generate-doc` again.

akalenyu:
> @akalenyu, you'll need to rebase and to run `make generate-doc` again.

pushed a rebase, will take a look at the review comments too

api/v1beta1/conversion.go
nunnatsa, line 63: Please check if we can call `convertPersistentReservationV1beta1ToV1()` from within `convertStorageV1beta1ToV1()`

api/v1beta1/conversion.go
nunnatsa, line 97: Please check if we can call `convertPersistentReservationV1ToV1beta1()` from `convertStorageV1ToV1beta1()`

controllers/handlers/kubevirt.go
```diff
@@ -646,6 +646,26 @@ func toKvMediatedDevicesConfiguration(hc *hcov1.HyperConverged) *kubevirtcorev1.
 	return kvMdev
 }
 
+func toKvPersistentReservationConfiguration(hc *hcov1.HyperConverged) *kubevirtcorev1.PersistentReservationConfiguration {
+	var enabled *bool
+
+	if hc.Spec.Storage != nil &&
+		hc.Spec.Storage.PersistentReservationConfiguration != nil &&
+		hc.Spec.Storage.PersistentReservationConfiguration.Enabled != nil {
+		enabled = hc.Spec.Storage.PersistentReservationConfiguration.Enabled
```
nunnatsa: I think it's safer to copy the value instead of relaying a pointer we don't control anymore. e.g.
```go
enabled = new(*hc.Spec.Storage.PersistentReservationConfiguration.Enabled)
```

pkg/webhooks/mutator/hyperConvergedMutator.go
nunnatsa: Please combine with `dropMdevFG`:
rename to something like `dropFeatureGate()` and add a new `fgName` string parameter. the rest is the same.

pkg/webhooks/validator/v1beta1_validator_test.go
```diff
@@ -504,6 +504,11 @@ var _ = Describe("v1beta1 webhooks validator", func() {
 				Entry("should trigger a warning if the disableMDevConfiguration=true FG exists in the CR",
 					v1beta1.HyperConvergedFeatureGates{DisableMDevConfiguration: ptr.To(true)}, "disableMDevConfiguration"),
 
+				Entry("should trigger a warning if the persistentReservation=false FG exists in the CR",
+					v1beta1.HyperConvergedFeatureGates{PersistentReservation: ptr.To(false)}, "persistentReservation"),
```
nunnatsa: please use the golang builtin `new`, instead of `ptr.To`

pkg/webhooks/validator/v1beta1_validator_test.go
```diff
@@ -504,6 +504,11 @@ var _ = Describe("v1beta1 webhooks validator", func() {
...
+				Entry("should trigger a warning if the persistentReservation=true FG exists in the CR",
+					v1beta1.HyperConvergedFeatureGates{PersistentReservation: ptr.To(true)}, "persistentReservation"),
```
nunnatsa: same

pkg/webhooks/validator/v1beta1_validator_test.go
```diff
@@ -1460,6 +1465,11 @@ var _ = Describe("v1beta1 webhooks validator", func() {
...
```
nunnatsa: same

docs/cluster-configuration-v1beta1-api.md
```diff
@@ -154,29 +154,14 @@ For additional information, see here: [KubeSecondaryDNS](https://github.com/kube
...
-    devices:
-      disks:
-      - name: mypvcdisk
-        lun:
-          reservations: true
-```
-**Note**: An important aspect of this feature is that the SCSI persistent reservation doesn't support migration. Even if you apply the reservation to an RWX PVC provisioning SCSI devices, the restriction is due to the reservation done by the initiator on the node. The VM could be migrated but not the reservation.
-
-**Note**: this feature is in Developer Preview.
+The `persistentReservation` feature gate is deprecated. Use
```
nunnatsa: Please keep the old documentation here, with the deprecation warning and the reference to v1. This change leave us with no documentation for this feature in v1beta1.

nunnatsa (commented):
Thanks @akalenyu . The PR looks good.

My concern is that it adds a lot of code that is very similar to the mdev case (except for the mdev FG reverse logic), and to a least one additional feature gate we're going to move to a field in this release.

I wonder how much of this code can be generalized, to allow reuse. I know that nested optional fields are hard to combine into a common function, as well as different default values.

Anyway, we can do it in a followup PR.

v1beta1 documentation should stay there, I think. We just need to add a deprecation message and a reference to the new v1 field.

Please also add some coverage in `api/v1beta1/conversion_fuzz_test.go`, and run
```shell
make test-fuzz-api-conversion
```

api/v1beta1/conversion.go
akalenyu: both storage conversion functions have early returns that would get in the way, anyway, I believe keeping each converted chunk in its own function is a common practice in projects with conversion webhooks

pkg/webhooks/validator/v1beta1_validator_test.go
akalenyu: done

pkg/webhooks/mutator/hyperConvergedMutator.go
akalenyu: done

controllers/handlers/kubevirt.go
akalenyu: true, done

docs/cluster-configuration-v1beta1-api.md
akalenyu: makes sense!

api/v1beta1/conversion.go
nunnatsa: The conversion functions become longer and longer...

controllers/handlers/kubevirt_test.go
```diff
@@ -1999,23 +1998,6 @@ Version: 1.2.3`)
 							Expect(kv.Annotations).ToNot(HaveKey(kubevirtcorev1.EmulatorThreadCompleteToEvenParity))
 						},
 					),
-					// PersistentReservation
```
nunnatsa, line 2002: Isn't this test still valid? the deprecated FG is going to be dropped, but it's still supported.

api/v1beta1/conversion.go
akalenyu: yeah I think it has to do with the problem definition of a conversion webhook: https://github.com/kubernetes-sigs/cluster-api/blob/main/api/core/v1beta1/conversion.go

controllers/handlers/kubevirt_test.go
akalenyu: nope, it always translates to the proper config field. it is supported for backwards compatibility so users that don't know better (like HCO) can keep getting their expectation satisfied

nunnatsa (approved):
/approve

Thanks @akalenyu !!

PR #4392, opened 2026-07-12
DOC: Update the cluster-configuration.md: FG names are case-insensitive

nunnatsa:
Follow-up #4384: update the user guide to include the fact that the feature gate names are now case-insensitive.

Also, in order to match to the cluster-configuration files, and to reflect that the default API version is now `v1`, rename the `docs/api.md` file to `docs/api-v1beta1.md`, and `docs/api-v1.md`, to `docs/api.md`.

**Release note**:
```release-note
None
```


docs/cluster-configuration.md
```diff
@@ -69,6 +69,25 @@ spec:
...
+  - name: AlignCPUs
+  - name: DecentralizedLiveMigration
+    state: Disabled
+```
+
+> **Note**: The feature gate names in the `v1beta1` API version are *case-sensitive*.
+> 
+> **Note**: It is not allowed to create more than one feature gate with the same name (case-insensitive). This kind of
+> change will be rejected.
+
```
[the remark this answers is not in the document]
nunnatsa: @coderabbitai, the v1beta1 API version documentation is done in a different file.

docs/cluster-configuration.md
```diff
@@ -69,6 +69,25 @@ spec:
...
+  name: kubevirt-hyperconverged
+  namespace: kubevirt-hyperconverged
+spec:
+  featureGates:
+  - name: AlignCPUs
+  - name: DecentralizedLiveMigration
+    state: Disabled
+```
+
+> **Note**: The feature gate names in the `v1beta1` API version are *case-sensitive*.
```
orenc1: if this file is only about `v1` api version, why do we have a note here about `v1beta1`?
it can confuse the readers.

docs/cluster-configuration.md
```diff
@@ -69,6 +69,25 @@ spec:
...
+> 
+> **Note**: It is not allowed to create more than one feature gate with the same name (case-insensitive). This kind of
```
orenc1: the word `create` here is a bit inaccurate, since the CEL validation applies for any action, such as `patch` or `update`, not just `create`.
maybe:

```suggestion
> **Note**: It is not allowed to have more than one feature gate with the same name (case-insensitive). This kind of
```
or:

```suggestion
> **Note**: Feature gate names must be unique (case-insensitive).
```

orenc1 (commented):
nice.
please see couple of comments below

docs/cluster-configuration.md
nunnatsa: removed

docs/cluster-configuration.md
nunnatsa: Changed to the 2nd option.

orenc1 (approved):
/approve

docs/api.md
```diff
@@ -94,67 +172,88 @@ HyperConvergedList contains a list of HyperConverged
...
+## HyperConvergedSpec
 
-HyperConvergedObsoleteCPUs allows avoiding scheduling of VMs for obsolete CPU models
+HyperConvergedSpec defines the desired state of HyperConverged
 
 | Field | Description | Scheme | Default | Required |
 | ----- | ----------- | ------ | ------- | -------- |
-| minCPUModel | MinCPUModel is not in use Deprecated: This field is not in use and is ignored. | string |  | false |
-| cpuModels | CPUModels is a list of obsolete CPU models. When the node-labeller obtains the list of obsolete CPU models, it eliminates those CPU models and creates labels for valid CPU models. The default values for this field is nil, however, HCO uses opinionated values, and adding values to this list will add them to the opinionated values. | []string |  | false |
+| featureGates | For feature gate details, see [here](#hco-feature-gates) | featuregates.HyperConvergedFeatureGates |  | false |
```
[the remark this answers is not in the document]
nunnatsa: @coderabbitai - this is an auto generated file. ignore it.

PR #4394, opened 2026-07-13
fix panic alert to fire from first panic

avlitman:
Fix VMNonRecoverableOSPanic and ClusterVMPanicDetected to detect the first panic, and eliminates the fractional values.

**Release note**:

```release-note
none
```


michalskrivanek:
this query seems to work ok for cases when it didn't exist and a panic happened, correctly showing "1".
/lgtm

PR #4401, opened 2026-07-14
Deploy observability-controller

avlitman:
Adds the kubevirt-observability-controller as a standalone operand deployed by HCO behind the deployObservabilityController feature flag. HCO creates the ServiceAccount, ClusterRole, ClusterRoleBinding, Deployment, and metrics Service. The controller manages KubeVirt core PrometheusRules and exposes metrics.

**Jira Ticket**:

```jira-ticket
https://redhat.atlassian.net/browse/CNV-87589
```

**Release note**:

```release-note
Deploy virt-observability-controller as an HCO-managed operand, enabling independent shipping of KubeVirt monitoring rules and metrics without requiring KubeVirt control plane updates.
```

## Summary by CodeRabbit

* **New Features**
  * Added optional deployment of the observability controller through the `deployObservabilityController` configuration setting.
  * The controller exposes KubeVirt metrics and manages Prometheus monitoring rules independently.
  * Added automatic creation and reconciliation of required permissions, service account, and deployment resources.
* **Configuration**
  * Installation manifests now support configuring the observability controller image and version.
* **Compatibility**
  * Preserved the new setting during API version conversion and round trips.
* **Security**
  * Applied hardened pod and container security settings.


controllers/handlers/observabilitycontroller/deployment.go
```diff
@@ -0,0 +1,232 @@
...
+
+	selectorLabels := map[string]string{
+		hcoutil.AppLabel:          hcoutil.HyperConvergedName,
+		hcoutil.AppLabelComponent: string(hcoutil.AppComponentObservability),
+	}
+
+	podLabels := operands.GetLabels(hcoutil.AppComponentObservability)
+
+	affinity := &corev1.Affinity{
+		NodeAffinity: &corev1.NodeAffinity{
```
nunnatsa: Is this a requirement? Should the pod run on a control plane node? why?

controllers/handlers/observabilitycontroller/deployment.go
```diff
@@ -0,0 +1,232 @@
...
+		},
+	}
+}
+
+func newDeployment(_ *hcov1.HyperConverged) *appsv1.Deployment {
+	image := os.Getenv(hcoutil.ObservabilityControllerImageEnvV)
+
+	var replicas int32
+	if nodeinfo.IsInfrastructureHighlyAvailable() {
+		replicas = int32(2)
```
nunnatsa: Is it really needed?

controllers/handlers/observabilitycontroller/deployment.go
```diff
@@ -0,0 +1,232 @@
...
+		Replicas: &replicas,
+		Selector: &metav1.LabelSelector{
+			MatchLabels: selectorLabels,
+		},
+		Template: corev1.PodTemplateSpec{
+			ObjectMeta: metav1.ObjectMeta{
+				Labels: podLabels,
+				Annotations: map[string]string{
+					"openshift.io/required-scc": "restricted-v2",
+				},
```
nunnatsa: is this needed? why?

controllers/handlers/observabilitycontroller/deployment.go
```diff
@@ -0,0 +1,232 @@
...
+			},
+			Spec: corev1.PodSpec{
+				ServiceAccountName: serviceAccountName,
+				PriorityClassName:  "system-cluster-critical",
+				Affinity:           affinity,
+				NodeSelector: map[string]string{
+					"kubernetes.io/os": "linux",
+				},
```
nunnatsa: why?

controllers/handlers/observabilitycontroller/deployment.go
```diff
@@ -0,0 +1,232 @@
...
+						Ports: []corev1.ContainerPort{
+							{
+								ContainerPort: 8080,
+								Name:          "metrics",
+								Protocol:      corev1.ProtocolTCP,
+							},
+							{
+								ContainerPort: 8081,
+								Name:          "health",
+								Protocol:      corev1.ProtocolTCP,
```
nunnatsa: the endpoints must use https. if the new server does not do it, it must be implemented. then at least the metrics endpoint should move to 8433 or something like that.

controllers/handlers/observabilitycontroller/deployment.go
```diff
@@ -0,0 +1,232 @@
...
+								corev1.ResourceMemory: resourceMemoryRequest,
+							},
+							Limits: corev1.ResourceList{
+								corev1.ResourceMemory: resourceMemoryLimit,
+							},
+						},
+						SecurityContext: &corev1.SecurityContext{
+							AllowPrivilegeEscalation: new(bool),
+							ReadOnlyRootFilesystem:   boolPtr(true),
+							RunAsNonRoot:             boolPtr(true),
```
nunnatsa: why? can't it run as a regular user?

controllers/handlers/observabilitycontroller/deployment.go
```diff
@@ -0,0 +1,232 @@
...
+						},
+						TerminationMessagePath:   corev1.TerminationMessagePathDefault,
+						TerminationMessagePolicy: corev1.TerminationMessageFallbackToLogsOnError,
+					},
+				},
+			},
+		},
+	}
+
+	*dep.Spec.Template.Spec.SecurityContext.RunAsNonRoot = true
```
nunnatsa: why?

controllers/handlers/observabilitycontroller/rbac.go
```diff
@@ -0,0 +1,89 @@
...
+			{
+				APIGroups: []string{"monitoring.coreos.com"},
+				Resources: []string{"prometheusrules"},
+				Verbs:     []string{"get", "list", "watch", "create", "update", "delete"},
+			},
+			{
+				APIGroups: []string{""},
+				Resources: []string{"configmaps"},
+				Verbs:     []string{"get"},
+			},
```
nunnatsa: As far as I know, an entity, like this operator, may only grant permissions if it has it for itself. HCO does not have these permissions.

controllers/handlers/observabilitycontroller/service.go
```diff
@@ -0,0 +1,40 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      serviceName,
+			Namespace: hcoutil.GetOperatorNamespaceFromEnv(),
+			Labels:    operands.GetLabels(hcoutil.AppComponentObservability),
+		},
+		Spec: corev1.ServiceSpec{
+			Ports: []corev1.ServicePort{
+				{
+					Port:       8080,
+					TargetPort: intstr.FromInt32(8080),
```
nunnatsa: reminder to move to https

hack/config
```diff
@@ -12,6 +12,7 @@ INFLIGHT_OPERATIONS_VERSION="v0.0.6"
 WASP_AGENT_VERSION="v1.2.2"
 AIE_WEBHOOK_VERSION="v1.1.0"
 IOMMUFD_DEVICE_PLUGIN_VERSION="v0.0.1"
+OBSERVABILITY_CONTROLLER_VERSION="v0.0.1"
```
nunnatsa: Please also handle this in automation/release-bumper/release-bumper.sh

nunnatsa (commented):
Thanks @avlitman - added a few comments and questions inline.

Also missing the TLS configurations for the new servers. Please make sure to implement it in the server images, and then here.

hack/config
avlitman: done

controllers/handlers/observabilitycontroller/rbac.go
avlitman: removed

controllers/handlers/observabilitycontroller/deployment.go
avlitman: replaced with boolPtr(true)

controllers/handlers/observabilitycontroller/deployment.go
avlitman: removed

controllers/handlers/observabilitycontroller/deployment.go
avlitman: removed

controllers/handlers/observabilitycontroller/deployment.go
avlitman: removed

controllers/handlers/observabilitycontroller/deployment.go
avlitman: removed

controllers/handlers/observabilitycontroller/deployment.go
avlitman: TLS support needs to be implemented in the observability-controller first no? cc @machadovilaca

controllers/handlers/observabilitycontroller/service.go
avlitman: TLS support needs to be implemented in the observability-controller first no? cc @machadovilaca

controllers/handlers/observabilitycontroller/deployment.go
avlitman: it does run as a regular user, these settings enforce exactly that no?

controllers/handlers/observabilitycontroller/deployment.go
nunnatsa: Right, but it's a must. I would even say a pre-requisite.

machadovilaca: you can just configure

```
"--metrics-bind-address=:8080",
"--metrics-secure=false",`
```

to 8443 and 'true'

nunnatsa: cool. We'll also need the option to set the min TLS version and the TLS ciphers. In the next version we'll also need to be able to set the TLS curves.

controllers/handlers/observabilitycontroller/rbac.go
```diff
@@ -0,0 +1,117 @@
...
+		{
+			APIGroups: []string{""},
+			Resources: []string{"configmaps"},
+			Verbs:     []string{"get"},
+		},
+		{
+			APIGroups: []string{""},
+			Resources: []string{"services", "secrets"},
+			Verbs:     []string{"get", "list", "watch", "create", "update", "delete"},
+		},
```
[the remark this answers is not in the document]
avlitman: the controller NEEDS them for MetricsResourcesReconciler no?

controllers/handlers/observabilitycontroller/conditional_test.go
```diff
@@ -0,0 +1,206 @@
...
+
+	hcov1 "github.com/kubevirt/hyperconverged-cluster-operator/api/v1"
+	"github.com/kubevirt/hyperconverged-cluster-operator/controllers/common"
+	"github.com/kubevirt/hyperconverged-cluster-operator/controllers/commontestutils"
+	"github.com/kubevirt/hyperconverged-cluster-operator/controllers/operands"
+	hcoutil "github.com/kubevirt/hyperconverged-cluster-operator/pkg/util"
+)
+
+type handlerTestEntry struct {
+	name       string
```
nunnatsa: name => kind

controllers/handlers/observabilitycontroller/conditional_test.go
```diff
@@ -0,0 +1,206 @@
...
+		newHandler: func(cl client.Client, s *runtime.Scheme) operands.Operand { return NewClusterRoleBindingHandler(cl, s) },
+		existing:   func() client.Object { return newClusterRoleBinding() },
+		getKey:     func(_ *hcov1.HyperConverged) client.ObjectKey { return client.ObjectKey{Name: clusterRoleBindingName} },
+	},
+	{
+		name:       "ServiceAccount",
+		newHandler: func(cl client.Client, s *runtime.Scheme) operands.Operand { return NewServiceAccountHandler(cl, s) },
+		existing: func() client.Object {
+			sa := newServiceAccount()
+			sa.Namespace = commontestutils.NewHco().Namespace
```
nunnatsa: Please use the `commontestutils.Namespace` const, instead of creating a large object only for one of its fields.

controllers/handlers/observabilitycontroller/conditional_test.go
```diff
@@ -0,0 +1,206 @@
...
+		req *common.HcoRequest
+	)
+
+	BeforeEach(func() {
+		hco = commontestutils.NewHco()
+		req = commontestutils.NewReq(hco)
+		setupImageEnv()
+	})
+
+	for _, e := range handlerEntries {
```
nunnatsa: Please consider using [`DescribeTableSubtree`](https://onsi.github.io/ginkgo/#generating-subtree-tables) instead.

controllers/handlers/observabilitycontroller/deployment.go
```diff
@@ -0,0 +1,176 @@
...
+			Namespace: hcoutil.GetOperatorNamespaceFromEnv(),
+			Labels:    operands.GetLabels(hcoutil.AppComponentObservability),
+		},
+	}
+}
+
+func newDeployment(_ *hcov1.HyperConverged) *appsv1.Deployment {
+	image := os.Getenv(hcoutil.ObservabilityControllerImageEnvV)
+
+	var replicas int32 = 1
```
nunnatsa: if this is not going to change (e.g. in other deployments we check if the cluster highly available, them make it two replicas), please remove the variable.

Instead, set the field directly with `new(int32(1))`. (line 59)

controllers/handlers/observabilitycontroller/deployment.go
```diff
@@ -0,0 +1,176 @@
...
+			},
+		},
+	}
+
+	return dep
+}
+
+func boolPtr(b bool) *bool {
+	return &b
+}
```
nunnatsa: Not needed. please use the builtin `new` function instead. i.e. `new(true)` or `new(false)`.

controllers/handlers/observabilitycontroller/deployment.go
```diff
@@ -0,0 +1,176 @@
...
+				ServiceAccountName: serviceAccountName,
+				PriorityClassName:  "system-cluster-critical",
+				Tolerations: []corev1.Toleration{
+					{
+						Key:      "CriticalAddonsOnly",
+						Operator: corev1.TolerationOpExists,
+					},
+				},
+				SecurityContext: &corev1.PodSecurityContext{
+					RunAsNonRoot: boolPtr(true),
```
nunnatsa: use `new`

controllers/handlers/observabilitycontroller/deployment.go
```diff
@@ -0,0 +1,176 @@
...
+								corev1.ResourceMemory: resourceMemoryRequest,
+							},
+							Limits: corev1.ResourceList{
+								corev1.ResourceMemory: resourceMemoryLimit,
+							},
+						},
+						SecurityContext: &corev1.SecurityContext{
+							AllowPrivilegeEscalation: boolPtr(false),
+							ReadOnlyRootFilesystem:   boolPtr(true),
+							RunAsNonRoot:             boolPtr(true),
```
nunnatsa: use `new` instead of `boolPtr`

controllers/handlers/observabilitycontroller/deployment.go
```diff
@@ -0,0 +1,176 @@
...
+				Containers: []corev1.Container{
+					{
+					Name:  "manager",
+					Image: image,
+					Args: []string{
+						"--metrics-bind-address=:8443",
+						"--metrics-secure=true",
+						"--health-probe-bind-address=:8081",
+						"--leader-elect=false",
+					},
```
nunnatsa: Please handle TLS min version and TLS ciphers. This maybe requires changes in the controller code, but it's a must.

controllers/handlers/observabilitycontroller/rbac.go
```diff
@@ -0,0 +1,127 @@
...
+	cr.Rules = []rbacv1.PolicyRule{
+		{
+			APIGroups: []string{"kubevirt.io"},
+			Resources: []string{"virtualmachines", "virtualmachineinstances", "virtualmachineinstancemigrations", "kubevirts"},
+			Verbs:     []string{"get", "list", "watch"},
+		},
+		{
+			APIGroups: []string{"instancetype.kubevirt.io"},
+			Resources: []string{"virtualmachineinstancetypes", "virtualmachineclusterinstancetypes", "virtualmachinepreferences", "virtualmachineclusterpreferences"},
+			Verbs:     []string{"get", "list", "watch"},
```
nunnatsa, line 58: Not sure this will work. HCO does not have these permissions, so it cant grant it to other entities.

deploy/images.env
```diff
@@ -43,6 +43,7 @@ MIGRATION_OPERATOR_IMAGE="${MIGRATION_OPERATOR_IMAGE:-quay.io/kubevirt/kubevirt-
 MIGRATION_CONTROLLER_IMAGE="${MIGRATION_CONTROLLER_IMAGE:-quay.io/kubevirt/kubevirt-migration-controller@sha256:d235178862451a14826894d9ddf8af96898fc098ae6e7af56987f694cfd57c5b}"
 AUTOPILOT_IMAGE="${AUTOPILOT_IMAGE:-quay.io/openshift-virtualization/virt-platform-autopilot@sha256:ba94138eb74ebf3508dd1f2722773d0f1091aa683947ff1a1a5690168fed6384}"
 INFLIGHT_OPERATIONS_IMAGE="${INFLIGHT_OPERATIONS_IMAGE:-quay.io/ifo-operator/controller@sha256:3dff33743351c00c9d2b7fc990e8820ac4db88690f6754533bcc23f9bbaa7eaf}"
+OBSERVABILITY_CONTROLLER_IMAGE="${OBSERVABILITY_CONTROLLER_IMAGE:-quay.io/kubevirt/virt-observability-controller:${OBSERVABILITY_CONTROLLER_VERSION}}"
```
nunnatsa: is it a manual change? this file is auto generated.

nunnatsa (commented):
Thanks @avlitman 

Added some more inline comments. 

One critical comment is about TLS settings, which are mandatory.

controllers/handlers/observabilitycontroller/rbac.go
avlitman: Added the missing permissions to HCO's own ClusterRole

deploy/images.env
avlitman: now auto-generated

controllers/handlers/observabilitycontroller/deployment.go
avlitman: fixed, @machadovilaca need to add --tls-min-version and --tls-cipher-suites flags to the upstream controller

controllers/handlers/observabilitycontroller/conditional_test.go
```diff
@@ -0,0 +1,209 @@
...
+			})
+
+			It("should not create when deployObservabilityController feature gate is not set", func() {
+				cl := commontestutils.InitClient([]client.Object{hco})
+
+				handler := entry.newHandler(cl, commontestutils.GetScheme())
+				res := handler.Ensure(req)
+
+				Expect(res.Err).ToNot(HaveOccurred())
+				Expect(res.Created).To(BeFalse())
```
nunnatsa, line 72: I would try to read the CR, to make sure it's not there. not critical.

controllers/handlers/observabilitycontroller/conditional_test.go
```diff
@@ -0,0 +1,209 @@
...
+				cl := commontestutils.InitClient([]client.Object{hco, entry.existing()})
+
+				handler := entry.newHandler(cl, commontestutils.GetScheme())
+				res := handler.Ensure(req)
+
+				Expect(res.Err).ToNot(HaveOccurred())
+				Expect(res.Deleted).To(BeTrue())
+
+				obj := entry.existing()
+				err := cl.Get(context.Background(), entry.getKey(hco), obj)
```
nunnatsa: This will not work. this line is trying to read a resource with the name and the namespace of the HyperConverged CR.

nunnatsa: WDYT about something like this?
```go
			It("should delete when deployObservabilityController feature gate is disabled", func() {
				hco.Spec.FeatureGates.Disable(featureGateName)
				cr := entry.existing()
				crKey := entry.getKey(cr)
				cl := commontestutils.InitClient([]client.Object{hco, cr})

				handler := entry.newHandler(cl, commontestutils.GetScheme())
				res := handler.Ensure(req)

				Expect(res.Err).ToNot(HaveOccurred())
				Expect(res.Deleted).To(BeTrue())

				obj := entry.existing()
				err := cl.Get(context.Background(), entry.getKey(crKey), obj)
```

controllers/handlers/observabilitycontroller/conditional_test.go
```diff
@@ -0,0 +1,209 @@
...
+				Expect(err).To(HaveOccurred())
+				Expect(err).To(MatchError(apierrors.IsNotFound, "not found error"))
```
nunnatsa: the `MatchError` matcher also checks that the error is not nil. no need to use `HaveOccurred()` if your're going to use it. line 125 may be dropped. This is cosmetic and not critical.

controllers/handlers/observabilitycontroller/conditional_test.go
```diff
@@ -0,0 +1,209 @@
...
+			})
+
+			It("should reconcile labels while preserving user labels", func() {
+				hco.Spec.FeatureGates.Enable(featureGateName)
+				obj := entry.existing()
+				labels := obj.GetLabels()
+				expectedLabels := maps.Clone(labels)
+				delete(labels, hcoutil.AppLabelComponent)
+				labels["user-added-label"] = "user-value"
+				obj.SetLabels(labels)
```
nunnatsa: this test should validate that the CR exists when the FG is enabled. Testing the label logic is not part of it. I would remove these line to keep the test focused.

controllers/handlers/observabilitycontroller/conditional_test.go
```diff
@@ -0,0 +1,209 @@
...
+				Expect(res.Err).ToNot(HaveOccurred())
+				Expect(res.Updated).To(BeTrue())
+
+				found := entry.existing()
+				Expect(cl.Get(context.Background(), entry.getKey(hco), found)).To(Succeed())
+
+				for key, value := range expectedLabels {
+					Expect(found.GetLabels()).To(HaveKeyWithValue(key, value))
+				}
+				Expect(found.GetLabels()).To(HaveKeyWithValue("user-added-label", "user-value"))
```
nunnatsa: same

controllers/handlers/observabilitycontroller/conditional_test.go
```diff
@@ -0,0 +1,209 @@
...
+				cl := commontestutils.InitClient([]client.Object{hco, obj})
+
+				handler := entry.newHandler(cl, commontestutils.GetScheme())
+				res := handler.Ensure(req)
+
+				Expect(res.Err).ToNot(HaveOccurred())
+				Expect(res.Updated).To(BeTrue())
+
+				found := entry.existing()
+				Expect(cl.Get(context.Background(), entry.getKey(hco), found)).To(Succeed())
```
nunnatsa: the key is wrong. as above.

controllers/handlers/observabilitycontroller/conditional_test.go
```diff
@@ -0,0 +1,209 @@
...
+				for key, value := range expectedLabels {
+					Expect(found.GetLabels()).To(HaveKeyWithValue(key, value))
+				}
+				Expect(found.GetLabels()).To(HaveKeyWithValue("user-added-label", "user-value"))
+			})
+		},
+		Entry("ClusterRole", handlerEntries[0]),
+		Entry("ClusterRoleBinding", handlerEntries[1]),
+		Entry("ServiceAccount", handlerEntries[2]),
+		Entry("Deployment", handlerEntries[3]),
```
nunnatsa: if we're only using the handlerEntries slice with direct index access, and we never loop over it, I think we are not benefit from the slice itself. 

I would either define 4 vards with meaningful names, or adding their content to the `Entry()` itself, e.g.
```go
Entry("ClusterRole", handlerTestEntry{
		kind:       "ClusterRole",
		newHandler: func(cl client.Client, s *runtime.Scheme) operands.Operand { return NewClusterRoleHandler(cl, s) },
		existing:   func() client.Object { return newClusterRole() },
		getKey:     func(_ *hcov1.HyperConverged) client.ObjectKey { return client.ObjectKey{Name: clusterRoleName} },
	})
```

controllers/handlers/observabilitycontroller/rbac.go
nunnatsa: This is not great... HCO can access resource it does not need. But if there is no other choice, we must do that

docs/api.md
```diff
@@ -497,6 +497,7 @@ A feature gate may be in the following phases:
 | alignCPUs | Enable KubeVirt to request up to two additional dedicated CPUs in order to complete the total CPU count to an even parity when using emulator thread isolation. Note: this feature is in Developer Preview. | alpha |
 | containerPathVolumes | ContainerPathVolumes enables the use of container paths as volumes in KubeVirt. This allows VMs to access files and directories from the virt-launcher pod's filesystem via virtiofs. | alpha |
 | deployKubeSecondaryDNS | Deploy KubeSecondaryDNS by CNAO | alpha |
+| deployObservabilityController | Deploy the virt-observability-controller component. When enabled, the controller exposes KubeVirt metrics and manages PrometheusRule resources independently from the KubeVirt control plane. | alpha |
```
nunnatsa, line 512: This seems like a leftover. Please run `make generate-docs` and commit the result.

nunnatsa (commented):
Thanks for the fixes @avlitman 

To general comments:
1. Again - setting the TLS min-version and TLS ciphers is mandatory for the deployment.
2. Please add human-readable, user-guide style documentation in docs/cluster-configuration.md

nunnatsa:
another general comment: we need a simple, minimal e2e test, to make sure the resources are deployed when we set the FG

avlitman:
[> This is not great... HCO can access resource it does not need. But if there is no other choice, we must do that](https://github.com/kubevirt/hyperconverged-cluster-operator/pull/4401#discussion_r3623553998)


cc @machadovilaca wdyt?

controllers/handlers/observabilitycontroller/conditional_test.go
avlitman: thanks!

controllers/handlers/observabilitycontroller/conditional_test.go
avlitman: thank you!

controllers/handlers/observabilitycontroller/rbac.go
machadovilaca: how does kubevirt/kubevirt manage the rbac permissions they need?

controllers/handlers/observabilitycontroller/conditional_test.go
```diff
@@ -0,0 +1,166 @@
...
+	"github.com/kubevirt/hyperconverged-cluster-operator/controllers/operands"
+	hcoutil "github.com/kubevirt/hyperconverged-cluster-operator/pkg/util"
+)
+
+type handlerTestEntry struct {
+	newHandler func(client.Client, *runtime.Scheme) operands.Operand
+	existing   func() client.Object
+}
+
+
```
nunnatsa: This is failing in sanity, because it runs `make goimport`, that reformat all the go files, and it removes the extra line.

Please run `make goinport` locally, before pushing the PR.

tests/func-tests/observability_controller_test.go
```diff
@@ -2,151 +2,116 @@ package tests_test
...
+const (
+	observabilityControllerName       = "virt-observability-controller"
+	observabilityControllerCRBName    = "virt-observability-controller-rolebinding"
+	observabilityControllerFGName     = "deployObservabilityController"
+)
+
+var _ = Describe("Observability Controller", Label("observability-controller"), Serial, Ordered, func() {
+	tests.FlagParse()
 
-var _ = Describe("Observability Controller", Label(tests.OpenshiftLabel, testName), func() {
```
nunnatsa, line 28: Why was the old test dropped?

tests/func-tests/observability_controller_test.go
```diff
@@ -2,151 +2,116 @@ package tests_test
...
 )
 
-const testName = "observability_controller"
+const (
+	observabilityControllerName       = "virt-observability-controller"
+	observabilityControllerCRBName    = "virt-observability-controller-rolebinding"
+	observabilityControllerFGName     = "deployObservabilityController"
+)
+
+var _ = Describe("Observability Controller", Label("observability-controller"), Serial, Ordered, func() {
```
nunnatsa: For now, add also the `tests.OpenshiftLabel` label. We don't have Prometheus installed in the kubevirtci lanes, and so this test should not run there. 

Later we could try to install Prometheus as part of the deployment and re-enable it for non-openshift clusters.

tests/func-tests/observability_controller_test.go
```diff
@@ -2,151 +2,116 @@ package tests_test
...
+		It("should deploy observability controller resources", func(ctx context.Context) {
+			By("enabling the deployObservabilityController feature gate")
+			Expect(tests.EnableFG(ctx, cli, observabilityControllerFGName)).To(Succeed())
+
+			By("checking the ServiceAccount is created")
+			Eventually(func(ctx context.Context) error {
+				sa := &corev1.ServiceAccount{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      observabilityControllerName,
+						Namespace: tests.InstallNamespace,
```
nunnatsa, line 139: The test is failing now. I thin this is because the cluster role is a cluster scoped resource. Please remove the namespace field.

nunnatsa (commented):
@thanks @avlitman - it looks great.

Added a few comment to fix several issues in CI.

I hope fixing them will solve all remining issues.

tests/func-tests/observability_controller_test.go
avlitman: Fixed

tests/func-tests/observability_controller_test.go
avlitman: added tests.OpenshiftLabel

tests/func-tests/observability_controller_test.go
avlitman: but the namespace field here is on serviceAccount and needed no?

nunnatsa: yap. Sorry for that. I'm trying to understand why it is not found in the functional test

avlitman: yes me too

nunnatsa:
/approve

Let's wait for a new version of the controller, because currently it's not working, and we'll need to fix the image here anyway when a new version lands.

/hold

avlitman:
> /approve
> 
> Let's wait for a new version of the controller, because currently it's not working, and we'll need to fix the image here anyway when a new version lands.
> 
> /hold

I upgraded to the new version of the controller with the fix, lets see if it pass

nunnatsa:
hco-e2e-consecutive-operator-sdk-upgrades-aws passed

/override ci/prow/hco-e2e-consecutive-operator-sdk-upgrades-azure

nunnatsa:
hco-e2e-operator-sdk-gcp lane passed
/override ci/prow/hco-e2e-operator-sdk-aws ci/prow/hco-e2e-operator-sdk-azure

PR #4413, opened 2026-07-19
CNV-47050: Allow downloading of the virtio-win iso image from the artifacts server

nunnatsa:

**Jira Ticket**:

```jira-ticket
https://redhat.atlassian.net/browse/CNV-47050
```

**Release note**:

```release-note
```

### DIFF
--- a/controllers/handlers/virtio_win_ConfigMap.go
+++ b/controllers/handlers/virtio_win_ConfigMap.go
@@ -114,7 +114,7 @@ func NewVirtioWinCmReaderRoleBinding() *rbacv1.RoleBinding {
 }
 
 func getVirtioImageName() (string, error) {
-	virtiowinContainer := os.Getenv("VIRTIOWIN_CONTAINER")
+	virtiowinContainer := os.Getenv(hcoutil.VirtioWinImageEnvV)
 	if virtiowinContainer == "" {
 		return "", errors.New("kv-virtiowin-image-name was not specified")
 	}
--- a/controllers/handlers/virtio_win_ConfigMap.go
+++ b/controllers/handlers/virtio_win_ConfigMap.go
@@ -2,6 +2,7 @@ package handlers
 
 import (
 	"errors"
+	"net/url"
 	"os"
 
 	corev1 "k8s.io/api/core/v1"
@@ -10,7 +11,9 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	hcov1 "github.com/kubevirt/hyperconverged-cluster-operator/api/v1"
 	"github.com/kubevirt/hyperconverged-cluster-operator/controllers/operands"
+	"github.com/kubevirt/hyperconverged-cluster-operator/pkg/downloadhost"
 	hcoutil "github.com/kubevirt/hyperconverged-cluster-operator/pkg/util"
 )
 
@@ -18,11 +21,12 @@ const virtioWinCmName = "virtio-win"
 
 // NewVirtioWinCmHandler creates the Virtio-Win ConfigMap Handler
 func NewVirtioWinCmHandler(cli client.Client, Scheme *runtime.Scheme) (operands.Operand, error) {
-	virtioWincm, err := NewVirtioWinCm()
+	_, err := getVirtioImageName()
 	if err != nil {
 		return nil, err
 	}
-	return operands.NewCmHandler(cli, Scheme, virtioWincm), nil
+
+	return operands.NewDynamicCmHandler(cli, Scheme, NewVirtioWinCm), nil
 }
 
 // NewVirtioWinCmReaderRoleHandler creates the Virtio-Win ConfigMap Role Handler
@@ -35,10 +39,28 @@ func NewVirtioWinCmReaderRoleBindingHandler(cli client.Client, Scheme *runtime.S
 	return operands.NewRoleBindingHandler(cli, Scheme, NewVirtioWinCmReaderRoleBinding())
 }
 
-func NewVirtioWinCm() (*corev1.ConfigMap, error) {
-	virtiowinContainer := os.Getenv("VIRTIOWIN_CONTAINER")
-	if virtiowinContainer == "" {
-		return nil, errors.New("kv-virtiowin-image-name was not specified")
+const (
+	virtioWinImageKey   = "virtio-win-image"
+	virtioWinImageDLKey = "virtio-win-image-download-url"
+)
+
+func NewVirtioWinCm(_ *hcov1.HyperConverged) (*corev1.ConfigMap, error) {
+	virtiowinContainer, err := getVirtioImageName()
+	if err != nil {
+		return nil, err
+	}
+
+	data := map[string]string{
+		virtioWinImageKey: virtiowinContainer,
+	}
+
+	if imageDLFilePath, envFound := os.LookupEnv(hcoutil.VirtIOWinDataFileEnvV); envFound && imageDLFilePath != "" {
+		downloadURL := url.URL{
+			Scheme: "https",
+			Host:   string(downloadhost.Get().CurrentHost),
+			Path:   imageDLFilePath,
+		}
+		data[virtioWinImageDLKey] = downloadURL.String()
 	}
 
 	return &corev1.ConfigMap{
@@ -47,9 +69,7 @@ func NewVirtioWinCm() (*corev1.ConfigMap, error) {
 			Labels:    operands.GetLabels(hcoutil.AppComponentDeployment),
 			Namespace: hcoutil.GetOperatorNamespaceFromEnv(),
 		},
-		Data: map[string]string{
-			"virtio-win-image": virtiowinContainer,
-		},
+		Data: data,
 	}, nil
 }
 
@@ -91,4 +111,13 @@ func NewVirtioWinCmReaderRoleBinding() *rbacv1.RoleBinding {
 			},
 		},
 	}
+}
+
+func getVirtioImageName() (string, error) {
+	virtiowinContainer := os.Getenv("VIRTIOWIN_CONTAINER")
+	if virtiowinContainer == "" {
+		return "", errors.New("kv-virtiowin-image-name was not specified")
+	}
+
+	return virtiowinContainer, nil
 }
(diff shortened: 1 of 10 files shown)


tests/func-tests/cli_download_test.go
```diff
@@ -160,6 +169,48 @@ var _ = Describe("[rfe_id:5100][crit:medium][vendor:cnv-qe@redhat.com][level:sys
...
+	Expect(cli.Get(ctx, client.ObjectKey{Namespace: tests.InstallNamespace, Name: "virtio-win"}, virtioWinCM)).To(Succeed())
+
+	virtioWinDLPathURL, urlExists := virtioWinCM.Data["virtio-win-image-download-url"]
+	Expect(urlExists).To(BeTrue())
+	Expect(virtioWinDLPathURL).To(ContainSubstring(cliDLHost))
+	Expect(virtioWinDLPathURL).To(HaveSuffix(virtioWinDLPath))
+	res, err := httpClient.Head(virtioWinDLPathURL)
+	Expect(err).NotTo(HaveOccurred(), "HEAD failed for %s", virtioWinDLPathURL)
+	Expect(res.StatusCode).To(Equal(http.StatusOK), "non OK response for %s", virtioWinDLPathURL)
+}
```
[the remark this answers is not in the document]
nunnatsa: I think this is a valid case for doing it. We need the same complex, long validation in two different places. It's a validation helper. So I think it's ok.

tests/func-tests/cli_download_test.go
```diff
@@ -59,23 +62,28 @@ var _ = Describe("[rfe_id:5100][crit:medium][vendor:cnv-qe@redhat.com][level:sys
...
 				continue
 			}
 			By("Checking links. Link:" + link.Href)
-			client := &http.Client{Transport: &http.Transport{
-				// ssl of the route is irrelevant
-				TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
-			}}
-			resp, err := client.Get(link.Href)
+			resp, err := httpClient.Head(link.Href)
 			_ = resp.Body.Close()
```
orenc1: the error returned by `httpClient.Head(link.Href)` is not checked before calling `Close()`, this might cause a nil pointer dereference if there's a network failure in the e2e test suite.
please move body close after the error check, or use defer after the nil check.

tests/func-tests/cli_download_test.go
```diff
@@ -59,23 +62,28 @@ var _ = Describe("[rfe_id:5100][crit:medium][vendor:cnv-qe@redhat.com][level:sys
...
+			Expect(err).NotTo(HaveOccurred(), "HEAD failed for %s", link.Href)
+			Expect(resp.StatusCode).To(Equal(http.StatusOK), "non OK response for %s", link.Href)
 
 			ExpectWithOffset(1, err).ToNot(HaveOccurred())
 			ExpectWithOffset(1, resp).To(HaveHTTPStatus(http.StatusOK))
```
orenc1, line 80: assertion duplication with lines 76-77, that are asserting the same thing.

pkg/components/components.go
```diff
@@ -303,6 +313,40 @@ func GetDeploymentSpecCliDownloads(params *DeploymentOperatorParams) appsv1.Depl
...
+func addVirtIOVolume(spec *appsv1.DeploymentSpec, params *DeploymentOperatorParams) {
+	if params.VirtIOWinDataFile == "" || params.VirtIOWinMountPath == "" {
+		return
+	}
+
+	spec.Template.Spec.Volumes = append(spec.Template.Spec.Volumes, corev1.Volume{
+		Name:         "virtiowin-data",
+		VolumeSource: corev1.VolumeSource{EmptyDir: &corev1.EmptyDirVolumeSource{}},
+	})
+	spec.Template.Spec.InitContainers = []corev1.Container{
```
orenc1: best practice is to use `append` rather than overwriting the init container, to be future-safe.

pkg/components/components.go
```diff
@@ -303,6 +313,40 @@ func GetDeploymentSpecCliDownloads(params *DeploymentOperatorParams) appsv1.Depl
...
+			Image:           params.VirtIOWinContainer,
+			ImagePullPolicy: corev1.PullPolicy(params.ImagePullPolicy),
+			Command:         []string{"cp", params.VirtIOWinDataFile, "/mnt/virtiowin-data/"},
+			VolumeMounts: []corev1.VolumeMount{
+				{Name: "virtiowin-data", MountPath: "/mnt/virtiowin-data"},
+			},
+			SecurityContext:          getVirtIOWinInitContainerSecurityContext(),
+			TerminationMessagePolicy: corev1.TerminationMessageFallbackToLogsOnError,
+		},
+	}
```
orenc1: the new init container is missing `Resources` (requests/limits) unlike other components. this can cause issues when using resource quotas and LimitRanges on the cluster.

pkg/components/components.go
```diff
@@ -303,6 +313,40 @@ func GetDeploymentSpecCliDownloads(params *DeploymentOperatorParams) appsv1.Depl
...
+	spec.Template.Spec.Volumes = append(spec.Template.Spec.Volumes, corev1.Volume{
+		Name:         "virtiowin-data",
+		VolumeSource: corev1.VolumeSource{EmptyDir: &corev1.EmptyDirVolumeSource{}},
+	})
+	spec.Template.Spec.InitContainers = []corev1.Container{
+		{
+			Name:            "virtiowin-data-loader",
+			Image:           params.VirtIOWinContainer,
+			ImagePullPolicy: corev1.PullPolicy(params.ImagePullPolicy),
+			Command:         []string{"cp", params.VirtIOWinDataFile, "/mnt/virtiowin-data/"},
```
orenc1: the virtio-win image is a data-only image.
i checked `quay.io/kubevirt/virtio-container-disk:latest`, and it doesn't have the `cp` command.

controllers/handlers/virtio_win_ConfigMap.go
```diff
@@ -92,3 +112,12 @@ func NewVirtioWinCmReaderRoleBinding() *rbacv1.RoleBinding {
 		},
 	}
 }
+
+func getVirtioImageName() (string, error) {
+	virtiowinContainer := os.Getenv("VIRTIOWIN_CONTAINER")
```
orenc1: consider defining `VirtIOWinContainerEnvV = "VIRTIOWIN_CONTAINER"` as a shared constant in `pkg/utils/consts.go`

pkg/components/components.go
```diff
@@ -303,6 +313,40 @@ func GetDeploymentSpecCliDownloads(params *DeploymentOperatorParams) appsv1.Depl
...
+	return spec
+}
+
+func addVirtIOVolume(spec *appsv1.DeploymentSpec, params *DeploymentOperatorParams) {
+	if params.VirtIOWinDataFile == "" || params.VirtIOWinMountPath == "" {
+		return
+	}
+
+	spec.Template.Spec.Volumes = append(spec.Template.Spec.Volumes, corev1.Volume{
+		Name:         "virtiowin-data",
```
orenc1: `virtiowin-data` magic string is hard-coded 3 times.
it's better to use a const.

tests/func-tests/cli_download_test.go
```diff
@@ -160,6 +169,48 @@ var _ = Describe("[rfe_id:5100][crit:medium][vendor:cnv-qe@redhat.com][level:sys
...
+	virtioWinCM := &corev1.ConfigMap{}
+	Expect(cli.Get(ctx, client.ObjectKey{Namespace: tests.InstallNamespace, Name: "virtio-win"}, virtioWinCM)).To(Succeed())
+
+	virtioWinDLPathURL, urlExists := virtioWinCM.Data["virtio-win-image-download-url"]
+	Expect(urlExists).To(BeTrue())
+	Expect(virtioWinDLPathURL).To(ContainSubstring(cliDLHost))
+	Expect(virtioWinDLPathURL).To(HaveSuffix(virtioWinDLPath))
+	res, err := httpClient.Head(virtioWinDLPathURL)
+	Expect(err).NotTo(HaveOccurred(), "HEAD failed for %s", virtioWinDLPathURL)
+	Expect(res.StatusCode).To(Equal(http.StatusOK), "non OK response for %s", virtioWinDLPathURL)
```
orenc1, line 219: Add defer `res.Body.Close()` after the error checks, to avoid leaking tcp connections.

controllers/handlers/virtio_win_ConfigMap_test.go
```diff
@@ -81,14 +84,14 @@ var _ = Describe("VirtioWin", func() {
 		})
 
 		It("should reconcile according to env values and HCO CR", func() {
-			virtiowink := "virtio-win-image"
+			virtiowink := virtioWinImageKey
```
orenc1: why we need this assignment, and not using `virtioWinImageKey` later on?

tests/func-tests/cli_download_test.go
```diff
@@ -160,6 +169,48 @@ var _ = Describe("[rfe_id:5100][crit:medium][vendor:cnv-qe@redhat.com][level:sys
...
+
+			checkVirtioWinDownload(ctx, cli, httpClient, newCLIDLHost)
 		})
 	})
 })
+
+func getHTTPClient() *http.Client {
+	customTransport := http.DefaultTransport.(*http.Transport).Clone()
+	customTransport.TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
+	return &http.Client{Transport: customTransport, Timeout: time.Second * 1}
```
orenc1: timeout of 1 seconds is a bit low, can cause CI failures in case of transient network issues or if the cluster is loaded.

orenc1 (commented):
thanks @nunnatsa ,
i have a few comments below.

also, the requirement was to make the download URL of the virtio-win iso file accessible to anyone, but it's currently being stored in the `virtio-win` config map. regular users can `get` it, but only if they know the exact name, and they can't list config maps in the installation namespace (kubevirt-hyperconverged / openshift-cnv), so it's not really accessible to regular users via the UI.

tests/func-tests/cli_download_test.go
nunnatsa: this comment was done on an old version of the code...

tests/func-tests/cli_download_test.go
nunnatsa: 1. not accurate: err represent a golang error. the 2nd line checks the HTTP response - not the same thing. you can get 404 with no golang error.
2. This is not part of the PR.

pkg/components/components.go
nunnatsa: This is why it's controlled by the parameters.

controllers/handlers/virtio_win_ConfigMap.go
nunnatsa: This will cause many changes in many other files. but ok.

tests/func-tests/cli_download_test.go
nunnatsa: not needed for HEAD requests. no body.

nunnatsa:
The coverage dropped a bit, for some new code in the build code, not in production code (pkg/components)

/override "coverage/coveralls"

PR #4416, opened 2026-07-20
Split the components package

nunnatsa:
This package is used in the production code, but most of it is used in build time, to create the manifests.

This creates coupling that make it hard to maintain this pretty big package.

This PR takes the parts that are only relevant for the manifest generation, out to the new `tools/manifests` package, leaving only a small part of the original package, that used by the production or the test code.

**Jira Ticket**:

```jira-ticket
None
```

**Release note**:
```release-note
None
```


pkg/util/consts.go
```diff
@@ -104,7 +104,8 @@ const (
 	DeschedulerCRName    = "cluster"
 	DeschedulerNamespace = "openshift-kube-descheduler-operator"
 
-	DataImportCronEnabledAnnotation = "dataimportcrontemplate.kubevirt.io/enable"
+	DataImportCronEnabledAnnotation  = "dataimportcrontemplate.kubevirt.io/enable"
+	DisableOperandDeletionAnnotation = "console.openshift.io/disable-operand-delete"
```
orenc1, line 108: was this annotation added on purpose?
why we want to disable it?

nunnatsa: There is no code change here. only moving code around.

nunnatsa: Specifically about this const, it was moved from the `components` package, to allow dropping this dependency, when it only used for this constant.

orenc1: you're right.
/approve
/lgtm

PR #4419, opened 2026-07-21
[release-1.18] Allow downloading of the virtio-win iso image from the artifacts server

nunnatsa:

*This is a manual cherry pick of #4413*

**Jira Ticket**:

```jira-ticket
https://redhat.atlassian.net/browse/CNV-47050
```

**Release note**:

```release-note
```

### DIFF
--- a/controllers/handlers/virtio_win_ConfigMap.go
+++ b/controllers/handlers/virtio_win_ConfigMap.go
@@ -114,7 +114,7 @@ func NewVirtioWinCmReaderRoleBinding() *rbacv1.RoleBinding {
 }
 
 func getVirtioImageName() (string, error) {
-	virtiowinContainer := os.Getenv("VIRTIOWIN_CONTAINER")
+	virtiowinContainer := os.Getenv(hcoutil.VirtioWinImageEnvV)
 	if virtiowinContainer == "" {
 		return "", errors.New("kv-virtiowin-image-name was not specified")
 	}
--- a/controllers/handlers/virtio_win_ConfigMap.go
+++ b/controllers/handlers/virtio_win_ConfigMap.go
@@ -2,6 +2,7 @@ package handlers
 
 import (
 	"errors"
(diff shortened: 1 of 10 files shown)


controllers/handlers/virtio_win_ConfigMap.go
```diff
@@ -92,3 +112,12 @@ func NewVirtioWinCmReaderRoleBinding() *rbacv1.RoleBinding {
 		},
 	}
 }
+
+func getVirtioImageName() (string, error) {
+	virtiowinContainer := os.Getenv("VIRTIOWIN_CONTAINER")
```
orenc1: `VIRTIOWIN_CONTAINER` should be a const (`VirtioWinImageEnvV`) same as in the main PR.

controllers/handlers/virtio_win_ConfigMap_test.go
```diff
@@ -17,17 +17,20 @@ import (
...
+	const virtioImage = "new-virtiowin-container-value"
+
 	Context("Virtio-Win ConfigMap", func() {
 
 		var hco *hcov1beta1.HyperConverged
 		var req *common.HcoRequest
 
 		BeforeEach(func() {
-			os.Setenv("VIRTIOWIN_CONTAINER", "new-virtiowin-container-value")
+			os.Setenv("VIRTIOWIN_CONTAINER", virtioImage)
```
orenc1: this lacks the `Expect(...)` wrapper that was used in main:
https://github.com/kubevirt/hyperconverged-cluster-operator/blob/b57a0f9849a079008f947cf39a889a2d3ee29bb8/controllers/handlers/virtio_win_ConfigMap_test.go#L33

orenc1 (commented):
i found two issues that were ok in main and weren't backported here:

orenc1:
/lgtm
/approve
:+1:

PR #4423, opened 2026-07-21
Expose AllowWorkloadDisruption in LiveMigrationConfig

SamAlber:
Add the AllowWorkloadDisruption option to the HyperConverged LiveMigrationConfig API and pass it through to the KubeVirt MigrationConfiguration.

When enabled, the migration controller will not cancel migrations that exceed the acceptable completion time. Instead, it will switch to post-copy (if AllowPostCopy is true) or pause the VMI to allow the migration to complete.

Defaults to false, preserving existing behavior.

See: https://github.com/kubevirt/kubevirt/pull/11833

Adds the `allowWorkloadDisruption` field to the HyperConverged `liveMigrationConfig` API and passes it through to the KubeVirt `MigrationConfiguration`.

`AllowWorkloadDisruption` controls what happens when a live migration exceeds its acceptable completion time (determined by `completionTimeoutPerGiB` and the VM size):

**When disabled (default):**
- The migration is canceled when the completion timeout is reached.
- This is the current default behavior - no change for existing users.
- Note: `allowPostCopy` has no effect when `allowWorkloadDisruption` is explicitly set to `false`- post-copy will **not** trigger on timeout, even if `allowPostCopy: true`.

**When enabled:**
- The migration controller is allowed to disrupt the workload to force migration completion.
- The action taken depends on the `AllowPostCopy` setting:

| `allowWorkloadDisruption` | `allowPostCopy` | Behavior on timeout |
|---|---|---|
| `false` (default) | any | Migration is **aborted** (post-copy will **not** kick in) |
| `true` | `false` | VMI is **paused** until migration completes (no risk of data loss, but temporary workload inactivity) |
| `true` | `true` | Migration switches to **post-copy** mode (faster completion, but a network failure during post-copy can crash the VMI) |

**Backward compatibility:**

KubeVirt includes a backward-compatibility shim: when `allowWorkloadDisruption` is not set (nil), it defaults to the value of `allowPostCopy`. This means existing users with `allowPostCopy: true` will continue to get post-copy behavior without needing to explicitly set `allowWorkloadDisruption: true`. However, if `allowWorkloadDisruption` is explicitly set to `false`, it takes precedence and post-copy will not activate.

**Jira Ticket**:
```jira-ticket
https://redhat.atlassian.net/browse/CNV-81274
```

**Release Note**:

```release-note
Added `allowWorkloadDisruption` field to `liveMigrationConfig`. When enabled, the migration controller will force migration completion by either switching to post-copy mode (if `allowPostCopy` is true) or pausing the VMI, instead of aborting migrations that exceed the completion timeout.
```

Assisted-by: Claude [noreply@anthropic.com](mailto:noreply@anthropic.com)

### DIFF
--- a/api/v1/hyperconverged_types.go
+++ b/api/v1/hyperconverged_types.go
@@ -145,7 +145,7 @@ type HyperConvergedSpec struct {
 	FeatureGates featuregates.HyperConvergedFeatureGates `json:"featureGates,omitempty"`
 
 	// Virtualization contains all the configurations for virtualization
-	// +kubebuilder:default={"liveMigrationConfig": {"completionTimeoutPerGiB": 150, "parallelMigrationsPerCluster": 5, "parallelOutboundMigrationsPerNode": 2, "progressTimeout": 150, "allowAutoConverge": false, "allowPostCopy": false}, "virtualMachineOptions": {"disableFreePageReporting": false, "disableSerialConsoleLog": false}, "vmiCPUAllocationRatio": 10}
+	// +kubebuilder:default={"liveMigrationConfig": {"completionTimeoutPerGiB": 150, "parallelMigrationsPerCluster": 5, "parallelOutboundMigrationsPerNode": 2, "progressTimeout": 150, "allowAutoConverge": false, "allowPostCopy": false, "allowWorkloadDisruption": false}, "virtualMachineOptions": {"disableFreePageReporting": false, "disableSerialConsoleLog": false}, "vmiCPUAllocationRatio": 10}
 	// +kubebuilder:validation:XValidation:rule="!has(self.vmiCPUAllocationRatio) || self.vmiCPUAllocationRatio > 0",message="vmiCPUAllocationRatio must be greater than 0"
 	// +k8s:conversion-gen=false
 	Virtualization VirtualizationConfig `json:"virtualization,omitempty"`
@@ -187,7 +187,7 @@ type VirtualizationConfig struct {
 
 	// Live migration limits and timeouts are applied so that migration processes do not
 	// overwhelm the cluster.
-	// +kubebuilder:default={"completionTimeoutPerGiB": 150, "parallelMigrationsPerCluster": 5, "parallelOutboundMigrationsPerNode": 2, "progressTimeout": 150, "allowAutoConverge": false, "allowPostCopy": false}
+	// +kubebuilder:default={"completionTimeoutPerGiB": 150, "parallelMigrationsPerCluster": 5, "parallelOutboundMigrationsPerNode": 2, "progressTimeout": 150, "allowAutoConverge": false, "allowPostCopy": false, "allowWorkloadDisruption": false}
 	// +optional
 	LiveMigrationConfig LiveMigrationConfigurations `json:"liveMigrationConfig,omitempty"`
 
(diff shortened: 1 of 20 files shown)


api/v1/hyperconverged_types.go
nunnatsa: Please add the default value of the `allowWorkloadDisruption` filed here. The field default is not enough.

nunnatsa: Please add the default value of the `allowWorkloadDisruption` filed here. The field default is not enough.

controllers/handlers/kubevirt_test.go
```diff
@@ -805,6 +808,7 @@ Version: 1.2.3`)
 			hco.Spec.Virtualization.LiveMigrationConfig.Network = ptr.To(network)
 			hco.Spec.Virtualization.LiveMigrationConfig.AllowAutoConverge = ptr.To(true)
 			hco.Spec.Virtualization.LiveMigrationConfig.AllowPostCopy = ptr.To(true)
+			hco.Spec.Virtualization.LiveMigrationConfig.AllowWorkloadDisruption = ptr.To(true)
```
nunnatsa: `ptr.To` => `new`

controllers/handlers/kubevirt_test.go
```diff
@@ -4363,6 +4368,7 @@ Version: 1.2.3`)
 				Network:                           ptr.To(network),
 				AllowAutoConverge:                 ptr.To(true),
 				AllowPostCopy:                     ptr.To(true),
+				AllowWorkloadDisruption:           ptr.To(true),
```
nunnatsa: `ptr.To` => `new`

controllers/handlers/kubevirt_test.go
```diff
@@ -589,6 +591,7 @@ Version: 1.2.3`)
 			Expect(mc.Network).To(BeNil())
 			Expect(*mc.AllowAutoConverge).To(BeFalse())
 			Expect(*mc.AllowPostCopy).To(BeFalse())
+			Expect(*mc.AllowWorkloadDisruption).To(BeFalse())
```
nunnatsa: Please use the `HaveValue` matcher: i.e. `To(HaveValue(BeFalse())`

controllers/handlers/kubevirt_test.go
```diff
@@ -4402,6 +4410,7 @@ Version: 1.2.3`)
 				Network:                           ptr.To(network),
 				AllowAutoConverge:                 ptr.To(true),
 				AllowPostCopy:                     ptr.To(true),
+				AllowWorkloadDisruption:           ptr.To(true),
```
nunnatsa: `ptr.To` => `new`

nunnatsa (dismissed):
Thanks for the PR @SamAlber !

Added a few inline comments.

Some general comments:
1. Please run `make build-manifests` and add the changes in the deploy/ directory
2. Please run `make generate-docs` and commit the changes from the docs/ dir
3. Please add a human readable, user-guide style documentation to [docs/cluster-configuration.md](https://github.com/kubevirt/hyperconverged-cluster-operator/blob/main/docs/cluster-configuration.md#virtualization-configurations) + update the `Example` section.

nunnatsa (commented):
@SamAlber - also - the commit is not signed.

SamAlber:
Thanks for the review @nunnatsa! I addressed the review, please have another look.

nunnatsa (approved):
/approve

PR #4426, opened 2026-07-22
change migration defaults

michalskrivanek:

Optimize defaults for current kubevirt to improve migration convergence

Update completionTimeoutPerGiB from 150 to 20 and
parallelOutboundMigrationsPerNode from 2 to 1 in kubebuilder
annotations, upgrade patch CR, tests, and manually-maintained docs.
The new defaults apply to both new installs and upgrades (unless
the user has explicitly overridden the values).

The current values have multiple issues - completionTimeoutPerGiB is designed to be universal, but these days it's too conservative. That's not a bad thing on its own, but currently it drives the PostCopy or Paused migration decision too far out. With an assumed 500Mbps speed available the VM spends very little time for the initial copy (what the variable was supposed to guard) and very long time when it tries to converge with the (unfortunately) hardcoded 300ms downtime requirement. The suggested new default is more suitable for 500Mbps and higher speeds, allowing faster switch over and less transfer wasted, while still giving ample time for the initial pass.

As for the parallel migrations per node: we need as much bandwidth as possible to try to get lucky and hit the low activity moment in the VM and enough bandwidth to fit into the 300ms window to converge. When you have two migrations ongoing - typically on node drain - they both go on for ages because the PerGiB is too large and the downtime so low, so for a very long time there's half the bandwidth consumed by the other VM.
Overall it shouldn't matter in most cases since twice as much bandwidth means twice as fast migration of each VM (if you disregard the downtime aspect). Except the following corner cases where changing the default is advised:
- there's enough bandwidth for 2 migrations at maximum speeds. But with multifd on by default with 8 fds we would be maxing out single migration at ~50-75 Gbps which is ....not impossible, but quite uncommon
- the VM size and dirty rate are small, large number of VMs to evacuate. This is the worst case scenario where the per-migration k8s and kubrevirt  overhead dominates over a fast memory transfer. The overhead now doubles. In practice it seems to be around 10s per VM..This corresponds to the observed measurements: (25Gb network, 8GB VMs, idle):
  - 10 VMs.......10/2/1 parallel -> 13s/52s/110s
  - 50 VMs......20/2/1 parallel -> 35s/280s/570s

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://redhat.atlassian.net/browse/CNV-74882
https://redhat.atlassian.net/browse/CNV-72388
https://redhat.atlassian.net/browse/VIRTSTRAT-643
```

**Release note**:

```release-note
default migration timeout per GiB has been lowered to 20, i.e. maximum transfer time per 1 GiB of VM memory to be 20s. Faster timeout means earlier abort when VM doesn't converge, or a more timely switch to PostCopy or Paused handover to help migration convergence when enabled. 
To further help on congested networks the number of parallel outbound migrations have been lowered to 1, effectively migrating one VM at a time giving it full available bandwidth.
Expected minimum bandwidth requirement for the default values to work as intended is 500Mbps
```


api/v1beta1/hyperconverged_types.go
```diff
@@ -74,7 +74,7 @@ type HyperConvergedSpec struct {
 
 	// Live migration limits and timeouts are applied so that migration processes do not
 	// overwhelm the cluster.
-	// +kubebuilder:default={"completionTimeoutPerGiB": 150, "parallelMigrationsPerCluster": 5, "parallelOutboundMigrationsPerNode": 2, "progressTimeout": 150, "allowAutoConverge": false, "allowPostCopy": false}
+	// +kubebuilder:default={"completionTimeoutPerGiB": 20, "parallelMigrationsPerCluster": 5, "parallelOutboundMigrationsPerNode": 1, "progressTimeout": 150, "allowAutoConverge": false, "allowPostCopy": false}
```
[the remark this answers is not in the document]
michalskrivanek: v1beta1 and v1 share the same underlying type, so defaults must stay aligned. Conversion and fuzz tests use explicit non-default values and are unaffected

docs/cluster-configuration-v1beta1-api.md
```diff
@@ -383,10 +383,10 @@ metadata:
   name: kubevirt-hyperconverged
 spec:
   liveMigrationConfig:
-    completionTimeoutPerGiB: 150
+    completionTimeoutPerGiB: 20
     network: migration-network
     parallelMigrationsPerCluster: 5
-    parallelOutboundMigrationsPerNode: 2
+    parallelOutboundMigrationsPerNode: 1
```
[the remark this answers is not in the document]
michalskrivanek: I'd rather avoid documenting exact value as it is a mere suggestion. There are many more variables that influence convergence

nunnatsa:
Thanks for the PR @michalskrivanek 

Please fill the PR description. It is now empty.

The current defaults follow KubeVirt default values. Are we sure we want to modify them?

Several additional comments: 
* Please run `make build-manifests` and add the changes. Please avoid modifying auto generated files.
* Please run `make generate-docs` and add the changes.
* Please fix the "defaults" functional test

nunnatsa:
@iholder101 - are the new default values make sense to you?

iholder101:
> @iholder101 - are the new default values make sense to you?

Hey!

Two thoughts:
* I don't see any backing data / rationale / motivation for this change. Are we sure it's the right choice? What is it based on?
* If these are better defaults - why not change them at the Kubevirt level?

michalskrivanek:
> > @iholder101 - are the new default values make sense to you?
> 
> Hey!
> 
> Two thoughts:
> 
>     * I don't see any backing data / rationale / motivation for this change. Are we sure it's the right choice? What is it based on?

it's in the linked tickets

>     * If these are better defaults - why not change them at the Kubevirt level?

they don't necessarily make sense for non-opinionated use case.

iholder101:
As said above, I think this PR lacks details.
The two questions above still remain unanswered. Let me try to emphasize what I'm missing.

> > > @iholder101 - are the new default values make sense to you?
> > 
> > 
> > Hey!
> > Two thoughts:
> > ```
> > * I don't see any backing data / rationale / motivation for this change. Are we sure it's the right choice? What is it based on?
> > ```
> 
> it's in the linked tickets

Would be nice to summarize the reasoning in the PR as well, especially since these are Red Hat tickets.

w.r.t. reasoning that is attached to this PR:
```
To further help on congested networks the number of parallel outbound
migrations have been lowered to 1, effectively migrating one VM at a time
giving it full available bandwidth.
```

What if the bandwidth is high enough to migrate two VMs at once? Sounds like a pretty specific use-case from a glance.

> > ```
> > * If these are better defaults - why not change them at the Kubevirt level?
> > ```
> 
> they don't necessarily make sense for non-opinionated use case.

Do they make sense for ALL cases in which Kubevirt is deployed with HCO, but NEVER to non-HCO deployments? This is a strong statement that needs some rationale in order to be convincing. There's a real benefit of HCO and Kubevirt using the same defaults.

Was it considered for these specific users to change configurations?

michalskrivanek:
> As said above, I think this PR lacks details. The two questions above still remain unanswered. Let me try to emphasize what I'm missing.
> 
> > > > @iholder101 - are the new default values make sense to you?
> > > 
> > > 
> > > Hey!
> > > Two thoughts:
> > > ```
> > > * I don't see any backing data / rationale / motivation for this change. Are we sure it's the right choice? What is it based on?
> > > ```
> > 
> > 
> > it's in the linked tickets
> 
> Would be nice to summarize the reasoning in the PR as well, especially since these are Red Hat tickets.

what makes them special? AFAICT they're visible all right
 
> w.r.t. reasoning that is attached to this PR:
> 
> ```
> To further help on congested networks the number of parallel outbound
> migrations have been lowered to 1, effectively migrating one VM at a time
> giving it full available bandwidth.
> ```
> 
> What if the bandwidth is high enough to migrate two VMs at once? Sounds like a pretty specific use-case from a glance.

It's possible, however far less likely than the opposite - hence the change. Even with a 25Gbps network a single regular VMs on a commodity HW saturates it good enough with multifd, at around 7Gbps per fd. The typical deployment is using pod network and that further limits the capacity.
Even if he capacity is higher (e.g. in multifd disabled case) the overall drain time is going to be longer, but it will _not_ be more likely to succeed.


> > > ```
> > > * If these are better defaults - why not change them at the Kubevirt level?
> > > ```
> > 
> > 
> > they don't necessarily make sense for non-opinionated use case.
> 
> Do they make sense for ALL cases in which Kubevirt is deployed with HCO, but NEVER to non-HCO deployments? This is a strong statement that needs some rationale in order to be convincing. There's a real benefit of HCO and Kubevirt using the same defaults.

The observed behavior that motivates the change is based on HCO-based deployment of OCP-V. For these it makes sense for MOST cases, i.e. they make sense as a default.
I cannot conduct tests on other HCO deployments, if any. I have no reason to believe it would behave differently unless such configuration is targeting <500Mbps networks.
As for Kubevirt defaults, I agree it would be desirable, however I do not have enough data for that. AFAICT the last change there doesn't have much backing data either. I do think the right direction is to push the live migration enhancements and get rid of this parameter, the stall detection work makes first steps towards that.



> Was it considered for these specific users to change configurations?

the goal is not to address specific configurations but the "default" or most prevalent configurations, and increase the success rate of node drains and migration's success in general.

jean-edouard:
@vladikr FYI

jean-edouard (approved):

michalskrivanek:
> Would be nice to summarize the reasoning in the PR as well, especially since these are Red Hat tickets.

updated PR description

michalskrivanek:
The current PR changes the defaults in annotations and hc.cr.json, but when upgrading from 1.18 (or earlier), the existing CR will already have completionTimeoutPerGiB: 150 and parallelOutboundMigrationsPerNode: 2 explicitly set. The CRD defaults won't override explicit values. We need upgradePatches.json entries with test+replace to conditionally update them — only if the user hasn't customized them.
We can further limit the upgrade to handle `completionTimeoutPerGiB` as it is highly unlikely to be a concious choice, whereas `parallelOutboundMigrationsPerNode` could be a valid customization (even if it's the previous default. It's prominently mentioned and e.g. displayed in https://github.com/kubevirt-ui/kubevirt-plugin for users to tweak)

api/v1/hyperconverged_types.go
```diff
@@ -567,8 +567,8 @@ type LiveMigrationConfigurations struct {
 
 	// Maximum number of outbound migrations per node.
```
tiraboschi: I'd suggest to expand the GoDoc section (It pops up in `oc explain` and the UI) explaining the rationale:

```
// ParallelOutboundMigrationsPerNode controls the maximum number of outbound 
// migrations allowed simultaneously per node.
// 
// Setting this to 1 (default) ensures maximum available bandwidth per migration. 
// Higher values accelerate node drains through parallel operations, but increase 
// the risk of busy VMs failing due to resource competition.
```

michalskrivanek: done

nunnatsa:
@michalskrivanek - please run `make build-manifests` and `make generate-docs` to make sure all the auto generated files are up-to-date with the changes in this PR. currently the sanity test fails for that.

docs/cluster-configuration-v1beta1-api.md
nunnatsa: Please fix the default value.

nunnatsa, line 357: Please update the description, to explain the logic and the risks, as done in the comment in v1.

nunnatsa: please update the default value

docs/cluster-configuration.md
nunnatsa: Please update the default value

nunnatsa: Please update the description, to explain the logic and the risks, as done in the comment in v1.

nunnatsa: Please update the default value

assets/upgradePatches.json
```diff
@@ -1,5 +1,21 @@
 {
-  "hcoCRPatchList": [],
+  "hcoCRPatchList": [
+    {
+      "semverRange": ">=1.18.0 <1.20.0",
```
nunnatsa: I would drop the higher boundry

controllers/hyperconverged/test-files/upgradePatches/upgradePatches.json
```diff
@@ -14,6 +14,21 @@
           "value": true
         }
       ]
+    },
+    {
+      "semverRange": ">=1.18.0 <1.20.0",
```
nunnatsa: same

assets/upgradePatches.json
nunnatsa: Also, please add unit test for that in `controllers/hyperconverged/upgrade_test.go`

nunnatsa (commented):
some more comments. Not sure why didn't they sent yesterday...

docs/cluster-configuration-v1beta1-api.md
michalskrivanek: AFAICT they are the same (speaking about `completionTimeoutPerGiB`, `progressTimeout` is not being touched)

docs/cluster-configuration.md
michalskrivanek: they're both the same (more or less empty) ,I'll add something to both...

controllers/hyperconverged/upgrade_test.go
```diff
@@ -601,6 +601,32 @@ var _ = Describe("Upgrade Mode", func() {
...
+			UpdateVersion(&expected.hco.Status, hcoVersionName, "1.18.5")
+			expected.hco.Spec.Virtualization.LiveMigrationConfig.CompletionTimeoutPerGiB = ptr.To[int64](800)
+
+			cl := expected.initClient()
+			_, reconciler, requeue := doReconcile(cl, expected.hco, nil)
+			Expect(requeue).To(BeTrue())
+			foundResource, _, requeue := doReconcile(cl, expected.hco, reconciler)
+			Expect(requeue).To(BeFalse())
+			Expect(foundResource.Spec.Virtualization.LiveMigrationConfig.CompletionTimeoutPerGiB).To(HaveValue(BeEquivalentTo(800)))
+		})
```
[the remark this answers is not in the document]
michalskrivanek: that's the intended behavior

assets/upgradePatches.json
```diff
@@ -1,5 +1,21 @@
 {
-  "hcoCRPatchList": [],
+  "hcoCRPatchList": [
+    {
+      "semverRange": ">=1.18.0",
```
nunnatsa: should be `<1.19.0`

controllers/hyperconverged/test-files/upgradePatches/upgradePatches.json
```diff
@@ -14,6 +14,21 @@
           "value": true
         }
       ]
+    },
+    {
+      "semverRange": ">=1.18.0",
```
nunnatsa: same

nunnatsa (commented):
Looks very good. last 2 comments (I hope) inline

michalskrivanek:
@nunnatsa thanks a lot for the review! let me know what else is missing

PR #4427, opened 2026-07-22
[release-1.16] Allow downloading of the virtio-win iso image from the artifacts server

nunnatsa:

*This is a manual cherry pick of #4424*

**Jira Ticket**:

```jira-ticket
None
```

**Release note**:

```release-note
```


orenc1 (approved):
/approve

PR #4421, opened 2026-07-21
handlers, netresinjector: Gate ExternalNetResourceInjection FG on NRI deployment readiness status

nirdothan:
**What this PR does / why we need it**:
Enable the ExternalNetResourceInjection KubeVirt feature gate when
deployNetworkResourcesInjector is set to true in the HCO CR, so the
injector deployment and its corresponding FG are controlled by the
same knob.

Use a NetworkResourcesInjectorReady status condition on the HCO CR
to signal readiness, with reasons DeploymentReady, DeploymentNotReady,
DeploymentNotFound, and DeploymentDisabled.

**Reviewer Checklist**

> Reviewers are supposed to review the PR for every aspect below one by one. To check an item means the PR is either "OK" or "Not Applicable" in terms of that item. All items are supposed to be checked before merging a PR. 

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility
- [ ] Troubleshooting Friendly

**Jira Ticket**:

```jira-ticket
https://redhat.atlassian.net/browse/CNV-93112
```

**Release note**:

```release-note
New CR condition indicating network-resources-injector deployment status
```

PR #4429, opened 2026-07-22
netresinjector: Re-enable deployment by default

nirdothan:
Reverts [1] which temporarily disabled network-resources-injector deployment by default. The underlying bug has been fixed in [2].

[1] https://github.com/kubevirt/hyperconverged-cluster-operator/pull/4377
[2] https://github.com/k8snetworkplumbingwg/network-resources-injector/pull/224
**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://redhat.atlassian.net/browse/CNV-93111
```

**Release note**:

```release-note
NONE
```

### DIFF
--- a/api/v1/hyperconverged_types.go
+++ b/api/v1/hyperconverged_types.go
@@ -169,7 +169,7 @@ type HyperConvergedSpec struct {
 	Security SecurityConfig `json:"security,omitempty"`
 
 	// Deployment contains all the configurations related to deployment of KubeVirt components
-	// +kubebuilder:default={"uninstallStrategy": "BlockUninstallIfWorkloadsExist", "deployVmConsoleProxy": false, "applicationAwareConfig": {"enable": false}}
+	// +kubebuilder:default={"uninstallStrategy": "BlockUninstallIfWorkloadsExist", "deployVmConsoleProxy": false, "deployNetworkResourcesInjector": true, "applicationAwareConfig": {"enable": false}}
 	// +optional
 	// +k8s:conversion-gen=false
 	Deployment DeploymentConfig `json:"deployment,omitempty"`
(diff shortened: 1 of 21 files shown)


orelmisan (commented):
Thanks @nirdothan.

I think we should demonstrate that the upgrade scenario works with the `ExternalNetResourceInjection` feature gate enabled before deploying the `network-resources-injector` by default.

Without that feature gate enabled, we would be adding a mutating webhook for all virt-launcher pods without actually taking advantage of its capabilities.

nunnatsa:
> Thanks @nirdothan.
> 
> I think we should demonstrate that the upgrade scenario works with the `ExternalNetResourceInjection` feature gate enabled before deploying the `network-resources-injector` by default.
> 
> Without that feature gate enabled, we would be adding a mutating webhook for all virt-launcher pods without actually taking advantage of its capabilities.

@orelmisan this functionality wasn't implemented in HCO, and so the FG is was never set.

nirdothan:
Covered by https://github.com/kubevirt/hyperconverged-cluster-operator/pull/4429

orelmisan:
> Covered by #4429

#4429 is this PR. Do you mean #4421?

nirdothan:
> > Covered by #4429
> 
> #4429 is this PR. Do you mean #4421?

Of course... clipboard... 
Yes it is covered by #4421

PR #4440, opened 2026-07-24
Migrate TLS profile check from shell script to Go functest

### LINKED ISSUE #4406: Migrate TLS profile check from shell/nmap to Go functest
**Is your feature request related to a problem? Please describe**:

The TLS security profile verification currently runs as a bash script (`hack/check_tlsprofile.sh`) that installs `nmap` at runtime inside the functest container. This has several problems:

1. The script has an x86_64-hardcoded RPM fallback (`nmap-7.94-1.x86_64.rpm`) that breaks on non-x86 architectures, which conflicts with ongoing multi-arch efforts.
2. It relies on 8 expected output text files (`hack/tlsprofiles/*.expected` and `*.expected.fips`) that need manual updates whenever cipher suites change.
3. It runs outside the Ginkgo test suite (invoked as a post-suite step from `hack/run-tests.sh`), so results don't appear in JUnit XML reports and failures are harder to triage.
4. It uses `sleep 2` for config propagation, which is fragile and can cause intermittent failures on slower clusters.

This was discussed in https://github.com/kubevirt/hyperconverged-cluster-operator/pull/4337#discussion_r3489569419

**Describe the solution you'd like**:

Replace `hack/check_tlsprofile.sh` with a native Go Ginkgo test that uses `crypto/tls` to probe the webhook endpoint directly. The test would:

- Patch the HCO CR to each TLS profile (Old, Intermediate, Modern, Custom) and verify cipher enforcement using `tls.DialWithDialer` handshakes instead of nmap.
- Derive expected ciphers from `openshiftconfigv1.TLSProfiles` at runtime instead of diffing text files.
- Handle FIPS automatically, since Go's crypto restricts ciphers in FIPS builds without needing separate expected files.
- Use `Eventually` polling instead of `sleep` for reliable config propagation detection.
- Run as part of the Ginkgo suite with a label filter, so results appear in JUnit XML reports.

All 7 existing checks would be preserved with identical coverage. No new test scenarios.

The expected file changes would be:

| Action | File |
|--------|------|
| Add | `tests/func-tests/tls_profile_test.go` (E2E test, 7 cases + helpers) |
| Add | `tests/func-tests/tls_enum_test.go` (unit tests for cipher enumeration) |
| Modify | `hack/run-tests.sh` (remove `check_tlsprofile.sh` invocation) |
| Delete | `hack/check_tlsprofile.sh` |
| Delete | `hack/tlsprofiles/*.expected` and `*.expected.fips` (8 files) |

**Describe alternatives you've considered**:

1. **Fix the nmap script for multi-arch**: Compute the RPM architecture from `uname -m` in the fallback. This would solve the arch issue but keeps the nmap runtime dependency, the fragile text-diffing, and the out-of-suite execution. Since nmap.org only publishes x86_64 RPMs, this path is limited.

2. **Keep the shell script but install nmap via package manager only**: Remove the RPM fallback entirely and rely on `microdnf`/`dnf`/`yum`. This works for current UBI9-based containers but still carries all the other downsides (text diffing, no JUnit, sleep-based timing).

3. **Use a standalone Go binary instead of a Ginkgo test**: Write a Go CLI tool that replaces the shell script, still invoked from `hack/run-tests.sh`. This would eliminate nmap but wouldn't integrate into the Ginkgo suite and keeps the post-suite invocation pattern.

The Ginkgo integration approach is preferred because it aligns with how all other functest checks work and gives the best reporting and reliability.

**Additional context**:

- The existing `openshiftconfigv1.TLSProfiles` map and `openshift/library-go/pkg/crypto` package already provide the cipher definitions and OpenSSL-to-IANA name conversion needed, so no new dependencies would be required.
- The webhook already uses `GetConfigForClient` in `MutateTLSConfig` to apply TLS settings dynamically per-connection, so no restart is needed after patching -- `Eventually` polling would pick up the change reliably.
- Happy to submit a PR for this if the approach looks good.

carterpewpew:
## Summary

Replaces `hack/check_tlsprofile.sh` with a native Go Ginkgo test (`tls_profile_test.go`) that uses `crypto/tls` to probe the webhook server directly.

- **Eliminates nmap dependency**: The shell script installed `nmap` at runtime and had x86_64-only RPM fallbacks, making it fail on s390x/ppc64le. The Go test uses `crypto/tls` for direct TLS handshake probing, working on all architectures.
- **Removes static expected files**: The 8 `hack/tlsprofiles/*.expected{,.fips}` files are replaced by programmatic derivation from `openshiftconfigv1.TLSProfiles` and `library-go/pkg/crypto`, so tests stay correct when upstream profile definitions change.
- **Handles FIPS implicitly**: Go's `crypto/tls` restricts cipher suites automatically under FIPS mode, eliminating the need for separate `.expected.fips` files.
- **Improves reliability**: Fixed `sleep` delays in the shell script are replaced with Gomega `Eventually` polling for webhook config propagation.
- **Integrates with Ginkgo**: TLS verification now appears in JUnit reports and benefits from the Ginkgo test framework (`--ginkgo.label-filter="tls"`).

## Changes

| File | Action |
|------|--------|
| `tests/func-tests/tls_profile_test.go` | New Ginkgo E2E test (7 specs: Old, dry-run, Intermediate, Modern, Custom rejection, Custom enforcement, removal default) |
| `tests/func-tests/tls_enum_test.go` | New unit tests for TLS helper functions using local TLS server |
| `hack/run-tests.sh` | Remove `check_tlsprofile.sh` invocation |
| `hack/check_tlsprofile.sh` | Deleted |
| `hack/tlsprofiles/*` | Deleted (8 files) |

## Test plan

- [x] `go vet ./tests/func-tests/` passes
- [x] `go test -run "^TestTLS" ./tests/func-tests/` - all 14 unit tests pass
- [x] `ginkgo build ./tests/func-tests/` compiles successfully
- [x] E2E tested against live OpenShift cluster - all 7 specs pass (Old, dry-run, Intermediate, Modern, Custom rejection, Custom enforcement, removal default)

Fixes: https://github.com/kubevirt/hyperconverged-cluster-operator/issues/4406

```release-note
Migrate TLS profile functional test from shell/nmap to native Go
```

This PR was created with the help of AI, but every change has been manually verified and tested


tests/func-tests/tls_profile_test.go
```diff
@@ -0,0 +1,402 @@
...
+
+		for _, c := range deploy.Status.Conditions {
+			if c.Type == appsv1.DeploymentAvailable {
+				g.Expect(c.Status).To(Equal(corev1.ConditionTrue),
+					"ssp-operator deployment is not available")
+				return
+			}
+		}
+		Fail("ssp-operator deployment missing Available condition")
+	}).WithTimeout(sspReadyTimeout).WithPolling(sspReadyPolling).WithContext(ctx).Should(Succeed())
```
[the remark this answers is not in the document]
carterpewpew: Addressed on the latest commit

tests/func-tests/tls_profile_test.go
```diff
@@ -0,0 +1,402 @@
...
+		return net.JoinHostPort(pod.Status.PodIP, strconv.Itoa(webhookPort)), func() {}, nil
+	}
+
+	serviceName, err := getWebhookServiceName(ctx, k8sCli, namespace)
+	if err != nil {
+		return "", nil, err
+	}
+
+	return startPortForward(namespace, serviceName)
+}
```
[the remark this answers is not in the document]
carterpewpew: Addressed

tests/func-tests/tls_profile_test.go
```diff
@@ -0,0 +1,402 @@
...
+	}).WithTimeout(30 * time.Second).WithPolling(time.Second).Should(Succeed(),
+		"port-forward to %s did not become ready", serviceName)
+
+	return addr, func() {
+		if cmd.Process != nil {
+			_ = cmd.Process.Kill()
+			_ = cmd.Wait()
+		}
+	}, nil
+}
```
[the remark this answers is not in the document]
carterpewpew: Addressed

nunnatsa (changes requested):
Thanks for the PR @carterpewpew 
This id a great improvement. 

I'll take a deeper look next week, but one big general comment: we are using the gomega and ginkgo packages for tests in this project, and not pure go. Please convert the test accordingly.

Puneethl56 (approved):

tests/func-tests/tls_profile_test.go
```diff
@@ -0,0 +1,428 @@
...
+	case openshiftconfigv1.VersionTLS11:
+		return tls.VersionTLS11
+	case openshiftconfigv1.VersionTLS12:
+		return tls.VersionTLS12
+	case openshiftconfigv1.VersionTLS13:
+		return tls.VersionTLS13
+	default:
+		return tls.VersionTLS12
+	}
+}
```
nunnatsa: `crypto.TLSVersion` does the same

tests/func-tests/tls_enum_test.go
```diff
@@ -0,0 +1,131 @@
+package tests_test
```
nunnatsa: Why do we need this file at all? is it a unit test to the functional test? We can't run it as unit test. Anyway, it does not test the HCO code, and just adds complexity.

tests/func-tests/tls_profile_test.go
```diff
@@ -0,0 +1,428 @@
...
+		supportsVersion := false
+		for _, v := range suite.SupportedVersions {
+			if v == tlsVersion {
+				supportsVersion = true
+				break
+			}
+		}
+		if !supportsVersion {
+			continue
+		}
```
nunnatsa: WDYT about this option?

```suggestion
		if !slices.Contains(suite.SupportedVersions, tlsVersion) {
			continue
		}
```

tests/func-tests/tls_profile_test.go
```diff
@@ -0,0 +1,428 @@
...
+		allowedSet := make(map[uint16]bool, len(allowedTLS12Ciphers))
+		for _, c := range allowedTLS12Ciphers {
+			allowedSet[c] = true
+		}
+
+		for _, c := range serverCiphers {
+			g.Expect(allowedSet[c]).To(BeTrue(),
+				"server accepts cipher %s (0x%04x) which is not in the allowed set for this profile",
+				tls.CipherSuiteName(c), c)
+		}
```
nunnatsa: nit: bool maps are usually not needed, and this case is not an exception. The bet practice is to use an empty struct (size == 0 bytes) as a value. WDYT about something like this?

```suggestion
		allowedSet := make(map[uint16]struct{}, len(allowedTLS12Ciphers))
		for _, c := range allowedTLS12Ciphers {
			allowedSet[c] = struct{}{}
		}

		for _, c := range serverCiphers {
			g.Expect(allowedSet).To(HaveKey(c)),
				"server accepts cipher %s (0x%04x) which is not in the allowed set for this profile",
				tls.CipherSuiteName(c), c)
		}
```

tests/func-tests/tls_profile_test.go
```diff
@@ -0,0 +1,428 @@
...
+	}()
+
+	Eventually(func() error {
+		conn, dialErr := net.DialTimeout("tcp", addr, time.Second)
+		if dialErr != nil {
+			return dialErr
+		}
+		conn.Close()
+		return nil
+	}).WithTimeout(30 * time.Second).WithPolling(time.Second).Should(Succeed(),
```
nunnatsa: polling and timeout are the same. I would make polling to be 2 sec, just for safe.

nunnatsa (commented):
Thanks.

Added a few comments inline.

Some general comment:

if possible (e.g. when not using `Eventually`), please prefer returning regular errors from helper functions, over using gomega matchers to for non-validation logic, then validate these errors from the test.

Also, please move the test itself, the main dish, to the beginning of the file. It's a bit hard to understand.

Last thing that I also commented inline: the purpose of the tls_enum_test.go file is not clear to me. Please explain why is it good for.

tests/func-tests/tls_profile_test.go
```diff
@@ -0,0 +1,428 @@
...
+		found := false
+		for _, c := range deploy.Status.Conditions {
+			if c.Type == appsv1.DeploymentAvailable {
+				g.Expect(c.Status).To(Equal(corev1.ConditionTrue),
+					"ssp-operator deployment is not available")
+				found = true
+				break
+			}
+		}
+		g.Expect(found).To(BeTrue(), "ssp-operator deployment missing Available condition")
```
nunnatsa: Can be simplified:

```suggestion
		available := slices.ContainsFunc(deploy.Status.Conditions, func(condition appsv1.DeploymentCondition) bool {
		return condition.Type == appsv1.DeploymentAvailable && condition.Status == corev1.ConditionTrue
	})

	g.Expect(available).To(BeTrue(), "ssp-operator deployment missing Available condition")
```

tests/func-tests/tls_profile_test.go
```diff
@@ -0,0 +1,428 @@
...
+		conn.Close()
+		return nil
+	}).WithTimeout(30 * time.Second).WithPolling(time.Second).Should(Succeed(),
+		"port-forward to %s did not become ready", serviceName)
+
+	ready = true
+	return addr, cleanup, nil
+}
+
+func waitForSSPReady(ctx context.Context, k8sCli *kubernetes.Clientset) {
```
nunnatsa: Call this function from the `AfterEach` function, instead of calling it in every test.

tests/func-tests/tls_profile_test.go
```diff
@@ -0,0 +1,428 @@
...
+	GinkgoHelper()
+
+	_, err := k8sCli.AppsV1().Deployments(tests.InstallNamespace).Get(ctx, "ssp-operator", metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		return
+	}
+	Expect(err).NotTo(HaveOccurred())
```
nunnatsa: This duplicates the code below. We can skip this part if SSP is not deployed.

I would drop this code, and call this function from the `AfterEach` function, only when running on OpenShift. Something like this:
```
AfterEach(func(ctx context.Context) {
	if Label(tests.OpenshiftLabel).MatchesLabelFilter(GinkgoLabelFilter()) {
		waitForSSPReady(ctx, k8sCli)
	}
})
```

tests/func-tests/tls_enum_test.go
carterpewpew: Removed. We originally added it to test the cipher enumeration helpers in isolation with a local TLS server, but you're right -- they're already exercised through the E2E specs and it doesn't test any HCO code.

tests/func-tests/tls_profile_test.go
carterpewpew: Done, much cleaner.

tests/func-tests/tls_profile_test.go
carterpewpew: Good catch, switched to `crypto.TLSVersionOrDie`. Didn't realize library-go already had this.

tests/func-tests/tls_profile_test.go
carterpewpew: Switched, thanks.

tests/func-tests/tls_profile_test.go
carterpewpew: Bumped to 2 seconds.

tests/func-tests/tls_profile_test.go
carterpewpew: Done, nice simplification.

tests/func-tests/tls_profile_test.go
carterpewpew: Moved SSP readiness to `AfterEach`. We kept it unconditional since `waitForSSPReady` already does an early return if the ssp-operator deployment doesn't exist, so it's a no-op on non-OpenShift clusters without needing the label check.

tests/func-tests/tls_profile_test.go
carterpewpew: Moved to `AfterEach`, removed the individual calls from each spec.

carterpewpew:
@nunnatsa Thanks for the thorough review! Addressed all comments on the latest push:

- Removed `goTLSVersion`, using `crypto.TLSVersionOrDie` from library-go
- Deleted `tls_enum_test.go`
- `slices.Contains` for version check, `slices.ContainsFunc` for SSP condition
- `map[uint16]struct{}` + `HaveKey` for cipher set
- Polling bumped to 2 seconds
- `Describe` block moved to the top of the file
- `waitForSSPReady` moved to `AfterEach` (kept unconditional since it already no-ops when ssp-operator isn't deployed)
- Helper functions already return errors; remaining `Expect` is only in validation logic

All 7 specs passing against a live cluster.

tests/func-tests/tls_profile_test.go
nunnatsa: This early return is not the best check.

SSP must run on OpenShift. Let's do this instead:
```
	isOpenShift, err := tests.IsOpenShift(ctx, cli)

	if isOpenShift {
		return
	}
```

nunnatsa:
@carterpewpew thanks for the fixes.

I reply to your comment here: https://github.com/kubevirt/hyperconverged-cluster-operator/pull/4440#discussion_r3664370714

tests/func-tests/tls_profile_test.go
carterpewpew: Done. Used `if !isOpenShift { return }` to skip the SSP readiness wait on non-OpenShift clusters where SSP won't be deployed. Verified all 7 TLS tests passing on OpenShift.

tests/func-tests/tls_profile_test.go
```diff
@@ -0,0 +1,393 @@
...
+		Eventually(func() error {
+			conn, dialErr := net.DialTimeout("tcp", addr, time.Second)
+			if dialErr != nil {
+				return dialErr
+			}
+			conn.Close()
+			return nil
+		}).WithTimeout(30 * time.Second).WithPolling(2 * time.Second).Should(Succeed(),
+			"webhook at %s did not become ready", addr)
+		return addr, func() {}, nil
```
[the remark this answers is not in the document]
carterpewpew: The existing functest codebase uses `GinkgoHelper()` + `Eventually` inside helpers throughout (e.g. `validateAIEDeleted`, `validateNetResInjectorDeleted`, `getEtcdPod`). The maintainer's guidance was to prefer returning errors "when not using `Eventually`". Keeping `Eventually` here is consistent with the project's established pattern. Added `GinkgoHelper()` to both functions for proper line attribution.

tests/func-tests/tls_profile_test.go
```diff
@@ -0,0 +1,395 @@
...
+			MinVersion:         version,
+			MaxVersion:         version,
+		},
+	)
+	if err != nil {
+		return false
+	}
+	conn.Close()
+	return true
+}
```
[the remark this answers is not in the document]
nunnatsa: @carterpewpew - the bot has a point here... Please address.

nunnatsa:
@carterpewpew - the sanity test (also) runs `make goimport`. The goimports tool also reformat (gofmt) all the files. Please run `make goimport` locally, so the sanity test will pass.

tests/func-tests/tls_profile_test.go
```diff
@@ -0,0 +1,395 @@
...
+		allowedSet := make(map[uint16]struct{}, len(allowedTLS12Ciphers))
+		for _, c := range allowedTLS12Ciphers {
+			allowedSet[c] = struct{}{}
+		}
+
+		for _, c := range serverCiphers {
+			g.Expect(allowedSet).To(HaveKey(c),
+				"server accepts cipher %s (0x%04x) which is not in the allowed set for this profile",
+				tls.CipherSuiteName(c), c)
+		}
```
tiraboschi, line 265: This is checking just in one direction: every cipher the server accepts is in the allowed set (no unauthorized ciphers), but it doesn't check the reverse (every cipher in the allowed set is actually accepted by the server).
What about:
  `g.Expect(serverCiphers).To(ConsistOf(allowedTLS12Ciphers), "server ciphers should match the profile exactly")`
?

tiraboschi (changes requested):
The advantages of a pure Go implementation are evident (multi-arch support, no external dependencies, programmatic expected values, and proper Ginkgo integration). However, we are completely losing visibility on the elliptic curves negotiated during the TLS handshake, which nmap's ssl-enum-ciphers was reporting (e.g., `secp256r1`, `ecdh_x25519`), while Go's crypto/tls public API does not provide a straightforward way to enumerate server-supported curves. This gap becomes particularly concerning as the platform is working on PQC adoption, where the key exchange groups are the primary surface that changes (e.g., `hybrid X25519+ML-KEM-768`). Accepting this without a plan to restore group/curve verification could leave us blind to PQC compliance issues in the TLS profiles.

tiraboschi:
Did we investigate the opportunity of replacing the nmap based scan with Openshift TLS compliance scanner ( https://github.com/openshift/tls-scanner ).
That tool should also enforce post-quantum (ML-KEM) readiness checks.
On the other side it's just another external tool and not a pure golang implementation.

carterpewpew:
Hey @tiraboschi, good question. We did look into tls-scanner but it would bring back the same external binary problem we're trying to get rid of here.

On the curve/key exchange visibility I dug into this and it turns out we're not losing any validation there. The `TLSProfileSpec` only configures `Ciphers` and `MinTLSVersion` there's no curve/group field in the API. And HCO's `MutateTLSConfig` in `pkg/internal/tlssecprofile/tls_security_profile.go` only sets `config.MinVersion` and `config.CipherSuites`, never `CurvePreferences`. So the curves nmap was reporting (`secp256r1`, `ecdh_x25519`) are just Go's hardcoded defaults they show up identically in every profile's expected file (Old, Intermediate, Modern) which confirms no profile is actually controlling them.

For PQC/ML-KEM - Go 1.26 already includes `X25519MLKEM768`, `SecP256r1MLKEM768`, and `SecP384r1MLKEM1024` in its default CurvePreferences (see [src/crypto/tls/defaults.go](https://github.com/golang/go/blob/2403e594/src/crypto/tls/defaults.go)). Since HCO never overrides `CurvePreferences`, the webhook already accepts all three PQ hybrid key exchanges out of the box I have verified this on a live cluster. If the platform ever needs to restrict which groups are offered, that would need a new field in `TLSProfileSpec`, and new test assertions to go with it regardless of tool choice.

tiraboschi:
> Hey @tiraboschi, good question. We did look into tls-scanner but it would bring back the same external binary problem we're trying to get rid of here.

I know, on the other side I'm not so sure we can easily have a pure golang implementation covering also curves/groups.

> On the curve/key exchange visibility I dug into this and it turns out we're not losing any validation there. The `TLSProfileSpec` only configures `Ciphers` and `MinTLSVersion` there's no curve/group field in the API. 

`TLSSecurityProfile` on HCO API is just `*openshiftconfigv1.TLSSecurityProfile`  from Openshift API and https://github.com/openshift/api/pull/2583  got merged there about 2 months ago adding a Groups field to TLSProfileSpec for configuring TLS key exchange groups (including X25519MLKEM768 for PQC). So on the next bump it will be here as well.

tiraboschi:
Please see also: https://github.com/kubevirt/hyperconverged-cluster-operator/pull/4464

carterpewpew:
@nunnatsa @tiraboschi - would you be okay with landing this as-is and tracking the groups/PQC work as a follow-up? Or would you prefer we hold this until the vendor bump and groups support are in place? Thanks again for all the feedback on this PR

tiraboschi:
> @nunnatsa @tiraboschi - would you be okay with landing this as-is and tracking the groups/PQC work as a follow-up? Or would you prefer we hold this until the vendor bump and groups support are in place? Thanks again for all the feedback on this PR

I explored integrating openshift/tls-scanner as an alternative since it's the platform tool for OpenShift TLS auditing, wrapping testssl.sh, with built-in PQC readiness checking (-pqc-check for ML-KEM).
It's also a golang tool, however, all its Go packages are under internal/, so it can only be used as an external binary (or partially forked here).
To avoid network dependencies at build time we'd need a Go tool directive + a git submodule for testssl.sh (~3MB). Doable but probably overkill for scanning a single webhook endpoint.

Your crypto/tls approach sounds as a better call here.
On the other side I'm still a bit concerned that it could be a dead path once we will have to cover PQC readiness.
Could you please double check what we can really cover trying handshakes with tls.X25519MLKEM768 (available since Go 1.24)? If a single TLS handshake attempt with that curve preference would verify whether the server negotiates only ML-KEM without the need for external tools this will be enough in my opinion.

carterpewpew:
> > @nunnatsa @tiraboschi - would you be okay with landing this as-is and tracking the groups/PQC work as a follow-up? Or would you prefer we hold this until the vendor bump and groups support are in place? Thanks again for all the feedback on this PR
> 
> I explored integrating openshift/tls-scanner as an alternative since it's the platform tool for OpenShift TLS auditing, wrapping testssl.sh, with built-in PQC readiness checking (-pqc-check for ML-KEM). It's also a golang tool, however, all its Go packages are under internal/, so it can only be used as an external binary (or partially forked here). To avoid network dependencies at build time we'd need a Go tool directive + a git submodule for testssl.sh (~3MB). Doable but probably overkill for scanning a single webhook endpoint.
> 
> Your crypto/tls approach sounds as a better call here. On the other side I'm still a bit concerned that it could be a dead path once we will have to cover PQC readiness. Could you please double check what we can really cover trying handshakes with tls.X25519MLKEM768 (available since Go 1.24)? If a single TLS handshake attempt with that curve preference would verify whether the server negotiates only ML-KEM without the need for external tools this will be enough in my opinion.

Yes, I verified this on a live cluster. A single `tls.DialWithDialer` with `CurvePreferences []tls.CurveID{tls.X25519MLKEM768}` works. I tested all three PQ hybrid groups and they all succeed:

- `X25519MLKEM768` - accepted
- `SecP256r1MLKEM768` - accepted
- `SecP384r1MLKEM1024` - accepted

So it's definitely not a dead path. The pure Go approach covers PQC probing without any external tools. Makes sense to add the PQC assertion alongside the groups work after the vendor bump. WDYT ?

nunnatsa:
@carterpewpew - let's wait with it for the next realse, when we'll adopt TLS groups (carves). WDYT?

carterpewpew:
> @carterpewpew - let's wait with it for the next realse, when we'll adopt TLS groups (carves). WDYT?

Sounds good to me @nunnatsa , should I close this PR for now then ?

PR #4450, opened 2026-07-28
OCPBUGS-100031: Don't force single-replica KubeVirt infra on 2-node control planes

fonta-rh:
## Context

Currently HCO Forces `spec.infra.replicas: 1` on the KubeVirt CR (virt-api, virt-controller, virt-exportproxy, virt-synchronization-controller) whenever `IsControlPlaneHighlyAvailable()` is false. That flag requires 3+ control plane nodes, or 2 nodes plus an arbiter — so a 2-node control plane without an arbiter (Two Node OpenShift with Fencing / `DualReplica` topology) gets treated like a single-node cluster, even though there are 2 real nodes to spread replicas across.

Single `virt-api` replica removes redundancy for the eviction-interceptor webhook: draining the node hosting the only `virt-api` pod evicts it before the webhook can redirect `virt-launcher` eviction into a live migration. Eviction fails (`no endpoints available for service "virt-api"`) or force-kills the VM instead of migrating it.

This is a replica-count bug, not an "HCO doesn't know TNF is HA" bug. `IsControlPlaneHighlyAvailable()` is untouched — its degraded-cluster semantics (3→2 nodes still reports false) stay as-is. HCO has no visibility into fencing/quorum health; not claiming it now understands TNF topology.

### **Fix:** 
`hcoConfig2KvConfig`'s replica decision never needed "is the control plane HA" — it needs "how many nodes can we spread infra pods across." Added `nodeinfo.IsControlPlaneMultiNode()` (`cpNodeCount >= 2`) as that narrower signal and swapped it in for `IsControlPlaneHighlyAvailable()` at that call site.

Renamed the parameter (`controlPlaneHighlyAvailable` → `controlPlaneMultiNode`) in the same commit as the behavior change, since the old name no longer matched what it gated. Flagging that explicitly: a rename plus behavior change in one diff hides the actual delta on skim. The delta is: `Replicas` is no longer forced to `1` on exactly 2 control-plane nodes without an arbiter.

**Tests:** 4 new cases via `DualReplicaNodeInfoMock`, covering all Infra/Workloads placement combos in `kubevirt_test.go`. `go build`, `go test ./pkg/internal/nodeinfo/... ./controllers/handlers/...`, `gofmt -l`, `go vet` all clean.

EDIT: Verified on a TNT 4.22.7 cluster over release-1.16. 
2 replicas are created, and drain works properly while maintaining availability

```release-note
Fixed KubeVirt infrastructure components (virt-api, virt-controller) being forced to a single replica on two-node control planes, which removed redundancy for the eviction-interceptor webhook and could cause VM eviction failures during node drains.
```

### DIFF
--- a/controllers/handlers/kubevirt.go
+++ b/controllers/handlers/kubevirt.go
@@ -279,7 +279,7 @@ func NewKubeVirt(hc *hcov1.HyperConverged, opts ...string) (*kubevirtcorev1.Kube
 
 	kvCertConfig := hcoCertConfig2KvCertificateRotateStrategy(hc.Spec.Security.CertConfig)
 
-	controlPlaneHighlyAvailable := nodeinfo.IsControlPlaneHighlyAvailable()
+	controlPlaneMultiNode := nodeinfo.IsControlPlaneMultiNode()
 	controlPlaneNodeExists := nodeinfo.IsControlPlaneNodeExists()
 	infraHighlyAvailable := nodeinfo.IsInfrastructureHighlyAvailable()
 
@@ -293,7 +293,7 @@ func NewKubeVirt(hc *hcov1.HyperConverged, opts ...string) (*kubevirtcorev1.Kube
 		infra = np.Infra
 		workload = np.Workload
 	}
-	kvInfra := hcoConfig2KvConfig(infra, infraHighlyAvailable, controlPlaneHighlyAvailable, controlPlaneNodeExists)
+	kvInfra := hcoConfig2KvConfig(infra, infraHighlyAvailable, controlPlaneMultiNode, controlPlaneNodeExists)
 	kvWorkloads := hcoConfig2KvConfig(workload, true, true, true)
 
 	spec := kubevirtcorev1.KubeVirtSpec{
@@ -887,8 +887,8 @@ func NewKubeVirtWithNameOnly() *kubevirtcorev1.KubeVirt {
 }
 
 func hcoConfig2KvConfig(
-	nodePlacement *api.NodePlacement, infraHighlyAvailable, controlPlaneHighlyAvailable, controlPlaneNodeExists bool) *kubevirtcorev1.ComponentConfig {
-	if nodePlacement == nil && controlPlaneHighlyAvailable {
+	nodePlacement *api.NodePlacement, infraHighlyAvailable, controlPlaneMultiNode, controlPlaneNodeExists bool) *kubevirtcorev1.ComponentConfig {
+	if nodePlacement == nil && controlPlaneMultiNode {
 		return nil
 	}
 
@@ -907,7 +907,7 @@ func hcoConfig2KvConfig(
 		return kvConfig
 	}
 
-	if !controlPlaneHighlyAvailable {
+	if !controlPlaneMultiNode {
 		kvConfig.Replicas = ptr.To[uint8](1)
 	}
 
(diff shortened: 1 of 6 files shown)


controllers/handlers/kubevirt.go
```diff
@@ -907,7 +907,7 @@ func hcoConfig2KvConfig(
 		return kvConfig
 	}
 
-	if !controlPlaneHighlyAvailable {
+	if !controlPlaneMultiNode {
 		kvConfig.Replicas = ptr.To[uint8](1)
```
nunnatsa:
```suggestion
		kvConfig.Replicas = new(uint8(1))
```

pkg/internal/nodeinfo/nodeinfo.go
```diff
@@ -71,6 +71,9 @@ func processNodeInfo(nodes []corev1.Node, hc *hcov1.HyperConverged) bool {
 	newValue := cpNodeCount >= 3 || (cpNodeCount >= 2 && arbiterNodeCount >= 1)
 	changed := controlPlaneHighlyAvailable.Swap(newValue) != newValue
 
+	newValue = cpNodeCount >= 2
+	changed = controlPlaneMultiNode.Swap(newValue) != newValue || changed
```
orenc1, line 75: if a two control plane nodes cluster is a valid cluster configuration, without an arbiter node, wouldn't it be simpler just to set the condition as:
```
newValue := cpNodeCount >= 2
```
?
and remove the arbiter logic?

controllers/handlers/kubevirt.go
fonta-rh: Hey Nahshon! I... think that this is not part of my PR, and I'm not sure it will actually work, but in any case it would be out of scope for my change :)

pkg/internal/nodeinfo/nodeinfo.go
fonta-rh: Hello Oren!

Let's see if I can do a better job of explaining myself, sorry about the confusion!
Part of the proposed change is that the code was conflating **multi-node CP** with **highly-available CP**. In this part of the code, HCO only cares about number of nodes (to set the number of pod replicas), not if the control plane is actually highly available. 

As to why I didn't use the simplification you propose: 2 node CP is not highly available by default, only if fencing is installed and set up, while 2 node with arbiter is. 
I didn't want to cause issues in other current or potential uses of the IsControlPlaneHighlyAvailable function.

As part of looking into this bug, we realized that there might be future changes needed to have HCO be aware of a highly available 2-cp topology, but for now I just wanted to handle giving space for 2 replicas of virt-api when 2 CPs are available.

controllers/handlers/kubevirt.go
nunnatsa: it will work... Trying to move slowly to this format. but that fair. forget about it.

fonta-rh:
Working on testing this in a cluster right now, moved to ready-for-review for CI testing

fonta-rh:
Verified on a TNT 4.22.7 cluster over HCO release-1.16.
2 replicas are created, and drain works properly while maintaining availability

controllers/handlers/kubevirt.go
fonta-rh: I'll apply it then, you guys know your code better!

nunnatsa (approved):
/approve

jaypoulz:
@nunnatsa @orenc1 which versions support 4.22? we want to backport this to those versions

PR #4452, opened 2026-07-28
increase memory for the artifact DL init container

nunnatsa:
This is a copy of the fix from #4451 

The pod sometimes fails to start. It seems that increasing the limit and required memory for the init container helps.

**Jira Ticket**:

```jira-ticket
None
```

**Release note**:

```release-note
None
```


avlitman:
typo in commit description - “increasig”  - “increasing”

/approve

PR #4471, opened 2026-08-02
CNV-94341: validating wh: don't mask errors with warnings

nunnatsa:
In some cases, when warnings are returned from the validating webhook, they mask real errors, allowing wrong setting applied to the HyperConverged CR.

For example, when both a deprecated FG and a wrong jsonpatch annotation are set, the HCO Webhook will return a warning for the FG, but will allow the request. The result is a HyperConverged CR with a wrong annotation, that should be rejected.

This PR makes sure that warnings do not mask real errors.

**Jira Ticket**:

```jira-ticket
https://redhat.atlassian.net/browse/CNV-94341
```

**Release note**:

```release-note
None
```

### DIFF
--- a/pkg/webhooks/validator/v1beta1_validator.go
+++ b/pkg/webhooks/validator/v1beta1_validator.go
@@ -13,7 +13,6 @@ import (
 	"github.com/go-logr/logr"
 	openshiftconfigv1 "github.com/openshift/api/config/v1"
 	admissionv1 "k8s.io/api/admission/v1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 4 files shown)


pkg/webhooks/validator/v1beta1_validator.go
```diff
@@ -90,137 +91,117 @@ func (wh *WebhookV1Beta1Handler) Handle(ctx context.Context, req admission.Reque
...
+	warnings, err := wh.validateUpdateHyperConverged(requested, exists)
+	if err != nil {
+		return warnings, err
+	}
+
+	// If no change is detected in the spec nor the annotations - nothing to validate
+	if reflect.DeepEqual(exists.Spec, requested.Spec) &&
+		reflect.DeepEqual(exists.Annotations, requested.Annotations) {
+		return warnings, nil
 	}
```
avlitman, line 148: the order is not the same as v1 so even if not change it can now fail

nunnatsa: This is a behavior change I prefer not to do for v1beta1.

PR #4486, opened 2026-08-06
CNV-94424: enable readOnlyRootFilesystem in HCO pods

nunnatsa:
The artifact download server didn't work after adding the readOnlyRootFilesystem setting to its pod. 

This PR also added a new volume to the this pod, mount it to `/tmp`, and move all its write operations to write to `/tmp`. The logs are written to stdout/stderr. now, the server never writes to the root filesysetem, and it's working properly.

To ease this, the nginx.conf file is now added to the repo and copied to the image, instead of generating it from the Dockerfile.

It also allows to bump the nginx base image, because before that, the `sed` command in the Dockerfile, failed for the new base image. Now, when the file is pre-prepared, the server runs with no issues.

**Jira**:
```jira-ticket
https://redhat.atlassian.net/browse/CNV-94424
```

**Release note**:

```release-note
None
```

### DIFF
--- a/build/Dockerfile.artifacts
+++ b/build/Dockerfile.artifacts
@@ -36,16 +36,16 @@ RUN eval $(cat /tmp/config  |grep KUBEVIRT_VERSION=) && \
         done; \
     done
 
-FROM --platform=${TARGETPLATFORM} registry.access.redhat.com/ubi9/nginx-124
+FROM --platform=${TARGETPLATFORM} registry.access.redhat.com/ubi9/nginx-126
 ARG TARGETPLATFORM
 
 WORKDIR /opt/app-root/src
 
 COPY --from=downloader ./bin/ ./
+COPY build/assets/artifacts/nginx.conf /etc/nginx/nginx.conf
 
 USER 0
-RUN sed -i -E '/^ +location.*$/a\        }\n\n        location = /health {\n            access_log off;\n            return 200;' /etc/nginx/nginx.conf && \
-    sed '/^\s*listen\s*\[::\]:8080/d' /etc/nginx/nginx.conf > /etc/nginx/nginx.conf.ipv4 && \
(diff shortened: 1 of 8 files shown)


build/Dockerfile.artifacts
```diff
@@ -59,13 +59,13 @@ LABEL multi.GIT_URL=${git_url} \
 
 CMD if [[ -d "/proc/sys/net/ipv4" && -d "/proc/sys/net/ipv6" ]]; \
     then \
-      nginx -g "daemon off;"; \
+      exec nginx -e /dev/stderr -g "daemon off;"; \
```
orenc1, line 62: what is the benefit of adding `exec` to the command here?

nunnatsa: nginx will run as process 1. it will get and handle sigs directly.

tools/manifests/manifests.go
```diff
@@ -307,6 +307,15 @@ func GetDeploymentSpecCliDownloads(params *DeploymentOperatorParams) appsv1.Depl
...
 						TerminationMessagePolicy: corev1.TerminationMessageFallbackToLogsOnError,
+						VolumeMounts: []corev1.VolumeMount{
+							{Name: "nginx-tmp", MountPath: "/tmp"},
+						},
+					},
+				},
+				Volumes: []corev1.Volume{
+					{
+						Name:         "nginx-tmp",
+						VolumeSource: corev1.VolumeSource{EmptyDir: &corev1.EmptyDirVolumeSource{}},
```
orenc1: there is no `sizeLimit` for the `nginx-tmp` emptyDir volume, and it can allow the nginx process to consume unlimited node ephemeral storage

orenc1: although it's only a further security hardening, and the risk here is minimal

nunnatsa: fixed

orenc1 (approved):
/approve

PR #4493, opened 2026-08-11
CNV-94727: fix bad conversion on upgrade

nunnatsa:

For some unknown reasons, sometimes during an upgrade from v1.18, the featureGates field becomes an empty object instead of an empty array.

Trying to recover from this edge case by removing the featureGates field.

**Jira Ticket**:

```jira-ticket
```

**Release note**:

```release-note
None
```


orenc1 (approved):
/approve

PR #4500, opened 2026-08-12
Skip observability controller deployment when image is not configured

carterpewpew:

When `OBSERVABILITY_CONTROLLER_IMAGE` is set to an empty string in the operator pod (as is the case in downstream builds that do not yet ship the alpha observability controller image), enabling the `deployObservabilityController` feature gate causes the operator to attempt creating a deployment with an empty image field. The deployment object is never actually created, and the functional tests time out waiting for it (120s per test).

This PR adds an image availability check to `shouldDeploy` so the operator gracefully skips deployment when the image is not configured. It also updates the functional tests to detect this condition upfront and skip with a clear message, avoiding unnecessary timeouts.

Reproduced on both x86_64 (Jul 30) and s390x (Aug 5) downstream CI runs.

This PR was created with the help of AI, but every change has been manually tested and verified

**Reviewer Checklist**

- [x] PR Message
- [x] Commit Messages
- [x] How to test - Run func tests with `--ginkgo.label-filter="observability-controller"` against a cluster where `OBSERVABILITY_CONTROLLER_IMAGE` is empty in the operator deployment. Tests should skip gracefully instead of timing out.
- [x] Unit Tests - Added test case for "should not create when image env var is not set"
- [x] Functional Tests - Updated existing tests to skip when image is not configured
- [ ] User Documentation -> N/A
- [ ] Developer Documentation -> N/A
- [ ] Upgrade Scenario -> N/A (graceful degradation, no state change)
- [ ] Uninstallation Scenario -> N/A
- [ ] Backward Compatibility - Fully backward compatible; no behavior change when image is configured
- [ ] Troubleshooting Friendly -> N/A

**Jira Ticket**:
```jira-ticket
NONE
```

**Release note**:
```release-note
NONE
```

### DIFF
--- a/tests/func-tests/observability_controller_test.go
+++ b/tests/func-tests/observability_controller_test.go
@@ -5,6 +5,7 @@ import (
 	"crypto/tls"
 	"fmt"
 	"net/http"
+	"slices"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
@@ -120,10 +121,15 @@ var _ = Describe("Observability Controller Deployment", Label(tests.OpenshiftLab
 
 	BeforeAll(func(ctx context.Context) {
 		cli = tests.GetControllerRuntimeClient()
-	})
 
-	AfterAll(func(ctx context.Context) {
-		tests.RestoreDefaultFeatureGates(ctx, cli)
+		By("checking the observability controller image is configured")
+		configured, err := isObservabilityImageConfigured(ctx, cli)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(configured).To(BeTrueBecause(`the OBSERVABILITY_CONTROLLER_IMAGE environment variable should be configured in operator deployment. Use the "!observability-controller" label filter in order to skip this test`))
+
+		DeferCleanup(func(ctx context.Context) {
+			tests.RestoreDefaultFeatureGates(ctx, cli)
+		})
 	})
 
 	When("deployObservabilityController feature gate is enabled", func() {
@@ -216,8 +222,18 @@ var _ = Describe("Observability Controller Allowlist Configuration", Label(tests
 	BeforeAll(func(ctx context.Context) {
 		cli = tests.GetControllerRuntimeClient()
 
+		By("checking the observability controller image is configured")
+		configured, err := isObservabilityImageConfigured(ctx, cli)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(configured).To(BeTrueBecause(`the OBSERVABILITY_CONTROLLER_IMAGE environment variable should be configured in operator deployment. Use the "!observability-controller" label filter in order to skip this test`))
+
 		By("enabling the deployObservabilityController feature gate")
 		Expect(tests.EnableFG(ctx, cli, observabilityControllerFGName)).To(Succeed())
+
+		DeferCleanup(func(ctx context.Context) {
+			tests.PatchHCO(ctx, cli, []byte(`[{"op": "remove", "path": "/spec/observability"}]`))
+			tests.RestoreDefaultFeatureGates(ctx, cli)
+		})
 
 		By("waiting for the deployment to be created")
 		Eventually(func(ctx context.Context) error {
@@ -229,11 +245,6 @@ var _ = Describe("Observability Controller Allowlist Configuration", Label(tests
 			}
 			return cli.Get(ctx, client.ObjectKeyFromObject(dep), dep)
 		}).WithTimeout(2 * time.Minute).WithPolling(time.Second).WithContext(ctx).Should(Succeed())
-	})
-
-	AfterAll(func(ctx context.Context) {
-		tests.PatchHCO(ctx, cli, []byte(`[{"op": "remove", "path": "/spec/observability"}]`))
-		tests.RestoreDefaultFeatureGates(ctx, cli)
 	})
 
 	getDeploymentArgs := func(ctx context.Context) ([]string, error) {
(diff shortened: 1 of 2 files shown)


nunnatsa (dismissed):
Thanks for the PR @carterpewpew. It solves a temporary issue that should be fixed in a few days.

But to be on the safe side, let's do it. I would use a different approach, though.

The environment variable can't and won't be changed on runtime. There is no point to re-check if it set in every reconcile look. I would keep the code under `controllers/handlers/observabilitycontroller/` as is.

Instead, I would add a condition in https://github.com/kubevirt/hyperconverged-cluster-operator/blob/d42018c01e86cb328744d70ea110f6e25562bf9b/controllers/operandhandler/operandHandler.go#L77; e.g.
```go
	if ci.IsMonitoringAvailable()  && os.Getenv(hcoutil.ObservabilityControllerImageEnvV) != "" {
		operandList = append(operandList, []operands.Operand{
			observabilitycontroller.NewServiceAccountHandler(client, scheme),
			observabilitycontroller.NewClusterRoleHandler(client, scheme),
			observabilitycontroller.NewClusterRoleBindingHandler(client, scheme),
			observabilitycontroller.NewDeploymentHandler(client, scheme),
		}...)
	}

```

That way, the controller handlers won't even exist if the environment variable is not set.

carterpewpew:
> Thanks for the PR @carterpewpew. It solves a temporary issue that should be fixed in a few days.
> 
> But to be on the safe side, let's do it. I would use a different approach, though.
> 
> The environment variable can't and won't be changed on runtime. There is no point to re-check if it set in every reconcile look. I would keep the code under `controllers/handlers/observabilitycontroller/` as is.
> 
> Instead, I would add a condition in
> 
> https://github.com/kubevirt/hyperconverged-cluster-operator/blob/d42018c01e86cb328744d70ea110f6e25562bf9b/controllers/operandhandler/operandHandler.go#L77
> 
> ; e.g.
> ```go
> 	if ci.IsMonitoringAvailable()  && os.Getenv(hcoutil.ObservabilityControllerImageEnvV) != "" {
> 		operandList = append(operandList, []operands.Operand{
> 			observabilitycontroller.NewServiceAccountHandler(client, scheme),
> 			observabilitycontroller.NewClusterRoleHandler(client, scheme),
> 			observabilitycontroller.NewClusterRoleBindingHandler(client, scheme),
> 			observabilitycontroller.NewDeploymentHandler(client, scheme),
> 		}...)
> 	}
> ```
> 
> That way, the controller handlers won't even exist if the environment variable is not set.

Thanks for the review! That makes total sense, I'll move the check there.

tests/func-tests/observability_controller_test.go
```diff
@@ -120,10 +120,17 @@ var _ = Describe("Observability Controller Deployment", Label(tests.OpenshiftLab
...
-	})
 
-	AfterAll(func(ctx context.Context) {
-		tests.RestoreDefaultFeatureGates(ctx, cli)
+		By("checking the observability controller image is configured")
+		configured, err := isObservabilityImageConfigured(ctx, cli)
+		Expect(err).ToNot(HaveOccurred())
+		if !configured {
+			Skip("OBSERVABILITY_CONTROLLER_IMAGE not configured in operator deployment")
+		}
```
nunnatsa: We are trying to avoid `Skip` in our functional tests. Instead, we have labels that can be used to skip tests according to the specific environment.

To enforce this, we fail tests that are running in the an environment they should not run. Please change this to something like 

```suggestion
		if !configured {
			Fail("OBSERVABILITY_CONTROLLER_IMAGE not configured in operator deployment")
		}
```
Or

```suggestion
Expect(configured).To(BeTrueBecause("OBSERVABILITY_CONTROLLER_IMAGE should be configured in operator deployment"))
```

tests/func-tests/observability_controller_test.go
```diff
@@ -216,6 +223,13 @@ var _ = Describe("Observability Controller Allowlist Configuration", Label(tests
 	BeforeAll(func(ctx context.Context) {
 		cli = tests.GetControllerRuntimeClient()
 
+		By("checking the observability controller image is configured")
+		configured, err := isObservabilityImageConfigured(ctx, cli)
+		Expect(err).ToNot(HaveOccurred())
+		if !configured {
```
nunnatsa: same

tests/func-tests/observability_controller_test.go
```diff
@@ -327,3 +341,23 @@ func getServiceAccountToken(ctx context.Context) (string, error) {
...
+	depList := &appsv1.DeploymentList{}
+	err := cli.List(ctx, depList, client.InNamespace(tests.InstallNamespace), client.MatchingLabels{"name": "hyperconverged-cluster-operator"})
+	if err != nil {
+		return false, fmt.Errorf("failed to list operator deployments: %w", err)
+	}
+	if len(depList.Items) == 0 {
+		return false, fmt.Errorf("no hyperconverged-cluster-operator deployment found in namespace %s", tests.InstallNamespace)
+	}
+
+	for _, container := range depList.Items[0].Spec.Template.Spec.Containers {
```
nunnatsa: you can assume that the operator run on the first (and only) container. e.g.

```suggestion
	if (depList.Items[0].Spec.Template.Spec.Containers) == 0 {
		// should never get here
		return fmt.Errorf(...
	}
	container := depList.Items[0].Spec.Template.Spec.Containers[0]
```

tests/func-tests/observability_controller_test.go
```diff
@@ -327,3 +341,23 @@ func getServiceAccountToken(ctx context.Context) (string, error) {
...
+		for _, env := range container.Env {
+			if env.Name == "OBSERVABILITY_CONTROLLER_IMAGE" && env.Value != "" {
+				return true, nil
+			}
```
nunnatsa: please use the standard library for that; e.g.

```suggestion
		if slices.ContainsFunc(pod.Spec.Containers[0].Env, func(envVar corev1.EnvVar) bool {
		return envVar.Name == hcoutil.VirtIOWinDataFileEnvV && envVar.Value != ""
	}) {
				return true, nil
		}
```

nunnatsa (commented):
Thanks. Added a few comment inline.

tests/func-tests/observability_controller_test.go
nunnatsa: correction: add the label name to the error message, e.g.
```suggestion
		Expect(configured).To(BeTrueBecause(`the OBSERVABILITY_CONTROLLER_IMAGE environment variable should be configured in operator deployment. Use the "!observability_controller" label filter in order to skip this test`))
```

carterpewpew: Done. Switched to Expect(configured).To(BeTrueBecause(...)) in both suites.

tests/func-tests/observability_controller_test.go
carterpewpew: Fixed as well.

tests/func-tests/observability_controller_test.go
carterpewpew: Addressed

tests/func-tests/observability_controller_test.go
carterpewpew: switched to slices.ContainsFunc

carterpewpew:
Thank you @nunnatsa for the review, I have addressed all the comments PTAL.

nunnatsa (approved):
/approve

PR #4516, opened 2026-08-19
CNV-95652: [1.18] Don't force single-replica KubeVirt infra on 2-node control planes

nunnatsa:

**Note**: this is a manual cherry pick of PR #4450

Currently HCO Forces `spec.infra.replicas: 1` on the KubeVirt CR (virt-api, virt-controller, virt-exportproxy, virt-synchronization-controller) whenever `IsControlPlaneHighlyAvailable()` is false. That flag requires 3+ control plane nodes, or 2 nodes plus an arbiter — so a 2-node control plane without an arbiter (Two Node OpenShift with Fencing / `DualReplica` topology) gets treated like a single-node cluster, even though there are 2 real nodes to spread replicas across.

Single `virt-api` replica removes redundancy for the eviction-interceptor webhook: draining the node hosting the only `virt-api` pod evicts it before the webhook can redirect `virt-launcher` eviction into a live migration. Eviction fails (`no endpoints available for service "virt-api"`) or force-kills the VM instead of migrating it.

This is a replica-count bug, not an "HCO doesn't know TNF is HA" bug. `IsControlPlaneHighlyAvailable()` is untouched — its degraded-cluster semantics (3→2 nodes still reports false) stay as-is. HCO has no visibility into fencing/quorum health; not claiming it now understands TNF topology.

`hcoConfig2KvConfig`'s replica decision never needed "is the control plane HA" — it needs "how many nodes can we spread infra pods across." Added `nodeinfo.IsControlPlaneMultiNode()` (`cpNodeCount >= 2`) as that narrower signal and swapped it in for `IsControlPlaneHighlyAvailable()` at that call site.

**Changes in this PR**:
Renamed the parameter (`controlPlaneHighlyAvailable` → `controlPlaneMultiNode`) in the same commit as the behavior change, since the old name no longer matched what it gated. Flagging that explicitly: a rename plus behavior change in one diff hides the actual delta on skim. The delta is: `Replicas` is no longer forced to `1` on exactly 2 control-plane nodes without an arbiter.

**Jira Ticket**:

```jira-ticket
https://redhat.atlassian.net/browse/CNV-95652
```

**Release note**:
```release-note
Don't force single-replica KubeVirt infra on 2-node control planes
```

### DIFF
--- a/controllers/commontestutils/nodeinfomocks.go
+++ b/controllers/commontestutils/nodeinfomocks.go
@@ -4,6 +4,7 @@ import "github.com/kubevirt/hyperconverged-cluster-operator/pkg/nodeinfo"
 
 var (
 	origIsControlPlaneHighlyAvailable = nodeinfo.IsControlPlaneHighlyAvailable
+	origIsControlPlaneMultiNode       = nodeinfo.IsControlPlaneMultiNode
 	origIsControlPlaneNodeExists      = nodeinfo.IsControlPlaneNodeExists
 	origIsInfraHighlyAvailable        = nodeinfo.IsInfrastructureHighlyAvailable
 	origGetControlPlaneArchitectures  = nodeinfo.GetControlPlaneArchitectures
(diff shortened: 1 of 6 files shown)


orenc1 (approved):
/approve

PR #4520, opened 2026-08-20
Fix KubeVirt infra replicas forced to 1 on HA HyperShift clusters

orenc1:
On HyperShift clusters with no control plane nodes, hcoConfig2KvConfig unconditionally set Replicas=1 when !controlPlaneMultiNode. This flag is always false on HyperShift (zero CP nodes), so HA clusters with 2+ workers were incorrectly pinned to a single replica.

Split the condition so that Replicas=1 is only forced when a single CP node actually exists, or when there are no CP nodes and the infrastructure is not highly available (single worker). HA HyperShift clusters now leave replicas unset, letting KubeVirt default to 2.

**Reviewer Checklist**

- [ ] PR Message
- [ ] Commit Messages
- [ ] How to test
- [ ] Unit Tests
- [ ] Functional Tests
- [ ] User Documentation
- [ ] Developer Documentation
- [ ] Upgrade Scenario
- [ ] Uninstallation Scenario
- [ ] Backward Compatibility

**Jira Ticket**:

```jira-ticket
https://redhat.atlassian.net/browse/CNV-95245
```

**Release note**:

```release-note
Fix KubeVirt infra replicas forced to 1 on HA HyperShift clusters
```

### DIFF
--- a/controllers/handlers/kubevirt.go
+++ b/controllers/handlers/kubevirt.go
@@ -890,21 +890,16 @@ func hcoConfig2KvConfig(
 
 	kvConfig := &kubevirtcorev1.ComponentConfig{}
 
+	if shouldHaveVirtInfraSingleReplica(infraHighlyAvailable, controlPlaneMultiNode, controlPlaneNodeExists) {
+		kvConfig.Replicas = new(uint8(1))
+	}
+
 	// In case there are no control plane / master nodes in the cluster, we're setting
 	// an empty struct for NodePlacement so that kubevirt control plane pods won't have
 	// any affinity rules, and they could get scheduled onto worker nodes.
 	if nodePlacement == nil && !controlPlaneNodeExists {
 		kvConfig.NodePlacement = &kubevirtcorev1.NodePlacement{}
-		if !infraHighlyAvailable {
-			// if there is only one worker node and no control plane nodes,
-			// set the kubevirt control plane replica count to 1.
-			kvConfig.Replicas = new(uint8(1))
-		}
 		return kvConfig
-	}
-
-	if !controlPlaneMultiNode {
-		kvConfig.Replicas = new(uint8(1))
 	}
 
 	if nodePlacement == nil {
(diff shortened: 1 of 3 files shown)


controllers/handlers/kubevirt.go
```diff
@@ -909,7 +909,9 @@ func hcoConfig2KvConfig(
 		return kvConfig
 	}
 
-	if !controlPlaneMultiNode {
+	if controlPlaneNodeExists && !controlPlaneMultiNode {
+		kvConfig.Replicas = new(uint8(1))
+	} else if !controlPlaneNodeExists && !infraHighlyAvailable {
 		kvConfig.Replicas = new(uint8(1))
 	}
```
nunnatsa: We can simplify a bit to prevent code duplication:

```suggestion
	if controlPlaneNodeExists && !controlPlaneMultiNode {
		kvConfig.Replicas = new(uint8(1))
	} else if !controlPlaneNodeExists && !infraHighlyAvailable {
		kvConfig.Replicas = new(uint8(1))
	}

	// In case there are no control plane / master nodes in the cluster, we're setting
	// an empty struct for NodePlacement so that kubevirt control plane pods won't have
	// any affinity rules, and they could get scheduled onto worker nodes.
	if nodePlacement == nil && !controlPlaneNodeExists {
		kvConfig.NodePlacement = &kubevirtcorev1.NodePlacement{}

		return kvConfig
	}
```

nunnatsa: This code is right. But **personally** it read strange to me, because two conditions cause the same outcome. It looks like a bug at the first glace.

What do you think about using or condition, or adding a function to check it?
e.g. 

```suggestion
	if (controlPlaneNodeExists && !controlPlaneMultiNode) ||!controlPlaneNodeExists && !infraHighlyAvailable) {
kvConfig.Replicas = new(uint8(1))
}
```

OR
```go
func shouldHaveVirtInfraSingleReplica(infraHighlyAvailable, controlPlaneMultiNode, controlPlaneNodeExists bool) bool {
	return (controlPlaneNodeExists && !controlPlaneMultiNode) || (!controlPlaneNodeExists && !infraHighlyAvailable)
}
```

and then


```suggestion
	if shouldHaveVirtInfraSingleReplica(infraHighlyAvailable, controlPlaneMultiNode, controlPlaneNodeExists) {
		kvConfig.Replicas = new(uint8(1))
	}
```

orenc1: done.
the logic has been refactored to make it more clear and remove duplication.
added a helper function of `shouldHaveVirtInfraSingleReplica()`

nunnatsa (approved):
/approve 
/cherry-pick release-1.19

PR #4522, opened 2026-08-20
CNV-95686: document Subscription vs HyperConverged node placement

harsh-thakare:
## Summary
- The previous approach (labeling classic OCP infra nodes with `node-role.kubevirt.io/control-plane` so `virt-operator` could leave the control plane) is **not valid**. That label exists for **Hosted Control Plane**, where there are no Kubernetes control-plane nodes. When CP nodes exist, `virt-operator` must run there.
- This PR documents the mismatch between Subscription placement and HyperConverged placement, including which pods each API actually owns.

## What the docs now say
**OLM Subscription** places operator pods. These exist as soon as the operator is installed, even without a HyperConverged CR:
`aaq-operator`, `cdi-operator`, `cluster-network-addons-operator`, `hco-operator`, `hco-webhook`, `hostpath-provisioner-operator`, `hyperconverged-cluster-cli-download`, `kubevirt-migration-operator`, `ssp-operator`, `virt-operator`.

**HyperConverged CR** places operand pods. These exist only after the HyperConverged CR is created:
`cdi-apiserver`, `cdi-deployment`, `cdi-uploadproxy`, `kubemacpool-cert-manager`, `kubemacpool-mac-controller-manager`, `kubevirt-console-proxy`, `kubevirt-console-plugin`, `kubevirt-ipam-controller-manager`, `kubevirt-migration-controller`, `virt-api`, `virt-controller`, `virt-exportproxy`, `virt-template-validator`.

`virt-operator` is created by OLM, but required control-plane affinity means Subscription `spec.config` cannot move it off CP nodes on classic OpenShift.

## Related product docs PR
OpenShift manuals [1] and [2] are in `openshift/openshift-docs`, not this repo:
https://github.com/openshift/openshift-docs/pull/118704

## Test plan
- [x] Nodes controller unit tests still pass (`go test ./controllers/nodes`)
- [x] Product docs PR opened for the 4.22 node-placement assembly

Jira: https://issues.redhat.com/browse/CNV-95686


controllers/nodes/nodes_controller.go
```diff
@@ -161,11 +185,33 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+		return err
+	}
+
+	return c.Watch(
+		source.Kind[*csvv1alpha1.Subscription](
+			mgr.GetCache(), &csvv1alpha1.Subscription{},
+			handler.TypedEnqueueRequestsFromMapFunc(func(_ context.Context, _ *csvv1alpha1.Subscription) []reconcile.Request {
+				return []reconcile.Request{placementReq}
+			}),
+			subscriptionConfigPredicate{},
```
nunnatsa: we can't watch the subscription. First, there is no point watching it, because we're already watching the result - the deployment. But more important - this will probably make HCO crash on non-openshift clusters. I'll add ok-to-test to check it.

And finally, we must not add new code referring OLM v0 resources. They won't be available in OLM v1 anyway.

tools/csv-merger/csv-merger.go
```diff
@@ -815,6 +818,9 @@ func overwriteDeploymentSpecLabels(specs []csvv1alpha1.StrategyDeploymentSpec, c
 		}
 		overwriteWithStandardLabels(specs[i].Spec.Template.Labels, *hcoKvIoVersion, component)
 		overwriteWithStandardLabels(specs[i].Label, *hcoKvIoVersion, component)
+		if specs[i].Name == virtOperatorDeploymentName {
```
nunnatsa: This should come from  KubeVirt repo.

nunnatsa:
Not sure about this solution. First, watching the subscription is not an option. Need to dig a bit deeper.

/hold

Also, need to check if this solution is working on a plain kkubernetest cluster.
/ok-to-test

controllers/nodes/nodes_controller.go
harsh-thakare: Thanks — agreed. The Subscription watch is gone.

HCO no longer lists or watches OLM v0 Subscription objects. Operator placement is read from HCO's own Deployment `nodeSelector` (OLM applies the same Subscription config to every CSV operator Deployment, including this one). That Deployment is already in HCO's restricted cache, unlike `virt-operator`.

The extra Deployment watch is registered only on classic OpenShift (`IsOpenshift() && !IsHyperShiftManaged()`), so this path is not taken on plain Kubernetes or HCP.

tools/csv-merger/csv-merger.go
harsh-thakare: Removed. HCO no longer injects kubevirt control-plane affinity/toleration into the virt-operator CSV deployment. That belongs in the KubeVirt repo (same label HCO already applies on HCP workers).

harsh-thakare:
Thanks for the review @nunnatsa — the feedback is addressed in 41e40a8.

**Subscription watch (and OLM v0 APIs)**
- Removed the Subscription watch and all Subscription listing/fallback.
- Operator placement is taken from HCO's own Deployment `nodeSelector`. OLM copies Subscription `spec.config` onto every CSV operator Deployment, HCO has no required control-plane affinity so it can schedule first, and that Deployment is already in the restricted cache (`app=kubevirt-hyperconverged`). `virt-operator` is not.
- The Deployment watch is registered only on classic OpenShift (`IsOpenshift() && !IsHyperShiftManaged()`). On plain Kubernetes the nodes controller still only watches Nodes + HyperConverged, so it will not crash looking up a missing CRD.

**CSV affinity injection**
- Reverted the csv-merger change and the generated CSV dumps. virt-operator affinity stays owned by the KubeVirt repo.

**Tests**
```
go test ./pkg/internal/nodeinfo ./controllers/nodes
ok
```
Coverage now includes idempotent labeled/unlabeled nodes, missing HCO Deployment, Deployment Get errors, and the operator-deployment predicate. No Subscription fixtures remain.

Could you take another look, and `/unhold` if this addresses your concerns?

harsh-thakare:
@nunnatsa We dug deeper on the solution (including the “watching the Subscription is not an option” concern, and whether this actually unblocks `virt-operator` on classic OCP / plain Kubernetes). Summary below.

## Verdict

The **HCO-side design is the right one**, and it is the same contract HCP already uses:

1. Do **not** watch `Subscription` (OLM v0; missing on plain Kubernetes / OLM v1; would crash HCO).
2. Do **not** inject virt-operator affinity in HCO’s csv-merger (that belongs in kubevirt).
3. **Do** label selected non-kubernetes-control-plane nodes with `node-role.kubevirt.io/control-plane` so they can satisfy virt-operator’s affinity **without** being counted as Kubernetes control-plane (HCO HA logic keys off `node-role.kubernetes.io/control-plane` / `master` only).

That labeling is exactly what [#4303](https://github.com/kubevirt/hyperconverged-cluster-operator/pull/4303) already does for HCP workers. This PR extends it to classic OCP, but **only** for nodes that match operator `nodeSelector`.

## Why watching Subscription was wrong, and what we watch instead

OLM copies `Subscription.spec.config.nodeSelector` onto **every** CSV operator Deployment, including HCO and virt-operator. So the Deployment is the result; there is no need to watch the Subscription object.

Two details that were wrong in the previous revision and are fixed in `af92f3e`:

- The live OLM Deployment is named **`hco-operator`** (`hcoutil.OperatorName` / csv-merger `hcoDeploymentName`), **not** `hyperconverged-cluster-operator` (that is the SA / container name). Looking up the SA name would always `NotFound` on a real cluster, so labeling would never run.
- That Deployment is **not** in HCO’s restricted cache (`app=kubevirt-hyperconverged`). Operand Deployments HCO creates have that label; the OLM-managed `hco-operator` Deployment does not (CSV labels are `name=hyperconverged-cluster-operator` + `app.kubernetes.io/*`). Watching `mgr.GetCache()` for Deployments would never see it.

So this revision:

- Reads `nodeSelector` from **`hco-operator`** via the **API reader** (not the restricted cache).
- Watches that Deployment with a **dedicated namespaced informer** (`metadata.name=hco-operator`), classic OCP only (`IsOpenshift() && !IsHyperShiftManaged()`).
- Still does not list/watch Subscription.

On plain Kubernetes the extra watch is not registered, so HCO does not depend on the Subscription CRD.

## Why labeling alone is necessary but not sufficient

`virt-operator` required affinity is still:

```text
node-role.kubernetes.io/control-plane Exists
OR
node-role.kubernetes.io/master Exists
```

OLM `Subscription.spec.config` can set `nodeSelector` / `tolerations`, **not** replace required affinity. Infra nodes typically have neither kubernetes CP label, so:

`nodeSelector=infra` **AND** `required affinity=kubernetes CP` → `FailedScheduling`.

KubeVirt already owns the missing OR term: [`kubevirt#17763`](https://github.com/kubevirt/kubevirt/pull/17763) (you LGTM’d) adds `--with-kubevirt-control-plane-label` to csv-generator, which appends:

```text
OR node-role.kubevirt.io/control-plane Exists
```

plus a matching toleration. HCO’s current `create_virt_csv()` was **not** passing that flag, and the dumped 1.20 CSV still only has the kubernetes CP/master terms. HCP labeling in #4303 has the same dependency.

This PR now passes `--with-kubevirt-control-plane-label` in `hack/build-manifests.sh` `create_virt_csv()`. The affinity term then comes from **kubevirt’s csv-generator**, not HCO csv-merger. The committed CSV dump will pick it up on the next `make build-manifests` / component bump (we did not hand-edit the generated CSV).

## What stays unchanged

- Default install (only `kubernetes.io/os=linux` on the operator Deployment): no extra node labels; virt-operator still requires real kubernetes control-plane/master.
- Real CP/master nodes are never labeled.
- HCP worker labeling is unchanged.
- Empty-string labels use key-exists matching (missing key is not `""`).

Happy to adjust if you’d rather the kubevirt CSV flag live in a follow-up, or if you want a different way to observe `hco-operator` than the dedicated informer.

harsh-thakare:
The product manuals in [1] and [2] cannot be changed from this repository. This PR stays in `kubevirt/hyperconverged-cluster-operator` (HCO user docs). The OpenShift Virtualization manuals live in `openshift/openshift-docs`, so that content is in a separate PR:

https://github.com/openshift/openshift-docs/pull/118704

That docs PR adds the Subscription vs HyperConverged pod lists and the `virt-operator` control-plane affinity note to the 4.22 node-placement assembly.

harsh-thakare:
Docs content check for the related product-manuals PR is posted here:

https://github.com/openshift/openshift-docs/pull/118704#issuecomment-new

Summary: the 10 Subscription pods and 13 HyperConverged pods are present, plus the `virt-operator` control-plane affinity note.

Rendered Netlify preview is waiting on `/ok-to-test` on the openshift-docs PR. Expected URL after that:

https://118704--ocpdocs-pr.netlify.app/openshift-enterprise/latest/virt/post_installation_configuration/virt-node-placement-virt-components.html

harsh-thakare:
Follow-up to the placement docs: OLM **does** support `spec.config.affinity` (since ~v0.22), and that field overrides CSV/deployment `nodeAffinity`.

The earlier wording that Subscription config cannot replace `virt-operator` affinity was wrong. `nodeSelector` still cannot replace affinity, which is why a selector-only override leaves `virt-operator` unschedulable.

This commit documents `spec.config.affinity.nodeAffinity` as the workaround for placing `virt-operator` on infrastructure nodes, with a Subscription YAML example.

Product docs: https://github.com/openshift/openshift-docs/pull/118704

harsh-thakare:
## Cluster test: Subscription `affinity` workaround

Spun up a 3-node Kind cluster and reproduced the `virt-operator` placement case from this docs change.

Setup:
- `cnv-95686-control-plane` (`node-role.kubernetes.io/control-plane`)
- `cnv-95686-worker` labeled/tainted as infra (`example.io/example-infra-key=example-infra-value`)
- `cnv-95686-worker2` unlabeled worker

This is the scheduler-equivalent of OLM `Subscription.spec.config`: OLM copies `nodeSelector` / `tolerations` onto the CSV deployment and **replaces** `affinity` when `spec.config.affinity` is set.

### Case 1: `nodeSelector` + CSV control-plane affinity (does **not** work)

`virt-operator` required affinity for `control-plane`/`master` **and** infra `nodeSelector` + toleration:

```
0/3 nodes are available: 1 node(s) had untolerated taint {node-role.kubernetes.io/control-plane: }, 2 node(s) didn't match Pod's node affinity/selector.
```

Pod stayed `Pending`. `nodeSelector` does not replace `nodeAffinity`.

### Case 2: `spec.config.affinity.nodeAffinity` override (workaround)

Same infra `nodeSelector` + toleration, with `nodeAffinity` replaced by the infra label (as OLM does when `spec.config.affinity` is set):

```
virt-operator-affinity-override   1/1   Running   cnv-95686-worker
```

Pod scheduled on the infra node.

This matches the updated docs: use Subscription `spec.config.affinity.nodeAffinity` to place `virt-operator` on infra nodes; a `nodeSelector`-only override cannot.

nunnatsa:
@harsh-thakare - revisiting this PR, I think the documentation changes are fit better into the main README.md file. At the top of this document, there is a similar, but out-of date documentation.

PR #4528, opened 2026-08-24
CNV-95803: Drop the wasp-agent handlers

nunnatsa:

Auto-pilot is now enabled by default, and is responsible for WASP.

This PR removes the wasp resources creation from HCO, and also removes the leftover resources on upgrade.

**Jira Ticket**:
```jira-ticket
https://redhat.atlassian.net/browse/CNV-95803
```

**Release note**:
-->
```release-note
Stop deploying wasp agent resources.
```


enp0s3:
/lgtm

@nunnatsa Thank you!

PR #4534, opened 2026-08-25
test: guard KubeVirt maintenance and Descheduler documentation

fanzy618:
## Summary
- add a destructive HA functional test that cordons a worker, evicts a virt-launcher through the Kubernetes Eviction API, and verifies live migration with VMIM/VMI/launcher evidence
- add a served/storage CRD schema compatibility test for the HCO and VM evictionStrategy paths used by the node-maintenance howto
- add optional Descheduler compatibility checks for KubeVirt profiles, policy YAML, policy volume wiring, and EvictionsInBackground
- accept the graduated and legacy KubeVirt Descheduler profile names while keeping the documented fields version-aware

## Validation
- `gofmt -d` (clean)
- `go test ./tests/func-tests -run '^$'` (package compile passed; no live tests run)\n- `go vet ./tests/func-tests` (passed)\n- `git diff --check` (passed)\n\nThe live functional tests require a configured HA workload cluster with KubeVirt and, for the optional checks, Descheduler installed. This environment does not currently have a valid workload-cluster context, so no live migration or Descheduler rebalance result is claimed.

### DIFF
--- a/tests/func-tests/descheduler_kubevirt_compatibility_test.go
+++ b/tests/func-tests/descheduler_kubevirt_compatibility_test.go
@@ -4,67 +4,218 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/yaml"
 
 	tests "github.com/kubevirt/hyperconverged-cluster-operator/tests/func-tests"
 )
 
 const (
-	deschedulerCRDName = "kubedeschedulers.operator.openshift.io"
-	graduatedProfile   = "KubeVirtRelieveAndMigrate"
-	legacyProfile      = "DevKubeVirtRelieveAndMigrate"
-	deschedulerMode    = "Automatic"
+	deschedulerCRDName       = "kubedeschedulers.operator.openshift.io"
+	deschedulerLabelSelector = "app.kubernetes.io/name=descheduler"
+	rawPolicyAPIVersion      = "descheduler/v1alpha2"
+	graduatedProfile         = "KubeVirtRelieveAndMigrate"
+	legacyProfile            = "DevKubeVirtRelieveAndMigrate"
+	deschedulerMode          = "Automatic"
 )
+
+type deschedulerPolicyDocument struct {
+	APIVersion string                   `json:"apiVersion"`
+	Kind       string                   `json:"kind"`
+	Profiles   []map[string]interface{} `json:"profiles"`
+}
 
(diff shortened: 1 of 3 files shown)


nunnatsa:
Thanks for the PR @fanzy618.

I can't find out why these test should be in this repository. It seems to me that the right place is in the KubeVirt repository.

Please close this PR.

/hold

fanzy618:
Thanks for the review. The requested compatibility guard is intentionally placed in HCO's func-tests because it protects the HCO/VM API schema and the node-maintenance runbook's documented eviction contract. The destructive case is an HCO functional test: it cordons a worker, evicts the virt-launcher through the Kubernetes Eviction API, and verifies VMIM/VMI/launcher evidence.

The Descheduler check is optional and skips when the ACP plugin is absent; it guards the ACP-specific policy/configuration contract referenced by the companion ACP documentation. It does not claim a live rebalance result. If the project does not want this ACP integration contract in HCO, I can split that file into the ACP integration test repository while retaining the HCO schema and node-maintenance tests here.

nunnatsa:
> Thanks for the review. The requested compatibility guard is intentionally placed in HCO's func-tests because it protects the HCO/VM API schema and the node-maintenance runbook's documented eviction contract. 

1. VM schema is KubeVirt's, not HCO's. Same as the runbook.
2. Even if we want to check that a specific HCO field matches some kind of standard, we can't do that for fields that their types are defined elsewhere, because then we have no way to fix a failure.
3. In the case described in point 2 above, the field structure is taken from the vendor folder. This is not something that ever changed in a regular PR, but only when we bump the version of the dependency (KubeVirt API in this case), so it should not be tested, but even if it would, it should be a static check, not an e2e test.

> The destructive case is an HCO functional test: it cordons a worker, evicts the virt-launcher through the Kubernetes Eviction API, and verifies VMIM/VMI/launcher evidence.

All of this is implemented in KubeVirt and should tested there.

> The Descheduler check is optional and skips when the ACP plugin is absent; 

This is always the case: ACP is not known in this repo. Is that something you want to change? Let's discuss it in a new VEP.

tests/func-tests/descheduler_kubevirt_compatibility_test.go
```diff
@@ -0,0 +1,333 @@
...
+	deschedulerMode          = "Automatic"
+)
+
+type deschedulerPolicyDocument struct {
+	APIVersion string                   `json:"apiVersion"`
+	Kind       string                   `json:"kind"`
+	Profiles   []map[string]interface{} `json:"profiles"`
+}
+
+var _ = Describe("KubeVirt Descheduler documentation compatibility", Label("deschedulerCompatibility"), func() {
```
nunnatsa, line 37: The tests in this file are testing other component's logic, not HCO logic. The test must not be here.

tests/func-tests/kubevirt_documentation_compatibility_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+	tests "github.com/kubevirt/hyperconverged-cluster-operator/tests/func-tests"
+)
+
+const (
+	hcoCRDName = "hyperconvergeds.hco.kubevirt.io"
+	vmCRDName  = "virtualmachines.kubevirt.io"
+)
+
+var _ = Describe("KubeVirt documentation API compatibility", Label("documentationCompatibility"), func() {
```
nunnatsa, line 21: This test checks a dependency content, that can't be fixed here, and so must not be tested here.

tests/func-tests/node_maintenance_test.go
```diff
@@ -0,0 +1,344 @@
+package tests_test
```
nunnatsa, line 1: the test in this file validates the KubeVirt behavior that is not controlled and can't be fixed in this repository, and so it must not run here.

PR #4552, opened 2026-08-30
Fix conversion webhook crash on legacy featureGates empty-object shape

### LINKED ISSUE #4549: v1beta1→v1 conversion webhook fails on spec.featureGates: "cannot unmarshal object into Go struct field HyperConvergedSpec.spec.featureGates"
## Summary

On the `nightly` community-operator catalog tag (`quay.io/kubevirt/hyperconverged-cluster-index:nightly`), every `Get`/watch of the `HyperConverged` CR (created via the `hco.kubevirt.io/v1beta1` API, `spec: {}`) intermittently fails with:

```
Error getting HCO: conversion webhook for hco.kubevirt.io/v1, Kind=HyperConverged failed: json: cannot unmarshal object into Go struct field HyperConvergedSpec.spec.featureGates of type featuregates.HyperConvergedFeatureGates
```

This persists continuously (every ~5s poll, for 6+ minutes straight) until the caller's own client-side timeout gives up. The CSV in play both times this was observed was `kubevirt-hyperconverged-operator.v1.20.0`.

## Impact

Any client using the `v1beta1` HyperConverged API (e.g. via a dynamic/unstructured client) is blocked entirely from reading the object back after creation, since every read goes through the `v1beta1`→`v1` (hub) conversion webhook and that webhook itself errors on `spec.featureGates`.

## Observed behavior

- Confirmed live in CI, twice, testing against the same `nightly` catalog tag (`candidate-v1.20` channel, `kubevirt-hyperconverged-operator.v1.20.0`).
- The first occurrence self-resolved on a later retest (a different pull of the same `nightly` tag apparently didn't reproduce it), suggesting this may be tied to a specific build of the index/operator image rather than every `v1.20.0` CSV instance -- but it has now recurred, so it isn't a one-off fluke either.
- The `HyperConverged` CR is created minimally (`spec: {}`), so any `spec.featureGates` shape mismatch must come from a mutating/defaulting webhook populating it in a form the `v1` conversion function doesn't expect, or a genuine incompatibility between the `v1beta1` and `v1` `featureGates` schemas in this particular build.

## What we'd appreciate

- Confirmation of whether this is a known/transient issue with a specific `nightly` build, or an actual schema regression between `v1beta1` and `v1` `featureGates`.
- If it's the latter, a fix to the conversion webhook (or to whatever webhook populates `spec.featureGates` by default) so a freshly-created, empty-spec `HyperConverged` CR can always be read back without error.

Happy to provide additional CI logs/artifacts on request.

> [!Note]
> Responses generated with Claude

kaovilai:
## Summary

Root cause of #4549: some objects persisted before v1.19 have `spec.featureGates` stored as an empty JSON object (`{}`) instead of an empty array — a legacy artifact of an older, struct-based representation.

The mutating webhook already has a narrow recovery for this shape (`recoverBadFeatureGates` in `pkg/webhooks/mutator/hyperConvergedMutator.go`), but only when it hits an old-object decode failure during an `Update` admission request.

The CRD **conversion** webhook has no equivalent safety net: `controller-runtime` decodes the raw stored (hub, `v1`) bytes into `hcov1.HyperConverged` before `ConvertFrom`/`ConvertTo` ever run (see `vendor/sigs.k8s.io/controller-runtime/pkg/webhook/conversion/conversion.go`'s `handleConvertRequest`), so any `HyperConverged` object with this legacy shape becomes permanently unreadable via the `v1beta1` API — reproducing the exact error from the issue on every read.

## Fix

Give `HyperConvergedFeatureGates` (`api/v1/featuregates/feature_gates.go`) a custom `UnmarshalJSON` that treats an empty JSON object as an empty list, and returns a real error for a non-empty object (so genuine corruption isn't silently swallowed). Since this is a method on the type itself, it fixes every decode path uniformly (conversion webhook, mutating webhook, reconciler, apiserver) instead of only the one admission-webhook code path that already had a workaround.

## Testing

- `api/v1/featuregates/feature_gates_test.go`: unmarshal of `{}` (bare and nested in a struct field, mirroring the webhook decode) now succeeds as an empty list; a non-empty object is still rejected.
- `api/v1beta1/conversion_test.go`: reproduces the exact issue scenario — `json.Unmarshal` of raw hub-version bytes with `"featureGates":{}"` now succeeds, and the object converts via `ConvertFrom` cleanly.
- Verified both new tests fail with the exact quoted error from the issue on `main`, and pass with this fix.

Fixes #4549

> [!Note]
> Responses generated with Claude

### DIFF
--- a/api/v1/featuregates/feature_gates.go
+++ b/api/v1/featuregates/feature_gates.go
@@ -1,7 +1,9 @@
 package featuregates
 
 import (
+	"bytes"
 	"encoding/json"
+	"fmt"
 	"slices"
 	"strings"
 
@@ -72,6 +74,38 @@ func (fg *FeatureGate) UnmarshalJSON(bytes []byte) error {
 // +kubebuilder:validation:MaxItems=64
 // +kubebuilder:validation:XValidation:rule="self.all(x, self.exists_one(y, x.name.lowerAscii() == y.name.lowerAscii()))",message="feature gate names must be unique (case-insensitive)"
 type HyperConvergedFeatureGates []FeatureGate
+
+// UnmarshalJSON decodes a JSON array into the feature gate list.
+//
+// Some objects persisted before v1.19 have spec.featureGates stored as an empty
+// JSON object ("{}") instead of an empty array, a legacy artifact of an older,
+// struct-based representation. That shape can't be produced by any current
+// code path, but it can still be read back from etcd, and a strict array
+// decode of it fails every caller that unmarshals the object - including the
+// CRD conversion webhook, which has no chance to recover since it decodes the
+// raw bytes before ConvertFrom/ConvertTo ever run (see
+// pkg/webhooks/mutator/hyperConvergedMutator.go's recoverBadFeatureGates for
+// the equivalent admission-webhook-only recovery). Treat an empty object as
+// an empty list so a stale record like that can still be read.
+func (fgs *HyperConvergedFeatureGates) UnmarshalJSON(data []byte) error {
+	trimmed := bytes.TrimSpace(data)
+	if len(trimmed) > 0 && trimmed[0] == '{' {
+		var obj map[string]json.RawMessage
+		if err := json.Unmarshal(trimmed, &obj); err != nil {
+			return err
+		}
+
+		if len(obj) != 0 {
+			return fmt.Errorf("featureGates: got an object with %d field(s), expected a JSON array", len(obj))
+		}
+
+		*fgs = nil
+		return nil
+	}
+
+	type plain HyperConvergedFeatureGates
+	return json.Unmarshal(data, (*plain)(fgs))
+}
 
 // Enable enables a feature gate by its name
 func (fgs *HyperConvergedFeatureGates) Enable(name string) {
--- a/api/v1/featuregates/feature_gates_test.go
+++ b/api/v1/featuregates/feature_gates_test.go
@@ -157,6 +157,32 @@ var _ = Describe("FeatureGate", func() {
 			))
 		})
 
+		It("should unmarshal a legacy empty-object shape as an empty list (issue #4549)", func() {
+			fgs := featuregates.HyperConvergedFeatureGates{
+				{Name: "shouldBeCleared"},
+			}
+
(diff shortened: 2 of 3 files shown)


nunnatsa:
/ok-to-test

Thanks you very much for this PR @kaovilai !

I'm hunting this issue for a long time now. I couldn't find the root cause. I mean - yes, somehow the stored v1 object is with `spec.featureGates` of and empty object, but what caused it? I prefer to find the actual root cause before applying your fix, which makes sense, but it does not solve the actual issue, that I'm still looking for. For now, I couldn't find the root cause in release-1.18 not in release-1.19 branches. If the root cause is in some other component that somehow bypasses the conversion logic, then I'll merge this PR. 

In the meantime, please fix the commit message and add the missing release-notes to the PR description, because currently the PR is blocked.

nunnatsa:
@kaovilai - it seems that the fix in #4493 is not needed after your fix, and that fails the unit tests. we need to check if we can fully remove the logic added in #4493.

nunnatsa:
> @kaovilai - it seems that the fix in #4493 is not needed after your fix, and that fails the unit tests. we need to check if we can fully remove the logic added in #4493.

OK, the logic added to pkg/webhooks/mutator/hyperConvergedMutator.go in #4493 is now redundant and should be removed. 

However, let's keep the `"should recover from a bad featureGate format" ` unit test in pkg/webhooks/mutator/hyperConvergedMutator_test.go, but change it to expect res.Patches to be empty; i.e.
modify pkg/webhooks/mutator/hyperConvergedMutator_test.go lines 608 to `Expect(res.Patches).To(BeEmpty())`, and delete line 609.

nunnatsa:
I found the actual root cause. Details are here: https://github.com/kubevirt/hyperconverged-cluster-operator/issues/4549#issuecomment-5489625017

Giving this new information, the fix here is not enough, and will hide other issues.

nunnatsa:
Opened #4557 to try to solve this issue.

kaovilai:
Ok. thanks for review.

kaovilai:
Superceded by 4557

PR #4557, opened 2026-09-01
CNV-94727: Add the conversion webhook settings to the CRD

### LINKED ISSUE #4549: v1beta1→v1 conversion webhook fails on spec.featureGates: "cannot unmarshal object into Go struct field HyperConvergedSpec.spec.featureGates"

nunnatsa:

We have a race condition on upgrade, where OLM replaces the CRD with the new version's one. That removes the conversion webhook configurations from the CRD, until the webhook restores them. Since we moved to v1 as the stored API version, any update to the HyperConverged CR at this point, will skip the conversion webhook, and will corrupt the CR.

To fix this, we now adding the webhook configurations to the CRD.

We must use different service name, and to allow modifiying the namespace. For that, the csv generator tool accepts two new parameters. When generating CRD for the CSV generator, we're using the CSV webhook deployment name.

Fixes: #4549

**Jira Ticket**:
```jira-ticket
```

**Release note**:
```release-note
Fix CNV-94727
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,8 @@ else
 	TEMP_ARCH := $(UNAME_ARCH)
 endif
 
+INSTALLATION_NAMESPACE ?= kubevirt-hyperconverged
+
 ARCH ?= $(TEMP_ARCH)
 
 # Prow doesn't have docker command
@@ -37,7 +39,7 @@ DO=eval
 export JOB_TYPE=prow
 endif
 
-sanity: generate gogenerate prepare-tools-crd go-fix generate-doc validate-no-offensive-lang goimport lint-metrics lint-monitoring update-kv-fg-file
+sanity: generate gogenerate generate-crd generate-crd-for-csv go-fix generate-doc validate-no-offensive-lang goimport lint-metrics lint-monitoring update-kv-fg-file
 	go version
 	go fmt ./...
 	go mod tidy -v
@@ -62,7 +64,7 @@ build: build-operator build-csv-merger build-webhook
 build-operator: $(SOURCES) ## Build binary from source
 	go build -ldflags="${LDFLAGS}" -o _out/hyperconverged-cluster-operator ./cmd/hyperconverged-cluster-operator
 
-build-csv-merger: ## Build binary from source
+build-csv-merger: generate-crd-for-csv ## Build binary from source
 	go build -ldflags="${LDFLAGS}" -o _out/csv-merger ./tools/csv-merger
 
 build-manifest-templator: ## Build binary from source
@@ -77,7 +79,7 @@ sort-feature-gates:
 	jq 'sort_by((if .phase == "beta" then 0 elif .phase == "alpha" then 1 else 2 end), .name)' pkg/featuregatedetails/feature-gates.json > pkg/featuregatedetails/feature-gates.temp
 	mv pkg/featuregatedetails/feature-gates.temp pkg/featuregatedetails/feature-gates.json
 
-build-crd-creator: generate
+build-crd-creator: # this should run from a builder image. and so it assumes that everything is pre generated
 	go build -ldflags="${LDFLAGS}" -o _out/crd-creator ./tools/crd-creator
 
 build-manifest-splitter:
@@ -86,7 +88,7 @@ build-manifest-splitter:
 build-webhook: $(SOURCES) ## Build binary from source
 	go build -ldflags="${LDFLAGS}" -o _out/hyperconverged-cluster-webhook ./cmd/hyperconverged-cluster-webhook
 
-build-manifests: prepare-tools-crd build-csv-merger build-manifest-splitter build-manifest-templator
+build-manifests: generate-crd generate-crd-for-csv build-csv-merger build-manifest-splitter build-manifest-templator
 	DUMP_NETWORK_POLICIES=$(DUMP_NETWORK_POLICIES) ./hack/build-manifests.sh
 
 build-manifests-prev:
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -77,7 +79,7 @@ sort-feature-gates:
 	jq 'sort_by((if .phase == "beta" then 0 elif .phase == "alpha" then 1 else 2 end), .name)' pkg/featuregatedetails/feature-gates.json > pkg/featuregatedetails/feature-gates.temp
 	mv pkg/featuregatedetails/feature-gates.temp pkg/featuregatedetails/feature-gates.json
 
-build-crd-creator: generate
+build-crd-creator: # this should run from a builder image. and so it assumes that everything is pre generated
```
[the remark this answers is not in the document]
nunnatsa: As explained in the comment, this is now should running as part of image build. the `generate` make target can't run in this environment, and so I made sure it only called from targets that requires `generate`

Makefile
```diff
@@ -37,7 +39,7 @@ DO=eval
 export JOB_TYPE=prow
 endif
 
-sanity: generate gogenerate prepare-tools-crd go-fix generate-doc validate-no-offensive-lang goimport lint-metrics lint-monitoring update-kv-fg-file
+sanity: generate gogenerate generate-crd generate-crd-for-csv go-fix generate-doc validate-no-offensive-lang goimport lint-metrics lint-monitoring update-kv-fg-file
```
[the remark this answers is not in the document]
nunnatsa: not relevant here

tools/crd-creator/crd-creator.go
```diff
@@ -22,14 +23,21 @@ import (
...
-	fileName string
+	fileName    string
+	namespace   string
+	webhookName string
 )
 
 func init() {
 	flag.StringVar(&fileName, "output-file", "", "CRD output file name")
+	flag.StringVar(&namespace, "namespace", defaultNamespace, "the installation namespace")
+	flag.StringVar(&webhookName, "webhook-name", defaultWebhookName, "the webhook deployment name, to generate the service name out of it")
```
orenc1, line 40: No input validation on `--webhook-name` and `--namespace` flags.
Passing `--webhook-name=""` produces a service name of just "-service", and `--namespace=""` produces an empty namespace in the CRD

tools/crd-creator/crd-creator.go
```diff
@@ -101,11 +109,27 @@ func getOperatorCRD() (*extv1.CustomResourceDefinition, error) {
...
+				Service: &extv1.ServiceReference{
+					Namespace: namespace,
+					Name:      webhookName + "-service",
+					Path:      new(hcoutil.HCOConversionWebhookPath),
+					Port:      new(int32(hcoutil.WebhookPort)),
+				},
+			},
+			ConversionReviewVersions: []string{hcov1.APIVersionV1, hcov1beta1.APIVersionBeta},
+		},
+	}
```
orenc1, line 131: the webhook creation logic has no unit test coverage

Makefile
```diff
@@ -254,10 +256,12 @@ gogenerate: generate
 generate-crd: generate build-crd-creator
 	./_out/crd-creator --output-file=config/crd/bases/hco.kubevirt.io_hyperconvergeds.yaml
 	@echo "the CRD file was generated in config/crd/bases/hco.kubevirt.io_hyperconvergeds.yaml"
-
-prepare-tools-crd: generate-crd
-	cp config/crd/bases/hco.kubevirt.io_hyperconvergeds.yaml ./tools/csv-merger/generated-crd.yaml
 	cp config/crd/bases/hco.kubevirt.io_hyperconvergeds.yaml ./tools/manifest-templator/generated-crd.yaml
+	@echo "the CRD file was generated in tools/manifest-templator/generated-crd.yaml"
+
+generate-crd-for-csv: build-crd-creator # this should run from a builder image. and so it assumes that everything is pre generated
```
orenc1, line 262: `make build` target reaches `generate-crd-for-csv` without generate in the dependency chain, which might produce a CRD with an outdated schema

orenc1 (commented):
looks good.
couple of comments below

tools/crd-creator/crd-creator.go
nunnatsa: I see what you mean. But this is an internal tool. If someone will use it like that, well, it their problem.

tools/crd-creator/crd-creator.go
nunnatsa: no part of this tool is covered by a unit test. there are no tests for tools in general. This PR won't start it.

Makefile
nunnatsa: This is why I added the comment. This make target is meant to run in a builder image. The generate target will fail if running there, while when running, all the relevant (generated) code is assumed to be up-to-date.

nunnatsa: Running this make target directly makes no sense. And again - it's the user problem if they do that. it can fail in CI if they do that

tools/crd-creator/crd-creator.go
orenc1: right.
maybe worth thinking about adding such in future to mitigate potential regressions on critical functions.
/lgtm
/approve

PR #4558, opened 2026-09-01
CNV-94727: [1.19] Add the conversion webhook settings to the CRD

nunnatsa:

**Note**: this is a manual cherry-pick of #4557

We have a race condition on upgrade, where OLM replaces the CRD with the new version's one. That removes the conversion webhook configurations from the CRD, until the webhook restores them. Since we moved to v1 as the stored API version, any update to the HyperConverged CR at this point, will skip the conversion webhook, and will corrupt the CR.

To fix this, we now adding the webhook configurations to the CRD.

We must use different service name, and to allow modifiying the namespace. For that, the csv generator tool accepts two new parameters. When generating CRD for the CSV generator, we're using the CSV webhook deployment name.

Fixes: #4549

**Jira Ticket**:
```jira-ticket
```

**Release note**:
```release-note
Fix CNV-94727
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,8 @@ else
 	TEMP_ARCH := $(UNAME_ARCH)
 endif
 
+INSTALLATION_NAMESPACE ?= kubevirt-hyperconverged
+
 ARCH ?= $(TEMP_ARCH)
 
 # Prow doesn't have docker command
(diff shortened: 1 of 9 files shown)


orenc1 (approved):
/approve

PR #4569, opened 2026-09-06
CNV-96764: [1.19] fix observability controller TLS settings

nunnatsa:

HCO sets the TLS cipher configurations in the observability controller deployment, with the wrong command line argument, getting the observability controller pod into crash loop.

This PR fixes the issue by using the correct command line argument of `--tls-ciphers`.

**Jira Ticket**:
```jira-ticket
https://redhat.atlassian.net/browse/CNV-96764
```

**Release note**:

```release-note
None
```


orenc1 (approved):
/approve

PR #4604, opened 2026-09-22
Refactor the optional KV FG logic

nunnatsa:
The current logic of adding optional KubeVirt feature gates is long and confusing. Each time a feature gate is added, removed, or become enabled by default, it is required to add an explicit logic to it, and fix a const in the tests.

In addition, sonarCloud and the linter keeps adding warning about the complicity of the `getFeatureGateChecks()` function, that became very long, indeed.

This PR refactors the `getFeatureGateChecks()` by adding the two new the `kvExposedBetaFGs` and `kvExposedAlphaFGs` slices, and loop over them to add conditional feature gates that are exposed in the HyperConverged CR.

Now it's easier to add, remove or change the default state of an exposed feature gate, by just adding the FG name to one of these slices.

**Jira Ticket**:
```jira-ticket
None
```

**Release note**:
```release-note
None
```


michalskrivanek:
Is it useful to keep the Alpha/Beta distinction? It has no meaning for CNV, just two lists of gates that are on and off would be clearer. Ideally the complete set could be validated against kubevirt code

nunnatsa:
> Is it useful to keep the Alpha/Beta distinction? It has no meaning for CNV, just two lists of gates that are on and off would be clearer. Ideally the complete set could be validated against kubevirt code

Interesting. Because eventually we use the `featureGates.IsEnable()` methods that knows the HCO default for the specific FG. It make sense to unify the lists. 

However, splitting the lists is also important for the the unit testing, so the contributor that wants to add a feature gate, can add it in one and only one place.

EDIT: also I'm afraid there is a confusion here. These are **HCO** feature gates, not KubeVirt's.

Several of HCO feature gates, not all of them, are "proxy FGs" for KubeVirt. We only expose a subset of KubeVirt feature gates in HCO, when we have a good reason for that. This is not 1:1 relationship and we don't just set FG in KubeVirt CR if they have the same name.

michalskrivanek:
> > Is it useful to keep the Alpha/Beta distinction? It has no meaning for CNV, just two lists of gates that are on and off would be clearer. Ideally the complete set could be validated against kubevirt code
> 
> Interesting. Because eventually we use the `featureGates.IsEnable()` methods that knows the HCO default for the specific FG. It make sense to unify the lists.
> 
> However, splitting the lists is also important for the the unit testing, so the contributor that wants to add a feature gate, can add it in one and only one place.
> 
> EDIT: also I'm afraid there is a confusion here. These are **HCO** feature gates, not KubeVirt's.
> Several of HCO feature gates, not all of them, are "proxy FGs" for KubeVirt. 

I thought ll of them are kubevirt's. Which ones are not, and what's the purpose for them? 

> We only expose a subset of KubeVirt feature gates in HCO, when we have a good reason for that. This is not 1:1 relationship and we don't just set FG in KubeVirt CR if they have the same name.

And that's the frequent source of problems. IMHO all of them should be exposed to avoid the json patch instructions.
My point is - it would have been a lot more straightforward to simply expose _all_ of the kubevirt feature gates, and a simple "these are on", and "these are off lists, crosschecked so we automatically add and clean up from kubevirt. If we need a CNV dev preview feature to be highlighted separately that can be done via docs.

nunnatsa:
> I thought ll of them are kubevirt's. Which ones are not, and what's the purpose for them?

It's all in the documentation... see https://github.com/kubevirt/hyperconverged-cluster-operator/blob/main/docs/cluster-configuration.md#featuregates

> And that's the frequent source of problems

Sorry, but I disagree. This is one of the core concepts of HCO. HCO is a opinionated packaging and a single configuration place of the KubeVirt "sub-system" (not only KV, BTW). HCO is not and it won't be just a KubeVirt proxy - it wasn't designed for that.

We don't expose everything from KubeVirt API (as well as CDI, CANO, SSP and so on). We are trying to expose as less as we can and only for a good reason.

nunnatsa:
> to avoid the json patch instructions

This is an important point. But it does not change anything. If there is a real need for a specific KV FG, it can be added. Specially now when we moved to the v1 API version, it's super easy.

controllers/handlers/kubevirt_test.go
```diff
@@ -2093,6 +2059,7 @@ Version: 1.2.3`)
 						},
 						And(Not(ContainElement(kvGraceIOVirtualization)), Not(ContainElement(kvIOMMUFD)), Not(ContainElement(kvPCINUMAAwareTopology))),
 					),
+					// DeclarativeHotplugVolumes / HotplugVolumes
```
avlitman, line 2062: why it's needed?

controllers/handlers/kubevirt_test.go
```diff
@@ -2109,22 +2076,7 @@ Version: 1.2.3`)
...
-					),
-					Entry("should not add the ObjectGraph feature gate if ObjectGraph is false in HyperConverged CR",
-						func(hc *hcov1.HyperConverged) {
-							hc.Spec.FeatureGates = featuregates.HyperConvergedFeatureGates{
-								{Name: "objectGraph", State: new(featuregates.Disabled)},
-							}
-						},
-						Not(ContainElement(kvObjectGraph)),
-					),
+					// IncrementalBackup + UtilityVolume
```
avlitman, line 2079: same

controllers/handlers/kubevirt_test.go
nunnatsa: This comment explains the following entries

controllers/handlers/kubevirt_test.go
nunnatsa: This comment explains the following entries

PR #4605, opened 2026-09-22
CNV-97810: fix a blocker bug in upgrade

nunnatsa:
When upgrading, if the webhook controller did not fix the caBundle filed in the CRD before the v1 restore process started, the retry does not applied and the webhook crashes after only one attempt. That gets the webhook and the operator into a non-recoverable crash loop.

This PR fixes it by fixing the retry mechanism, to retry for any error, no matter what kind of error.

In addition, if the KV, CDI, CNA or SSP CRs are missing on upgrade validation request, the webhook will not exit with error. This is an edge case that will probably never happen, but if it does (for example in debug process), it prevents the recovery of the webhook. So it is safer to log this error instead of rejecting the request.

**Jira Ticket**:
```jira-ticket
https://redhat.atlassian.net/browse/CNV-97810
```

**Release note**:

```release-note
fix a blocker bug in upgrade
```


cmd/hyperconverged-cluster-webhook/main.go
```diff
@@ -313,33 +313,50 @@ func ensureAPIv1(ctx context.Context, cli client.Client, namespace string, logge
...
-		retryLogger.Info("re-storing the HyperConverged CR in API v1 format")
-		err = cli.Update(ctx, hc, client.FieldValidation("Ignore"))
-		if err != nil {
-			return fmt.Errorf("failed to re-store the HyperConverged CR in API v1 format; %w", err)
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-time.After(duration):
+			duration *= factor
 		}
```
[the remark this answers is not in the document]
nunnatsa: Fixed. @coderabbitai - please take another look

orenc1 (commented):
looks good and addresses the certificate error of the caBundle of the webhook upon upgrade.
since the sanity has failed and anyway you need to fix it, i added two minor comments below.

cmd/hyperconverged-cluster-webhook/main.go
```diff
@@ -313,33 +313,57 @@ func ensureAPIv1(ctx context.Context, cli client.Client, namespace string, logge
...
+			return ctx.Err()
+		case <-time.After(duration):
+			duration *= factor
+		}
+	}
+
+	return fmt.Errorf("failed to re-store the HyperConverged CR in API v1 format; %w", err)
+}
+
+func reStoreHCV1Imp(ctx context.Context, cli client.Client, namespace string, logger logr.Logger) error {
```
orenc1: naming inconsistency of `V1` vs. `v1` in `reStoreHCv1`.
they should be both upper case or lower case, with a preference to upper case.

cmd/hyperconverged-cluster-webhook/main.go
```diff
@@ -313,33 +313,57 @@ func ensureAPIv1(ctx context.Context, cli client.Client, namespace string, logge
...
-		}()
-
-		retryLogger := logger.WithValues("attempt", attempt)
-		retryLogger.Info("reading the HyperConverged CR in API v1 format")
-		hc, err := getHyperConvergedV1(ctx, cli, namespace, retryLogger)
-		if err != nil {
-			return err
-		}
+	const (
+		factor   = time.Duration(5)
```
orenc1: note that it's 5 nanoseconds. is that intended? or should it be milliseconds?

orenc1 (commented):
looks good and addresses the certificate error of the caBundle of the webhook upon upgrade.
since the sanity has failed and anyway you need to fix it, i added two minor comments below.

cmd/hyperconverged-cluster-webhook/main.go
nunnatsa: It's just because go won't let us multiply duration by int. We can move the casting to the actual expression if it more readable

nunnatsa: ho! it seems go1.26 allows that. It wasn't allowed before.

orenc1 (approved):
/approve

